PR #22, opened 2023-07-17
Kdata

ckolbPTB:
Basic implementation of KData class. Needs further work.


src/mrpro/data/raw.py
```diff
@@ -0,0 +1,426 @@
...
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+import os
+import re
+from abc import ABC
+from abc import abstractmethod
+from pathlib import Path
+from typing import Dict
+from typing import List
```
fzimmermann89: As we limit compatibility to newer then 3.10 anyways, we can just use list and dict instead of typing.List and typing.Dict

src/mrpro/data/raw.py
```diff
@@ -0,0 +1,426 @@
...
+             'ACQ_USER2',
+             'ACQ_USER3',
+             'ACQ_USER4',
+             'ACQ_USER5',
+             'ACQ_USER6',
+             'ACQ_USER7',
+             'ACQ_USER8')
+
+
+class ELimits():
```
fzimmermann89: This could also be a dataclass  in Python 3.10:

@dataclass(slots=True)
class Foo:
    bar: int
    baz: int

src/mrpro/data/raw.py
```diff
@@ -0,0 +1,426 @@
...
+        self.slice_dir = torch.zeros(
+            (num_acq, ismrmrd.constants.DIRECTION_LENGTH), dtype=torch.float32)
+        self.trajectory_dimensions = torch.zeros((num_acq,), dtype=torch.int64)
+        self.user_float = torch.zeros(
+            (num_acq, ismrmrd.constants.USER_FLOATS), dtype=torch.float32)
+        self.user_int = torch.zeros(
+            (num_acq, ismrmrd.constants.USER_INTS), dtype=torch.int64)
+        self.version = torch.zeros((num_acq,), dtype=torch.int64)
+
+    def from_ismrmrd_acq_header(self, curr_idx: int, acq: ismrmrd.Acquisition) -> None:
```
fzimmermann89: This is named like a class method but is an instance method. Could we make it a class method?

src/mrpro/data/raw.py
```diff
@@ -0,0 +1,426 @@
...
+from typing import List
+
+import ismrmrd
+import numpy as np
+import torch
+
+from mrpro.data.traj import KTrajectory
+
+# Acquisition flags labelling each readout. Readouts can have multiple flags.
+ACQ_FLAGS = ('ACQ_NO_FLAG',
```
fzimmermann89: Maybe there are better structs to represent flags than a tuple of strings? Enum?

src/mrpro/data/raw.py
```diff
@@ -0,0 +1,426 @@
...
+    return (_return_par_tensor(par, array_attr=('minimum', 'maximum', 'center')))
+
+
+def return_acc_factor_tensor(par: ismrmrd.xsd.ismrmrdschema.ismrmrd.accelerationFactorType) -> torch.Tensor | None:
+    return (_return_par_tensor(par, array_attr=('kspace_encoding_step_1', 'kspace_encoding_step_2')))
+
+
+def bitmask_flag_to_strings(flag: int):
+    if flag > 0:
+        bmask = '{0:064b}'.format(flag)
```
fzimmermann89: This looks interesting, is there no easier way to unpack to bitfieled than using a regex?

src/mrpro/data/raw.py
```diff
@@ -0,0 +1,426 @@
...
+    if par is None:
+        return None
+    else:
+        par_tensor = []
+        for attr in array_attr:
+            par_tensor.append(getattr(par, attr))
+        return (torch.tensor(par_tensor))
+
+
+def return_par_matrix_tensor(par: ismrmrd.xsd.ismrmrdschema.ismrmrd.matrixSizeType) -> torch.Tensor | None:
```
fzimmermann89: Personally not a big fan of method names "return... "
Also, these helper functions might be candidates for static methods of the classes using them

src/mrpro/data/raw.py
```diff
@@ -0,0 +1,426 @@
...
+                curr_attr[curr_idx, ...] = torch.tensor(
+                    getattr(acq.idx, slot), dtype=curr_attr.dtype)
+
+            else:
+                curr_attr[curr_idx, ...] = torch.tensor(
+                    getattr(acq, slot), dtype=curr_attr.dtype)
+            setattr(self, slot, curr_attr)
+
+
+def _return_par_tensor(par, array_attr) -> torch.Tensor | None:
```
fzimmermann89, line 21: We should add a Docstring here at some point

src/mrpro/data/raw.py
```diff
@@ -0,0 +1,426 @@
...
+                 'interleave_dim', 'multiband', 'traj_type', 'traj_description', 'device_id', 'device_sn', 'coil_label', 'station_name',
+                 'h1freq', 'ref_frame_uid', 'series_num', 'meas_depend', 'ref_im_seq', 'rel_table_pos', 'seq_name', 'series_date',
+                 'series_description', 'series_uid', 'series_time', 'te', 'ti', 'tr_siemens', 'fa', 'tr', 'seq_type', 'diffusion',
+                 'diffusion_dim', 'diffusion_scheme', 'accession_number', 'body_part', 'ref_physician', 'study_date', 'study_description',
+                 'study_id', 'study_uid', 'study_time')
+
+    def __init__(self) -> None:
+        pass
+
+    def from_ismrmrd_header(self, header: ismrmrd.xsd.ismrmrdschema.ismrmrd.ismrmrdHeader, num_acq: int) -> None:
```
fzimmermann89: This also looks like a class method, or?

src/mrpro/data/raw.py
```diff
@@ -0,0 +1,426 @@
...
+
+        sort_ki = np.stack((kheader.acq_info.kspace_encode_step_1, kheader.acq_info.kspace_encode_step_2, kheader.acq_info.average,
+                            kheader.acq_info.slice, kheader.acq_info.contrast, kheader.acq_info.phase,
+                            kheader.acq_info.repetition, kheader.acq_info.set), axis=0)
+        sort_idx = np.lexsort(sort_ki)
+
+        new_shape = (np.prod(kdim_num[2:]), kdim_num[1], kdim_num[0],)
+        kdata = torch.reshape(kdata[sort_idx, :, :],
+                              new_shape + kdata.shape[1:])
+        kdata = torch.moveaxis(kdata, (0, 1, 2, 3, 4), (0, 2, 3, 1, 4))
```
fzimmermann89: I want to advertise einops for all torch moveaxis/permute operations.
https://einops.rocks/api/rearrange/
This makes it way more explicit what is being done there.

src/mrpro/data/raw.py
```diff
@@ -0,0 +1,426 @@
...
+
+    def __init__(self) -> None:
+        pass
+
+    def from_ismrmrd_header(self, header: ismrmrd.xsd.ismrmrdschema.ismrmrd.ismrmrdHeader, num_acq: int) -> None:
+
+        # Encoding
+        assert len(header.encoding) == 1, 'Multiple encodings are not supported.'
+        enc = header.encoding[0]
+        self.etl = enc.echoTrainLength
```
fzimmermann89: Are we sure that we want to use these abbreviations instead of the more descriptive longer names?

src/mrpro/data/raw.py
schuenke: Looks interesting indeed, but I personally prefer not to use too many third party packages if they are not required for functionality.

fzimmermann89: I'd argue that the readability benefits for this common operation (and pack/unpack/reduce/mix) greatly outweighs the burden of having a python-only dependency...

src/mrpro/data/raw.py
ckolbPTB: Probably, happy for any suggestions.

src/mrpro/data/raw.py
ckolbPTB: Wouldn't rearrange just change this to kdata = rearrange(kdata, (a,b,c,d,e), (a,d,b,c,e))? Maybe I have been using numpy.moveaxis for too long, but I don't find reading this easier. Happy to hear other opinions.

src/mrpro/data/raw.py
ckolbPTB: I find the ISMRMRD names just far too long. I think having names like "encoded_space_matrix_size" would make the code very difficult to read (especially with a 79 character limit)

src/mrpro/data/raw.py
ckolbPTB: Could you suggest an alternative?

src/mrpro/data/raw.py
fzimmermann89: ... The idea would be to name the dimensions to have self-documenting code that makes it more obvious why the axes are moved and what the old order and the new order of axes is :D 
So naming them a b c d e does not help, yes. :D

src/mrpro/data/raw.py
fzimmermann89: You have more experience with MR data, to me it seems we should maybe aim for a compromise
and keep them a bit easier to understand? 
For example, ref_im_seq and ref_physician are kind of confusing. Many of the abbreviation are not obvious. It should be possible for a PhD student  with basic MR understanding to guess the meaning of most variables I'd argue..

src/mrpro/data/raw.py
fzimmermann89: If we want to stick with stdlib functions:
```
from enum import Flag, IntFlag, auto

class AcqFlags(Flag): #or IntFlags if we need more int-like behaviour
    FLAG1 = auto() # auto() will assign the next power of 2 to the flag.
    FLAG2 = auto() # so FLAG2 will be 2, FLAG3 will be 4, FLAG4 will be 8 etc.
    FLAG3 = auto() # we can also assign values manually, e.g.
    FLAG4 = 8
# We can create Flags either by using the name of the flags:    
a=AcqFlags.FLAG1|AcqFlags.FLAG2
# or by using the value of the flags (as we would do it when reading from file):
b=AcqFlags(5) # 5 = 1+4, so FLAG1 and FLAG3 are set
print("a", a, a.value)
print("b", b,b.value)

# To check if a flag is set, we can use the "in" operator:
print(AcqFlags.FLAG2 in b)

# We can also combine flags etc as we would expect:
c=a|b
print(c)

#The only issue i know:
if c.FLAG4: # will execute even though FLAG4 is not set!
    print(c.FLAG4) # will not print False but  AcqFlags.FLAG4 ..
# .. so we need to use the "in" operator.
```

src/mrpro/data/raw.py
fzimmermann89: see enum.Flags above maybe?

src/mrpro/data/raw.py
fzimmermann89: Or if we use a library (py-flags):
```
#! pip install py-flags

from flags import Flags

class AcqFlags(Flags): 
    FLAG1 = () # () will assign the next power of 2 to the flag.
    FLAG2 = () # so FLAG2 will be 2, FLAG3 will be 4, FLAG4 will be 8 etc.
    FLAG3 = () # we can also assign values manually, e.g.
    FLAG4 = 8
    
# We can create Flags either by using the name of the flags:    
a=AcqFlags.FLAG1|AcqFlags.FLAG2
# or by using the value of the flags (as we would do it when reading from file):
b=AcqFlags(5) # 5 = 1+4, so FLAG1 and FLAG3 are set
print("a", a)
print("b", b)

# To check if a flag is set, we can use the "in" operator:
print(AcqFlags.FLAG2 in b)

# We can also combine flags etc as we would expect:
c=a|b
print(c)

#This works now:
if c.FLAG4: # will not execute as FLAG4 is not set
    print(c.FLAG4)
else:
    print("FLAG4 not set")

```
I am split. py-flags is actually quite a bit higher quality IMHO compared to the current stdlib implementation, but even the author recommends to check out if the stdlib flags already fulfills the requirements ..

src/mrpro/data/raw.py
ckolbPTB: Yes to both. What about `par_matrix_to_tensor`?

src/mrpro/data/raw.py
ckolbPTB: Looks good. py-flags seems to do things better than enum but also looks unmaintained. Maybe safer to go with enum.

src/mrpro/data/raw.py
ckolbPTB: Can you make some suggestions on how you would improve the names?

src/mrpro/data/raw.py
ckolbPTB: This is part of code which should definitely be improved. Currently this function extracts the information about the position of a single acquisition (i.e. readout + coils) and adds it at a certain position given by `curr_idx` into the AcqInfo tensors. It is therefore not a class method. Maybe needs renaming.

src/mrpro/data/raw.py
ckolbPTB: Yes, this should be a class method

ckolbPTB: Maybe we should also change the order and
1) create AcqInfo acq_info
2) fill acq_info e.g. `acq_info.from_ismrmrd_acq_header(idx, acq)`
3) create KHeader with `from_ismrmrd(cls, header: ismrmrd.xsd.ismrmrdschema.ismrmrd.ismrmrdHeader, acq_info: AcqInfo) -> KHeader`

src/mrpro/data/raw.py
ckolbPTB: E.g. kdata = rearrange(kdata, (dim4, k2, k1, coils, k0), (dim4, coils, k2, k1, k0))?

fzimmermann89:
I just rebased Patrick's and my work so we only have to discuss in one PR the remaining changes...

(And sorry for the 'Update'.. commits, I did not notice that whenever I tried to accept one of Christophs changes it created a new commit with this unhelpful commit message)

fzimmermann89:
superseded by #29

PR #24, opened 2023-07-19
Kdata

schuenke:


src/mrpro/data/raw.py
```diff
@@ -367,46 +482,94 @@ def from_file(cls,
...
+            ktraj[sort_idx, :, :], new_shape + ktraj.shape[1:]
+        )
         ktraj = torch.moveaxis(ktraj, (0, 1, 2, 3, 4), (0, 2, 3, 1, 4))
 
         for slot in kheader.acq_info.__slots__:
             curr_attr = getattr(kheader.acq_info, slot)
+            # TODO: Check for correct dimensionality in test function!
             if curr_attr.ndim == 2:
-                curr_shape = new_shape + (curr_attr.shape[1],)
+                curr_shape: Any = new_shape + (curr_attr.shape[1],)
```
schuenke, line 565: This is to prevent mypy from further checking the variable, which seems to be problematic when iterating over the different slots that might have different dimensionality.

Issue #28, opened 2023-07-21
Supported Versions of Dependencies


fzimmermann89:
We should consider the versions of the dependencies we support, specify them and test at least the bounds in our tests.
For example, we maybe want to use
 - pytorch>=2.0,<3.0 (our tests should test 2.0.0 and latest)
 - numpy>=1.20 (??  numpy typing was introduced at 1.20 I believe)
 
 etc.
 This is not urgent, and should be reevaluated from time to time

fzimmermann89:
Especially if we want to depend on pulseq (#78), we have to track which versions are compatible and what features we can use.

fzimmermann89:
we now limit xsd as ismrmrd python has a bug in there dependencies.
we will have to remove this later on!

ckolbPTB:
utilise e.g. https://pypi.org/project/pessimist/ to check requirements. 

run once every week as an github action?

ckolbPTB:
maybe we could also enable automatic update PRs:

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates

schuenke:
I just set up dependabot version updates for all python packages and all gh actions for the [BMCTool Repo](https://github.com/schuenke/BMCTool). A simple config that allready groups the updates into single PR for each package ecosystem looks like this:

```yaml
version: 2
updates:
- package-ecosystem: pip
  directory: "/"
  schedule:
    interval: daily
  groups:
    python-packages:
      patterns:
        - "*"

- package-ecosystem: "github-actions"
  directory: "/"
  schedule:
    interval: "daily"
  groups:
    all-actions:
      patterns: [ "*" ]

```

This is how an example PR looks like: https://github.com/schuenke/BMCTool/pull/31


NOTE: for the coverage report to work also in the dependabot PRs, we have to add the GIST secret to the dependabot secrets as well. Probably need to create a new token, if @ckolbPTB didn't save it somewhere. But it's done in a minute.

lrlunin:
Resolved by #709, I guess? @fzimmermann89

Issue #23, opened 2023-07-18
Siemens Special Tab


mxlutz:
It would be nice if we could load in the attruibutes from the Siemens special tab automatically.

Would it be easiest to add a Dictionary Attribute to the KData Header?

Also to save the Special Tab in the ismrmrd files some small changes in the xls and xlm of siemens_to_ismrmrd are necessary, is that something we want to standardize?

fzimmermann89:
Good point!

Can you provide an example of an xsl and an ismrmrd file? (Does not have to be public)

I believe it would be good to consider this already now when were are designing the header class

mxlutz:
Yes sure! I did that for VB17, so probably for the new scanner we have to adapt it again (or maybe its already in the default one for XA or what file siemens_to_ismrmrd will use). I had to change the filendings to txt to attach it here

In the xsl the following lines are important:
```
                <xsl:for-each select="siemens/MEAS/sWipMemBlock/alFree">
                    <xsl:if test=". &gt; 0">
                      <userParameterDouble>
                            <name>alWipMemBlock</name>
                        <value>
                            <xsl:value-of select=". " />
                        </value>
                        </userParameterDouble>
                    </xsl:if>
                </xsl:for-each> 

                <xsl:for-each select="siemens/MEAS/sWipMemBlock/adFree">
                    <xsl:if test=". &gt; 0">
                        <userParameterDouble>
                            <name>adWipMemBlock</name>
                            <value>
                                <xsl:value-of select=". " />
                            </value>
                        </userParameterDouble>    
                    </xsl:if>
                </xsl:for-each> 
```
The filoper.py shows how to pass the custom files to siemens_to_ismrmrd

I think the xml was actually not important, I also added some lines to read out the RF Amplifer reference amplitude that is important for me but usually not read, I dont know whether we want to add that?

[fileoper.py.txt](https://github.com/ckolbPTB/mrpro/files/12077822/fileoper.py.txt)
[IsmrmrdParameterMap_Siemens_custom_Max.xsl.txt](https://github.com/ckolbPTB/mrpro/files/12077824/IsmrmrdParameterMap_Siemens_custom_Max.xsl.txt)
[IsmrmrdParameterMap_Siemens_VB17_custom_Max.xml.txt](https://github.com/ckolbPTB/mrpro/files/12077826/IsmrmrdParameterMap_Siemens_VB17_custom_Max.xml.txt)

mxlutz:
[...]

Issue #21, opened 2023-07-12
Create test data


ckolbPTB:
Create test data for unit tests.

Check list:
- [ ]     Create branch
- [ ]     Implement code
- [ ]     Add comments
- [ ]     Implement tests
- [ ]     Squash commmits
- [ ]     Create PR
- [ ]     Assign reviewer(s) to PR

ckolbPTB:
Can we somehow use test data creation functions of ismrmrd: e.g. https://github.com/ismrmrd/ismrmrd-python/blob/master/tests/test_common.py  -> No. Copy and modify code.

fzimmermann89:
I want to suggest using pytest for testing and fixtures to generate the data necessary

https://docs.pytest.org/en/6.2.x/fixture.html#quick-example

@schuenke @schote (or whoever else has opinions regarding tests): What do you think?

schuenke:
I am not 100% sure if I understood the advantage of the fixtures, but I also don't see any disadvantages aside from the additional third party package. So for me using pytest would be fine. Maybe somebody else (e.g. @JoHa0811) has a stronger opinion?

fzimmermann89:
> Can we somehow use test data creation functions of ismrmrd: e.g. https://github.com/ismrmrd/ismrmrd-python/blob/master/tests/test_common.py -> No. Copy and modify code.

The file above only creates random acquisition. 
This is in 88da941556112074d71a59f3242f255c568c1719 and used by 51bf7e6eba4032e871f24d66f99b92ed8f13a6d7

Biggest issue is the header.
Here we should missing information to get good coverage.
Or maybe fixup some existing 'create random XML given the xsd schema' code to fuzz the header parsing.


Also maybe useful:
https://github.com/ismrmrd/ismrmrd-python-tools/blob/17fe6fbf1bb645112e2ad0023eb4f42afb36421e/generate_cartesian_shepp_logan_dataset.py
which creates matching information and useful data

fzimmermann89:
... which, btw, is also copied at https://github.com/kspaceKelvin/python-ismrmrd-server/blob/master/generate_cartesian_shepp_logan_dataset.py

and might be what @ckolbPTB had in mind.

I'd argue to copy it to our repo as well and make it a pytest fixture :D

ckolbPTB:
I was thinking of this: https://github.com/ismrmrd/ismrmrd/blob/master/utilities/generate_cartesian_shepp_logan.cpp
[...]

PR #29, opened 2023-07-26
Initial work on KData and ISMRMRD loading

### LINKED ISSUE #1: KData class
Dataclass for k-space data

```
class KData():
    KData.header -> ismrmrd header provided by pyismrmrd module
    KData.traj -> torch.Tensor, 5D: [everything_else, 3, k2, k1, k0]
    KData.data -> torch.Tensor, 5D: [everything_else, coils, k2, k1, k0]
```

Check list:
- [x]     Create branch
- [x]     Implement code
- [x]     Add comments
- [x]     Implement tests
- [x]     Squash commmits
- [x]     Create PR
- [x]     Assign reviewer(s) to PR

### LINKED ISSUE #26: Units of class attributes
Currently, some of our class attributes (e.g. KHeader.recon_fov) are neither in SI units, nor the used unit is added to the name of the attribute. 

For the Inhommcor project we decided (https://github.com/ckolbPTB/InhomCorr/issues/15) that we:

-     always use SI units and mention it in the documentation
-     keep the attribute names without units, but mention the units in all docstrings
-     in our code, add some comments about units where it might be useful (e.g. if dB0 is in Hz or ppm)

IMO, we should do the same for the MRpro project

### LINKED ISSUE #27: Structured Data VS Lists or Tensors
There are some parts of the KData code where Tensors or Lists are used to save values, and the different positions in the tensor have different meaning.

An example are/were the Encoding Limits. Patrick and I just changed Tensors of [min, max, center] to use a dataclass instead. So you would write EncodingLimit.max instead of having the magic index number 1.

Other places currently are FOV and similar. These have .x, .y, .z fields in ismrmrd but we currently convert them to Tensors.

I'd suggest to either reuse the ismrmrd dataclass or our own dataclass for these as well.

(If we really want to be able to use them like tensors, we could add a __as_tensor__ or astuple to the dataclass definition. But IMHO code is easier to read if you don't have to remember at which position in a list or tensor what information is stored but instead use names to access the data.)

This is a general decision to be made as this question will come up in the future for different kinds of data.
What do you think? Any opinions in one or the other direction?

fzimmermann89:
Working initial version of the KData/KHeader and related functionality

After discussing with @schuenke (and remembering what we discussed in one of the kick-off meetings), 
I limited the amount of information copied over from the ismrmrd header. 
The idea is, that all these fields are now guaranteed to exist. We will test these fields. Our code only relies on these fields.
If we need more fields, we shall add them later on. Everything else is dumped in a "misc" dictionary for now.

Fixes #1 

Related #28 : Fixed numpy to >1.20 and pytorch to >2.0
Related #21 : Some draft test functions are included
Related #23  : The information should end up in misc
Fixes #26
Fixes #27 
Closes #22


fzimmermann89:
There is also 589f643958b4440edb6e28aed5665156eac6d247 which might be faster but looks a bit more ugly...

schuenke (approved):
Just squash some commits and than it should be ready to merge imo 👍

PR #30, opened 2023-08-02
Initial IData and IHeader implementations

### LINKED ISSUE #2: IData class
```
class IData():
    IData.header (based on ismrmrd) -> check in functions if certain parameters are available
    IData.data: torch.Tensor 5D: [everything_else, coils, z, y, x]

from_kdata()
   from_ismrmrd()
   from_nifty()
   from_dicom()
```

Check list:
- [x]     Create branch
- [x]     Implement code
- [x]     Add comments
- [x]     Implement tests
- [x]     Squash commmits
- [x]     Create PR
- [x]     Assign reviewer(s) to PR

fzimmermann89:
Closes #2


fzimmermann89:
Not sure who will be working on this, but we might consider replacing our custom recursive function with either one of `iterall` or `walk `from pydicom:
```python

def iterall(self) -> Iterator[DataElement]:
        """Iterate through the :class:`Dataset`, yielding all the elements.

        Unlike ``iter(Dataset)``, this *does* recurse into sequences,
        and so yields all elements as if dataset were "flattened".
        ...
"""
def walk(
        self, callback: Callable[["Dataset", DataElement], None], recursive: bool = True
    ) -> None:
        """Iterate through the :class:`Dataset's<Dataset>` elements and run
        `callback` on each.

        Visit all elements in the :class:`Dataset`, possibly recursing into
        sequences and their items. The `callback` function is called for each
        :class:`~pydicom.dataelem.DataElement` (including elements
        with a VR of 'SQ'). Can be used to perform an operation on certain
        types of elements.
        ...
        """
```

ckolbPTB:
Not sure how to test `from_tensor_and_kheader`:

Shall we create a `TestHeader: KHeader` + random tensor or shall we create a `IsmrmrdRawTestData` object and pass this on?

fzimmermann89:
@schuenke :
Can you remind me what the reason for the custom lookup of the tags was?
Why not ds.RepetitionTime ?  (InversionTime , EffectiveEchoTime / EchoTime, FlipAngle)

fzimmermann89:
> Not sure how to test `from_tensor_and_kheader`:
> 
> Shall we create a `TestHeader: KHeader` + random tensor or shall we create a `IsmrmrdRawTestData` object and pass this on?

I'd slightly prefer option 1 if I understood it correctly (Reasoning: Keep the scope of the tests simple and somewhat contained.)

Maybe we can make a fixture ` random_kheader(full_header)` that just creates a kheader using the random ismrmrd header fixture? (there we could also use some renaming of full_header to random_full_ismrmrd header...). This might be useful for other tests as well.

If we ever add more logic to the` from_tensor_and_kheader`: that needs to be covered, we might reconsider.

schuenke:
> @schuenke : Can you remind me what the reason for the custom lookup of the tags was? Why not ds.RepetitionTime ? (InversionTime , EffectiveEchoTime / EchoTime, FlipAngle)

At least for my example file/script these attributes were not available, but ofc it would be much easier/nicer like this. Feel free to double check if they are available (now) or if I was just to stupid to find/access them 😄

fzimmermann89:
Sorry, I got confused by the test: 
For echo:/Volumes/users/allgemein/projects/8_13/MRPro/example_data/dicoms/siemens_XA61_FLASH_2D.dcm
The FlipAngle is in` ds.SharedFunctionalGroupsSequence[0].MRTimingAndRelatedParametersSequence[0].FlipAngle`
 I am not 100% sure if this structure is fixed or vendor dependent. 
 So, there is the reason.
 
 I just pushed a suggestions feel free to undo it..:
 - Moving the dicom->header stuff inside IHeader.
 -  Using iterall and String Names instead of tag numbers.

tests/data/conftest.py
```diff
@@ -154,3 +156,40 @@ def random_ismrmrd_file(random_acquisition, random_noise_acquisition, full_heade
...
+    """Random (not necessarily valid) KHeader."""
+    seed = request.param['seed']
+    generator = RandomGenerator(seed)
+    ktraj = generate_random_trajectory(generator)
+    kheader = KHeader.from_ismrmrd(random_full_ismrmrd_header, acq_info=random_acq_info, defaults={'trajectory': ktraj})
+    return kheader
+
+
+@pytest.fixture(params=({'seed': 0, 'Nd4': 2, 'Ncoils': 16, 'Nz': 32, 'Ny': 128, 'Nx': 256},))
+def random_kheader_and_data(request, random_kheader):
```
fzimmermann89: I'd maybe split this.
The random image data and the header a currently completely unrelated, to they can just be two separate fixtures, or?

schuenke (approved):
I adjusted all imports within mrpro to fulfill the same style. At some point we might want to discuss if it makes sense to keep all imports in the __init__ or if we want to limit ourselves to "public" classes / functions here, but at least everything uses the same style now. 

**EDIT: I will fix the tests soon**

PR #31, opened 2023-08-02
fix names in from_ismrmrd_encodingLimitsType()

schuenke:
Adjust naming in EncodingLimits from ISMRMRD to MRPro naming convention


fzimmermann89 (approved):
LGTM

PR #36, opened 2023-08-10
Make KData a frozen dataclass

### LINKED ISSUE #35: KData should be a dataclass
@SherineBrahma just noticed that KData is not a dataclass.

I think this is an oversight by @schuenke  and me, we should fix this.

https://github.com/ckolbPTB/mrpro/blob/a1002e1e4ae4f89009fc57bb55689ca0d5a483cc/src/mrpro/data/_KData.py#L45

fzimmermann89:
KData can be a dataclass.

Seeting frozen makes the attributes read-only, i.e. data and header can only be set at init. This matches the previous behavior.

We might want to keep this as a regular class depening on how much we want to to in the getters or setters. Or make it a dataclass for now and maybe switch to something more flexible later on..

Fixes #35

### DIFF
--- a/src/mrpro/data/_KData.py
+++ b/src/mrpro/data/_KData.py
@@ -42,17 +42,8 @@ KDIM_SORT_LABELS = (
 )
 
 
+@dataclasses.dataclass(slots=True, frozen=True)
 class KData:
-    def __init__(
-        self,
-        header: KHeader,
-        data: torch.Tensor,
-        traj: torch.Tensor,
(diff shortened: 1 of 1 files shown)


schuenke (approved):
I agree that (a frozen) dataclass makes sense for KData. I will do the same changes for IData (PR #30).

SherineBrahma (approved):
My first time reviewing on Github haha. Looks cool!

PR #37, opened 2023-08-10
2D radial trajectory

SherineBrahma:
Hi,

I have written a preliminary code for a 2D radial trajectory for golden radial trajectory which I would not like to be merged right away.

If possible, I would like to have feedback on:

1. Whether it makes sense to use dataclass decorator to provide specific information.
2. In the [documentation](https://ismrmrd.readthedocs.io/en/latest/mrd_raw_data.html#encodingcounters), I found that trajectory information is optional and it is possible the trajectory coordinates may already be present. In that case, the same can be obtained by looping over the "**header.acq_info.trajectory_dimensions**" variable.
3. Feel free to roast my coding style and content.

### DIFF
--- a/src/mrpro/data/_KTrajectory.py
+++ b/src/mrpro/data/_KTrajectory.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 import dataclasses
 from abc import ABC
 from abc import abstractmethod
+from dataclasses import dataclass
 
 import numpy as np
 import torch
@@ -67,3 +68,51 @@ class DummyTrajectory(KTrajectory):
         3, k2, k1, k0)
         """
         return torch.zeros(self._get_shape(header))
+
+
+@dataclass
+class RadialKTraj2D(KTrajectory):
+    """Returns 2D radial trajectory."""
+
+    golden_angle: bool = True
+    multi_slice: bool = False
+
+    def calc_traj(self, header: KHeader) -> torch.Tensor:
+        """Calculate the 2D radial trajectory for the given header."""
+
+        # # I found that the following in the header file can have some
+        # # potential use but the same information can be derived from
+        # # self._get_shape(header)
+        # spoke_counter = header.acq_info.idx.k1
+        # sample_counter = header.acq_info.number_of_samples
+
+        n_spokes, n_readout = self._get_shape(header)[-2], self._get_shape(header)[-1]
+
+        # length of each spoke;
+        spoke_length = n_readout
+
+        # golden angle radial increment
+        if self.golden_angle is True:
+            del_theta = np.deg2rad(180 / ((1 + np.sqrt(5)) / 2))
+
+        # create radial coordinates
+        ktheta = torch.arange(0, del_theta*n_spokes, del_theta)
+        kradius = torch.linspace(-np.pi, np.pi, spoke_length)  # format of trajectories kbNUFFT package expects to be
+
+        # construct transformation matrices
+        ktraj = torch.zeros(self._get_shape(header))
+        ktraj_init = torch.stack((kradius, torch.zeros(n_readout), torch.zeros(n_readout)))
+        for i in range(n_spokes):
+            theta = ktheta[i]
+            rot_mat = torch.tensor([[torch.cos(theta), -torch.sin(theta), 0],
+                                    [torch.sin(theta), torch.cos(theta), 0],
+                                    [0, 0, 1]])
+            ktraj[..., i, :] = torch.matmul(rot_mat, ktraj_init).unsqueeze(1).unsqueeze(0)
+
+        # skip if all optional trajectory information are missing else do something
+        if torch.all(header.acq_info.trajectory_dimensions == 0):
+            pass
+        else:
+            pass  # do something: replace the calculated trajectory information with the already present one
+
+        return ktraj
--- a/tests/data/test_ktraj.py
+++ b/tests/data/test_ktraj.py
@@ -1,0 +1,13 @@
+# from pathlib import Path
+
+from mrpro.data import KData
+from mrpro.data._KTrajectory import RadialKTraj2D
+
+
+def test_KTrajectory_calc(random_ismrmrd_file):
+    k = KData.from_file(random_ismrmrd_file, RadialKTraj2D(golden_angle=True))
+    assert k is not None
+
+# fpath = Path('/data/brahma01/Datasets/MRpro_data/meas_MID00020_FID02083_FLASH.mrd')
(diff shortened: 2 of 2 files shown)


src/mrpro/data/_KTrajectory.py
```diff
@@ -67,3 +68,51 @@ def calc_traj(self, header: KHeader) -> torch.Tensor:
...
+        # sample_counter = header.acq_info.number_of_samples
+
+        n_spokes, n_readout = self._get_shape(header)[-2], self._get_shape(header)[-1]
+
+        # length of each spoke;
+        spoke_length = n_readout
+
+        # golden angle radial increment
+        if self.golden_angle is True:
+            del_theta = np.deg2rad(180 / ((1 + np.sqrt(5)) / 2))
```
fzimmermann89, line 96: What if golden angle is False?

Maybe instead have the increment as a __init__ parameter and default to golden angle?

src/mrpro/data/_KTrajectory.py
```diff
@@ -67,3 +68,51 @@ def calc_traj(self, header: KHeader) -> torch.Tensor:
...
+    def calc_traj(self, header: KHeader) -> torch.Tensor:
+        """Calculate the 2D radial trajectory for the given header."""
+
+        # # I found that the following in the header file can have some
+        # # potential use but the same information can be derived from
+        # # self._get_shape(header)
+        # spoke_counter = header.acq_info.idx.k1
+        # sample_counter = header.acq_info.number_of_samples
+
+        n_spokes, n_readout = self._get_shape(header)[-2], self._get_shape(header)[-1]
```
fzimmermann89, line 89: We should check if the _get_shape logic works for different files.
We should especially make sure that pulseq sequences and sequences with over/undersampling/partial readout would work. 
When I implemented the _get_shape logic, I did not consider these and just selected some header information that looked reasonable.
This is something @schuenke  was volunteered to think about ;)

src/mrpro/data/_KTrajectory.py
```diff
@@ -67,3 +68,51 @@ def calc_traj(self, header: KHeader) -> torch.Tensor:
...
+            del_theta = np.deg2rad(180 / ((1 + np.sqrt(5)) / 2))
+
+        # create radial coordinates
+        ktheta = torch.arange(0, del_theta*n_spokes, del_theta)
+        kradius = torch.linspace(-np.pi, np.pi, spoke_length)  # format of trajectories kbNUFFT package expects to be
+
+        # construct transformation matrices
+        ktraj = torch.zeros(self._get_shape(header))
+        ktraj_init = torch.stack((kradius, torch.zeros(n_readout), torch.zeros(n_readout)))
+        for i in range(n_spokes):
```
fzimmermann89, line 105: can we vectorize this section?

src/mrpro/data/_KTrajectory.py
```diff
@@ -67,3 +68,51 @@ def calc_traj(self, header: KHeader) -> torch.Tensor:
...
+            theta = ktheta[i]
+            rot_mat = torch.tensor([[torch.cos(theta), -torch.sin(theta), 0],
+                                    [torch.sin(theta), torch.cos(theta), 0],
+                                    [0, 0, 1]])
+            ktraj[..., i, :] = torch.matmul(rot_mat, ktraj_init).unsqueeze(1).unsqueeze(0)
+
+        # skip if all optional trajectory information are missing else do something
+        if torch.all(header.acq_info.trajectory_dimensions == 0):
```
fzimmermann89, line 113: #38

src/mrpro/data/_KTrajectory.py
SherineBrahma: Yes, I can do that.

src/mrpro/data/_KTrajectory.py
SherineBrahma: Do you mean like getting rid of the loop?

I thought about it but decided against it as I felt that using the loop was more readable. I am not sure how much we need to optimize the code to be fast at this stage.

src/mrpro/data/_KTrajectory.py
```diff
@@ -67,3 +68,51 @@ def calc_traj(self, header: KHeader) -> torch.Tensor:
         3, k2, k1, k0)
         """
         return torch.zeros(self._get_shape(header))
+
+
+@dataclass
+class RadialKTraj2D(KTrajectory):
```
ckolbPTB, line 74: Thanks for making a start on this. 

I would suggest to call this function KTrajectoryRadial2D

src/mrpro/data/_KTrajectory.py
```diff
@@ -67,3 +68,51 @@ def calc_traj(self, header: KHeader) -> torch.Tensor:
...
+    """Returns 2D radial trajectory."""
+
+    golden_angle: bool = True
```
ckolbPTB, line 77: I would allow the user to set the `angle_increment` rather than using a Boolean. The `angle_increment` has to be exactly like it was calculated in the sequence code. Even tiny rounding errors can lead to large errors for large numbers of angles. 

This would also mean this function works for non-Golden-Angle trajectories. Otherwise we would have to rename it to KTrajectoryGoldenRadial2D.

src/mrpro/data/_KTrajectory.py
```diff
@@ -67,3 +68,51 @@ def calc_traj(self, header: KHeader) -> torch.Tensor:
...
+    multi_slice: bool = False
```
ckolbPTB, line 78: I don't think this is needed, because the trajectory is the same for all slices.

src/mrpro/data/_KTrajectory.py
ckolbPTB: I think it would be better to use the information from `AcqInfo` and `AcqIdx`. See also comment about ktheta and kradius below.

src/mrpro/data/_KTrajectory.py
ckolbPTB: `kradius` should be calculated using `AcqInfo.number_of_samples` and `AcqInfo.center_sample`. In your version you assume a full spoke is acquired but this does not have to be the case. It can also be partial Fourier or half-spoke-acquisitions. 

Similarly for `ktheta` (which I would call `kangle` to fit with `angle_increment`). Here we should use AcqIdx.k1. 

I also agree with Feliz, this should be vectorised. No need for a rotation matrix. kx = krad * cos(kangle) and ky = krad * sin(kangle).

fzimmermann89:
Sorry for closing this PR, not sure how this happend!

PR #51, opened 2023-08-26
Ismrmrd test data

### LINKED ISSUE #21: Create test data

ckolbPTB:
Might need some more test cases but hopefully a start.


ckolbPTB:
Currently the tests are not discovered by github actions. Should we change the github actions also to pytest?

tests/data/IsmrmrdRawData.py
```diff
@@ -0,0 +1,362 @@
...
+        scaling factor for noise level, by default 0.00005
+    trajectory_type
+        cartesian, by default cartesian
+    sampling_order
+        order how phase encoding points (ky) are obtained, by default linear
+    phantom
+        phantom with different ellipses
+    """
+
+    filename: str | os.PathLike
```
schuenke: in the other classes we always use pathlib.Path instead of os.PathLike.
I would change this to :  `filename: str | Path,`

tests/data/IsmrmrdRawData.py
```diff
@@ -0,0 +1,362 @@
...
+    centre_y: float
+    radius_x: float
+    radius_y: float
+    intensity: float
+
+
+class EllipsePhantom():
+    def __init__(self):
+        # Create three circles with different intensity
+        self.ellipses: list[ellipse_pars] = [ellipse_pars(centre_x=0.2, centre_y=0.2,
```
fzimmermann89: Personally, I do not like this to much. Is the EllipsePhantom anything else than a `List[ellipse_pars]`? We don't have any methods here that use the state. If we really want to be specific, we could Alias the List to something called EllipsePhantom. But I'd prefer to keep it as a list.

tests/data/IsmrmrdRawData.py
```diff
@@ -0,0 +1,362 @@
...
+from dataclasses import dataclass
+
+import ismrmrd
+import ismrmrd.xsd
+import numpy as np
+import scipy.special as sp_special
+
+
+@dataclass(slots=True)
+class ellipse_pars():
```
fzimmermann89: Class Names usually are capitalized

tests/data/IsmrmrdRawData.py
```diff
@@ -0,0 +1,362 @@
...
+        # Create three circles with different intensity
+        self.ellipses: list[ellipse_pars] = [ellipse_pars(centre_x=0.2, centre_y=0.2,
+                                                          radius_x=0.2, radius_y=0.2, intensity=1),
+                                             ellipse_pars(centre_x=0.1, centre_y=-0.1,
+                                                          radius_x=0.2, radius_y=0.2, intensity=2),
+                                             ellipse_pars(centre_x=-0.2, centre_y=0.1, radius_x=0.3,
+                                                          radius_y=0.3, intensity=4)]
+
+
+def k2i(kdat, axes=(0, 1)):
```
fzimmermann89: Make axes defaulting two last two?

tests/data/IsmrmrdRawData.py
```diff
@@ -0,0 +1,362 @@
...
+    kdat
+        k-space data on Cartesian grid
+    axes, optional
+        axes along which FFT is applied, by default (0, 1)
+
+    Returns
+    -------
+        FFT of kdat
+    """
+    return (np.fft.fftshift(np.fft.fftn(np.fft.fftshift(kdat, axes=axes), axes=axes), axes=axes))
```
fzimmermann89: Isn't it usually 
```
k = fftshift(fft(ifftshift(img)));
img=fftshift(ifft(ifftshift(k)));
```
to move between centered and shifted ffts?

This only makes a difference for odd-numbered samples, which is uncommon for MRI, but still..

fzimmermann89:
Just making sure I got it correctly:

I thought we wanted to create analytic image space data such a a shepp-logan phantom, create k-space data for it and then put that into an ISMRMRD file. 

But if I am not mistaken, this code creates ellipses in **k space** instead? And then considered the iFFT of that as the ground truth?

Edit: I am mistaken (?)
```
 arg = np.sqrt(el.radius_x ** 2 * kx ** 2 + el.radius_y ** 2 * ky ** 2)
arg[arg < 1e-6] = 1e-6  # avoid zeros
cdat = el.radius_x * el.radius_y * 0.5 * sp_special.jv(1, np.pi * arg) / arg
```

But I'd need some more comments what is going on there...

And why don't we use this to create the image space data as well?

tests/data/IsmrmrdRawData.py
```diff
@@ -0,0 +1,362 @@
...
+    # kx and ky have to be of same shape
+    if kx.shape != ky.shape:
+        raise ValueError(f'shape mismatch between kx {kx.shape} and ky {ky.shape}')
+
+    kdat = 0
+    for el in phantom.ellipses:
+        arg = np.sqrt(el.radius_x ** 2 * kx ** 2 + el.radius_y ** 2 * ky ** 2)
+        arg[arg < 1e-6] = 1e-6  # avoid zeros
+
+        cdat = el.radius_x * el.radius_y * 0.5 * sp_special.jv(1, np.pi * arg) / arg
```
fzimmermann89: Sorry, I need an explanation to understand what is going on here...

ckolbPTB:
> But if I am not mistaken, this code creates ellipses in **k space** instead? And then considered the iFFT of that as the ground truth?

no, it creates ellipses in image space by using their analytic description in k-space (Bessel functions). Therefore, you can create ground truth k-space data for any trajectory. If we started in image space then we would have to use nufft for non-Cartesian and this might then create a bias towards a certain nufft package. With this approach we have an analytic description in image and k-space.

Currently I am using FFT to get the reference image, this could be improved by calculating the ellipses analytically in image space but I was too lazy to think about the correct dimensions and orientations. I leave that for somebody else....

tests/data/IsmrmrdRawData.py
```diff
@@ -0,0 +1,362 @@
...
+    kx_idx = range(-nkx//2, nkx//2)
+    ky_idx = range(-nky//2, nky//2)
+    [kx, ky] = np.meshgrid(kx_idx, ky_idx)
+    kfull = analytic_2d_kspace(ky, kx, phantom)
+    imref = k2i(kfull)
+
+    return (ktrue, imref)
+
+
+def calc_phase_encoding_steps(nky: int, acceleration: int = 1, sampling_order: str = 'linear'):
```
fzimmermann89: sampling_order: Literal["linear", "low_high", "high_low"]

tests/data/IsmrmrdRawData.py
```diff
@@ -0,0 +1,362 @@
...
+    ky = np.concatenate((ky_neg, ky_pos), axis=0)
+
+    if sampling_order == 'linear':
+        ky = np.sort(ky)
+    elif sampling_order == 'low_high' or sampling_order == 'high_low':
+        idx = np.argsort(np.abs(ky), kind='stable')
+        ky = ky[idx]
+    else:
+        raise ValueError(f'sampling order {sampling_order} not supported.')
+    if sampling_order == 'high_low':
```
fzimmermann89: I'd prefer to just 
 ```
elif sampling_order == 'low_high':
        idx = np.argsort(np.abs(ky), kind='stable')
        ky = ky[idx]
elif sampling_order == 'high_low'
        idx = np.argsort(-np.abs(ky), kind='stable')
        ky = ky[idx]
else:        
```
instead of the "afterthought" of switching the order after the ValueError. Currently, those two steps which are logically one are separated by the exception, which makes it less clear what is going on.. Also, it would be easier to see at a glance what the different options are

File: tests/data/IsmrmrdRawData.py [hunk 9]
```diff
@@ -0,0 +1,362 @@
...
+        ky = ky[::-1]
+    return (ky)
+
+
+@dataclass(slots=True)
+class IsmrmrdRawData():
```
fzimmermann89: You always do `IsmrmrdRawData(...) ` followed by an immediate `.create`

schuenke:
> Currently the tests are not discovered by github actions. Should we change the github actions also to pytest?

Switching to pytest is easy and works for me, but I will try to tackle the code coverage calculation/report as well and create a separate branch/PR for this.

tests/data/IsmrmrdRawData.py
```diff
@@ -0,0 +1,362 @@
...
+
+    filename: str | os.PathLike
+    matrix_size: int = 256
+    ncoils: int = 8
+    oversampling: int = 2
+    repetitions: int = 1
+    flag_invalid_reps: bool = False
+    acceleration: int = 1
+    noise_level: float = 0.00005
+    trajectory_type:  str = 'cartesian'
```
fzimmermann89: later on fixed to cartesian

tests/data/IsmrmrdRawData.py
```diff
@@ -0,0 +1,298 @@
...
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+from pathlib import Path
+
+import ismrmrd
+import ismrmrd.xsd
+import numpy as np
+from Phantoms import EllipsePhantom
```
fzimmermann89: - This is missing?
- why is phantoms capitalized?
- Is this supposed to be a completely different library? 
or mrpro.phantoms?

tests/data/test_kdata.py
```diff
@@ -1,7 +1,83 @@
...
+    k = KData.from_file(ismrmrd_cart.filename, DummyTrajectory())
     assert k is not None
+
+
+def test_KData_kspace(ismrmrd_cart):
+    # Read in data and verify k-space by comparing reconstructed image
+    k = KData.from_file(ismrmrd_cart.filename, DummyTrajectory())
+    irec = k2i(k.data, axes=(3, 4))
+    idiff = np.mean(np.abs(irec[0, 0, 0, ...] - ismrmrd_cart.imref))
+    imean = 0.5*np.mean(np.abs(irec[0, 0, 0, ...]) + np.abs(ismrmrd_cart.imref))
```
fzimmermann89: Please add a comment why we do this.
Also, maybe we can consider a better criterion. First step could be to ignore the border?

tests/data/test_kdata.py
```diff
@@ -1,7 +1,83 @@
...
+    assert idiff <= imean*0.05
+
+
+def test_KData_from_file_undersampled(ismrmrd_cart_us4):
+    # Expected to fail with DummyTrajectory
```
fzimmermann89: Not sure if this is a useful test.
It fails with the current Dummy Trajectory. But is this a specification that is MUST always fail with the DummyTracectory? I.e. would it be wrong to fix dummy trajectory?

tests/data/test_kdata.py
```diff
@@ -1,7 +1,83 @@
...
+def test_KData_from_file(ismrmrd_cart):
+    # Read in data from file
+    k = KData.from_file(ismrmrd_cart.filename, DummyTrajectory())
     assert k is not None
+
+
+def test_KData_kspace(ismrmrd_cart):
+    # Read in data and verify k-space by comparing reconstructed image
+    k = KData.from_file(ismrmrd_cart.filename, DummyTrajectory())
+    irec = k2i(k.data, axes=(3, 4))
```
fzimmermann89: Consider axes=(-1,-2) ?

tests/data/IsmrmrdRawData.py
```diff
@@ -0,0 +1,298 @@
...
+        limits1.maximum = ny - 1
+        limits.kspace_encoding_step_1 = limits1
+
+        limits_rep = ismrmrd.xsd.limitType()
+        limits_rep.minimum = 0
+        limits_rep.center = round(self.repetitions / 2)
+        limits_rep.maximum = self.repetitions - 1
+        limits.repetition = limits_rep
+
+        limits_rest = ismrmrd.xsd.limitType()
```
fzimmermann89: These do not have to be set. 
We should test what happens if they are missing.

fzimmermann89 (changes requested):

tests/data/IsmrmrdRawData.py
fzimmermann89: Remove superfluous parenthesis?

tests/data/test_kdata.py
```diff
@@ -1,7 +1,83 @@
...
+    return (ismrmrd_kdat)
+
+
+@pytest.fixture(scope='session')
+def ismrmrd_cart_invalid_reps(tmp_path_factory):
+    # Fully sampled cartesian data set
+    ismrmrd_filename = tmp_path_factory.mktemp('mrpro') / 'ismrmrd_cart.h5'
+    ismrmrd_kdat = IsmrmrdRawData(filename=ismrmrd_filename, noise_level=0.0, repetitions=3, flag_invalid_reps=True)
+    ismrmrd_kdat.create()
+    return (ismrmrd_kdat)
```
fzimmermann89: Remove superfluous parenthesis?

tests/data/IsmrmrdRawData.py
ckolbPTB: sorry, forgot to add a file to the last commit

tests/data/test_kdata.py
ckolbPTB: comment added, but implementing edge detection filters seems a bit over the top here

tests/data/test_kdata.py
ckolbPTB: I added a comment that this should be adapted as soon as we have a valid TrajectoryCalculator for Cartesian trajectories - ok?

tests/data/IsmrmrdRawData.py
ckolbPTB: Also works without them -> removed

tests/data/IsmrmrdRawData.py
ckolbPTB: added create to init

tests/data/IsmrmrdRawData.py
fzimmermann89: We should reconsider our naming conventions. 
I'd suggest 
- Capitalized class names and file names that only contain one class with the same name but underscored
- lowercase file names for files containing multiple classes.

I.e. phantoms.py or _EllipsePhanton.py
or, even better mrpro/phantoms/_EllipsePhanton.py ?

tests/data/IsmrmrdRawData.py
```diff
@@ -0,0 +1,306 @@
...
+
+        # Multi-coil acquisition
+        # TODO: proper application of coils
+        ktrue = np.tile(ktrue[None, ...], (self.ncoils, 1, 1))
+
+        # Open the dataset
+        dset = ismrmrd.Dataset(self.filename, 'dataset', create_if_needed=True)
+
+        # Create the XML header and write it to the file
+        header = ismrmrd.xsd.ismrmrdHeader()
```
fzimmermann89: If we feel this init is now to long, we might consider at some point to refactor it into  something along the lines of (pseudocode..)
```
def __init__(...)
  header = self.createHeader(...)
  noiseacqs = self.createNoiseAcq()
  acqs =  self.createAcquisitions(...)
....
  ds.writeHeader(header)
  for a in noiseacqs:
   ds.append_acquisition(a)
   for a in acqs:
    ds.append_acquisition(a)
 
```

tests/data/IsmrmrdRawData.py
```diff
@@ -0,0 +1,306 @@
...
+        # Set encoded and recon spaces
+        encoding.encodedSpace = espace
+        encoding.reconSpace = rspace
+
+        # Encoding limits
+        limits = ismrmrd.xsd.encodingLimitsType()
+
+        limits1 = ismrmrd.xsd.limitType()
+        limits1.minimum = 0
+        limits1.center = round(ny / 2)
```
fzimmermann89: Q: Is this correct for odd samples? Is it round(ny/2) or ny//2 ?

tests/data/Phantoms.py
```diff
@@ -0,0 +1,95 @@
...
+
+from dataclasses import dataclass
+
+import numpy as np
+import scipy.special as sp_special
+
+
+@dataclass(slots=True)
+class EllipsePars():
+    centre_x: float
```
fzimmermann89: numpy uses AE, scipy uses AE, center has at least 5x more matches on github.... :D

tests/data/Phantoms.py
```diff
@@ -0,0 +1,95 @@
...
+        for el in self.ellipses:
+            arg = np.sqrt((el.radius_x * 2) ** 2 * kx ** 2 + (el.radius_y * 2) ** 2 * ky ** 2)
+            arg[arg < 1e-6] = 1e-6  # avoid zeros
+
+            cdat = 2 * 2 * el.radius_x * el.radius_y * 0.5 * sp_special.jv(1, np.pi * arg) / arg
+            kdat += (np.exp(-1j * 2 * np.pi * (el.centre_x * kx + el.centre_y * ky))
+                     * cdat * el.intensity).astype(np.complex64)
+
+        # Scale k-space data by factor 1/sqrt(number of points) to ensure correct scaling after FFT with
+        # normalisation "ortho". See e.g.
```
fzimmermann89: See eg. ?

tests/data/test_kdata.py
fzimmermann89: nah, eiher we figure out how it is smoothed (I will take a look)
Or I'd maybe add a .image_valid or (or similar named) function to the phantom that  returns a  mask, masking out 1 pixel in each direction around the ellipse border?

tests/data/test_kdata.py
fzimmermann89: We could do test driven development and let the test fail for now, either  with an empty CartesianTrajectory or the DummyTrajectory.

My issue here is, that we do not want to 'enforce' the currently wrong behavior. I.e. to assert an exception there seems to me to be just the opposite of what we want.

A failing test is not an issue, we should write test that fail :D

fzimmermann89 (approved):
some small comments, but looks great already.

tests/data/IsmrmrdRawData.py
ckolbPTB: naming conventions -> yes (in my local branch it is correct ;) - on MacOS (and probably Windows) git ignores file name case changes by default) 

As numerical phantoms are very useful outside of testing, I think it would be a good idea to move it to mrpro/src/phantoms/_EllipsePhantom.py

tests/data/test_kdata.py
ckolbPTB: I would suggest to create an issue on that and leave it for later

tests/data/test_kdata.py
ckolbPTB: For a PR the assumption is that all tests pass, so we should remove this test altogether

tests/data/IsmrmrdRawData.py
schote, line 1: I personally think the filename is a bit confusing.
Written out this would be "Ismrm raw data RawData".
Or does the second RawData have a different meaning?

tests/data/IsmrmrdRawData.py
```diff
@@ -0,0 +1,306 @@
...
+        ky = ky[idx]
+    elif sampling_order == 'high_low':
+        idx = np.argsort(-np.abs(ky), kind='stable')
+        ky = ky[idx]
+    else:
+        raise ValueError(f'sampling order {sampling_order} not supported.')
+    return ky
+
+
+class IsmrmrdRawData:
```
schote: I would call it `IsmrmRawData`, see comment above

tests/data/IsmrmrdRawData.py
```diff
@@ -0,0 +1,306 @@
...
+    """ISMRMR raw data object.
```
schote: ISMRM raw data

tests/data/IsmrmrdRawData.py
```diff
@@ -0,0 +1,306 @@
...
+    'cartesian',
+    'epi',
+    'radial',
+    'goldenangle',
+    'spiral',
+    'other',
+)
+
+
+def k2i(kdat, axes=(-1, -2)):
```
schote: Is this function name clear? In the `Phantoms` function names like `kspace()` or `image_space()` are used. I would go with `kspace_to_image` here, even if it is a bit longer

tests/data/Phantoms.py
```diff
@@ -0,0 +1,95 @@
...
+                                                        radius_x=0.18, radius_y=0.25, intensity=4)]
+
+    def kspace(self, ky: np.ndarray, kx: np.ndarray):
+        """Create 2D analytic kspace data based on given k-space locations.
+
+        For a corresponding image with 256 x 256 voxel, the k-space locations should be defined within [-128, 127]
+
+        The Fourier representation of ellipses can be analytically described by Bessel functions. Further information
+        and derivations can be found e.g. here: https://doi.org/10.1002/mrm.21292
+
```
schote: Note:
For the documentation we could also add examples here which appear as formatted code in the documentation.
Would be something like the following which you add to the docstring.
```
Example
-------
>>> ro = np.linspace(-63, 64, 128)
>>> pe = np.linspace(-63, 64, 128)
>>> grid_ro, grid_pe = np.meshgrid(ro, pe)
>>> kspace(grid_ro, grid_pe)
```
_(Not sure if the above example is correct)_

schote (commented):
Remarks regarding some naming conventions, suggestion for code documetation

tests/data/IsmrmrdRawData.py
ckolbPTB: To me Ismrmrd is the data format, so _IsmrmrdRawData_ is like saying _DicomImageData_. Does this make sense or would you prefer _IsmrmRawData_ (I am happy with both).

schote: I think for _Ismrmrd_ it's clear, that it contains raw data, isn't it? A _Dicom_ file can in principle also contain other stuff, so here it would make sense to further specify it. Any other optinions @schuenke @fzimmermann89 ?

fzimmermann89: Also, not 100% convinced by the name: This is not general ISMRMRD data. It it not a container for general raw data.
It is synthetic a ISMRMRD factory for tests.

ISMRMRDTestData ?
ISMRMRDPhantom?
SyntheticISMRMRD?

def generateISMRMRD() :D

tests/data/test_kdata.py
fzimmermann89: I agree with removing it.
I disagree with the assumption that a PR can only add passing tests.
A PR should not break existing tests, but I would encourage ppl to add new tests in PR for missing functionality, not yet handled edge cases etc...

tests/data/IsmrmrdRawData.py
fzimmermann89: today i learned..

tests/data/IsmrmrdRawData.py
schote: I would prefer `ISMRMRDTestData` but I don't have a string opinion with it

ckolbPTB: > I think for _Ismrmrd_ it's clear, that it contains raw data, isn't it?

No. ISMRMRD can also contain image data: https://ismrmrd.readthedocs.io/en/latest/mrd_image_data.html

ckolbPTB: > This is not general ISMRMRD data.

Why?

fzimmermann89: It is not someting like KData or IData, that represents arbitrary ISMRMRD-RawData (the latter would be one of the objects in the ismrmrd-python library). 
But instead only simulated data with a particular structure useful for testing.

schote: > > I think for _Ismrmrd_ it's clear, that it contains raw data, isn't it?
> 
> No. ISMRMRD can also contain image data: https://ismrmrd.readthedocs.io/en/latest/mrd_image_data.html

Ok, but this sounds more like _ISMRMRD_ is deprecated and renamed to _MRD_ which stands for _Magnetic Resonance Data_ (https://ismrmrd.readthedocs.io/en/latest/overview.html). Also on their index page they introduce it as _ISMRM Data Format (MRD)_ (https://ismrmrd.readthedocs.io/en/latest/index.html). 

I think `MRDRawData` and `MRDImageData` is totally fine.

tests/data/test_kdata.py
```diff
@@ -1,7 +1,73 @@
...
+
+
+@pytest.mark.parametrize(
+    'field,value',
+    [
+        ('b0', 11.3),
+        (
+            'tr',
+            [
+                24.3,
```
fzimmermann89: Remove the comma here after the values to avoid this strange black formatting.
Background: A trailing (unnecessary) comma is used to signify that you want a line break within the parameter list

ckolbPTB:
I restructured the code quite a lot and renamed things. @fzimmermann89 @schuenke @schote can you please have a look if this makes sense?

tests/utils.py
```diff
@@ -0,0 +1,54 @@
...
+
+    Returns
+    -------
+        FFT of kdat
+    """
+    return sp.fft.fftshift(sp.fft.ifftn(sp.fft.ifftshift(kdat, axes=axes), axes=axes, norm='ortho'), axes=axes)
+
+
+def rel_image_diff(im1, im2):
+    """Calculate relative difference between two images.
```
fzimmermann89: Mean absolute relative ... ;)

fzimmermann89:
Looks sensible. Good idea to move the data generation stuff into a new module. 

- There is a auto formatting issue in one of the fixture decorators (see comment)
- we should decide if we stick with the common ismrmrd or with MRD..
- Still prefered a class name that makes it clear that this is not a general Ismrm raw data class, but numerical Phantom data..

Otherwise, I'd suggest to squash (or rebase it into a  few logical commits) and get it into main :)

ckolbPTB:
> There is a auto formatting issue in one of the fixture decorators (see comment)

Sorry, where is the comment?

> we should decide if we stick with the common ismrmrd or with MRD..
Still prefered a class name that makes it clear that this is not a general Ismrm raw data class, but numerical Phantom data..

How about `IsmrmrdRawTestData` ? (as we have ismrmrd everywhere else, maybe a good idea to stick with it)

tests/data/test_kdata.py
fzimmermann89:
```
@pytest.mark.parametrize(
    "field,value",
    [
        ("b0", 11.3),
        ("tr", [24.3]),
    ],
)
```

fzimmermann89:
> > There is a auto formatting issue in one of the fixture decorators (see comment)
> 
> Sorry, where is the comment?

Sorry, my bad. I seem to have forgotten to click on save... in test_kdata, around line 57.
maybe:

```python
@pytest.mark.parametrize(
    "field,value",
    [
        ("b0", 11.3),
        ("tr", [24.3]),
    ],
)
```

> > we should decide if we stick with the common ismrmrd or with MRD..
> > Still prefered a class name that makes it clear that this is not a general Ismrm raw data class, but numerical Phantom data..
> 
> How about `IsmrmrdRawTestData` ? (as we have ismrmrd everywhere else, maybe a good idea to stick with it)

👍

tests/phantoms/test_phantoms.py
```diff
@@ -0,0 +1,51 @@
...
+
+import numpy as np
+
+from mrpro.phantoms import EllipsePars
+from mrpro.phantoms import EllipsePhantom
+from tests.utils import kspace_to_image
+from tests.utils import rel_image_diff
+
+
+def test_EllipsePhantom():
```
schuenke: Should we maybe make this a TestClass and than test for all the class methods. For the (simple) EllipsePhantom it's maybe not a must, but it could be used as a blueprint for more complex classes later on.

In this case it could look like this:
```python
class TestEllipsePhantom:
    nx = 512
    ny = 256
    [kx, ky] = np.meshgrid(range(-nx // 2, nx // 2), range(-ny // 2, ny // 2))

    # Define five ellipses
    test_ellipses = [
        EllipsePars(center_x=0.1, center_y=0.0, radius_x=0.1, radius_y=0.25, intensity=1),
        EllipsePars(center_x=0.3, center_y=0.3, radius_x=0.1, radius_y=0.1, intensity=2),
        EllipsePars(center_x=0.1, center_y=0.1, radius_x=0.1, radius_y=0.1, intensity=3),
        EllipsePars(center_x=-0.2, center_y=-0.2, radius_x=0.1, radius_y=0.1, intensity=4),
        EllipsePars(center_x=-0.3, center_y=-0.3, radius_x=0.1, radius_y=0.1, intensity=5),
    ]

    ph = EllipsePhantom(test_ellipses)

    def test_image_space(self):
        """Check if image space has correct shape."""
        im = self.ph.image_space(self.nx, self.ny)
        assert im.shape == (self.ny, self.nx)

    def test_kspace_correct_shape(self):
        """Check if kspace has correct shape."""
        kdat = self.ph.kspace(self.kx, self.ky)
        assert kdat.shape == (self.ny, self.nx)

    def test_kspace_raises_error(self):
        """Check if kspace raises error if kx and ky have different shapes."""
        [kx_, _] = np.meshgrid(range(-self.nx // 2, self.nx // 2), range(-self.ny // 2, self.ny // 2 + 1))
        with pytest.raises(ValueError) as excinfo:
            self.ph.kspace(kx_, self.ky)

    def test_kspace_to_image(self):
        """Check if kspace_to_image works."""
        kdat = self.ph.kspace(self.kx, self.ky)
        irec = kspace_to_image(kdat)
        # Due to discretisation artifacts the reconstructed image will be different to the reference image. Using standard
        # testing functions such as numpy.testing.assert_almost_equal fails because there are few voxels with high
        # differences along the edges of the elliptic objects.
        assert rel_image_diff(irec, self.ph.image_space(self.nx, self.ny)) <= 0.05
```

fzimmermann89: Not a fan tbh. I'd keep it as is.
if we want to group the test later into smaller logical packages, I'd just put them in different files and continue to use fixtures for shared resources such as the phantom.

ckolbPTB: Sounds good too me, except for `test_kspace_to_image`. This should not test the function `kspace_to_image()` but should verify that the kspace and the image of the phantom correspond assuming that `kspace_to_image()` is correct. 

Other opinions?

EDIT: sorry forgot to update before replying. I will add the suggested additional tests and then we can discuss on Tuesday if we want to use classes for tests or only fixtures.

schote (approved):
Thanks for the effort @ckolbPTB !

tests/phantoms/test_phantoms.py
schote: `test_phantom_verification` ?

or

Can't we do the test by just implementing a test for the functions `test_kspace_to_image` and `test_image_to_kspace`?
In these we could verify if the analytical phantom (kspace and image) korresponds to the transformations.

schuenke: Regarding the use of fixtures: 
as far as I understood it we would use fixtures to test different functions that use the same data. When testing the different methods of a single class and when we don't expect that the class/instance will be used by other tests, test classes are an elegant way to structure these tests imo. But my opinion is not super strong here.

Regarding `test_kspace_to_image`:
I agree that this is not a pure test of `kspace_to_image()`, but more like a minimal intergration test. We could rename it to something like `test_kspace_image_match` ?!
**UPDATE:** I just noticed that `kspace_to_image()` is a function in tests.utils now. Thus, we should have a separate test for the function outside the _TestEllipsePhantom_ class. Still, testing the match of the recon_image and ref_image within the class makes sense imo.

ckolbPTB:
I struggle with the pyfixture version for the phantom. This of course works:

```
@pytest.fixture(scope='session')
def ellipse_phantom():
    """Create ellipse phantom for testing."""
    # Define five ellipses
    test_ellipses = [
        EllipsePars(center_x=0.1, center_y=0.0, radius_x=0.1, radius_y=0.25, intensity=1),
        EllipsePars(center_x=0.3, center_y=0.3, radius_x=0.1, radius_y=0.1, intensity=2),
        EllipsePars(center_x=0.1, center_y=0.1, radius_x=0.1, radius_y=0.1, intensity=3),
        EllipsePars(center_x=-0.2, center_y=-0.2, radius_x=0.1, radius_y=0.1, intensity=4),
        EllipsePars(center_x=-0.3, center_y=-0.3, radius_x=0.1, radius_y=0.1, intensity=5),
    ]
    return EllipsePhantom(test_ellipses)

def test_image_space(ellipse_phantom):
    """Check if image space has correct shape."""
    nx = 512
    ny = 256
    im = ellipse_phantom.image_space(nx, ny)
    assert im.shape == (ny, nx)
```
but then `nx`, `ny` have to be defined in each test function which does not seem nice.

Of course something like `return (EllipsePhantom(test_ellipses), nx, ny, kx, ky)` would be possible but also not very nice.

One other option is to create a `EllipseTestPhantom` class which creates a phantom of a certain shape and call this within ellipse_phantom(). But then I don't see the advantage of the pytest.fixture approach and would then go with Patrick's approach.

fzimmermann89:
Maybe we should have a short discussion in the next meeting about what functions are tested where .. 

My understanding is that image_to_kspace are utilities and should just be used within the tests. Those are just wrappers around library functions.

Not sure if we need test functions for test functions.. 

What we test, is that the data is loaded correctly.

tests/phantoms/test_phantoms.py
fzimmermann89: > `test_phantom_verification` ?
> 
> or
> 
> Can't we do the test by just implementing a test for the functions `test_kspace_to_image` and `test_image_to_kspace`?
> In these we could verify if the analytical phantom (kspace and image) korresponds to the transformations. 

No. This is circlular. Those two functions are just wrappers around the FFT. We want to test the data loading. 

We might want to test the phantom as well.

But this is a separate issue.

We should avoid needing tests for tests. ;)

schuenke:
I agree that we should discuss this in more detail in the next meeting, but maybe we can find a solution for the current PR so that we can merge it and proceed with other open issues like the pytest integration etc. This would be my summary so far:

- we should not test test-functions. Thus a test for `kspace_to_image()` is not needed here.
- we intend to use fixtures, but don't try to enforce it if it doesn't make sense
- test classes can be an option to test the methods of a class and reduce overhead compared to independent test functions

For the current PR, I would suggest to change to the TestEllipsePhantom class, but rename the `test_kspace_to_image` method posted above to `test_kspace_image_match`, `test_phantom_verification` or something like this.

And just to make it clear again: this test is just specific for the current implementation of the EllipsePhantom. For other Phantoms we would maybe assert something else using numpy.testing.assert_almost_equal() etc...

fzimmermann89:
_**EDIT : Sorry, I got confused about the scope.  I mixed up a) using the phantom in a test and b) testing the phantom.**_


From the main page of pytest: "Having each test share the same class instance would be very detrimental to test isolation and would promote poor test practices." (https://docs.pytest.org/en/7.1.x/getting-started.html)
 
Re Christoph: The advantage is reusability and composability. The phantom can be used in different tests, jointly with different other fixtures. I would even consider going a step further and creating an ellipse_cartestian_img and ellipse_cartesian_kspace fixture to be used for the kdata tests.


> def test_image_space_shape(ellipse_phantom):
>     """Check if image space has correct shape."""
>     nx = 512
>     ny = 256
>     im = ellipse_phantom.image_space(nx, ny)
>     assert im.shape == (ny, nx)
> ```

This looks completely self contained and descriptive to me,.  I much prefer this style.

fzimmermann89:
Maybe we can split this into two PRs.

Frist, get the ismrmrd test data stuff merged ASAP with this PR.
Second, add tests for the phantom.

ckolbPTB:
Next try. The problem with pytest.fixtures and flake8 is now solved by excluding tests from flake8 in the pre-comit-config.yaml. 

Other options would be to 

- put all fixtures into conftest.py (maybe a bit messy) OR
- add flake8 exceptions to the specific lines (e.g. # noqa: F401)

fzimmermann89:
Question:
Is there an advantage of using scipy and numpy instead of torch?

On the other hand, it could be useful to have all functions differentiable  and able to handle Tensor input (and GPU tensors)

There is torch.special.bessel_j1  and torch.fft.

This is not urgent, but something to maybe consider for future work.

PR #59, opened 2023-09-04
switch CI pipeline from unittest to pytest

schuenke:


schuenke:
Unfortunately, the links in the *Coverage Report* do not work yet. However, this seems to be a general Issue with the **MishaKav/pytest-coverage-comment Action** because it doesn't work in their own repo, as well: https://github.com/MishaKav/pytest-coverage-comment/pull/142

fzimmermann89:
> Unfortunately, the links in the _Coverage Report_ do not work yet. However, this seems to be a general Issue with the **MishaKav/pytest-coverage-comment Action** because it doesn't work in their own repo, as well: [MishaKav/pytest-coverage-comment#142](https://github.com/MishaKav/pytest-coverage-comment/pull/142)


No. This s a different issue. The links will not work on their repo as they are mocked with some example code that is not in the action repo. But it works in different repos, https://github.com/ONSdigital/research-and-development/pull/13#issuecomment-1486615453 for example.

This issue is NOT caused by the action. But the coverage file containing these wrong paths. (See action execution log)
So something is wrong with how we call the coverage generation inside the github actions..

EDIT: Fixed. We need coveragerc paths to merge src and site-packages/...

fzimmermann89 (commented):
...

schuenke:
Works now ??!?? This badge should update itself everytime pytest is run...

badge

fzimmermann89 (approved):

ckolbPTB (approved):

fzimmermann89:
Nice :)

PR #60, opened 2023-09-05
Update VSCode settings & recommended extensions

schuenke:
Feel free to add/remove options and extensions or to pimp our README !!!

What I did so far:
- add black extension
- remove deprecated entries in vscode settings
- update README


schuenke:
If there are no other extensions / infos you would like to add, I would merge this PR today.

fzimmermann89:
The only thing I would like is to have these long copyright notices collapsed by default. Similar to what the editor.foldingImportsByDefault does for imports.

 I haven't found the time to figure out how to configure this. There are some extensions, that might work, but you need to specify what to fold by regexes and spent the time to try them out.. 


But this is super low priority and I can also be done at a later time.

PR #62, opened 2023-09-14
RPE trajectory

### LINKED ISSUE #10: RPE trajectory
Create trajectory for (Golden) RPE including Johannes' sunflower trajectory

Check list:
- [x]     Create branch
- [ ]     Implement code
- [ ]     Add comments
- [ ]     Implement tests
- [ ]     Squash commmits
- [ ]     Create PR

ckolbPTB:
closes #10

### DIFF
--- a/tests/data/test_ktraj.py
+++ b/tests/data/test_ktraj.py
@@ -1,0 +1,69 @@
+"""Tests for KTrajectory classes."""
+
+# Copyright 2023 Physikalisch-Technische Bundesanstalt
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#       http://www.apache.org/licenses/LICENSE-2.0
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+import numpy as np
+import pytest
+import torch
+
+from mrpro.data import KTrajectoryRpe
+from mrpro.data import KTrajectorySunflowerGoldenRpe
+from tests.data.conftest import random_kheader
+
+
+@pytest.fixture(scope='function')
+def valid_rpe_kheader(monkeypatch, random_kheader):
+    """KHeader with all necessary parameters for RPE trajectories."""
+    # K-space dimensions
+    nk0 = 200
+    nk1 = 20
+    nk2 = 10
+
+    # List of k1 and k2 indices in the shape (d4, k2, k1)
+    k1 = torch.linspace(0, nk1 - 1, nk1, dtype=torch.int32)
+    k2 = torch.linspace(0, nk2 - 1, nk2, dtype=torch.int32)
+    idx_k1, idx_k2 = torch.meshgrid(k1, k2)
+    idx_k1 = torch.reshape(idx_k1, (1, nk2, nk1))
+    idx_k2 = torch.reshape(idx_k2, (1, nk2, nk1))
+
+    # Set parameters for RPE trajectory
+    monkeypatch.setattr(random_kheader.acq_info, 'number_of_samples', torch.zeros_like(idx_k1) + nk0)
+    monkeypatch.setattr(random_kheader.acq_info, 'center_sample', torch.zeros_like(idx_k1) + nk0 // 2)
+    monkeypatch.setattr(random_kheader.acq_info.idx, 'k1', idx_k1)
+    monkeypatch.setattr(random_kheader.acq_info.idx, 'k2', idx_k2)
+    monkeypatch.setattr(random_kheader.encoding_limits.k1, 'center', int(nk1 // 2))
+    monkeypatch.setattr(random_kheader.encoding_limits.k1, 'max', int(nk1 - 1))
+    monkeypatch.setattr(random_kheader.encoding_limits.k2, 'center', int(nk2 // 2))
+    return random_kheader
+
+
+def rpe_traj_shape(valid_rpe_kheader):
+    """Expected shape of trajectory based on KHeader."""
+    nk0 = valid_rpe_kheader.acq_info.number_of_samples[0, 0, 0]
+    nk1 = valid_rpe_kheader.acq_info.idx.k1.shape[2]
+    nk2 = valid_rpe_kheader.acq_info.idx.k1.shape[1]
+    return torch.Size([1, 3, nk2, nk1, nk0])
+
+
+def test_KTrajectoryRpe(valid_rpe_kheader):
+    """Calculate RPE trajectory."""
+    ktrajectory = KTrajectoryRpe(angle=np.pi * 0.618034)
+    ktraj = ktrajectory.calc_traj(valid_rpe_kheader)
+    assert ktraj.shape == rpe_traj_shape(valid_rpe_kheader)
+
+
+def test_KTrajectorySunflowerGoldenRpe(valid_rpe_kheader):
+    """Calculate RPE Sunflower trajectory."""
+    ktrajectory = KTrajectorySunflowerGoldenRpe(rad_us_factor=1)
+    ktraj = ktrajectory.calc_traj(valid_rpe_kheader)
+    assert ktraj.shape == rpe_traj_shape(valid_rpe_kheader)
--- a/src/mrpro/data/_KData.py
+++ b/src/mrpro/data/_KData.py
@@ -26,9 +26,9 @@ from einops import rearrange
 
 from mrpro.data import AcqInfo
 from mrpro.data import KHeader
-from mrpro.data import KTrajectory
 from mrpro.data._EncodingLimits import Limits
 from mrpro.data.enums import AcqFlags
+from mrpro.data.traj_calculators import KTrajectory
 
 KDIM_SORT_LABELS = (
     'k1',
--- a/src/mrpro/data/_KHeader.py
+++ b/src/mrpro/data/_KHeader.py
@@ -31,7 +31,7 @@ from mrpro.data import enums
 
 if TYPE_CHECKING:
     # avoid circular imports by importing onlz when type checking
-    from mrpro.data import KTrajectory
+    from mrpro.data.traj_calculators import KTrajectory
 
 UNKNOWN = 'unknown'
 
--- a/src/mrpro/data/__init__.py
+++ b/src/mrpro/data/__init__.py
@@ -7,6 +7,8 @@ from mrpro.data._AcqInfo import AcqIdx
 from mrpro.data._AcqInfo import AcqInfo
 from mrpro.data._EncodingLimits import EncodingLimits
 from mrpro.data._KTrajectory import KTrajectory
+from mrpro.data._KTrajectory import KTrajectoryRpe
+from mrpro.data._KTrajectory import KTrajectorySunflowerGoldenRpe
 from mrpro.data._KHeader import KHeader
 from mrpro.data._KData import KData
 from mrpro.data._IHeader import IHeader
--- a/src/mrpro/data/__init__.py
+++ b/src/mrpro/data/__init__.py
@@ -6,9 +6,6 @@ from mrpro.data._TrajectoryDescription import TrajectoryDescription
 from mrpro.data._AcqInfo import AcqIdx
 from mrpro.data._AcqInfo import AcqInfo
 from mrpro.data._EncodingLimits import EncodingLimits
-from mrpro.data._KTrajectory import KTrajectory
-from mrpro.data._KTrajectory import KTrajectoryRpe
-from mrpro.data._KTrajectory import KTrajectorySunflowerGoldenRpe
 from mrpro.data._KHeader import KHeader
 from mrpro.data._KData import KData
 from mrpro.data._IHeader import IHeader
--- a/src/mrpro/data/traj_calculators/_KTrajectory.py
+++ b/src/mrpro/data/traj_calculators/_KTrajectory.py
@@ -1,4 +1,4 @@
-"""K-space trajectory classes."""
+"""K-space trajectory base class."""
 
 # Copyright 2023 Physikalisch-Technische Bundesanstalt
 #
@@ -29,10 +29,10 @@ class KTrajectory(ABC):
 
     @abstractmethod
     def calc_traj(self, header: KHeader) -> torch.Tensor:
-        """Calculate the trajectory for the given header.
+        """Calculate the trajectory for given KHeader.
 
-        The shape should broadcastable to (prod(all_other_dimensions),
-        3, k2, k1, k0)
+        The shape of the calculated trajectory must be broadcastable to
+        (prod(all_other_dimensions), 3, k2, k1, k0)
         """
         ...
 
@@ -40,13 +40,12 @@ class KTrajectory(ABC):
 class DummyTrajectory(KTrajectory):
     """Simple Dummy trajectory that returns zeros.
 
-    Shape will not fit to all data. Only used until we implement proper
-    trajectories
+    Shape will not fit to all data. Only used as dummy for testing.
     """
 
     @staticmethod
     def _get_shape(header: KHeader) -> tuple[int, ...]:
-        """Get the shape of a basic dummy trajectory for the given header.
+        """Get the shape of a basic dummy trajectory for the given KHeader.
 
         Assumes fully sampled data. Do not use outside of testing.
         """
@@ -68,9 +67,5 @@ class DummyTrajectory(KTrajectory):
         return shape
 
     def calc_traj(self, header: KHeader) -> torch.Tensor:
-        """Calculate the trajectory for the given header.
-
-        The shape should broadcastable to (prod(all_other_dimensions),
-        3, k2, k1, k0)
-        """
+        """Calculate dummy trajectory for given KHeader."""
         return torch.zeros(self._get_shape(header))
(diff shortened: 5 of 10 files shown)


src/mrpro/data/_KTrajectory.py
```diff
@@ -37,6 +37,225 @@ def calc_traj(self, header: KHeader) -> torch.Tensor:
         ...
 
 
+class KTrajectoryRpe(KTrajectory):
```
schuenke: I would suggest to move KTrajectoryRpe and KTrajectorySunflowerGoldenRpe into one or even two separate file(s). We will have several other trajectories and implementing them all in a single file might become messy.

src/mrpro/data/_KTrajectory.py
```diff
@@ -37,6 +37,225 @@ def calc_traj(self, header: KHeader) -> torch.Tensor:
...
+            MR raw data header (KHeader) containing required meta data
+        """
+        kang = kang.flatten()
+        _, indices = np.unique(kang, return_index=True)
+        ang_idx = np.argsort([kang[index] for index in sorted(indices)])
+        shift_idx = np.zeros(max(ang_idx) + 1)
+        shift_idx[ang_idx] = np.arange(len(ang_idx))
+
+        # Apply sunflower shift
+        GoldenCut = 0.5 * (np.sqrt(5) + 1)
```
schuenke: _GoldenCut_ should be all lower case and renamed to _golden_cut_ etc.

src/mrpro/data/_KTrajectory.py
```diff
@@ -37,6 +37,225 @@ def calc_traj(self, header: KHeader) -> torch.Tensor:
...
+    """
+
+    def __init__(
+        self,
+        angle: float,
+        shift_between_rpe_lines: torch.Tensor = torch.tensor([0, 2, 1, 3]),
+        rad_us_factor: float = 1.0,
+    ) -> None:
+        super().__init__()
+        self.angle: float = angle
```
schuenke: does it make sense to limit the angle between 0 and 2pi using `self.angle: float = angle % (2 * np.pi)` ?

src/mrpro/data/_KTrajectory.py
```diff
@@ -37,6 +37,225 @@ def calc_traj(self, header: KHeader) -> torch.Tensor:
...
+    https://doi.org/10.1118/1.4890095 (open access).
+
+    Parameters
+    ----------
+    angle
+        angle in rad between two radial phase encoding lines
+    shift_between_rpe_lines
+        shift between radial phase encoding lines along the radial direction.
+        See _apply_shifts_between_rpe_lines() for more details
+    rad_us_factor
```
schuenke: the unsersampling factor (`rad_us_factor`) is currently not used. Is this on purpose? Is it still work in progress? If we don't use it yet, I would vote to remove it to prevent confusion.

src/mrpro/data/_KTrajectory.py
```diff
@@ -37,6 +37,225 @@ def calc_traj(self, header: KHeader) -> torch.Tensor:
...
+
+        # Angles of phase encoding lines
+        kang = kheader.acq_info.idx.k2 * self.angle
+
+        # K-space locations along phase encoding lines
+        krad = (kheader.acq_info.idx.k1 - kheader.encoding_limits.k1.center).to(torch.float32)
+        krad = self._apply_shifts_between_rpe_lines(krad, (kheader.acq_info.idx.k2 - kheader.encoding_limits.k2.center))
+        krad *= 2 * torch.pi / kheader.encoding_limits.k1.max
+
+        return self._combine_to_3d_traj(krad, kang, k0)
```
schuenke: Would it make sense to store the calculated trajectory in our KTrajectory objects as well? Or should it be returned only? I was thinking about simple KTrajectory.plot() methods and currently such visualisation and/or other further processing methods would require to rerun the `calc_traj()` method

tests/data/test_ktraj.py
```diff
@@ -0,0 +1,69 @@
...
+
+def rpe_traj_shape(valid_rpe_kheader):
+    """Expected shape of trajectory based on KHeader."""
+    nk0 = valid_rpe_kheader.acq_info.number_of_samples[0, 0, 0]
+    nk1 = valid_rpe_kheader.acq_info.idx.k1.shape[2]
+    nk2 = valid_rpe_kheader.acq_info.idx.k1.shape[1]
+    return torch.Size([1, 3, nk2, nk1, nk0])
+
+
+def test_KTrajectoryRpe(valid_rpe_kheader):
```
schuenke: I still propose that we test all class methods and also as many initialization cases as possible. For KTrajectoryRpe I thus propose to introduce different tests that cover for example the shape, different `angle `between the spokes, different `shift_between_rpe_lines`, different `rad_us_factor` (if we keep it) etc...

This could look like this:

```python
def test_KTrajectoryRpe_shape(valid_rpe_kheader):
    """Calculate RPE trajectory."""
    ktrajectory = KTrajectoryRpe(angle=np.pi * 0.618034)
    ktraj = ktrajectory.calc_traj(valid_rpe_kheader)
    assert ktraj.shape == rpe_traj_shape(valid_rpe_kheader)


def test_KTrajectoryRpe_angle(valid_rpe_kheader):
    ktrajectory1 = KTrajectoryRpe(angle=np.pi * 0.618034)
    ktrajectory2 = KTrajectoryRpe(angle=np.pi * 0.618034 + 2 * np.pi)
    ktraj1 = ktrajectory1.calc_traj(valid_rpe_kheader)
    ktraj2 = ktrajectory2.calc_traj(valid_rpe_kheader)
    assert np.array_equal(ktraj1, ktraj2)

def test_KTrajectoryRpe_rad_us_factor(valid_rpe_kheader):
    ...

```

tests/data/test_ktraj.py
```diff
@@ -0,0 +1,69 @@
...
+    """KHeader with all necessary parameters for RPE trajectories."""
+    # K-space dimensions
+    nk0 = 200
+    nk1 = 20
+    nk2 = 10
+
+    # List of k1 and k2 indices in the shape (d4, k2, k1)
+    k1 = torch.linspace(0, nk1 - 1, nk1, dtype=torch.int32)
+    k2 = torch.linspace(0, nk2 - 1, nk2, dtype=torch.int32)
+    idx_k1, idx_k2 = torch.meshgrid(k1, k2)
```
schuenke: This should be `idx_k1, idx_k2 = torch.meshgrid(k1, k2, indexing='ij')` to avoid Warning:

**_UserWarning: torch.meshgrid: in an upcoming release, it will be required to pass the indexing argument. (Triggered internally at ..\aten\src\ATen\native\TensorShape.cpp:3484.)_**

schuenke (commented):
Please see my comments...

src/mrpro/data/_KTrajectory.py
fzimmermann89: No, the trajectory is not meant as a data storing class, it is only used to calculate trajectories. All stored state shall be settings for the trajectory. Not a concrete Tensor of kspace points.
As this can be different for different aquisitions.

The trajectory is saved only in Kdata.traj

So if we want to have a lot function, it should take in a Kdata object.

src/mrpro/data/_KTrajectory.py
ckolbPTB: I would not pose any restrictions on the angle because the trajectory calculation has to be the same as in the sequence. Although I cannot give a good reason why somebody would use _angle > 2 * pi_ it might be needed for something and then we should not change this here.

tests/data/test_ktraj.py
ckolbPTB: I will try to add some more tests. 

fyi "test_KTrajectoryRpe_angle" fails because pytorch does not seem to be too accurate. For me this
`torch.cos(torch.tensor(np.pi * 0.618034)) - torch.cos(torch.tensor(np.pi * 0.618034 + 2 * np.pi))`
yields `tensor(-5.9605e-08)` which would be fine when comparing with `almost_equal` but for 100 spokes: 
`torch.cos(100*torch.tensor(np.pi * 0.618034)) - torch.cos(100*torch.tensor(np.pi * 0.618034 + 2 * np.pi))` 
yields `tensor(1.0848e-05)`

`np.cos(np.pi * 0.618034) - np.cos(np.pi * 0.618034 + 2 * np.pi)` yields
`-4.996003610813204e-16`

schuenke: When I adjust the init to `self.angle: float = angle % (2 * np.pi)` (see comment above), this test actually passes even using `array_equal()`

src/mrpro/data/_KTrajectory.py
ckolbPTB: left over from an old version -> removed

tests/data/test_ktraj.py
ckolbPTB: yes, but if somebody used `torch.tensor(np.pi * 0.618034 + 2 * np.pi)` to calculate the angle in pypulseq then the trajectory would be wrong in mrpro if we used `self.angle: float = angle % (2 * np.pi)`

tests/data/test_ktraj.py
```diff
@@ -0,0 +1,85 @@
...
+
+def rpe_traj_shape(valid_rpe_kheader):
+    """Expected shape of trajectory based on KHeader."""
+    nk0 = valid_rpe_kheader.acq_info.number_of_samples[0, 0, 0]
+    nk1 = valid_rpe_kheader.acq_info.idx.k1.shape[2]
+    nk2 = valid_rpe_kheader.acq_info.idx.k1.shape[1]
+    return torch.Size([1, 3, nk2, nk1, nk0])
+
+
+def test_KTrajectoryRpe_uniform(valid_rpe_kheader):
```
schuenke: Is it worth testing the shape of both, a uniform and a golden angle distribution? There is no reason why the shapes should differ, is there?

I think it would make more sense to test for specific implementation details, for example assert that the k-space center is never shifted (e.g. `torch.all(ktraj[0, 0, :, :, center_idx] == 0)`

ckolbPTB: I modified this test to actually check ktraj pos for angle_gap and angle_gap//2 for matching RPE lines. (actually found a bug in the test code in the process  :) )

src/mrpro/data/_KTrajectoryRpe.py
```diff
@@ -0,0 +1,165 @@
...
+        angle in rad between two radial phase encoding lines
+    shift_between_rpe_lines
+        shift between radial phase encoding lines along the radial direction.
+        See _apply_shifts_between_rpe_lines() for more details
+    """
+
+    def __init__(
+        self,
+        angle: float,
+        shift_between_rpe_lines: torch.Tensor = torch.tensor([0, 2, 1, 3]),
```
fzimmermann89: We could consider making this a tuple of ints?
This can only be int, only 1d, never differentiable, right?

src/mrpro/data/_KTrajectoryRpe.py
```diff
@@ -0,0 +1,165 @@
...
+
+        Parameters
+        ----------
+        krad
+            k-space positions along each phase encoding line
+        kang_idx
+            indices of angles to be used for shift calculation
+        """
+        delta_shift = 1 / len(self.shift_between_rpe_lines)
+        for ind in range(len(self.shift_between_rpe_lines)):
```
fzimmermann89: for ind, shift in enumerate(self.shift_between_rpe_lines)

src/mrpro/data/_KTrajectoryRpe.py
```diff
@@ -0,0 +1,165 @@
...
+            curr_angle_idx = torch.nonzero(
+                torch.fmod(kang_idx, len(self.shift_between_rpe_lines)) == ind, as_tuple=True
+            )
+            curr_shift = self.shift_between_rpe_lines[ind] * delta_shift
+            curr_krad = krad[curr_angle_idx]
+
+            # Do not shift the k-space center
+            center_idx = curr_krad == 0
```
fzimmermann89: ==0 or approximately equal 0?

fzimmermann89: curr_krad +=  curr_shift*(curr_krad!=0) ?

tests/data/test_ktraj.py
```diff
@@ -0,0 +1,89 @@
...
+    """KHeader with all necessary parameters for RPE trajectories."""
+    # K-space dimensions
+    nk0 = 200
+    nk1 = 20
+    nk2 = 10
+
+    # List of k1 and k2 indices in the shape (d4, k2, k1)
+    k1 = torch.linspace(0, nk1 - 1, nk1, dtype=torch.int32)
+    k2 = torch.linspace(0, nk2 - 1, nk2, dtype=torch.int32)
+    idx_k2, idx_k1 = torch.meshgrid(k2, k1, indexing='ij')
```
fzimmermann89: consider `idx_k1, idx_k2 = torch.meshgrid(k1, k2, indexing='xy')`, as xy will become default in pytorch

src/mrpro/data/_KTrajectoryRpe.py
```diff
@@ -0,0 +1,165 @@
...
+            k-space points along readoud
+
+        Returns
+        -------
+            3D k-space trajectory
+        """
+        k1 = torch.repeat_interleave((krad * torch.cos(kang))[..., None], k0.shape[0], dim=3)
+        k2 = torch.repeat_interleave((krad * torch.sin(kang))[..., None], k0.shape[0], dim=3)
+        k0 = k0.expand(k1.shape)
+        return torch.concatenate([k0[:, None, :, :, :], k1[:, None, :, :, :], k2[:, None, :, :, :]], dim=1)
```
fzimmermann89: For comparison, an einops way would be:
```
k1=einops.repeat(krad * torch.cos(kang),"...->... k0",k0=k0.shape[-1])
k2=einops.repeat(krad * torch.sin(kang),"...->... k0",k0=k0.shape[-1])
k0 = k0.expand(k1.shape)
return einops.rearrange([k0,k1,k2],"dir other k2 k1 k0->other dir k2 k1 k0")
```
Which to me is easier to understand,..

src/mrpro/data/_KTrajectorySunflowerGoldenRpe.py
```diff
@@ -0,0 +1,90 @@
...
+        kang
+            angles of the radial phase encoding lines
+        kheader
+            MR raw data header (KHeader) containing required meta data
+        """
+        kang = kang.flatten()
+        _, indices = np.unique(kang, return_index=True)
+        ang_idx = np.argsort([kang[index] for index in sorted(indices)])
+        shift_idx = np.zeros(max(ang_idx) + 1)
+        shift_idx[ang_idx] = np.arange(len(ang_idx))
```
fzimmermann89: as np.unique (and torch.unique with default sorted=True) return sorted values,
`shift_idx = np.argsort(indices)` or am I missing some case? If I am wrong, please add a comment why it has to be this complicated...
Also, is there a reason to first allocate a numpy zero-array and then fill it?

src/mrpro/data/_KTrajectorySunflowerGoldenRpe.py
```diff
@@ -0,0 +1,90 @@
...
+
+        # Apply sunflower shift
+        golden_cut = 0.5 * (np.sqrt(5) + 1)
+        for ind in range(int(len(shift_idx))):
```
fzimmermann89:
```  
      for ind, shift in enumerate(shift_idx):
            krad[kheader.acq_info.idx.k2 == ind] +=((shift * golden_cut) % 1) - 0.5
 ```    
        
does the same.
        
Please check if the braces for `*2*0.5` are correct.
          
  Or is there a numerical reason why it has to be done that way? If yes, consider adding a comment for future reference

File: src/mrpro/data/_KTrajectorySunflowerGoldenRpe.py [hunk 16]
```diff
@@ -0,0 +1,90 @@
...
+            MR raw data header (KHeader) containing required meta data
+        """
+        kang = kang.flatten()
+        _, indices = np.unique(kang, return_index=True)
+        ang_idx = np.argsort([kang[index] for index in sorted(indices)])
+        shift_idx = np.zeros(max(ang_idx) + 1)
+        shift_idx[ang_idx] = np.arange(len(ang_idx))
+
+        # Apply sunflower shift
+        golden_cut = 0.5 * (np.sqrt(5) + 1)
```
fzimmermann89: This is the constant phi, aka golden_ratio , correct?
Maybe consider naming it golden_ratio or phi?
(Why is it named "cut" )

fzimmermann89 (commented):
some small comments and suggestions. Not sure I completely understood all of the logic. If I misunderstood it, you might consider adding some comments.

src/mrpro/data/_KTrajectoryRpe.py
ckolbPTB: tuple[float,...] because it does not have to be integer

fzimmermann89: If it is float, I'd expect to specify normalized values (0, 0.5, 0.25, 0.75).
The normalizatoin by the length confused me and made me think it has to be integer.
If it now can be float, we might consider ignoring my comment and leaving it as a tensor to be differentiable in the offset?

ckolbPTB: I changed it to normalized values as input rather than normalising with 1/length later

schuenke:
When we planned the package structure, the idea was that the "data" subfolder should contain all data classes (e.g., KData, IData, QData, CsmData, DcfData...). Should we move everything related to trajectories into a mrpro.traj subfolder or would you like it to stay in the data folder?

fzimmermann89 (approved):

PR #70, opened 2023-09-28
Suggestions for Refactor of DCF/Trajectory Logic

fzimmermann89:
I had the idea to refactor Kdata.traj into a KTrajectory dataclass with kx, ky, kz properties.

This was partially influenced by somebody asking how a trajectory shape interpreted, partially by the need for a special 'dcf_rpe_voronoi' - logic:

Now the RPE can have a trajectory with broadcasted kz dimension.
And the dcf voronoi logic can look if any of kx,ky,kz has two singleton dimensions. If this happens, then we can ignore this dimension in the calculation (I hope that's correct...)

Also, in a second pass, I refactored the DCF calculation a bit, mostly by simplifying the multiprocessing and sorting logic.
As scipy.spatial calls qhull with nogil from cython, we should be able to use multithreading. 
This is something I should test on one of the recon servers with a bit more demanding DCF example.

I am pretty sure that there are still logic issues in this code, and some style improvements can be made, but I want to open the discussion if this goes in the right direction or if we want to keep it as is.

@schuenke @ckolbPTB 

(If we want to go ahead, I can rebase and split the changes)

### DIFF
--- a/src/mrpro/data/_DcfData.py
+++ b/src/mrpro/data/_DcfData.py
@@ -21,11 +21,11 @@ from itertools import product
 
 import numpy as np
 import torch
-from mpro.utils import smap
 from scipy.spatial import ConvexHull
 from scipy.spatial import Voronoi
 
 from mrpro.data import KTrajectory
+from mrpro.utils import smap
 
 UNIQUE_ROUNDING_DECIMALS = 15
 
--- a/src/mrpro/data/_DcfData.py
+++ b/src/mrpro/data/_DcfData.py
@@ -103,10 +103,10 @@ class DcfData:
         Parameters
         ----------
         traj
-            torch.Tensor containing k-space points (d4, 2 or 3, k2, k1, k0).
+            torch.Tensor containing k-space points (other, 2 or 3, k2, k1, k0).
         """
 
-        ks = [traj.kx, traj.ky, traj.kz]
+        ks = [traj.kz, traj.ky, traj.kx]
         for i, k in enumerate(ks):
             if any(all(k.shape[ax] == 1 for ax in two_axes) for two_axes in [(-1, -2), (-1, -3), (-2, -3)]):
                 # Found a direction with at least two singleton dimensions, i.e. we have a 2D trajectory in 3D space
--- a/src/mrpro/data/_DcfData.py
+++ b/src/mrpro/data/_DcfData.py
@@ -13,29 +13,15 @@
 #   limitations under the License.
 
 from __future__ import annotations
-
 import dataclasses
-import itertools
-from functools import partial
-from multiprocessing import Pool
-
+from concurrent.futures import ThreadPoolExecutor
+from itertools import product
 import numpy as np
-import psutil
 import torch
-from einops import repeat
 from scipy.spatial import ConvexHull
 from scipy.spatial import Voronoi
 
-
-# Calculate volume/area of voronoi cells
-def calc_voronoi_volume_area(points_to_regions, verts, v_cell, idx):
-    """Calculate volume/area of voronoi cells."""
-    dcf_vol = np.zeros_like(idx, dtype=np.float64)
-    for n, id in enumerate(idx):
-        cell = v_cell[points_to_regions[id]]
-        vertices = [verts[j] for j in cell]
-        dcf_vol[n] = ConvexHull(vertices).volume
-    return dcf_vol
+from mrpro.data import KTrajectory
 
 
 @dataclasses.dataclass(slots=True, frozen=False)
@@ -60,49 +46,35 @@ class DcfData:
         -------
             density compensation values (1, k2, k1, k0)
         """
+        UNIQUE_ROUNDING_DECIMALS = 15
+
         # 2D and 3D trajectories supported
-        if traj.shape[0] not in (2, 3):
-            raise ValueError('Only 2D or 3D trajectories supported.')
+
+        dim = traj.shape[0]
+        if dim not in (2, 3):
+            raise ValueError(f'Only 2D or 3D trajectories supported, not {dim}D.')
 
         # Calculate dcf only for unique k-space positions
         traj_dim = traj.shape
-        traj = np.round(traj.numpy(), decimals=15)
-        traj = traj.reshape(traj_dim[0], -1)
+        traj = np.round(traj.numpy(), decimals=UNIQUE_ROUNDING_DECIMALS)
+        traj = traj.reshape(dim, -1)
         traj_unique, inverse, counts = np.unique(traj, return_inverse=True, return_counts=True, axis=1)
 
         # Especially in 3D, errors in the calculation of the convex hull can occur for edge points. To avoid this,
         # the corner points of a cube bounding box are added here. The bouding box is chosen very large to ensure these
         # edge points of the trajectory can still be accurately detected in the outlier detection further down.
         furthest_corner = np.max(np.abs(traj_unique))
-        corner_points = np.array(list(itertools.product([-1, 1], repeat=traj.shape[0]))) * furthest_corner * 10
-        traj_unique = np.concatenate((traj_unique, corner_points.transpose()), axis=1)
+        corner_points = np.array(list(product([-1, 1], repeat=dim))) * furthest_corner * 10
+        traj_extendend = np.concatenate((traj_unique, corner_points.transpose()), axis=1)
 
         # Carry out voronoi tessellation
-        vdiagram = Voronoi(traj_unique.transpose())
+        vdiagram = Voronoi(traj_extendend.transpose())
+        regions = [vdiagram.regions[r] for r in vdiagram.point_region[: -len(corner_points)]]  # Ignore corner points
+        vertices = [vdiagram.vertices[region] for region in regions]
 
-        # List of regions is now an array so we have fixed indices
-        v_cell = np.array(vdiagram.regions, dtype=object)
-
-        # Calculate area (2D) or volume (3D) of each voronoi cell but not the corner points added above
-        dcf = np.zeros(len(vdiagram.points) - len(corner_points), dtype=np.float64) - 10
-
-        # Split vertices for different workers (without the added corner points)
-        num_cpu = psutil.cpu_count(logical=False)
-        idx_cpu = []
-        for ind in range(num_cpu):
-            idx_cpu.append(np.asarray(range(ind, len(vdiagram.points) - len(corner_points), num_cpu)))
-
-        # Open Pool
-        pool = Pool(processes=num_cpu)
-        results = pool.starmap(
-            partial(calc_voronoi_volume_area, vdiagram.point_region, vdiagram.vertices, v_cell), zip(iter(idx_cpu))
-        )
-        pool.close()  # shut down the pool
-
-        # Sort in results from different cpus
-        for ind in range(num_cpu):
-            cidx = idx_cpu[ind]
-            dcf[cidx] = results[ind]
+        # Calculate volume/area of voronoi cells
+        future = ThreadPoolExecutor(max_workers=torch.get_num_threads()).map(lambda v: ConvexHull(v).volume, vertices)
+        dcf = np.array(list(future))
 
         # Get outliers (i.e. voronoi cell which are unbound) and set them to a reasonable value
         # Outliers are defined as values larger than 1.5 * inter quartile range of the values
@@ -111,34 +83,19 @@ class DcfData:
         q1, q3 = np.percentile(dcf_sorted, [25, 75])
         iqr = q3 - q1
         upper_bound = q3 + 1.5 * iqr
-        idx_outlier = np.where(dcf > upper_bound)
-        dcf_sorted_without_outliers = np.sort(np.delete(dcf, idx_outlier))
-        max_idx = int(len(dcf_sorted_without_outliers) * 0.99)
-        outlier_val = np.average(dcf_sorted_without_outliers[max_idx:])
-        dcf[idx_outlier] = outlier_val
+        idx_outliers = np.nonzero(dcf > upper_bound)
+        num_outliers = len(idx_outliers[0])
+        high_values_start = int(0.99 * (len(dcf_sorted) - num_outliers))
+        fill_value = np.average(dcf_sorted[high_values_start:-num_outliers])
+        dcf[idx_outliers] = fill_value
 
         # Sort dcf values back into the original order (i.e. before calling unique)
-        dcf = np.reshape((dcf / counts)[inverse], (1,) + traj_dim[1:])
+        dcf = np.reshape((dcf / counts)[inverse], traj_dim[1:])
 
         return torch.tensor(dcf, dtype=torch.float32)
 
-    @staticmethod
-    def _voronoi_dcf_for_each_d4(traj: torch.Tensor) -> torch.Tensor:
-        """Loop over all entries in d4 and calculate voronoi dcf.
-
-        Parameters
-        ----------
-        traj
-            torch.Tensor containing k-space points (d4, 2 or 3, k2, k1, k0).
-        """
-        # Calculate dcf for each dynamic, phase, average...
-        dcf = torch.zeros((traj.shape[0], 1) + traj.shape[2:], dtype=torch.float32)
-        for ind in range(traj.shape[0]):
-            dcf[ind, ...] = DcfData._dcf_using_voronoi(traj[ind, ...])
-        return dcf
-
     @classmethod
-    def from_traj_voronoi(cls, traj: torch.Tensor) -> DcfData:
+    def from_traj_voronoi(cls, traj: KTrajectory) -> DcfData:
         """Calculate dcf using voronoi approach for 2D or 3D trajectories.
 
         Parameters
@@ -146,27 +103,19 @@ class DcfData:
         traj
             torch.Tensor containing k-space points (d4, 2 or 3, k2, k1, k0).
         """
-        if traj.shape[1] != 2 and traj.shape[1] != 3:
-            raise ValueError('Trajectoy has to be 2D or 3D')
-        return cls(data=DcfData._voronoi_dcf_for_each_d4(traj))
 
-    @classmethod
-    def from_rpe_traj_voronoi(cls, traj: torch.Tensor) -> DcfData:
-        """Calculate dcf using voronoi approach for RPE trajectories.
+        ks = [traj.kx, traj.ky, traj.kz]
+        for i, k in enumerate(ks):
+            if any(all(k.shape[ax] == 1 for ax in two_axes) for two_axes in [(-1, -2), (-1, -3), (-2, -3)]):
+                # Found a direction with at least two singleton dimensions, i.e. we have a 2D trajectory in 3D space
+                # Remove this direction from the list of k-space points and calculate dcf for the remaining 2D trajectory
+                ks.pop(i)
+                new_traj = torch.stack(torch.broadcast_tensors(*ks), 1)
+                dcf = torch.stack([DcfData._dcf_using_voronoi(t) for t in new_traj])
+                dcf = dcf.expand(traj.broadcasted_shape)
+                break
+        else:
+            # Full 3D trajectory
+            dcf = torch.stack([DcfData._dcf_using_voronoi(t) for t in traj.as_tensor(1)])
 
-        For RPE the trajectoy is the same for each readout/frequency encoding position (k0), therefore it is
-        calculated in 2D for one k0 position and replicated. This is faster than calculating the trajectory in 3D.
-
-        Parameters
-        ----------
-        traj
-            torch.Tensor containing k-space points (d4, 3, k2, k1, k0).
-        """
-        if traj.shape[1] != 3:
-            raise ValueError('Trajectoy has to be 3D')
-
-        # Calculate trajectory in k1-k2 phase encoding plane for one frequency encoding "slice" and then copy to all
-        # other frequency encoding positions.
-        dcf = DcfData._voronoi_dcf_for_each_d4(traj[:, 1:, :, :, None, 0])
-        dcf = repeat(dcf, 'd4 dim k2 k1 k0->d4 dim k2 k1 (k0 nk0)', nk0=traj.shape[4])
         return cls(data=dcf)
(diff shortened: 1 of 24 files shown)


src/mrpro/data/_KTrajectory.py
```diff
@@ -0,0 +1,67 @@
...
+        Parameters
+        ----------
+        stack_dim:
+            The dimension to stack the tensor along.
+        """
+        shape = self.broadcasted_shape
+        return torch.stack([traj.expand(*shape) for traj in (self.kx, self.ky, self.kz)], dim=stack_dim)
+
+    @classmethod
+    def from_tensor(cls, tensor: torch.Tensor, stack_dim=0) -> KTrajectory:
```
schuenke: we should mention which order of kx, ky, kz is expected

src/mrpro/data/traj_calculators/_KTrajectoryCalculator.py
```diff
@@ -14,21 +14,20 @@
...
+from mrpro.data._KTrajectory import KTrajectory
 
 
-class KTrajectory(ABC):
+class KTrajectoryCalculator(ABC):
     """Base class for k-space trajectories."""
 
     @abstractmethod
-    def calc_traj(self, header: KHeader) -> torch.Tensor:
+    def __call__(self, header: KHeader) -> KTrajectory:
```
schuenke, line 30: I must confess that I find the callable variant kind of elegant 👍

ckolbPTB:
I like the split of the trajectory into kx ky kz - I think this really improves the readability.

If we want to stick with "KTrajectoryCalculator" then I would suggest to also rename "KTrajectoryCalculator" to "KTrajectoryCalculatorRpe".

I am not convinced by the broadcasting of the trajectory. This only applies to trajectories which haven been calculated with our current Calculators. If the trajectory is calculated via pulseq or read in via ISMRMRD then this won't work. For consistency I would suggest to stick with the current "full" trajectories.

schuenke (approved):
The implementation of KData.traj does not affect most of my personal use cases so that my opinion an this refactoring is not super strong. Anyway, I checked the code and everything seems to be correct and understandable. 

Maybe we can add some more infos about the (expected) dimensions. Not just for the new KTrajectory dataclass, but also for all the other data classes and methods/functions.

fzimmermann89:
The idea is to __allow__ broadcasting and use it when possible. All code should still work with 'full' trajectories, just not as efficient.
So if we use a raw ismrmrd trajectory (or something custom) as is, it would still work, just a bit slower in some cases.
Also, I'd say we try to use broadcasting also in our other readers. So for ismrmrd, we check if values are the same, and calculate the broadcasted version if they are. 
I'd rather in the beginning do some work (or have a RPE_ISMRMRD Trajectory calculator), then remember later on that I shall use a special _rpe function. 

I'd say, broadcasting is a pretty common  thing in the numpy/pytorch/etc world and makes it clear, that values are the same. 
This avoids a lot of np.unique calls or special cases that assume values are unique.

I prefer to have, for example, the DCF calculation check if dimensions are broadcasted and use a more efficient calculation, than having different functions for RPE etc (Similiar for the samefor the aquisions operators).
Also, this removes the need for a 2D trajectory, as it would just be a 3D trajectory with kz=zeros(1,1,1,1).

Next possible steps could be to:
- refactor the 2D-dimension in 3D-space logic from DCF into a utility function or as part of the KTrajectory to be able to use it in the acquisition operators.
-  keep the DCF also not expanded but broadcastable
- check the usage of k2,k1,k0 and kx,ky,kz to be consistent
- Decide if we want to add some reference to the calculator used into KTrajectory.

src/mrpro/data/_KTrajectory.py
fzimmermann89: Yes, I am also not 100% sure what order we expect currently.
Is it kx,ky,kz or kz,ky,kx in the 3 dimensions of Kdata.traj?

ckolbPTB:
> The idea is to **allow** broadcasting and use it when possible. All code should still work with 'full' trajectories, just not as efficient.
> So if we use a raw ismrmrd trajectory (or something custom) as is, it would still work, just a bit slower in some cases.

It is not just a bit slower. Even if it scaled linearly we are talking a factor of 100 or 200 for an RPE trajectory but it probably does not scale linearly because doing voronoi in 3D are much more complex than in 2D. So the only option would be to have `RPE_ISMRMRD Trajectory calculator` and `RPE_PULSEQ Trajectory calculator` which seems to take away all the advantages of having the trajectory information in the raw data file or in a pulseq file. For other trajectories like stack-of-stars it would even get more complicated because it can be the same but does not have to be, so you would need `stacks_of_stars_equal_kz_PULSEQ Trajectory calculator` and `stacks_of_stars_non_equal_kz_PULSEQ Trajectory calculator`.

fzimmermann89:
But isn't it a bit clearer to specify the trajectory ONCE and then have all the code use that information?
Instead of having to use dcf_rpe?
Your suggestion would still rely on the user knowing it's RPE and selecting all _rpe specific functions later on, right?
(And using the non_rpe functions would be a factor of 100 slower).

And the property the code use is nit RPE specific per-se - it is a repeated dimension.  So at least for me, it would be a lot clearer to check if this property is fulfilled then having the select 'rpe' in all functions.

Also, I'd argue we try to detect repeated values in the ismrmrd trajectory calculator without having to tell it it's RPE.

fzimmermann89:
Another Approach would be to to a bit more refactoring and  have, for example:
RPECalculator return an RPETrajectory(Child of Trajectory, Child of KZRepeatedInK0,Child of 3DTrajectory,...).
In all functions we do dispatching on the Trajectories
So the dcf_voronoi would check if the trajectory is a 'KZRepeatedInK0' (think of it as a trait/interface) and then do it's magic.

This is also something we can consider for other 'special cases'.
But for repeated values, I'd argue the most common technique in the python ecosystem is to signify it by broadcasting!

ckolbPTB:
> But for repeated values, I'd argue the most common technique in the python ecosystem is to signify it by broadcasting!

I agree - if we do this it should be via broadcasting and not via lots of different parent classes.

My main concern is that we end up with a whole zoo of TrajectoryCalculator functions but if we can detect repeated values outamatically (easy for ISMRMRD, might need to take rounding errors into account for pulseq) then it should be ok. 

I guess the main advantage would then be that we do not need different AcquisitionModels for Rpe, radial, stack-of-stars, 3D radial... but only one which automatically detects if there is an additional "Cartesian" dimension or not.

ckolbPTB:
>Now the RPE can have a trajectory with broadcasted kz dimension.
And the dcf voronoi logic can look if any of kx,ky,kz has two singleton dimensions. If this happens, then we can ignore this dimension in the calculation (I hope that's correct...)

If something can be broadcasted it not necessarily means that it is uniform. So it might need be taken into consideration during dcf calculation. A simple solution could be to calculate delta kx and use that as a dcf.

(RPE would have a broadcasted k**x** dimension)

fzimmermann89:
Maybe my refactor is currently wrong with regards to the naming of the dimensions. :D 

I assumed that the frequency-read-out (k0) dimension is in kz-direction. And this is the same for each 'spiral point' in kx,ky plane.  We should check all namings of the direction If we want to expliclty give them names. But this is wrong, isn't it?

If you can the following, I might be able to find most of the naming errors I made:
- So we have k2,k1,k0 as dimensions in of KData. And in cartesian encoding, these are aligned to kz,ky,kx-directions? I.e. kx changes along k0 and is the same for all k2,k1 etc?
- For RPE, the 'radial-like'  phase-enoded dimensions are in k2,k1 and in the kz,ky plane? and the frequency encoding is in kx?
(Of course, we could set up a Cartesian trajectory where in k0 the ktraj.ky is the value that changes, but this is not the common case)


Also, you are right, there is a bug in the DCF if the broadcasted dimension is not uniform -- this will most cause an issue for RPE has it is the frequency encoding dimension and most likely uniform. The code currently does pretty much the same logic your implementation I believe. But to be correct, after figuring out the 2D-dcf it should be multiplied by the removed dimension.

Example:
If (kx?) is of shape (other=1, k2,k1,k0=200)
We know that we are dealing with a stack of 2D trajectories in the k2,k1-plane at 200 different k0 points.
We can calculate the 2D dcf once, and scale it by the 1D-length of torch.diff(kx) (This is also fast)
This is what you meant, right?

The only issue I see is if we have  rotated trajectories where  the actual trajectory values are rotated. But I assume this will never happened, as it would also be a shitshow for cartesian. So rotations are always done by rotating the meaning of kx,ky,kz... right?

ckolbPTB:
> * So we have k2,k1,k0 as dimensions in of KData. And in cartesian encoding, these are aligned to kz,ky,kx-directions? I.e. kx changes along k0 and is the same for all k2,k1 etc?
> 
> * For RPE, the 'radial-like'  phase-enoded dimensions are in k2,k1 and in the kz,ky plane? and the frequency encoding is in kx?

yes

>(Of course, we could set up a Cartesian trajectory where in k0 the ktraj.ky is the value that changes, but this is not the common case)

no, because kx, ky and kz are defined as readout, phase encoding and slice encoding direction which are given by the directions of the gradients played out by the scanner. That is why we used k0,k1 and k2 for the data and the indices because e.g. for RPE k1,k2 and ky,kz do not correspond

> We can calculate the 2D dcf once, and scale it by the 1D-length of torch.diff(kx) (This is also fast)
This is what you meant, right?

I am not sure I fully understand it but we would need to scale each 2D dcf once by torch.diff(kx) of the "current" kx point. Nevertheless, for the time being we can also ignore this case because it only occurs for very special acquisitions or very special motion correction approaches. 

>The only issue I see is if we have rotated trajectories where the actual trajectory values are rotated. 

What do you mean with rotated trajectories?

fzimmermann89:
> > * So we have k2,k1,k0 as dimensions in of KData. And in cartesian encoding, these are aligned to kz,ky,kx-directions? I.e. kx changes along k0 and is the same for all k2,k1 etc?
> > * For RPE, the 'radial-like'  phase-enoded dimensions are in k2,k1 and in the kz,ky plane? and the frequency encoding is in kx?
> 
> yes
> 
> > (Of course, we could set up a Cartesian trajectory where in k0 the ktraj.ky is the value that changes, but this is not the common case)
> 
> no, because kx, ky and kz are defined as readout, phase encoding and slice encoding direction which are given by the directions of the gradients played out by the scanner. That is why we used k0,k1 and k2 for the data and the indices because e.g. for RPE k1,k2 and ky,kz do not correspond
> 
Ok, so far I think I can follow.

But we still have 'directions' (what I called kx,ky and kz) 
and indices, as in KData is of shape (... k2, k1,k0).

These sometimes align (Cartesian), but for radial, for example, k0 is along a spoke and k1 is 'which spoke' in the data 
(correct me if I am wrong)

So, shall we call the directions in KTrajectory (kz,ky,kx) or (k2,k1,k0) or (dir_phase2, dir_phase1, dir_frequency) ?
I would prefer not to use (k2,k1,k0) but different names, as they are only the same for cartesian.
 
And, for now assuming we go with kx,ky,kz, should it be:
`traj=KTrajectory(kz, ky, kx)`  (fits to the order of the axes in the data)
or `traj=KTrajectory(kx, ky, kz)`  (fits to the order in the alphabet)

> What do you mean with rotated trajectories?

I believe you said this will not happen by  your explanations.

fzimmermann89:
Also, as a suggestion for the ismrmrd/pulseq discussion, we could add the detection to KTrajectory.
So by default, we check if there dimensions which should be singleton because they are just all the same value

(As a draft, see latest commit)

ckolbPTB:
> So, shall we call the directions in KTrajectory (kz,ky,kx) or (k2,k1,k0) or (dir_phase2, dir_phase1, dir_frequency) ?
> I would prefer not to use (k2,k1,k0) but different names, as they are only the same for cartesian.

I would go for (kz,ky,kx) but just to clarify, we would then have something like KTrajectory.kx, KTrajectory.ky and KTrajectory.kz and for e.g. RPE the shape of kx would be (1,1,num_k0) and ky/kz would be (num_k2, num_k1, 1). Correct?

fzimmermann89:
Yes, exactly.

src/mrpro/data/_KTrajectory.py
fzimmermann89: kz,ky,kx

fzimmermann89:
Now it works with non-uniform broadcasted dimensions, even if points are repeated or they are not sorted.

As length of a segment in the 1D case I chose and somewhat documented a schema that should for interior points match voronoi and results in uniform dcf for uniform trajectory.

A nice example of the new logic is the comparison of the two 3d_cart test:
In the 'full' testcase, we disable the broadcasting and 3D voronoi is used.
In the other case, the trajectory is sparse/broadcasted and the dcf is calculated as the product of 3 1D DCFs.



Another discussion point would be where we store some utility functions, such as `smap` 

(`smap` is a more general version of @ckolbPTB's `apply-for-each-d4 logic` and a more stupid, serial, implementation of `torch.vmap`. `torch.vmap` only works with a limited set of pure pytorch functions and actually auto-vectorizes the function call, smap is just a loop, more like `numpy.vectorize`. But this stupid version is missing in pytorch..)

This might be of use in other parts of the code as well, so we might consider putting it in a general `mrpro.utils.tensortools` (??) namespace

fzimmermann89:
If we go with `(kz ky kx)` , I would suggest doing this order everything to be consistent. 
So we should name our variables like this if they (instead of, say,  k0, k1, k2)
and also have  `(ny, nx)` as the order for parameters in all functions (such as phantoms). etc. 

I am pretty sure I missed some uses of x, y or z in this PR --> A second pair of eyes would be welcome. :)

Also, @ckolbPTB , what do you now think about broadcasting and repeat detection?
Still against it / should I first implement the pulseq logic to convince you / would you be ok with the current draft being merged into your DCF branch?

fzimmermann89:
Sorry for missing the failing tests, I only looked at the green checkmark and assumed all tests passed. 
But I missed to commit one of the changes in the test files regarding the order of the dimensions.

src/mrpro/data/_DcfData.py
```diff
@@ -45,7 +37,38 @@ class DcfData:
...
+        """
+
+        traj_sorted, inverse, counts = torch.unique(
+            torch.round(traj, decimals=UNIQUE_ROUNDING_DECIMALS), sorted=True, return_inverse=True, return_counts=True
+        )
+
+        # For a sorted trajectory: x0 x1 x2 ... xN
+        # We assign the point at x0 the area (x1-x0) / 2 * 2
+        # We assign the point at x1 the area (x1-x0) / 2  + (x2-x1) / 2
+        # We assign the pint at xN the area (xN-XN-1) / 2 * 2
```
ckolbPTB: pint at xN the area (xN-XN-1) -> p**o**int at xN the area (xN-**x**N-1) (2 typos)

src/mrpro/data/_DcfData.py
```diff
@@ -45,7 +37,38 @@ class DcfData:
...
+        # This is done by central differences (-1,0,1). As the be complicated,
+        # we just append/prepend the correct values afterwards.
```
ckolbPTB: As the be complicated, we just append/prepend the correct values afterwards. -> Can be removed?

tests/data/test_dcf_data.py
```diff
@@ -16,25 +16,29 @@
...
-def example_traj_rad_2d(nkr, nka):
-    """Create 2D radial trajectory with uniform angular gap."""
+def example_traj_rpe(nkr, nka, nk0):
+    """Create RPE trajectory with uniform angular gap."""
     krad = torch.linspace(-nkr // 2, nkr // 2 - 1, nkr) / nkr
     kang = torch.linspace(0, nka - 1, nka) * (torch.pi / nka)
-    kx = torch.matmul(torch.cos(kang[:, None]), krad[None, :])[None, None, None, :, :]
-    ky = torch.matmul(torch.sin(kang[:, None]), krad[None, :])[None, None, None, :, :]
-    ktraj = torch.concatenate((kx, ky), dim=1)
+    kx = (torch.cos(kang[:, None]) * krad[None, :])[None, :, :, None]
```
ckolbPTB: for RPE kx should be linear, ky should be cos(angle) * krad and kz should be sin(angle) * krad

src/mrpro/data/traj_calculators/_KTrajectoryRpe.py
```diff
@@ -149,15 +160,17 @@ def calc_traj(self, kheader: KHeader) -> torch.Tensor:
...
+        kang = self._kang(kheader)
 
         # K-space locations along phase encoding lines
-        krad = (kheader.acq_info.idx.k1 - kheader.encoding_limits.k1.center).to(torch.float32)
-        krad = self._apply_shifts_between_rpe_lines(krad, kheader.acq_info.idx.k2)
-        krad *= 2 * torch.pi / kheader.encoding_limits.k1.max
+        krad = self._krad(kheader)
 
-        return self._combine_to_3d_traj(krad, kang, k0)
+        kz = (krad * torch.cos(kang))[..., None]
```
ckolbPTB: kz should be sin(kang) and ky should be cos(kang)

src/mrpro/phantoms/_EllipsePhantom.py
```diff
@@ -47,10 +47,10 @@ def kspace(self, kx: np.ndarray, ky: np.ndarray):
 
         Parameters
         ----------
-        kx
-            k-space locations in kx
         ky
-            k-space loations in ky. Same shape as kx.
+            k-space locations in ky
+        kx
+            k-space loations in kx (Frequency Encoding Direction). Same shape as ky.
```
ckolbPTB: frequency encoding direction - no capitals

src/mrpro/data/_DcfData.py
```diff
@@ -45,7 +37,38 @@ class DcfData:
...
+
+        kernel = torch.tensor([-1 / 2, 0, 1 / 2], dtype=torch.float32, device=traj.device).reshape(1, 1, 3)
+        central_diff = torch.nn.functional.conv1d(traj_sorted[None, None, :], kernel)[0, 0]
+        first = traj_sorted[1] - traj_sorted[0]
+        last = traj_sorted[-1] - traj_sorted[-2]
+        central_diff = torch.cat((first[None], central_diff, last[None]), -1)
+
+        # Repeated points are reduced by the number of repeats
+        dcf = torch.nan_to_num(1 / (central_diff * counts))[inverse]
```
ckolbPTB: Why 1/central_diff? Should it not be weighted by the area rather than 1/area?

tests/data/test_dcf_data.py
```diff
@@ -89,14 +112,7 @@ def test_dcf_rpe_traj_voronoi():
...
-    dcf_3d = DcfData.from_traj_voronoi(ktraj)
-
-    # Do not test outer points because they have to be approximated and cannot be calculated
-    # accurately using voronoi
-    dcf_rpe.data = dcf_rpe.data[:, :, :, 1:-1, 1:-1]
-    dcf_3d.data = dcf_3d.data[:, :, :, 1:-1, 1:-1]
-
-    # Compare normalized values
-    torch.testing.assert_close(dcf_rpe.data / torch.sum(dcf_rpe.data), dcf_3d.data / torch.sum(dcf_3d.data))
+    # TODO: Test against analytical solution
```
ckolbPTB: Not sure this is needed - the analytical solution is the same as for 2D radial so it won't add much more information.

src/mrpro/data/_KData.py
```diff
@@ -46,13 +47,13 @@
...
     data: torch.Tensor
-    traj: torch.Tensor
+    traj: KTrajectory
 
     @classmethod
     def from_file(
         cls,
         filename: str | Path,
-        ktrajectory: KTrajectory,
+        ktrajectory: KTrajectoryCalculator | KTrajectory,
```
ckolbPTB, line 56: Originally we had discussed that we always want to set the trajectory with a KTrajectoryCalculator to make sure (at least to a certain degree) that data and traj match. If we decide that Ktrajectory can be passed in that we need to have some checks regarding dimensions in line 147

ckolbPTB:
> Also, @ckolbPTB , what do you now think about broadcasting and repeat detection?

The current implementation is fine by me. Would be great if we can address pulseq during the hackathon.

ckolbPTB:
In https://github.com/ckolbPTB/mrpro/blob/478afd74923eda4dc55611a58960ab013560dbbb/src/mrpro/data/_SpatialDimension.py#L37 we currently have the convention (x, y, z). Shall we change it there too?

ckolbPTB:
> This might be of use in other parts of the code as well, so we might consider putting it in a general `mrpro.utils.tensortools` (??) namespace

yes

src/mrpro/data/_DcfData.py
fzimmermann89: ups. yes! good catch!

src/mrpro/data/_KHeader.py
```diff
@@ -31,7 +31,7 @@
 
 if TYPE_CHECKING:
     # avoid circular imports by importing onlz when type checking
```
fzimmermann89: typo

ckolbPTB:
Would it make sense to change `test_dcf_3d_cart_full_traj_voronoi` to use a irregular grid such that not everything is 1 and we compare 1D broadcasting to 3D voronoi?

fzimmermann89:
> Would it make sense to change `test_dcf_3d_cart_full_traj_voronoi` to use a irregular grid such that not everything is 1 and we compare 1D broadcasting to 3D voronoi?

I added a new test for this.

fzimmermann89:
> In
> 
> https://github.com/ckolbPTB/mrpro/blob/478afd74923eda4dc55611a58960ab013560dbbb/src/mrpro/data/_SpatialDimension.py#L37
> 
> we currently have the convention (x, y, z). Shall we change it there too?

I switched to order in SpatialDimension to follow the same dimension.
I renamed from_array to from_array_xyz to mark that this function reads arrays with the 'wrong' convention, as this is what we need when creating from an ismrmrd header.

fzimmermann89:
I would 

1) rebase+merge this into the DCF branch.

2) interactive rebase the DCF branch to split the changes into separate commits addressing
- the dimensions order convention and KTrajectory changes
- adding the DCF-functionality 
- fixing unrelated typos and other small issues.

I hope in the end to get somewhat logical commits into main.

PR #74, opened 2023-10-04
 Fix #73: Let test pipeline fail if tests fail

### LINKED ISSUE #73: Failing tests result in green checkmark
Not sure if feature or bug:

Failing pytest tests result in a green checkmark for the action.

At least for me, this was surprising, as it is different for the pre-commit hook checks.

Do we want to change it and have a red check mark if at least one test is failing?
Or do we want to leave the green checkmark to signify that the testing framework worked, even though we had test failures?

(Example https://github.com/ckolbPTB/mrpro/pull/70)

fzimmermann89:
Fixes #73

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -55,11 +55,10 @@ jobs:
           message: ${{ steps.coverageComment.outputs.coverage }}
           color: ${{ steps.coverageComment.outputs.color }}
           namedLogo: python
-      
+
       - name: Set pipeline status
         run: |
           if [[ ${{ steps.coverageComment.outputs.errors }} -ne 0 || ${{ steps.coverageComment.outputs.failures }} -ne 0 ]]; then
             echo "Errors or failures detected, marking pipeline as failure."
             exit 1
           fi
-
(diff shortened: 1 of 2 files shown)


fzimmermann89:
I added an additional step that fails if either test errors or test failures occurred.

Why we need this: 
 - We want to have the pipeline fail if a test fails
 - We still want to continue the pipeline if a test fails to create the comment
 - We use a pipeline in the test invocation to get the coverage
 
The other way to do it would be to
- set `pipefail` in the pipeline test, to mark this step as a failure if the pytest command failed in the pipeline
- add an` if: always()` to the commands that would run regardless of the test failure.

The issue with this approach would be, that the comment step etc would run even if the setup failed, not only if the test failed.

So I decided to treat test failures in the pytest step as success, and in the end, fail a dummy step if we have failures to mark the pipeline as a failure.

fzimmermann89:
Please squash before merging.

fzimmermann89:
@schuenke ping :)

schuenke (approved):
Useful change! Thanks 👍

PR #77, opened 2023-10-05
Change phantom to torch

### LINKED ISSUE #58: Pytorch functions for all phantoms
Currently phantoms uses numpy/scipy functions to create the phantom. 

TODO: replace functions with pytorch equivalents to make it compatible with (GPU) tensors and allow for differentiation

ckolbPTB:
Closes #58 

Change phantom and associated functions from numpy/scipy to torch

### DIFF
--- a/src/mrpro/phantoms/_EllipsePhantom.py
+++ b/src/mrpro/phantoms/_EllipsePhantom.py
@@ -13,7 +13,7 @@
 #   limitations under the License.
 
 import numpy as np
-import scipy.special as sp_special
+import torch
 
 from mrpro.phantoms.phantom_elements import EllipsePars
 
@@ -37,7 +37,7 @@ class EllipsePhantom:
     ):
         self.ellipses: list[EllipsePars] = ellipses
 
-    def kspace(self, ky: np.ndarray, kx: np.ndarray):
+    def kspace(self, ky: torch.Tensor, kx: torch.Tensor) -> torch.Tensor:
         """Create 2D analytic kspace data based on given k-space locations.
 
         For a corresponding image with 256 x 256 voxel, the k-space locations should be defined within [-128, 127]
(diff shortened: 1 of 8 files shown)


src/mrpro/phantoms/_EllipsePhantom.py
```diff
@@ -56,22 +56,22 @@ def kspace(self, ky: np.ndarray, kx: np.ndarray):
         if kx.shape != ky.shape:
             raise ValueError(f'shape mismatch between kx {kx.shape} and ky {ky.shape}')
 
-        kdat = np.zeros_like(kx).astype(np.complex64)
+        kdat = torch.zeros_like(kx).to(torch.complex64)
```
fzimmermann89: torch.zeros_like(kx, dtype=torch.complex64)

src/mrpro/phantoms/_EllipsePhantom.py
```diff
@@ -56,22 +56,22 @@ def kspace(self, ky: np.ndarray, kx: np.ndarray):
...
+            arg = torch.sqrt((el.radius_x * 2) ** 2 * kx**2 + (el.radius_y * 2) ** 2 * ky**2)
             arg[arg < 1e-6] = 1e-6  # avoid zeros
 
-            cdat = 2 * 2 * el.radius_x * el.radius_y * 0.5 * sp_special.jv(1, np.pi * arg) / arg
-            kdat += (np.exp(-1j * 2 * np.pi * (el.center_x * kx + el.center_y * ky)) * cdat * el.intensity).astype(
-                np.complex64
+            cdat = 2 * 2 * el.radius_x * el.radius_y * 0.5 * torch.special.bessel_j1(torch.pi * arg) / arg
+            kdat += (torch.exp(-1j * 2 * torch.pi * (el.center_x * kx + el.center_y * ky)) * cdat * el.intensity).to(
+                torch.complex64
             )
```
fzimmermann89: .to(torch.complex64) not needed, torch defaults to float32/complex64 and does not promote, i.e.
`kdat += torch.exp(-1j * 2 * torch.pi * (el.center_x * kx + el.center_y * ky)) * cdat * el.intensity` is fine

tests/data/_Dicom2DTestImage.py
```diff
@@ -41,7 +42,7 @@ def __init__(self, filename: str | Path, matrix_size: int = 256, phantom: Ellips
 
         # Create image
         self.imref = self.phantom.image_space(matrix_size, matrix_size)
-        self.imref = np.abs(self.imref * 2**16).astype(np.uint16)
+        self.imref = torch.abs(self.imref * 2**16).to(torch.int16)
```
fzimmermann89: this can be problematic:
`1.0*2**16` can not be stored in torch.int16, as int16 goes only to `2**15`
suggest to use torch.int32 as replacement for uint16

tests/data/_IsmrmrdRawTestData.py
```diff
@@ -102,7 +103,7 @@ def __init__(
         [kx, ky] = np.meshgrid(kx_idx, ky_idx)
 
         # Create analytic k-space and reference image
-        ktrue = self.phantom.kspace(ky, kx)
+        ktrue = self.phantom.kspace(torch.Tensor(ky), torch.Tensor(kx)).numpy()
```
fzimmermann89: we _could_ consider keeing this as a tensor and 
l111: `ktrue=einops.repeat(ktrue, '... -> coils ... ', coils=self.ncoils) `
l210/l222: torch's random generator
and in the end cast to numpy before writing to the file

This would be a bit more consistent with respect to numpy vs. torch functions. 
New contributes only have to learn one style and one library to understand all code.

But as this is inside a test, I'd say potato/potato.

tests/data/test_idata.py
```diff
@@ -33,7 +33,8 @@ def dcm_2d(ph_ellipse, tmp_path_factory):
 def test_IData_from_dcm_file(dcm_2d):
     """IData from single dicom file."""
     idat = IData.from_single_dicom(dcm_2d.filename)
-    np.testing.assert_almost_equal(np.abs(idat.data[0, 0, 0, ...]), np.moveaxis(dcm_2d.imref, (0, 1), (1, 0)))
+    im = torch.abs(idat.data[0, 0, 0, ...])
+    torch.testing.assert_close(im, torch.movedim(dcm_2d.imref.to(im.dtype), (0, 1), (1, 0)))
```
fzimmermann89: why are we moving the dim here? 
wouldn't it make sense to have dcm_2d.imref already in the order of axis that we expect?
Also, the dtype should already be the correct one.

Otherwise, if idat.data would be float16 or int8, we would cast our reference to the wrong type and the test would pass, even though should expect float32 here.

I'd make this test more strict by having imref already in the type we expect and then make use that assert_close also checks if the dtype matches our expectation.

fzimmermann89 (commented):
Small suggestions

tests/data/_IsmrmrdRawTestData.py
ckolbPTB: I agree, we should use torch functions as much as possible

tests/data/_Dicom2DTestImage.py
ckolbPTB: I would suggest to keep it complex64 as long as possible and only convert when writing to dicom-file

fzimmermann89: 👍 
sounds reasonable, as our IData.data is also supposed to be complex, so the dicom reader should return complex64 and we should match that with the ground truth imref (see my comment below)

tests/data/test_idata.py
ckolbPTB: True - I hope I managed to create the dicom image now in such a way that (a) it corresponds to the dicom standard and (b) is consistent with our (z)-y-x convention.

fzimmermann89 (approved):

PR #76, opened 2023-10-05
Refactor ktraj to a dataclass with kz,ky,kx & add DCF calculation

fzimmermann89:
Closes https://github.com/ckolbPTB/mrpro/issues/17 
Closes https://github.com/ckolbPTB/mrpro/issues/7

Closes #69

### We move from Kdata.traj as a tensor with x,y,z stacked in dimension 1 to a dataclass with kz,ky,kx.
We also change the order to z,y,x to match k2,k1,k0 in the KData.
We renamed the old KTrajectory -> KTrajectoryCalculato and made them callable.

#### Example: Cartesian 2D
in k0 is the length in frequency encoding direction. k1 is in the phase encoding dimension. (k2 is 1 as it is 2D, other is 1)
KData.data has shape `(other,coils,k2,k1,k0) = (1, coils,1,k1,k0)`
KData.traj has a broadcasted shape` (1,1,k1,k0) `
KData.traj.kx has shape `(1,1,1,k0)`
KData.traj.ky has shape `(1,1,k1,1)`
KData.traj.kz is `zeros(1,1,1,1)`

Also, we move the order for SpatialDimension, Phantoms, etc to always be (z),y,x.

### We add density compensation calculation:
The DCF function uses scipy.spatials wrapper around [qhull ](http://www.qhull.org/) to to a voronoi tesselation of the trajectory.
If one of the KTrajectory dimensions is broadcasted, we can use simpler 1D logic in this dimension and the final dcf is the product of the DCFs

#### Example: RPE
Along the frequency encoding direction, the points are always the same for all kz,ky points and uniform. Thus, we onlyneed to do a single 2D voronoi calculation.

### We moved some utility functions to a new mrpro.utils module. 
The structure of this module might still need some adjustments and reorganization. The idea is, to collect reusable, generic, small utility functions used in our code base somwhere.

PR #78, opened 2023-10-05
Pulseq Trajectory Calculator

### LINKED ISSUE #72: Pulseq Trajectory
As I haven't found an issue for this:

We plan to have a  seq-File+ismrmrd->KData path.

This would include using the .seq to calculator the trajectory (main task)
And maybe filling in some header information into misc.

@Stef-Martin  is currently looking into it.

fzimmermann89:
This is @Stef-Martin's initial draft for the pypulseq trajectory calculator and a rough sketch of a test.

The calculator is in the 'before the refactor of #76'-style and a good starting point.

This is a TODO for the Hackathon: 
port it the the current style (see TODO comments in the code)

Closes #72

### DIFF
--- a/src/mrpro/data/traj_calculators/_KTrajectoryPulseq.py
+++ b/src/mrpro/data/traj_calculators/_KTrajectoryPulseq.py
@@ -1,0 +1,89 @@
+"""K-space trajectory from pulseq seq file."""
+
+# Copyright 2023 Physikalisch-Technische Bundesanstalt
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#       http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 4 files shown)


fzimmermann89:
Another task (step0) will be to figure out a pypulseq version that works and can be installed via a requirement. :)

The next steps could be:

step1: add pypulseq to the pyproject.toml (This is the reason the tests currently do not even execute, as the import of pypulseq fails)

Step2: make the test run with a valid file but fail. Here it might be a good idea to change to test to use a valid header.

Step3: make the test succeed by implementing the calculator

Step4: test for edge cases, such as broken or missing seq file. This should already fail on init of the calculator.

As pypulseq upstream seems to break rather frequently, we could consider making pypulseq and optional requirement: We try to import it, if we fail on import, this calculator will not be available but the rest of the code works.

ckolbPTB:
This PR has been superseded by #111

Issue #64, opened 2023-09-14
Csm inati method


ckolbPTB:
could be implemented via a wrapper for https://github.com/ismrmrd/ismrmrd-python-tools/blob/17fe6fbf1bb645112e2ad0023eb4f42afb36421e/ismrmrdtools/coils.py#L92

fzimmermann89:
I believe the 'iterative' version produces different results then the one we currently use in ptbpyrecon.

- We should check it this is significant.
- Decide if we want to torchify the version (to be able to run it on a GPU for example) or keep it as numpy. 

I'd slightly prefer copying it over and torchifying over wrapping.

fzimmermann89:
for the iterative method, see 
https://gist.github.com/fzimmermann89/7e6a1a34cdfbd27903056fff6f717f1a
(untested)

I will try to make it reproduce the results as close as possible to those of our internal ptbpyrecon

fzimmermann89:
#161

fzimmermann89:
Here it would be nice to test it vs. ptbpyrecon...

Issue #34, opened 2023-08-04
Espirit CSM Calculation


fzimmermann89:
For Reference:

Pytorch :
https://github.com/NKI-AI/direct/blob/2a4c29342bc52a404aae097bc2654fb4323e1ac8/direct/algorithms/mri_algorithms.py#L18
 
Which is based on:
https://github.com/mikgroup/sigpy/blob/1817ff849d34d7cbbbcb503a1b310e7d8f95c242/sigpy/mri/app.py#L388-L491

PR #79, opened 2023-10-06
Sim coil sensitivities

ckolbPTB:
This numerical simulation of a birdcage coil is taken from: https://github.com/ismrmrd/ismrmrd-python-tools/blob/17fe6fbf1bb645112e2ad0023eb4f42afb36421e/ismrmrdtools/simulation.py#L53

We can then use this simulation to generate test cases for the csm-methods of #64, #34 and #5

### DIFF
--- a/src/mrpro/phantoms/coils.py
+++ b/src/mrpro/phantoms/coils.py
@@ -1,0 +1,100 @@
+"""Numerical coil simulations."""
+
+# Copyright 2023 Physikalisch-Technische Bundesanstalt
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#       http://www.apache.org/licenses/LICENSE-2.0
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+import numpy as np
+import torch
+from einops import repeat
+
+from mrpro.data import SpatialDimension
+
+
+def birdcage_2d(
+    number_of_coils: int,
+    image_dimensions: SpatialDimension[int],
+    relative_radius: float = 1.5,
+    normalize_with_rss: bool = True,
+) -> torch.Tensor:
+    """Numerical simulation of 2D Birdcage coil sensitivities.
+
+    Parameters
+    ----------
+        number_of_coils
+            Number of coil elements
+        image_dimensions
+            Number of voxels in the image.
+            This is a 2D simulation so the output will be (1 number_of_coils 1 image_dimensions.y image_dimensions.x)
+        relative_radius
+            Relative radius of birdcage
+        normalize_with_rss
+            If set to true, the calculated sensitivities are normalized by the root-sum-of-squares
+
+    This function is strongly inspired by https://github.com/ismrmrd/ismrmrd-python-tools. The associated license
+    information can be found at the end of this file.
+    """
+
+    dim = [number_of_coils, image_dimensions.y, image_dimensions.x]
+    x_co, y_co = torch.meshgrid(
+        torch.linspace(-dim[2] // 2, dim[2] // 2 - 1, dim[2]),
+        torch.linspace(-dim[1] // 2, dim[1] // 2 - 1, dim[1]),
+        indexing='xy',
+    )
+
+    c = torch.linspace(0, dim[0] - 1, dim[0])[:, None, None]
+    coil_center_x = dim[2] * relative_radius * np.cos(c * (2 * torch.pi / dim[0]))
+    coil_center_y = dim[1] * relative_radius * np.sin(c * (2 * torch.pi / dim[0]))
+    coil_phase = -c * (2 * torch.pi / dim[0])
+
+    rr = torch.sqrt((x_co[None, ...] - coil_center_x) ** 2 + (y_co[None, ...] - coil_center_y) ** 2)
+    phi = torch.arctan2((x_co[None, ...] - coil_center_x), -(y_co[None, ...] - coil_center_y)) + coil_phase
+    sensitivities = (1 / rr) * np.exp(1j * phi)
+
+    if normalize_with_rss:
+        rss = torch.sqrt(torch.sum(torch.abs(sensitivities) ** 2, 0))
+        # Normalize only where rss is > 0
+        sensitivities[:, rss > 0] /= rss[None, rss > 0]
+
+    return repeat(sensitivities, 'coil y x->other coil z y x', other=1, z=1)
+
+
+# License information from https://github.com/ismrmrd/ismrmrd-python-tools
+
+# ISMRMRD-PYTHON-TOOLS SOFTWARE LICENSE JULY 2016
+
+# PERMISSION IS HEREBY GRANTED, FREE OF CHARGE, TO ANY PERSON OBTAINING
+# A COPY OF THIS SOFTWARE AND ASSOCIATED DOCUMENTATION FILES (THE
+# "SOFTWARE"), TO DEAL IN THE SOFTWARE WITHOUT RESTRICTION, INCLUDING
+# WITHOUT LIMITATION THE RIGHTS TO USE, COPY, MODIFY, MERGE, PUBLISH,
+# DISTRIBUTE, SUBLICENSE, AND/OR SELL COPIES OF THE SOFTWARE, AND TO
+# PERMIT PERSONS TO WHOM THE SOFTWARE IS FURNISHED TO DO SO, SUBJECT TO
+# THE FOLLOWING CONDITIONS:
+
+# THE ABOVE COPYRIGHT NOTICE, THIS PERMISSION NOTICE, AND THE LIMITATION
+# OF LIABILITY BELOW SHALL BE INCLUDED IN ALL COPIES OR REDISTRIBUTIONS
+# OF SUBSTANTIAL PORTIONS OF THE SOFTWARE.
+
+# SOFTWARE IS BEING DEVELOPED IN PART AT THE NATIONAL HEART, LUNG, AND BLOOD
+# INSTITUTE, NATIONAL INSTITUTES OF HEALTH BY AN EMPLOYEE OF THE FEDERAL
+# GOVERNMENT IN THE COURSE OF HIS OFFICIAL DUTIES. PURSUANT TO TITLE 17,
+# SECTION 105 OF THE UNITED STATES CODE, THIS SOFTWARE IS NOT SUBJECT TO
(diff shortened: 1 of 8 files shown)


src/mrpro/phantoms/_CoilSensitivities.py
```diff
@@ -0,0 +1,110 @@
...
+        ----------
+            relative_radius: Relative radius of birdcage
+
+        This function is strongly inspired by https://github.com/ismrmrd/ismrmrd-python-tools. The associated license
+        information can be found at the end of this file.
+        """
+
+        dim = [self.number_of_coils, self.image_dimensions.y, self.image_dimensions.x]
+        sensitivities = torch.zeros(dim, dtype=torch.complex64)
+        y_co, x_co = torch.meshgrid(
```
fzimmermann89: x_co and y_co are inthe wrong order IMHO:

```
dim = [1, 5, 6]  # c y x
y_co, x_co = torch.meshgrid(
    torch.linspace(-dim[2] // 2, dim[2] // 2 - 1, dim[2]), #x
    torch.linspace(-dim[1] // 2, dim[1] // 2 - 1, dim[1]), #y
    indexing='xy',
)
print(y_co.shape) # torch.Size([5, 6])
print(y_co[:,0])  # tensor([-3., -3., -3., -3., -3.])
print(y_co[0,:])  # tensor([-3., -2., -1.,  0.,  1.,  2.])
```
-> y_co currently changes along the last dim, which would be x in our naming convention

src/mrpro/phantoms/_CoilSensitivities.py
```diff
@@ -0,0 +1,110 @@
...
+        image_dimensions
+            Number of voxels in the image
+        normalize_with_rss
+            If set to true, the calculated sensitivities are normalized by the root-sum-of-squares
+    """
+
+    def __init__(
+        self,
+        number_of_coils: int,
+        image_dimensions: SpatialDimension[int],
```
fzimmermann89: If we take a spatialdimension for the dimension, I would prefer that the return value is always of shape
(1,ncoils,nz.ny.nx), or broadcastable to this. It is surprising, that even though we can supply an z dimension, it will just be ignored and not match the expected shape.

So instead for a 2D birdcage, we would always return (1,ncoils,1, ny.nx).

src/mrpro/phantoms/_CoilSensitivities.py
```diff
@@ -0,0 +1,110 @@
...
+
+        dim = [self.number_of_coils, self.image_dimensions.y, self.image_dimensions.x]
+        sensitivities = torch.zeros(dim, dtype=torch.complex64)
+        y_co, x_co = torch.meshgrid(
+            torch.linspace(-dim[2] // 2, dim[2] // 2 - 1, dim[2]),
+            torch.linspace(-dim[1] // 2, dim[1] // 2 - 1, dim[1]),
+            indexing='xy',
+        )
+
+        for c in range(0, dim[0]):
```
fzimmermann89: This could maybe also be vectorized over the coils without the loop?

fzimmermann89:
I am not sure if this should be a class...
It just looks like a function? What is the advantage over having a birdcage function?

Why are image dimensions init parameters?

This might also be a question of preference, if something can be a pure function, I prefer to have it as a function.
So you can call coilsensitivities.birdcage(dim, width) instead of Coilsensititive(dim).birdcage(width).
In my head, objects are things with state, or "things", not collections of functions. this is what modules / .py files are for.

The only reason I could so to have it as a class, would be to have a CoilSensitivies(QData) class(as we plan to do)  and with a class method simulate_birdcage() that returns a simulated cls instance.


Remark for the future: We should at some point add random (within a given range) diameter, random phase (currently, the phase increases linearly with the coil number), and random orientation to be able to use this for NN training.

Also, there might be a bug with the meshgrid., see inline comment

ckolbPTB:
> I am not sure if this should be a class...
> It just looks like a function? What is the advantage over having a birdcage function?

Good point. It can also be a function.

> The only reason I could so to have it as a class, would be to have a CoilSensitivies(QData) class(as we plan to do) and with a class method simulate_birdcage() that returns a simulated cls instance.

I think I prefer keeping that separate. We will have a CsmData(QData) class but I would prefer to use something like
```
csm_sim = coils.birdcage()
csm = CsmData.from_tensor_and_header(csm_sim, header)
```
rather than
`csm = CsmData.simulate_birdcage()`
because this would suggest that we also have 
`IData.simulate_ellipse()` and `KData.simulate_ellipse()`

src/mrpro/phantoms/_CoilSensitivities.py
ckolbPTB: Agreed. Would it make more sense to have
1) image_dimensions: SpatialDimension[int] and return (1,ncoils,1, ny, nx)
or
2) have nx, ny as input (as we do for EllipsePhantom) and return (ncoils, ny, nx)?

fzimmermann89:
>I think I prefer keeping that separate.

I concur 👍

Should we (once we have CsmData implemented) change it, such that the function returns CsmData instead of a tensor, i.e.
 ```
csm:CsmData = phantoms.coils.birdcage(header, **othersettings)
csm_sim:Tensor = csm.data #if we need the tensor
 ```
and similar for the other phantoms?

Or will all of our phantom / simulation data return "naked" tensors that we then have to wrap?
 (Sorry if  we already have discussed and decided this when merging the ElipsePhantom)
I think I would be fine with either, if we make it a "rule"

Also, we should be consistent with regard to the SpatialDimension/(ny,nx) in the phantoms module.
Once again, I would be fine with either, as I see the advantage of having the phantoms a bit more "naked" such that they can be easily used outside our packages as well as the advantage of defining x,y properly.

What about accepting a SpatialDimension or a Tuple with order z,y,x ? Is this "too flexible" for our taste?

ckolbPTB:
> Or will all of our phantom / simulation data return "naked" tensors that we then have to wrap?
> (Sorry if we already have discussed and decided this when merging the ElipsePhantom)
> I think I would be fine with either, if we make it a "rule"

I would prefer to keep it as simple tensors. If we pass a header to the simulation functions then it is not immediately obvious to the user which parameters of the header are used and need to be changed in order to modify the simulation. 

> What about accepting a SpatialDimension or a Tuple with order z,y,x ? Is this "too flexible" for our taste?

I see the problem here that we introducted SpatialDimension in order to avoid having to "remember" in which order we provide the dimensions. Allowing now a tuple with order z,y,x seems like taking a step back. I would prefer to either use `birdcage_2d(ncoils, idim: SpatialDimension)` or `birdcage_2d(ncoils, ny, nx)`. 

I am happy with both - one small advantage of the second option is to me that it is clearer to the user what is used. In the first case should we throw an error if idim[2] > 1?

src/mrpro/phantoms/_CoilSensitivities.py
fzimmermann89: 3. have  `ncoils=1,nz=1, ny=1, nx=1` as input and return (1, ncoils, 1,ny,nx) ?
this way, all coil sims take the same signature, regardless of 2d/3d?

4.  have nx,ny,nz keyword-only arguments
```
def birdcage(*, nz=1, ny, nx, **othersettings): #
    print(nz, ny, nx)
birdcage(nx=1, ny=2, nz=3)  #3 2 1
birdcage(nx=3, ny=2) # 1 2 3
birdcage(nz=2, nx=1) # invalid, as ny is missing
birdcage(1,2,3) # confusing, what is the order. but invalid as nz,ny,nx are keyword-only arguments
```

Disadvantage of 2.):
we would need to have some special logic in` CSMData.__init__(data:tensor, header:QHeader)`  to be able to to` CSMData(csm_sim,header=header)` . 
Or we would have to do ` CSMData(csm_sim[None,:,None],header=header)` (or similar).

So I'd prefer to output (1, ncoils, 1,ny,nx). 

Not sure between 1 and 3 and if we should have an nz input for a 2D simulation.
Advantage: Same signature for different simulations
Disadvantage: Might be confusing why there is a nz if it is not used, and why birdcage(10,2,3) returns (ncoils=10,nz=1,ny=3,nx=1)

If we do nx,ny,nz as separate inputs, we might consider making them keyword-only arguments.
Advantage: order is clear
Disadvantage: more to write whenever we call it


Really not sure... 

(Either way, I assume we would at some point switch the EllipsePhantom to the same logic)

fzimmermann89:
> In the first case should we throw an error if idim[2] > 1?

I would always return (1, ncoils, nz=1,ny,nx) as this is broadcastable to the requested shape without throwing an error.

ckolbPTB:
Should we also change `EllipsePhantom.image_space(ny, nx)` to `EllipsePhantom.image_space(SpatialDimension)`?

fzimmermann89:
> Should we also change `EllipsePhantom.image_space(ny, nx)` to `EllipsePhantom.image_space(SpatialDimension)`?


Yes, good idea.
If you want to, you could include this as a second commit into this PR (i.e. squash the coil-related into one commit and the ellipse-related changes into one commit) or as a small second PR, whatever is easier for you.
(Or leave this as a todo if you don't want to do it yourself)

src/mrpro/phantoms/coils.py
```diff
@@ -0,0 +1,101 @@
...
+            Relative radius of birdcage
+        normalize_with_rss
+            If set to true, the calculated sensitivities are normalized by the root-sum-of-squares
+
+    This function is strongly inspired by https://github.com/ismrmrd/ismrmrd-python-tools. The associated license
+    information can be found at the end of this file.
+    """
+
+    dim = [number_of_coils, image_dimensions.y, image_dimensions.x]
+    sensitivities = torch.zeros(dim, dtype=torch.complex64)
```
fzimmermann89: not required, will be overwritten in l61

src/mrpro/phantoms/coils.py
```diff
@@ -0,0 +1,101 @@
...
+    rr = torch.sqrt((x_co[None, ...] - coil_center_x) ** 2 + (y_co[None, ...] - coil_center_y) ** 2)
+    phi = torch.arctan2((x_co[None, ...] - coil_center_x), -(y_co[None, ...] - coil_center_y)) + coil_phase
+    sensitivities = (1 / rr) * np.exp(1j * phi)
+
+    if normalize_with_rss:
+        rss = torch.sqrt(torch.sum(torch.abs(sensitivities) ** 2, 0))
+        # Normalize only where rss is > 0
+        sensitivities[:, rss > 0] /= rss[None, rss > 0]
+
+    return rearrange(sensitivities, '(other coil z) y x->other coil z y x', other=1, z=1)
```
fzimmermann89: could be
 `einops.repeat(sensitivities, 'coil y x->other coil z y x', other=1, z=1)`
  as I recently learned and prefer.
 
  I think I used rearrange like you did also in a couple of places and will at some point replace them with repeats, so  doesn't matter right now, but maybe for the future

fzimmermann89 (approved):
2 tiny changes that are not really important

ckolbPTB:
@fzimmermann89 could you please have another quick check I did not mess up anything with the SpatialDimensions? Thanks!

fzimmermann89:
lgtm

PR #40, opened 2023-08-15
QData and QHeader implementation

schote:
closes #3

PR #83, opened 2023-10-07
CsmData with Walsh-Method implementation

### LINKED ISSUE #16: CsmData class
```
class CsmData(IData):
    CsmData.header (based on ismrmrd) -> check in functions if certain parameters available
    CsmData.data: torch.Tensor 5D: [everything_else, coils, z, y, x]
```
       
Check list:
- [ ]     Create branch
- [ ]     Implement code
- [ ]     Add comments
- [ ]     Implement tests
- [ ]     Squash commmits
- [ ]     Create PR

ckolbPTB:
Closes #16 and #5

ToDo:

- Derive CsmData from QData once it is available
- Replace for-loop https://github.com/ckolbPTB/mrpro/blob/2ade47482d5be0e0622ad824b1550032d7dbca1b/src/mrpro/data/_CsmData.py#L111 with smap?
- Replace for-loops in _iterative_walsh_csm for speed-up

### DIFF
--- a/src/mrpro/data/_CsmData.py
+++ b/src/mrpro/data/_CsmData.py
@@ -14,8 +14,6 @@
 
 from __future__ import annotations
 
-import dataclasses
-
 import torch
 
 from mrpro.data import IData
@@ -24,7 +22,6 @@ from mrpro.data import SpatialDimension
 from mrpro.utils.filters import spatial_uniform_filter_3d
 
 
-@dataclasses.dataclass(init=False, slots=True, frozen=True)
 class CsmData(QData):
     """Coil sensitivity map class."""
 
@@ -66,6 +63,9 @@ class CsmData(QData):
         # nan_to_num does not work for complexfloat, boolean indexing not with vmap.
         csm_data = torch.where(torch.isfinite(csm_data), csm_data, 0.0)
         return csm_data
+
+    def test(self):
+        self.aaaa = 1
 
     @classmethod
     def from_idata_walsh(
--- a/src/mrpro/data/_CsmData.py
+++ b/src/mrpro/data/_CsmData.py
@@ -51,9 +51,6 @@ class CsmData(IData):
         coil_cov = torch.einsum('azyx,bzyx->abzyx', coil_images, coil_images.conj())
 
         # Smooth the covariance along y-x for 2D and z-y-x for 3D data
-        if coil_images.shape[-3] == 1:
-            # 2D case
-            smoothing_width = SpatialDimension(1, smoothing_width.y, smoothing_width.x)
         coil_cov = spatial_uniform_filter_3d(coil_cov, filter_width=smoothing_width)
 
         # At each point in the image, find the dominant eigenvector
@@ -92,7 +89,8 @@ class CsmData(IData):
             Default is None, which means that all elements are processed at once.
         """
         csm_fun = torch.vmap(
-            lambda img: CsmData._iterative_walsh_csm(img, smoothing_width, niter), chunk_size=chunk_size_otherdim
+            lambda img: CsmData._iterative_walsh_csm(img, smoothing_width, niter),
+            chunk_size=chunk_size_otherdim,
         )
         csm_data = csm_fun(idata.data)
 
--- a/src/mrpro/data/_CsmData.py
+++ b/src/mrpro/data/_CsmData.py
@@ -64,9 +64,6 @@ class CsmData(QData):
         csm_data = torch.where(torch.isfinite(csm_data), csm_data, 0.0)
         return csm_data
 
-    def test(self):
-        self.aaaa = 1
-
     @classmethod
     def from_idata_walsh(
         cls,
--- a/src/mrpro/data/_CsmData.py
+++ b/src/mrpro/data/_CsmData.py
@@ -48,15 +48,8 @@ class CsmData(IData):
         niter
             Number if iterations of Walsh method.
         """
-        ncoils, nz, ny, nx = coil_images.shape
-
-        # Compute the sample covariance pointwise
-        coil_cov = torch.zeros((ncoils, ncoils, nz, ny, nx), dtype=coil_images.dtype)
-        for c1_idx in range(ncoils):
-            for c2_idx in range(c1_idx):
-                coil_cov[c1_idx, c2_idx, ...] = coil_images[c1_idx, ...] * torch.conj(coil_images[c2_idx, ...])
-                coil_cov[c2_idx, c1_idx, ...] = torch.conj(coil_cov[c1_idx, c2_idx, ...].clone())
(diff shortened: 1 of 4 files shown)


fzimmermann89:
Nice starting point 👍

Replacing the inner for loops would imho also improve readability a lot.


...we might also be able to replace the batch/other-for-loop by torch.vmap.
This would allow us to write the inner function for just one sample and pytorch auto-vectorizes for the batch. If I understand the logic correctly, there is nothing that would not work with camp.

src/mrpro/data/_CsmData.py
```diff
@@ -0,0 +1,147 @@
...
+
+        # Make sure there are no inf and nan-values due to very small values in the covariance matrix
+        csm_data[torch.isnan(csm_data) | torch.isinf(csm_data)] = 0
+        return csm_data
+
+    @classmethod
+    def from_idata_walsh(
+        cls,
+        idata: IData,
+        downsampling_factor: SpatialDimension[int] = SpatialDimension(1, 1, 1),
```
fzimmermann89: https://docs.python-guide.org/writing/gotchas/#default-args

could this be an issue?

src/mrpro/data/_CsmData.py
```diff
@@ -0,0 +1,147 @@
...
+        coil_cov = torch.zeros((ncoils, ncoils, nz, ny, nx), dtype=coil_images.dtype)
+        for c1_idx in range(ncoils):
+            for c2_idx in range(c1_idx):
+                coil_cov[c1_idx, c2_idx, ...] = coil_images[c1_idx, ...] * torch.conj(coil_images[c2_idx, ...])
+                coil_cov[c2_idx, c1_idx, ...] = torch.conj(coil_cov[c1_idx, c2_idx, ...].clone())
+            coil_cov[c1_idx, c1_idx, ...] = coil_images[c1_idx, ...] * torch.conj(coil_images[c1_idx, ...])
+
+        # Smooth the covariance along y-x for 2D and z-y-x for 3D data
+        if nz == 1:
+            smoothing_width.z = 1
```
fzimmermann89: We should not change the input.

If we call the function twice with the same smoothing_width spatialdimension object, once with 2d data, once with 3d data, this will result in a bug.

Imho this also applies to the default argument.

src/mrpro/data/_CsmData.py
```diff
@@ -0,0 +1,147 @@
...
+        coil_cov = spatial_uniform_filter_3d(coil_cov, filter_width=smoothing_width)
+
+        # At each point in the image, find the dominant eigenvector
+        # and corresponding eigenvalue of the signal covariance
+        # matrix using the power method
+        csm_data = torch.zeros((ncoils, nz, ny, nx), dtype=coil_images.dtype)
```
fzimmermann89: Should also match the input device?

src/mrpro/data/_CsmData.py
```diff
@@ -0,0 +1,147 @@
...
+
+                        for _ in range(niter):
+                            v = torch.mm(coil_cov_curr_vox, v)
+                            lam = torch.linalg.norm(v)
+                            v = v / lam
+
+                    csm_data[:, z, y, x] = v[:, 0]
+
+        # Make sure there are no inf and nan-values due to very small values in the covariance matrix
+        csm_data[torch.isnan(csm_data) | torch.isinf(csm_data)] = 0
```
fzimmermann89: torch.nan_to_num ?

src/mrpro/data/_CsmData.py
```diff
@@ -0,0 +1,147 @@
...
+            Images for each coil element.
+        smoothing_width
+            Width smoothing filter.
+        niter
+            Number if iterations of Walsh method.
+        """
+        ncoils, nz, ny, nx = coil_images.shape
+
+        # Compute the sample covariance pointwise
+        coil_cov = torch.zeros((ncoils, ncoils, nz, ny, nx), dtype=coil_images.dtype)
```
fzimmermann89: I think this could be 

```suggestion
        coil_cov=torch.einsum('azyx,bzyx->abzyx', coil_image, coil_image.conj())
```
 or similar..

src/mrpro/data/_CsmData.py
fzimmermann89: correction: torch.nan_to_num  does not work for complex values.

fzimmermann89:
I think this code still results in the same output, but is vectorized and quite a bit shorter.
Feel free to undo the changes or force-push over them if you have a different direction in mind. 

We should think a bit about SpatialDimension as default parameters.
Either we have a FixedSpatialDimension, a SpatialDimension(1,1,1, mutable=False) option of we should avoid using it as a default value.
Or am I mistaken and this is fine? I wonder why non of the linters notices that issue..

fzimmermann89:
..or we could, instead of vmap-ing along the other-dimension and einsum-vectiorzing along the spatial dimensions, split the logic into three steps:

- Covariance (vmapped over all but the coil dimension) inner function: ncoils->ncoils-,ncoils

- filtering (batched  over the first 3 dimensions), 

- csm calculation (vmapped over all but the coil dimensions), inner function ncoils,ncoils->ncoils

Not sure which way I prefer.

ckolbPTB:
Great! Reads much better and is also a lot faster (factor 10 on my laptop for a 3D dataset). 

Regarding the chunck_size in https://github.com/ckolbPTB/mrpro/blob/73a8e233c904edc4a51adcd3e329667a2bad75f7/src/mrpro/data/_CsmData.py#L91

Commonly for 3D dataset Nother would be 1. Only multi-slice 2D acquisitions would have Nother > 1. So we could simply set chunk_size = Nother for Nz = 1 and else chunk_size = 1

fzimmermann89:
> Commonly for 3D dataset Nother would be 1. Only multi-slice 2D acquisitions would have Nother > 1. So we could simply set chunk_size = Nother for Nz = 1 and else chunk_size = 1

So if Nz=1 we would do chunk_size=Nother=None
If Nz!=1 we assume Nother=1, so chunk_size=Nother=None.

So a reasonable default would be, chunksize=None (all Nother samples in parallel)
with an parameter to reduce it.

fzimmermann89:
This is only waiting for #40 to be merged :)

fzimmermann89:
I moved the SpatialDimension issue to be discussed in #91

schuenke (approved):
👍

fzimmermann89:
Do we need to decorate the child class?

fzimmermann89:
We don't need it for QData children to achieve:
- frozen children. 
- children that cannot get new attributes by (for example) ´csm.realdata=1´ (the slots logic), even though without the decorator, a __dict__ will be created.


We could consider if we really want to make QData a 'slots dataclass': maybe children need to save something else as well?
(I would leave it as slots and reconsider if needed)

I'd prefer not to repeat the rather long decorator for each QData-child, as it also must match the paranets decorator or we would get bugs.

PR #85, opened 2023-10-10
(DRAFT) Fix #84

### LINKED ISSUE #84: Dcf calculation fails for single spiral arm
When the trajectory is a single spiral arm, then `DcfData.from_traj_voronoi` detects this as a single non-singleton dimension and carries out a 1D calculation. This fails for a spiral trajectory because it covers a 2D k-space. 

Possible solution: Check if points lie on a straight line, otherwise use voronoi
This will still fail for certain trajectories but this then needs to be handled by the user with a case by case solution.

fzimmermann89:
This PR should address #84.
For now, I just added a test case that shows the issue.

### DIFF
--- a/tests/data/test_dcf_data.py
+++ b/tests/data/test_dcf_data.py
@@ -14,6 +14,7 @@
 
 import math
 
+import pytest
 import torch
 
 from mrpro.data import DcfData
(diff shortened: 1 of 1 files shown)


fzimmermann89:
This tested the wrong case. Here, the issue is with the analytical solution in the test case, not the dcf calculation.

PR #86, opened 2023-10-10
Fixing DCF logic bugs

### LINKED ISSUE #84: Dcf calculation fails for single spiral arm

fzimmermann89:
The fix for #84 and related bugs.
First, I added some currently failing tests showing the issues @ckolbPTB noticed.
Next, The fix.

Closes #84
Supersedes #85

### DIFF
--- a/src/mrpro/data/_DcfData.py
+++ b/src/mrpro/data/_DcfData.py
@@ -58,10 +58,17 @@ class DcfData:
         # we assign the point at xN the area (xN - xN-1) / 2 + (xN - xN-1) / 2.
 
         kernel = torch.tensor([-1 / 2, 0, 1 / 2], dtype=torch.float32, device=traj.device).reshape(1, 1, 3)
-        central_diff = torch.nn.functional.conv1d(traj_sorted[None, None, :], kernel)[0, 0]
-        first = traj_sorted[1] - traj_sorted[0]
-        last = traj_sorted[-1] - traj_sorted[-2]
-        central_diff = torch.cat((first[None], central_diff, last[None]), -1)
+
+        if (elements := len(traj_sorted)) >= 3:
+            central_diff = torch.nn.functional.conv1d(traj_sorted[None, None, :], kernel)[0, 0]
+            first = traj_sorted[1] - traj_sorted[0]
+            last = traj_sorted[-1] - traj_sorted[-2]
+            central_diff = torch.cat((first[None], central_diff, last[None]), -1)
+        elif elements == 2:
+            diff = traj_sorted[1] - traj_sorted[0]
+            central_diff = torch.cat((diff[None], diff[None]), -1)
+        else:
+            central_diff = torch.ones_like(traj_sorted)
 
         # Repeated points are reduced by the number of repeats
         dcf = torch.nan_to_num(central_diff / counts)[inverse]
@@ -121,7 +128,8 @@ class DcfData:
         idx_outliers = np.nonzero(dcf > upper_bound)
         num_outliers = len(idx_outliers[0])
         high_values_start = int(0.99 * (len(dcf_sorted) - num_outliers))
-        fill_value = np.average(dcf_sorted[high_values_start:-num_outliers])
+        high_values_end = len(dcf_sorted) - num_outliers  # this works also for num_outliers==0
+        fill_value = np.average(dcf_sorted[high_values_start:high_values_end])
         dcf[idx_outliers] = fill_value
 
         # Sort dcf values back into the original order (i.e. before calling unique)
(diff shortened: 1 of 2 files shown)


fzimmermann89:
I also found some (mostly irrelevant) other bugs:
-  The 1D calculation failed for less then 3 unique values, this can be tested for with a 3D cartesian with (2,2,2)
- The voronoi-outlier detection had a bug for less then 10 samples.
and added testcases.

fzimmermann89:
..and, I hope, the fixes.

The broadcast logic is now correct: Instead of checking if in one of kx/ky/kz a dimension is 1, now the logic is:
For each of k2,k1,k0 it is checked if more than one of kx,ky,kz change. 

If yes, we need voronoi. 
Otherwise, we use the 1D calculation for the one that changes or ignore the dimension if neither of kx,ky,kz changes.

src/mrpro/data/_DcfData.py
```diff
@@ -58,10 +58,17 @@ def _dcf_1d(traj: torch.Tensor) -> torch.Tensor:
         # we assign the point at xN the area (xN - xN-1) / 2 + (xN - xN-1) / 2.
 
         kernel = torch.tensor([-1 / 2, 0, 1 / 2], dtype=torch.float32, device=traj.device).reshape(1, 1, 3)
-        central_diff = torch.nn.functional.conv1d(traj_sorted[None, None, :], kernel)[0, 0]
-        first = traj_sorted[1] - traj_sorted[0]
-        last = traj_sorted[-1] - traj_sorted[-2]
-        central_diff = torch.cat((first[None], central_diff, last[None]), -1)
+
+        if (elements := len(traj_sorted)) >= 3:
```
fzimmermann89, line 62: This is the common case.

The other two cases should realistically not be hit to often:
Only for k-dimensions with less then 3 unique values.

src/mrpro/data/_DcfData.py
```diff
@@ -136,26 +144,35 @@ def from_traj_voronoi(cls, traj: KTrajectory) -> DcfData:
...
         dcfs = []
-        for k in [traj.kz, traj.ky, traj.kx]:
-            non_singleton_dims = tuple(dim for dim in (-1, -2, -3) if k.shape[dim] != 1)
-            if len(non_singleton_dims) == 1:
-                # Found a direction with two singleton dimensions and one non-singleton dimension
+
+        ks = [traj.kz, traj.ky, traj.kx]
+        spatial_dims = (-3, -2, -1)
+        ks_needing_voronoi = set()
+        for dim in spatial_dims:
```
fzimmermann89, line 154: This is the main fix

src/mrpro/data/_DcfData.py
```diff
@@ -136,26 +144,35 @@ def from_traj_voronoi(cls, traj: KTrajectory) -> DcfData:
...
+            non_singleton_ks = [ax for ax in ks if ax.shape[dim] != 1]
+            if len(non_singleton_ks) == 1:
+                # Found a dimension only one non-singleton axes
```
ckolbPTB: Found a dimension WITH only one non-singleton axes

src/mrpro/data/_DcfData.py
```diff
@@ -136,26 +144,35 @@ def from_traj_voronoi(cls, traj: KTrajectory) -> DcfData:
...
                 # We can handle this as a 1D trajectory
-                dcfs.append(smap(DcfData._dcf_1d, k, non_singleton_dims))
-            elif len(non_singleton_dims) == 0:
-                # Found a direction with only singleton dimensions, i.e. constant dcf
-                dcfs.append(torch.ones_like(k))
-            else:
+                dcfs.append(smap(DcfData._dcf_1d, non_singleton_ks.pop(), (dim,)))
+            elif len(non_singleton_ks) > 0:
                 # A Full Dimension needing voronoi
```
ckolbPTB: A full dimension needing...

src/mrpro/data/_DcfData.py
```diff
@@ -136,26 +144,35 @@ def from_traj_voronoi(cls, traj: KTrajectory) -> DcfData:
...
-                ks_needing_voronoi.append(k)
+                ks_needing_voronoi |= set(non_singleton_ks)
+            else:
+                # A all 1-dimension, we dont need to do anything
```
ckolbPTB: All dimensions are singleton, we don't need to do anything

src/mrpro/data/_DcfData.py
```diff
@@ -136,26 +144,35 @@ def from_traj_voronoi(cls, traj: KTrajectory) -> DcfData:
...
-            dcfs.append(smap(DcfData._dcf_2d3d_voronoi, torch.stack(ks_needing_voronoi, -4), 4))
-        # Multiply all dcfs together
-        dcf = reduce(torch.mul, dcfs)
+            dcfs.append(smap(DcfData._dcf_2d3d_voronoi, torch.stack(list(ks_needing_voronoi), -4), 4))
+
+        if dcfs:
+            # Multiply all dcfs together
+            dcf = reduce(torch.mul, dcfs)
+        else:
+            # only singleton spatial dimensions
```
ckolbPTB: **O**nly

ckolbPTB (approved):
Just a few typos.

PR #89, opened 2023-10-12
Run tests in parallel

### LINKED ISSUE #87: Suggestion: Run tests in parallel
As the number of tests increases, it might be nice to run tests in parallel, both in VSCode as well as in our CI

https://code.visualstudio.com/docs/python/testing#_run-tests-in-parallel

schuenke:
Closes #87 

I tested it on my Laptop (i3 12. gen) and the runtime decreased (just) by about 15%, but probably it will become more significant in the future when the number of tests increases.

**EDIT: Currently, for our automated workflow on GitHub, no significant runtime reduction is measureable, because of the huge overhead for setting up the python env and dependencies. However, I see no disadvantage and still suggest to merge the PR as it already improves the performance when running the tests in VSCode.**

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,14 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
-test = ["coverage", "codecov", "pre-commit", "pytest", "pytest-cov"]
+test = [
+    "coverage",
+    "codecov",
+    "pre-commit",
+    "pytest",
+    "pytest-cov",
+    "pytest-xdist",
+]
 lint = ["mypy", "flake8", "isort", "pre-commit", "autopep8", "pydocstyle"]
 docs = ["sphinx", "sphinx_rtd_theme"]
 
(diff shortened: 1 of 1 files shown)


fzimmermann89:
Have you tested if less parallel processes (for example, auto or 4) is faster on your machine?

I was mostly annoyed by the runtime of the test on my similar specd development machine, the ci time is not that critical for me.

fzimmermann89 (approved):

schuenke:
> Have you tested if less parallel processes (for example, auto or 4) is faster on your machine?

I did now and the result is interesting:

- **default:** 30.7 ± 4.0 s
- **n 1:** 17.8 ± 0.3 s
- **n 4:** 19.1 ± 0.6 s
- **n auto:** 20.8 ± 0.6 s
- **n logical:** 22.6 ± 0.8 s

So everything is much better than the default (= no addopts), but for our current 27 simple tests, n=1 seems to be best. Maybe we pick `-n auto` as a compromise between current and expected future performance? 😄

PR #95, opened 2023-10-14
espirit csm

### LINKED ISSUE #34: Espirit CSM Calculation

fzimmermann89:
A work-in-progress for espirit algorithm for cartesian images.
This is blocked until we have 
- a Cartesian FFT Operator (the code currently does an ifft on kdata.data)  and 
- a notion of the ACS region ('calib' in the code). 

Other TODOs/Notes:
- image_shape should removed by using the shape of the result of the adjoint operator
- the code is completely untested and will most likely contain logic and syntax errors.
- not sure if the unfolds are correct.

The logic is derived from [sigpy ](https://sigpy.readthedocs.io/en/latest/_modules/sigpy/mri/app.html#EspiritCalib) 
we would have to check the license

Why implement it and not wrap sigpy?
Sigpy only works with Cartesian data and is sometimes annoying to install, as it uses cupy for the GPU implementation.

Closes #34


fzimmermann89:
https://onlinelibrary.wiley.com/doi/full/10.1002/mrm.28386

src/mrpro/data/_CsmData.py
```diff
@@ -93,3 +96,86 @@ def from_idata_walsh(
...
+        mat = calib
+        for ax in (1, 2, 3):
+            mat = mat.unfold(dimension=ax, size=min(calib.shape[ax], kernel_width), step=1)
+        mat = rearrange(mat, 'coils z y x c b a -> (z y x) (coils c b a)')
+
+        # Perform SVD on calibration matrix
+        _, S, VH = torch.linalg.svd(mat, full_matrices=False)
+
+        # Get kernels
+        kernels = rearrange(VH[S > thresh * S.max()], 'n (coils cba) -> n coils cba')
```
SherineBrahma: **Background:** 

[ESPIRiT paper](https://onlinelibrary.wiley.com/doi/epdf/10.1002/mrm.24751)

In eq.(9), we need to calculate V_||  that is associated with the row space of the Calibration matrix A. This is later used to construct the forward matrix W. In principle, V_|| is obtained from V by trimming off the row vectors corresponding to zero eigen-values. However, here, they have done it based on this mask: "S > thresh * S.max()". This makes sense as eigen-values may not go to 0 because of the presence of noise. 

**Problem:**
This trimming of VH[S > thresh * S.max()] based on the mask (S > thresh * S.max()) is not supported inside "vmap" operations. I have a feeling that operations like this change the size of the matrix, and that poses challenges to perform batch operations, given that all elements of the batches should have the same shape. 

**Possible Solution:**
If we want to adopt a parallelized approach like before, maybe it makes sense to perform vmap only till the SVD is calculated, since SVD is usually a computationally demanding algorithm for large matrices.  Thereafter, we can use loops to construct V_||'s, where different V_||'s can have different shapes based on the batched SVD results. For this, I can use the smap function defined [here](https://github.com/PTB-MR/mrpro/blob/c7d040cbd8348f554dfd7ce6bf19abdfcdd94734/src/mrpro/utils/_smap.py#L21). However, based on the layout of this file (define __iterative_walsh_csm_ and then _estimate_walsh_, define _estimate_espirit_ and then  __espirit_csm_) ,  I am not sure if opening up this code to perform first part in one place (SVD calculation using vmap), while performing the other part in another place (forward operation using loops), will break the stye of the code. 

If we want to retain the original coding style, I can use the smap function over the whole __espirit_csm_ function where the SVD is included, but then we would not be parallelizing the SVD.

@ckolbPTB

fzimmermann89: I would just switch to fully serial calculation of the csms for different others using smap.

PR #101, opened 2023-10-18
hotfix vscode pytest-xdist issues (#87)

fzimmermann89:
Adresses isses caused by #89, mentioned in #87 by disabling parallel testing for now.

We still test in paralell in the CI.
We still install pytest-xdist.

Any developer not having issues with parallel testing can add "-n auto" in their vscode pytest settings as additional argument to still use parallel test execution.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -36,7 +36,7 @@ jobs:
       - name: Run PyTest
 
         run: |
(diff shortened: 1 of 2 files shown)


fzimmermann89:
This is just a bandaid so that we can be sure that the tests run for everybody tomorrow.

schuenke (approved):

PR #107, opened 2023-10-20
Implementation of ktraj shaped like original raw data

ckolbPTB:


src/mrpro/data/_KTrajectoryRawShape.py
```diff
@@ -0,0 +1,80 @@
...
+    Shape of each of kx,ky,kz is ((other,k2,k1),k0) this means that e.g. slices, averages... have not yet been
+    separated from the phase and slice encoding dimensions. The trajectory is in the same shape as the raw data in the
+    raw data file.
+    """
+
+    kz: torch.Tensor  # ((other,k2,k1),k0) #phase encoding direction, k2 if Cartesian
+    ky: torch.Tensor  # ((other,k2,k1),k0) #phase encoding direction, k1 if Cartesian
+    kx: torch.Tensor  # ((other,k2,k1),k0) #frequency encoding direction, k0 if Cartesian
+
+    def __init__(self, kz: torch.Tensor, ky: torch.Tensor, kx: torch.Tensor):
```
fzimmermann89: I think we dont need the init here?
Can we just use default dataclass init?
I.e.  `@dataclass(slots=True)` and no init

fzimmermann89 (approved):
Consider changing the init to default dataclass init, but otherwise lgtm!

PR #112, opened 2023-10-20
add draft for b1+ mapping script

schuenke:
This is still work in progress.  Things that need to be done:

- change numpy to torch
- reformat code
- create QData from image tensor
- fix mypy type checks (currently mypy is set to ignore the file 😃 )


ckolbPTB:
superseded by #230

PR #111, opened 2023-10-20
KTrajectoryPulseq

JoHa0811:
Calculate trajectory from pulseq sequence file

Closes #72

PR #113, opened 2023-10-25
Ismrmrd traj

### LINKED ISSUE #39: ISMRMRD Trajectory
We could also consider having a Trajectory class that reads in/prepares the Trajectory information that might be already saved inside the ismrmd file.

We should check if any of the converters write something useful there.

ckolbPTB:
closes #39 

Trajectory information is read out from ISMRMRD raw data file

### DIFF
--- a/src/mrpro/data/_KData.py
+++ b/src/mrpro/data/_KData.py
@@ -27,9 +27,11 @@ from einops import rearrange
 from mrpro.data import AcqInfo
 from mrpro.data import KHeader
 from mrpro.data import KTrajectory
+from mrpro.data import KTrajectoryRawShape
 from mrpro.data import Limits
 from mrpro.data.enums import AcqFlags
 from mrpro.data.traj_calculators import KTrajectoryCalculator
+from mrpro.data.traj_calculators import KTrajectoryIsmrmrd
 
 KDIM_SORT_LABELS = (
     'k1',
@@ -53,7 +55,7 @@ class KData:
     def from_file(
         cls,
         filename: str | Path,
-        ktrajectory: KTrajectoryCalculator | KTrajectory,
+        ktrajectory: KTrajectoryCalculator | KTrajectory | KTrajectoryIsmrmrd,
         header_overwrites: dict[str, object] | None = None,
         dataset_idx: int = -1,
     ) -> KData:
@@ -141,12 +143,22 @@ class KData:
                     setattr(current, subfield.name, reshape_acq_data(subcurrent))
 
         # Calculate trajectory and check if it matches the kdata shape
-        if isinstance(ktrajectory, KTrajectoryCalculator):
-            ktraj = ktrajectory(kheader)
-        elif isinstance(ktrajectory, KTrajectory):
-            ktraj = ktrajectory
-        else:
-            raise TypeError(f'ktrajectory must be a KTrajectory or KTrajectoryCalculator, not {type(ktrajectory)}')
+        match ktrajectory:
+            case KTrajectoryIsmrmrd():
+                ktraj = ktrajectory(acquisitions).reshape(sort_idx, num_k2, num_k1)
+            case KTrajectoryCalculator():
+                ktraj_calc = ktrajectory(kheader)
+                if isinstance(ktraj_calc, KTrajectoryRawShape):
+                    ktraj = ktraj_calc.reshape(sort_idx, num_k2, num_k1)
+                else:
+                    ktraj = ktraj_calc
+            case KTrajectory():
+                ktraj = ktrajectory
+            case _:
+                raise TypeError(
+                    'ktrajectory must be KTrajectoryIsmrmrd, KTrajectory or KTrajectoryCalculator,'
+                    f'not {type(ktrajectory)}'
+                )
 
         try:
             shape = ktraj.broadcasted_shape
--- a/tests/data/_IsmrmrdRawTestData.py
+++ b/tests/data/_IsmrmrdRawTestData.py
@@ -12,6 +12,8 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
+from __future__ import annotations
+
 from pathlib import Path
 from typing import Literal
 
@@ -76,7 +78,7 @@ class IsmrmrdRawTestData:
         flag_invalid_reps: bool = False,
         acceleration: int = 1,
         noise_level: float = 0.00005,
-        trajectory_type: str = 'cartesian',
+        trajectory_type: Literal['cartesian', 'radial'] = 'cartesian',
         sampling_order: Literal['linear', 'low_high', 'high_low'] = 'linear',
         phantom: EllipsePhantom = EllipsePhantom(),
     ):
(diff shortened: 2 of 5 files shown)


src/mrpro/data/_KData.py
```diff
@@ -141,12 +142,17 @@ def reshape_acq_data(data):
                     setattr(current, subfield.name, reshape_acq_data(subcurrent))
 
         # Calculate trajectory and check if it matches the kdata shape
-        if isinstance(ktrajectory, KTrajectoryCalculator):
+        if isinstance(ktrajectory, KTrajectoryIsmrmrd):
```
ckolbPTB: It would be nicer to have something like in line 156:
```
if ktraj is KTrajectoryRawShape:
      ktraj = ktraj.reshape(sort_idx, num_k2, num_k1)
```
but mypy did not like that.

fzimmermann89: we could do structural pattern matching, https://peps.python.org/pep-0636/ 

https://stackoverflow.com/questions/67524641/convert-multiple-isinstance-checks-to-structural-pattern-matching

fzimmermann89: Also (orthogonal decision):
considering [#72](https://github.com/PTB-MR/mrpro/issues/72), it would make sense to first do the check if ktrajectory is an ISMRMRDCalculator (call with acquisitions and set ktrajectory as the result), a TrajectoryCalculator (call with header  and set ktrajectory as the result) or else. 
Then if  ktrajectory is Rawshape, reshape.

this would handle both cases where the trajectory is returned as raw shape in one reshaping block.

src/mrpro/data/_KData.py
```diff
@@ -141,12 +142,17 @@ def reshape_acq_data(data):
...
+            ktraj_raw = ktrajectory(acquisitions)
+            ktraj = ktraj_raw.reshape(sort_idx, num_k2, num_k1)
+        elif isinstance(ktrajectory, KTrajectoryCalculator):
             ktraj = ktrajectory(kheader)
         elif isinstance(ktrajectory, KTrajectory):
             ktraj = ktrajectory
         else:
-            raise TypeError(f'ktrajectory must be a KTrajectory or KTrajectoryCalculator, not {type(ktrajectory)}')
+            raise TypeError(
+                f'ktrajectory must be None, a KTrajectory or KTrajectoryCalculator, not {type(ktrajectory)}'
```
fzimmermann89: can it be None?
ISMMRMD would be a sensible default, but currently it does not seem like it is

tests/data/_IsmrmrdRawTestData.py
```diff
@@ -99,13 +99,24 @@ def __init__(
         nkx = self.oversampling * nx
         nky = ny
 
-        # Create Cartesian grid for k-space locations
-        ky_idx = self._calc_phase_encoding_steps(nky, self.acceleration, self.sampling_order)
-        kx_idx = range(-nkx // 2, nkx // 2)
-        [kx, ky] = np.meshgrid(kx_idx, ky_idx)
+        if trajectory_type == 'cartesian':
```
fzimmermann89, line 104: maybe consider replacing the "_calc_phase_encoding_steps" function by
a "_cartesian_trejectory" and a "_radial_trajectory", both returning ky, kx,pe_line_idx ?

tests/data/_IsmrmrdRawTestData.py
```diff
@@ -221,12 +232,12 @@ def __init__(
...
         # Loop over the repetitions, add noise and write to disk
         for rep in range(self.repetitions):
-            noise = self.noise_level * torch.randn(self.ncoils, nky // self.acceleration, nkx, dtype=torch.complex64)
+            noise = self.noise_level * torch.randn(self.ncoils, nkx, nky // self.acceleration, dtype=torch.complex64)
             # Here's where we would make the noise correlated
             K = ktrue + noise
             acq.idx.repetition = rep
-            for idx, line in enumerate(ky_idx):
-                if not self.flag_invalid_reps or rep == 0 or idx < len(ky_idx) // 2:  # fewer lines for rep > 0
+            for idpe, line in enumerate(pe_line_idx):
```
fzimmermann89: Here the naming of pe_line_idx, idpe, line, line_idx is a maybe bit confusing (or I am tired)..

fzimmermann89 (approved):
Some suggestions with regard to the test and the logic in KData.
We can postpone the KData logic decision until the last of both #72 and this PR are ready to be merged, as both with touch the same lines.

ckolbPTB:
> considering https://github.com/PTB-MR/mrpro/issues/72, it would make sense to first do the check if ktrajectory is an ISMRMRDCalculator (call with acquisitions and set ktrajectory as the result), a TrajectoryCalculator (call with header and set ktrajectory as the result) or else.
Then if ktrajectory is Rawshape, reshape.

I tried it like this:

```
        # Calculate trajectory and check if it matches the kdata shape
        if isinstance(ktrajectory, KTrajectoryIsmrmrd):
            ktraj = ktrajectory(acquisitions)
        elif isinstance(ktrajectory, KTrajectoryCalculator):
            ktraj = ktrajectory(kheader)
        elif isinstance(ktrajectory, KTrajectory):
            ktraj = ktrajectory
        else:
            raise TypeError(
                f'ktrajectory must be KTrajectoryIsmrmrd, KTrajectory or KTrajectoryCalculator, not {type(ktrajectory)}'
            )

        if ktraj is KTrajectoryRawShape:
            ktraj = ktraj.reshape(sort_idx, num_k2, num_k1)
```

but mypy does not like it:

src/mrpro/data/_KData.py:149: error: Incompatible types in assignment (expression has type "KTrajectory", variable has type "KTrajectoryRawShape")  [assignment]
src/mrpro/data/_KData.py:151: error: Incompatible types in assignment (expression has type "KTrajectory", variable has type "KTrajectoryRawShape")  [assignment]
src/mrpro/data/_KData.py:158: error: Incompatible types in assignment (expression has type "KTrajectory", variable has type "KTrajectoryRawShape")  [assignment]
src/mrpro/data/_KData.py:161: error: "KTrajectoryRawShape" has no attribute "broadcasted_shape"  [attr-defined]
src/mrpro/data/_KData.py:169: error: Argument 3 to "KData" has incompatible type "KTrajectoryRawShape"; expected "KTrajectory"  [arg-type]

src/mrpro/data/_KData.py
ckolbPTB: No it can't be none, should be KTrajectoryIsmrmrd, KTrajectory or KTrajectoryCalculator

I would not set a default for ktrajectory because a default to me should always work, but KTrajectoryIsmrmrd would fail if the trajectory is not in the file and would yield rubbish if the trajectory is just 0s in the file.

fzimmermann89: Ok, then just the error message is wrong

fzimmermann89:
```python
        ktraj_calc: KTrajectory | KTrajectoryRawShape
        match ktrajectory:
            case KTrajectoryIsmrmrd():
                ktraj_calc = ktrajectory(acquisitions)
            case KTrajectoryCalculator():
                ktraj_calc = ktrajectory(kheader)
            case KTrajectory():
                ktraj_calc = ktrajectory
            case _:
                raise TypeError(
                    f'ktrajectory must be KTrajectoryIsmrmrd, KTrajectory or KTrajectoryCalculator, not {type(ktrajectory)}'
                )
        if isinstance(ktraj_calc, KTrajectoryRawShape):
            ktraj = ktraj_calc.reshape(sort_idx, num_k2, num_k1)
        else:
            ktraj = ktraj_calc

```
could work. But I agree, this looks uglier and more confusing than
```python
    match ktrajectory:
            case KTrajectoryIsmrmrd():
                ktraj_calc= ktrajectory(acquisitions).reshape(sort_idx, num_k2, num_k1)
                ktraj = ktraj_raw(kheader).reshape(sort_idx, num_k2, num_k1)
            case KTrajectoryCalculator():
                ktraj_calc = ktrajectory(kheader)
                if isinstance(ktraj_calc, KTrajectoryRawShape):
                    ktraj = ktraj_calc.reshape(sort_idx, num_k2, num_k1)
                else:
                    ktraj = ktraj_calc
            case KTrajectory():
                ktraj = ktrajectory
            case _:
                raise TypeError(
                    f'ktrajectory must be KTrajectoryIsmrmrd, KTrajectory or KTrajectoryCalculator, not {type(ktrajectory)}'
                )

```

This is less confusing and maybe more explicit?

Or we make a KTrajectroyRawShapeCalculator(KTrajectroyCaclulator) class and fully move over the Java/C# class hierachy ;)

tests/data/_IsmrmrdRawTestData.py
ckolbPTB: Rewrote that part a bit, hope it is clearer now.

ckolbPTB:
> This is less confusing and maybe more explicit?

I prefer more explicit. If we change our mind we can always modify it for #111

Issue #65, opened 2023-09-14
Noise prewhitening


ckolbPTB:
could be implemented as a wrapper of 
https://github.com/ismrmrd/ismrmrd-python-tools/blob/17fe6fbf1bb645112e2ad0023eb4f42afb36421e/ismrmrdtools/coils.py#L9
and
https://github.com/ismrmrd/ismrmrd-python-tools/blob/17fe6fbf1bb645112e2ad0023eb4f42afb36421e/ismrmrdtools/coils.py#L28

ckolbPTB:
Closed by #114

PR #114, opened 2023-10-26
Noise measurement data class and prewhitening algo added

### LINKED ISSUE #109: Option to create KData from noise data
Currently we hard coded to ignore noise data when creating a KData instance from a ISMRMRD raw data file:

https://github.com/PTB-MR/mrpro/blob/101f1f57713b56182ce0e5915eb6d37b3b56cde7/src/mrpro/data/_KData.py#L85

However, for some applications (e.g. multi-coil B1+ mapping) we need this data. 

Introducing a flag in `KData.from_file()` could be an option imo.

ckolbPTB:
Closes #109 and #65


src/mrpro/algorithms/_prewhiten_kspace.py
```diff
@@ -0,0 +1,61 @@
...
+    noise = rearrange(knoise.data, 'other coils k2 k1 k0->coils (other k2 k1 k0)')
+
+    # Calculate noise covariance matrix which should ideally be a unity matrix, i.e. no noise correlation between coils
+    noise_cov = (1.0 / (noise.shape[1])) * torch.einsum('ax,bx->ab', noise, noise.conj())
+
+    # Calculate prewhitening matrix and scale
+    prew = torch.linalg.inv(torch.linalg.cholesky(noise_cov)) * np.sqrt(scale_factor)
+
+    # Apply prewhitening matrix
+    prew_data = torch.einsum('yx,axbcd->aybcd', prew, kdata.data)
```
fzimmermann89, line 64: https://www.johndcook.com/blog/2010/01/19/dont-invert-that-matrix/ ;)

torch.cholesky_solve ?

src/mrpro/data/_KNoise.py
```diff
@@ -0,0 +1,61 @@
...
+
+import ismrmrd
+import torch
+from einops import rearrange
+
+from mrpro.data.enums import AcqFlags
+
+
+@dataclasses.dataclass(slots=True, frozen=True)
+class KNoise:
```
fzimmermann89, line 28: We could consider making the Flags a parameter in KData.from_ismrmrd

And KNoise inherit from KData with changed default flag selection ( keep only noise) 

Advantage: "Nicer" type hierachy, we can have functions that work on both, noise and real data.

Disadvantage: KNoise would get a KHeader and Trajectroy, which are useless.

tests/algorithms/test_prewhiten_kspace.py
```diff
@@ -0,0 +1,50 @@
...
+    knoise = KNoise(data=knoise_data)
+
+    # Create KData from same data with random trajectory
+    ktraj = KTrajectory(torch.zeros(1, 1, 1, 1), torch.zeros(1, 1, 1, 1), torch.zeros(1, 1, 1, 1))
+    kdata = KData(header=random_kheader, data=knoise_data, traj=ktraj)
+
+    kdata = prewhiten_kspace(kdata, knoise)
+
+    # Need to add epsilon to matrices otherwise relative assert could fail due to div by 0
+    eps = torch.finfo(torch.float32).eps
```
fzimmermann89: afaik torch.close uses 
∣input−other∣≤atol+rtol×∣other∣ 
as the condition. 

So you could maybe just increase the atol?

fzimmermann89 (commented):
some comments

src/mrpro/data/_KNoise.py
ckolbPTB: I would prefer to keep it separate rather than KNoise having a header and trajectory which are invalid.

tests/algorithms/test_prewhiten_kspace.py
ckolbPTB: thanks, I had misunderstand what rtol mean. `assert_close` works fine as is.

src/mrpro/algorithms/_prewhiten_kspace.py
ckolbPTB: > https://www.johndcook.com/blog/2010/01/19/dont-invert-that-matrix/ ;)

Great post!

`cholesky_solve` does not work but `torch.linalg.solve_triangular` would work. Nevertheless, then I don't see a way of doing the einsum application to the data but would need to rearrange to (other, coil, (k2,k1,k0)), solve and rearrange back. Seems a bit overkill for a matrix size which is never more than 40 x 40.

fzimmermann89: ups. I completely misunderstood to goal of these lines. Sorry.

schuenke (approved):
lgtm 👍

PR #116, opened 2023-11-17
Remove oversampling and fft added

### LINKED ISSUE #4: remove_oversampling
implement RemoveOversRo in [Functions.RemoveOversRo.py](https://github.com/ckolbPTB/PtbPyRecon/blob/master/PTBRecon/Functions/RemoveOversRo.py "‌")

`KData = remove_oversampling(KData)`

Check list:
- [ ]     Create branch
- [ ]     Implement code
- [ ]     Add comments
- [ ]     Implement tests
- [ ]     Squash commmits
- [ ]     Create PR

ckolbPTB:
Closes #4 

- moved kspace_to_image to utils and added image_to_kspace
- added tests for both
- implemented remove oversampling similar to https://github.com/gadgetron/gadgetron-python/blob/1fa47cc3943542dffd184a541399f57d73347731/gadgetron/examples/recon_acquisitions.py#L49

I am not sure about the test for remove_readout_os. Currently I am only testing if the image is correct but not if all the dimensions and trajectory are correct. 

Maybe sufficient to replace kspace_to_images in the test with the FourierOperator once it is merged to ensure we can still reconstruct an image. This should then implicitly also verify that all dimensions needed for a simple image reconstruction are still correct.

### DIFF
--- a/src/mrpro/algorithms/__init__.py
+++ b/src/mrpro/algorithms/__init__.py
@@ -1,1 +1,2 @@
 from mrpro.algorithms._prewhiten_kspace import prewhiten_kspace
+from mrpro.algorithms._remove_readout_os import remove_readout_os
--- a/src/mrpro/algorithms/_remove_readout_os.py
+++ b/src/mrpro/algorithms/_remove_readout_os.py
@@ -37,7 +37,7 @@ def remove_readout_os(kdata: KData) -> KData:
     Raises
     ------
     ValueError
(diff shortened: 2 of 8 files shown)


schuenke (approved):
I corrected 2 minor mistakes in the docstrings, please double check these again. 

The rest looks fine to me. 

Additional tests might be added in the future, but I don't think they are mandatory.

PR #118, opened 2023-11-21
Framework py -> ipynb -> docu

### LINKED ISSUE #93: Framework for examples
Set up framework to 

- have example scripts (.py) which show application of mrpro on real data (e.g. downloaded from zenodo)
- automatically create example notebooks with results (e.g. reconstruction images) as github actions
- add these example notebooks to documentation (requires #92)

ckolbPTB:
Closes #93 

I added a script which downloads data automatically and reconstructs an image (some TODOs because we have not merged the Fourier operator yet). In the long run we would also include T1 mapping here.

I tried to run it p2j (https://github.com/remykarem/python2jupyter) on this script but the results were not great. p2j does not recognise the cell formating of VSC. 

Any other suggestions for the py -> ipynb welcome. Other option would be to adapt p2j to our needs - not ideal.


ckolbPTB:
Maybe an option: https://github.com/mwouts/jupytext

ckolbPTB:
The scripts in examples are now automatically translated into notebooks and commited to the current branch. Then the notebooks are run and the output it saved as html. Finally these html files are included in the docu in Examples.rst. 

All  of this happens via the docs-workflow, which I now enabled to run every time not just when pull request to main. 

Still need to work on the Binder and Colab badges - they work but mrpro is not installed and hence the notebooks don't run

ckolbPTB:
Colab does not seem to work for our case - useful to run individual notebooks but not all notebooks of a package. 

binder works fine, set-up to pick-up requirements in binder subfolder. requirements.txt is automatically created from toml file. MRpro still needs to be installed after opening binder but very quick because requirements are already present. There is an option to create a binder image via github action but not sure if this is necessary: https://github.com/jupyterhub/repo2docker-action#use-github-actions-to-cache-the-build-for-binderhub

ckolbPTB:
ToDo: binder badge currently runs on this branch. Needs to be changed to main before merge with main

schuenke (approved):
Everything should work now as expected. 

Minor remaining issue are the warnings in the ipynb files...

PR #121, opened 2023-12-04
docu for users and devs extended

### LINKED ISSUE #106: Documentation: Software/Repository Architecture Description
Explain the repository structure: Where can I find interfaces, abstract classes, implementation of abstract classes, etc.

- [ ] Interactons between components and classes
- [ ] Reconstruction flow diagram/workflow (maybe with concrete 
- [ ] Example workflow
- [ ] Add diagrams/descriptions/examples to documentation

ckolbPTB:
Closes #106

### DIFF
--- a/docs/source/contributer_guide.rst
+++ b/docs/source/contributer_guide.rst
@@ -1,4 +1,64 @@
-MRPro Contributer Guide
-=======================
+============================
+Guide for MRPro Contributors
+============================
 
-This is a contributer guide for the mrpro repository.
+Repo structure
+==============
+This repository uses a *pyproject.toml* file to specify all the requirements.
+If you need a "normal" *requirements.txt* file, please have a look in *binder*. There you find a *requirements.txt*
+automatically created from *pyproject.toml* using GitHub actions.
+
+**.github/workflows**
+    Definitions of GitHub action workflows to carry out formatting checks, run tests and automatically create this
+    documentation.
+
+**.vscode**
+    Configuration files for VS Code.
+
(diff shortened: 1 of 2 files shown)


ckolbPTB:
@schote I know this is very basic but can you think of anything which is missing but absolutely necessary (or easy to add)? If you have an example of a good documentation of another repo, this would also be very helpful.

schote:
> @schote I know this is very basic but can you think of anything which is missing but absolutely necessary (or easy to add)? If you have an example of a good documentation of another repo, this would also be very helpful.

In my opinion [scikit-learn](https://scikit-learn.org/stable/) has a good documentation, which we could use as a reference. I think as soon as we have our first release and a release strategy (in case this is planned at some point) we should add:
- versioning to indicate the version the documentation refers to
- changelog
- timeline to indicate what future implementations are planned (could be helpful if one needs to make a decision whether or not the package is suitable or will be suitable)

Besides that I would add some minimal examples. I think you also already have that in mind with the notebooks, right @ckolbPTB ?

ckolbPTB:
scikit-learn has a very nice documentation - might be a bit too ambitious for us but let's see.

Yes, the examples will be done via automatically creating jupyter notebooks PR #118 

I added an issue regarding the versionen and release strategy #125

@schote Could you do a quick review of these changes such that I can merge them before the hackathon next week? thanks!

schote (approved):
I think the documentation helps a lot to better understand the structure of the repository 👍🏼

PR #124, opened 2023-12-07
reading in of all dcm files from folder

ckolbPTB:


schuenke (approved):
Made some small modifications and tested the functionality with the qmri data for the 2nd hackathon. Worked fine so far. 

Maybe we can add some additional tests in the future, but imo it's not super important 😄

PR #128, opened 2023-12-11
Synthetic Sugar for Operators (needs python 3.11)

fzimmermann89:
Our goal is to be able to write:

```
objective = L2(kacq) @ FourierOp(ktraj) @ T1SaturationRecovery + lambda1 * L2(p_reg) + lambda2*TV
parameters = LBFGS(objective, p0)
```

This PR implements the @, +, * for operator and linear-operator.

Furthermore, it removes non-linear-operator, as we decied to only have Operator and the special case LinearOperator.


tests/operators/test_operators.py
```diff
@@ -0,0 +1,156 @@
+"""Tests for the operators module."""
+
+import pytest
```
schote, line 3: unused import

schote (approved):

fzimmermann89:
This is a little bit blocked until Patrick is back as we should talk about the typing for operators...

fzimmermann89:
This construct now requires python=3.11 and mypy 1.7 for type checking...

Operator is a Generic class.

For an example, see
https://mypy-play.net/?mypy=1.7.1&python=3.11&enable-incomplete-feature=TypeVarTuple%2CUnpack&gist=3b309c5207fd60196d815b3d84b3ac54

fzimmermann89:
Rel #135:

All operators are supposed to return tuples

fzimmermann89:
This now touches 32 files, as I had to change all usage of our operators in the tests...
So each 
`kdata=FOp(image) `
is now 
`(kdata,)=FOp(image) `

also, I increase the minimum python version to 3.11
and increase the torch version in the precommit mypy environment

ckolbPTB:
src/mrpro/operators/._Operator.py.swp and src/mrpro/operators/._Operator.py.swo should not be part of the commit, should they?

fzimmermann89:
Ops.

schuenke (approved):
I went through the code and everything looked fine.
Tests seem reasonable and everything is up to date with main. 

I am OK with merging it as it is 👍

PR #129, opened 2023-12-11
Signal models added: Inversion and saturation recovery, MOLLI

Stef-Martin:


src/mrpro/operators/models/InversionRecovery.py
```diff
@@ -0,0 +1,40 @@
...
+            params: (M0, T1)
+
+        Returns
+        -------
+            Image data tensor (other, c, z, y, x)
+        """
+        M0 = qdata[0].unsqueeze(0)
+        T1 = qdata[1].unsqueeze(0)
+        ti = rearrange(torch.Tensor(self.ti), 't -> t 1 1 1 1 1')
+        y = M0 * (1 - 2 * torch.exp(-torch.div(ti, T1)))
```
fzimmermann89: use / instead of torch.div

src/mrpro/operators/models/InversionRecovery.py
```diff
@@ -0,0 +1,40 @@
...
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+import torch
+from einops import rearrange
+
+from mrpro.operators import Operator
+
+
+class InversionRecovery(Operator):
+    def __init__(self, ti: list[float]):
```
fzimmermann89: ti should be a tensor here already

src/mrpro/operators/models/InversionRecovery.py
```diff
@@ -0,0 +1,40 @@
...
+        ----------
+        qdata
+            Quantitative parameter tensor (params, other, c, z, y, x)
+            params: (M0, T1)
+
+        Returns
+        -------
+            Image data tensor (other, c, z, y, x)
+        """
+        M0 = qdata[0].unsqueeze(0)
```
fzimmermann89: suggestion: M0, T1 = qdata

src/mrpro/operators/models/InversionRecovery.py
```diff
@@ -0,0 +1,40 @@
...
+        T1 = qdata[1].unsqueeze(0)
+        ti = rearrange(torch.Tensor(self.ti), 't -> t 1 1 1 1 1')
```
fzimmermann89: this should work for multiple dimensions in other, currently it does not.

src/mrpro/operators/models/Molli.py
```diff
@@ -0,0 +1,57 @@
...
+            Image data tensor (other, c, z, y, x)
+        """
+        A = qdata[0].unsqueeze(0)
+        B = qdata[1].unsqueeze(0)
+        T1 = qdata[2].unsqueeze(0)
+        t = torch.Tensor(self.ti) + (self.n - 1) * self.rr
+        t = rearrange(t, 't -> t 1 1 1 1 1')
+        T1_star = torch.div(T1, (torch.div(B, A) - 1))
+        y = A - B * torch.exp(-torch.div(t, T1_star))
+        res = rearrange(y, 't other c z y x -> (t other) c z y x')
```
fzimmermann89: other should  capture multiple dimensions, consider (...)

src/mrpro/operators/models/InversionRecovery.py
fzimmermann89: suggestion:

Either A)
 ti = self.ti((...,)+(None,)*T1.ndim) 

or similiar

Or B)
 (ti/T1[...,None]).moveaxis(-1,0)
or similiar
(This would be torch scriptable..)

src/mrpro/operators/models/Molli.py
```diff
@@ -0,0 +1,57 @@
...
+        Returns
+        -------
+            Image data tensor (other, c, z, y, x)
+        """
+        A = qdata[0].unsqueeze(0)
+        B = qdata[1].unsqueeze(0)
+        T1 = qdata[2].unsqueeze(0)
+        t = torch.Tensor(self.ti) + (self.n - 1) * self.rr
+        t = rearrange(t, 't -> t 1 1 1 1 1')
+        T1_star = torch.div(T1, (torch.div(B, A) - 1))
```
fzimmermann89: see above

src/mrpro/operators/models/InversionRecovery.py
fzimmermann89: super().__init__()

fzimmermann89 (dismissed):
I would use the shorthand for all math operations instead of torch.div etc.

I would try to have all the models work with multiple dimenens in other.

The ti should 
a) be a tensor
b) all parameters of the model should  torch.parameter. this allows easy move to gpu and gradient calculation.
So it should be 
`self.ti=torch.nn.Parameter(ti, requires_grad=ti.requires_grad)` (and similiar)

src/mrpro/operators/models/Molli.py
```diff
@@ -0,0 +1,57 @@
...
+        ti
+            inversion times.
+        n
+            image number within the Look-Locker experiment
+        rr
+            heartbeat interval
+
+        """
+        super().__init__()
+        self.ti = torch.nn.Parameter(ti, requires_grad=ti.requires_grad)
```
ckolbPTB: also the MOLLI signal model only requires ti as a parameter. These are the actual inversion times of the images already taking the timing of the inversion pulse and the cardiac cycle length into account.

src/mrpro/operators/models/Molli.py
```diff
@@ -0,0 +1,57 @@
...
+            params: (a, b, t1)
+
+        Returns
+        -------
+            Image data tensor (other, c, z, y, x)
+        """
+        a, b, t1 = qdata.unsqueeze(1)
+        t = self.ti + (self.n - 1) * self.rr
+        t = t[(...,) + (None,) * (qdata[0].ndim)]
+        t1_star = t1 / ((b / a) - 1)
```
ckolbPTB: a lot of potential div by 0 cases here, maybe also add the +1e-10 as above?

tests/operators/models/test_models.py
```diff
@@ -0,0 +1,182 @@
+import pytest
+import torch
+
+from mrpro.operators.models import InversionRecovery
+from mrpro.operators.models import Molli
+from mrpro.operators.models import SaturationRecovery
+from tests import RandomGenerator
+
+
+def test_saturation_recovery_t_0():
```
ckolbPTB: Would be great if we could parametrize these tests, such that we say something like:

```
@pytest.mark.parametrize(
    'ti, result',
    [
        # short ti
        ( 0, '0')
        # long ti
        ( 20, 'm0')
    ]
```

tests/operators/models/test_models.py
```diff
@@ -0,0 +1,182 @@
...
+
+    # Generate signal model and torch tensor for comparison
+    model = SaturationRecovery(ti)
+    image = model.forward(random_tensor)
+    zeros = torch.zeros_like(image)
+
+    # Assert closeness to zero
+    torch.testing.assert_close(image, zeros)
+
+    assert image is not None
```
ckolbPTB: We do not need this and the next assert because if image is None the above asserting will also fail. To ensure the correct size we could create a zeros-image with the correct shape rather than from image and then `torch.testing.assert_close(image, zeros)` will fail if the shape is wrong.

ckolbPTB (dismissed):
I think our convention is to use an underscore in the filename if the file only contains a single class def, e.g. `_Molli.py `rather than `Molli.py`

schuenke:
> I think our convention is to use an underscore in the filename if the file only contains a single class def, e.g. `_Molli.py `rather than `Molli.py`

The reason why we named the signal model files without the leading underscore was that these are classes that the enduser will access directly. However, I agree that we also included the leading underscore in classes like KData, which the enduser will use as well. 

So yeah, maybe we should add the "_" to the signal model filenames as well.

fzimmermann89:
for better visibility:

I would only do some eps-addition only if it removes the discontinuity, not just moves it.

Otherwise, I would either try to find a definition what our function should return for edgecases such as t1=0, and return that. Or return the nan if that is the best definition for these input values.
allowing t1=0 but failing for t1=-1e-10 is surprising.
constraining the inputs to only plausible values (such as t1>0) should imho not be done within the signal function forward.
(see also comment above)


If we stick with parameters as tensors ( still up for discussion...) I could see a use in adding a 'domain' to the signal models.
some meta information (in code, not only as a comment) that
- names of the parameters as string
- upper and lower bound of physically allowed values.

So, as an rough  example:
```python
@dataclass 
class BiophysicalParameterMeta
  name: str
  upper_bound: float
  lower_bound:float
  default = 1.

class Molli
   parameters = (BiophysicalParameterMeta('t1',0,np.inf,1.0),BiophysicalParameterMeta('m0',0,np.inf,1.0))
   ...
```
 or similar...
 this allows for automatic sensible defaults and constrains for our fitting...
 
 although this feels like typing without making use of pythons typing functions ;)

ckolbPTB:
> Otherwise, I would either try to find a definition what our function should return for edgecases such as t1=0, and return that. Or return the nan if that is the best definition for these input values.
allowing t1=0 but failing for t1=-1e-10 is surprising.

That is true. We need to ensure the edge case of T1 = 0 is handled properly. T1 < 0 should lead to whatever it leads to and should be a problem of the fitting algorithm / user.

ckolbPTB:
> class BiophysicalParameterMeta

I would not make this part of the model because this might lead to more confusion than it helps. A and B in in MOLLI e.g. are not really biophysical parameters but combine different parameters. Also the default value is not well defined. T1 = 1 as default only makes sense if the timings are done in seconds. By default the timings in the dicom header are in ms. Also M0 depends on the magnitude of the images which can be much larger than one when read from dicom. 

I would suggest to leave this out of the model and discuss it again when we finalise QData. There it would make more sense for me to have these dype of meta information.

JoHa0811 (approved):

PR #131, opened 2023-12-12
Lbfgs box

### LINKED ISSUE #115: Model fitting for qMRI
Need an algorithm to carry out (voxel-wise?) fitting of image data to obtain parametric maps

schuenke:
closes #115


fzimmermann89:
Quick question:  Do we know want to have a Nonlinear operator as a class?

I thought we decided to only have operator and the special case Linearoperator...

fzimmermann89:
Maybe we should copy the initial value tensor?

It might be unexpected that the tensor is changed during the runtime..

tests/algorithms/test_lbfgs.py
```diff
@@ -0,0 +1,77 @@
...
+
+import pytest
+import torch
+import torch.nn as nn
+
+from mrpro.algorithms import lbfgs
+from tests import RandomGenerator
+
+
+class Rosenbrock(nn.Module):
```
schuenke: should inherit from Operator

src/mrpro/algorithms/_lbfgs.py
```diff
@@ -0,0 +1,86 @@
...
+        line search algorithm, either ‘strong_wolfe’ or None,
+        by default "strong_wolfe"
+
+    Returns
+    -------
+        optimized parameter p
+    """
+
+    # enable gradient calculation
+    for p in params:
```
fzimmermann89: detach().clone().requiere_grad_() to create new tensors 
enabling require grad on the inputs is surprising and will break.

fzimmermann89: or its the job of the calling function (which will also be part of mrpro)

so either way, I would recommend not having this loop here.

src/mrpro/algorithms/_lbfgs.py
```diff
@@ -0,0 +1,86 @@
...
+        max_iter=max_iter,
+        max_eval=max_eval,
+        tolerance_grad=tolerance_grad,
+        tolerance_change=tolerance_change,
+        line_search_fn=line_search_fn,
+    )
+
+    def closure():
+        lbfgs_.zero_grad()
+        objective = f(p)
```
fzimmermann89: should be *params

tests/algorithms/test_lbfgs.py
```diff
@@ -0,0 +1,85 @@
...
+    def __init__(self, a=1, b=100):
+        super().__init__()
+        self.a = a
+        self.b = b
+
+    def forward(self, x1, x2):
+        fval = (self.a - x1) ** 2 + self.b * (x[:, 1] - x2**2) ** 2
```
fzimmermann89: x[:, 1]

is wrong

tests/algorithms/test_lbfgs.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+@pytest.mark.parametrize(
+    'a, b',
+    [(1.0, 100.0), (1.0, 50.0), (1.0, 25.0), (1.0, 2.0)],
+)
+def test_lbfgs_rosenbrock(a, b):
+    """Test lbfgs functionality."""
+
+    # with pytest.raises(ImportWarning):  # ToDo: remove this when fixed
```
ckolbPTB: In test_adam this is still needed? Pls remove any commented code before merging

tests/algorithms/test_lbfgs.py
```diff
@@ -0,0 +1,74 @@
...
+    )
+
+    # minimizer of Rosenbrock function
+    sol = torch.tensor([a, a**2])
+
+    # obtained solution
+    x = torch.tensor(params)
+
+    # check if they are close
+    tol = 1.0
```
ckolbPTB: This seems very arbitrary. Would it make sense to use a relative error?

src/mrpro/algorithms/_adam.py
```diff
@@ -0,0 +1,99 @@
+"""ADAM for solving non-linear minimization problems."""
+
+# Copyright 2023 Physikalisch-Technische Bundesanstalt
```
ckolbPTB: should actually be 2024...probably wrong in all files written this year...

src/mrpro/algorithms/_adam.py
```diff
@@ -0,0 +1,99 @@
...
+        whether the fused implementation (CUDA only) is used. Currently, torch.float64, torch.float32,
+        torch.float16, and torch.bfloat16 are supported., by default None
+
+    Returns
+    -------
+        list of optimized parameters
+    """
+
+    # define Adam routine
+    adam_ = Adam(
```
fzimmermann89: why is it adam_?

src/mrpro/operators/_ConstraintsOp.py
```diff
@@ -0,0 +1,129 @@
...
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+
+import torch
+
+from mrpro.operators import Operator
+
+
+# TODO: inheriting from Operator throws "_backward_hooks"-error
```
fzimmermann89: is this still true?

src/mrpro/operators/_ConstraintsOp.py
```diff
@@ -0,0 +1,129 @@
...
+
+        for lb, ub in bounds:
+            if (ub is not None and lb is not None) and lb > ub:
+                raise ValueError(
+                    f'bounds should be ( (a1,b1), (a2,b2), ...) with ai<=bi if neither ai or bi is None;\
+                        \nbound tuple {lb,ub} is invalid'
+                )
+
+    @staticmethod
+    def sigmoid_transf(x: torch.Tensor, bounds: tuple[float, float], beta: float = 1.0) -> torch.Tensor:
```
fzimmermann89: what means transf?

src/mrpro/operators/_ConstraintsOp.py
```diff
@@ -0,0 +1,129 @@
...
+    def sigmoid_transf_inv(x: torch.Tensor, bounds: tuple[float, float], beta: float = 1.0) -> torch.Tensor:
+        """Inverse of 'sigmoid_transf."""
+
+        return -torch.log(1.0 / ((x - bounds[0]) / (bounds[1] - bounds[0])) - 1.0) / beta
+
+    @staticmethod
+    def softplus_transf(x: torch.Tensor, lbound: float, beta: float = 0.1) -> torch.Tensor:
+        """Constrain x to be in (bound,infty)."""
+
+        return lbound + 1.0 / beta * torch.log(1 + torch.exp(beta * x))
```
fzimmermann89: I would prefer to use pytorch softplus as it can do
`beta * torch.log(1 + torch.exp(beta * x)` in one call, right?

src/mrpro/operators/_ConstraintsOp.py
```diff
@@ -0,0 +1,129 @@
...
+
+    @staticmethod
+    def softplus_transf_inv(x: torch.Tensor, lbound: float, beta: float = 0.1) -> torch.Tensor:
+        """Inverse of 'softplus_transf."""
+
+        return torch.log(torch.exp(beta * (x - lbound)) - 1.0) / beta
```
fzimmermann89: would expm1 make this more numerically stable var small values of beta * (x - lbound) ?

src/mrpro/operators/_ConstraintsOp.py
```diff
@@ -0,0 +1,129 @@
...
+                raise ValueError(
+                    f'bounds should be ( (a1,b1), (a2,b2), ...) with ai<=bi if neither ai or bi is None;\
+                        \nbound tuple {lb,ub} is invalid'
+                )
+
+    @staticmethod
+    def sigmoid_transf(x: torch.Tensor, bounds: tuple[float, float], beta: float = 1.0) -> torch.Tensor:
+        """Constraint x to be in the range given by 'bounds'."""
+
+        return bounds[0] + (bounds[1] - bounds[0]) * (1.0 / (1.0 + torch.exp(-beta * x)))
```
fzimmermann89: please consider using torch sigmoid for performance, nan handling and numerical stability

src/mrpro/operators/_ConstraintsOp.py
fzimmermann89: I think it is 
x + torch.log(-torch.expm1(-x))
to be stable instead of 
torch.log(torch.exp(x )) - 1.0)

src/mrpro/operators/functionals/_L2_data_discrepancy.py
```diff
@@ -0,0 +1,38 @@
...
+import torch
+import torch.nn.functional as F
+
+from mrpro.operators import Operator
+
+
+class L2_data_discrepancy(Operator):
+    def __init__(self, data: torch.Tensor) -> None:
+        super().__init__()
+        """Squared L2-norm loss function, i.e.
```
fzimmermann89: I think the descirption does not match what this function does.

It should not divide by the number of elements of it is supposed to be the L2 norm, right?

If I am not mistaken, it returns different values for 1. and 1.+0.j which should also not be the case ..

tests/algorithms/test_adam.py
```diff
@@ -0,0 +1,156 @@
...
+from tests import RandomGenerator
+from tests.operators._OptimizationTestFunctions import Rosenbrock
+from tests.phantoms._SaturationRecoveryEllipsePhantom import SaturationRecoveryEllipsePhantom2D
+
+
+@pytest.mark.parametrize(
+    'a, b',
+    [(1.0, 100.0), (2.0, 50.0), (5.0, 10.0), (10.0, 5.0)],
+)
+def test_adam_rosenbrock(a, b):
```
fzimmermann89: this is pretty much the same for adam and lbfgs, right?
can we make it one test parameterized by the optimisation function or at least refactor it such that we don't have that much duplicate code?

tests/algorithms/test_adam.py
```diff
@@ -0,0 +1,156 @@
...
+
+        # minimizer of Rosenbrock function
+        sol = torch.tensor([a, a**2])
+
+        # obtained solution
+        x = torch.tensor(params)
+
+        # check if they are close
+        tol = 1e-1
+        mse = F.mse_loss(sol, x)
```
fzimmermann89: please use torch.testing.assert_close or somehting similiar, instead of manually calculating the difference.

tests/algorithms/test_adam.py
```diff
@@ -0,0 +1,156 @@
...
+        lr = 1e-2
+        max_iter = 1000
+
+        random_generator = RandomGenerator(seed=0)
+
+        # generate two-dimensional test data
+        x1 = random_generator.float32_tensor(size=(1,))
+        x2 = random_generator.float32_tensor(size=(1,))
+
+        # enable gradient calculation
```
fzimmermann89: General Remark: I would prefer not to have comments that trivially describe what is being done if it is obvious.

        # define Rosenbrock function
                # call adam

tests/algorithms/test_adam.py
```diff
@@ -0,0 +1,156 @@
...
+        x1.requires_grad = True
+        x2.requires_grad = True
+
+        params = [x1, x2]
+
+        # define Rosenbrock function
+        f = Rosenbrock(a, b)
+
+        # call adam
+        params = adam(
```
fzimmermann89: why not call it result?

tests/algorithms/test_adam.py
```diff
@@ -0,0 +1,156 @@
...
+            params,
+            max_iter=max_iter,
+            lr=lr,
+        )
+
+        # minimizer of Rosenbrock function
+        sol = torch.tensor([a, a**2])
+
+        # obtained solution
+        x = torch.tensor(params)
```
fzimmermann89: why do we need this?
why is the comment required? could we instead rename params such that it is self documenting?

tests/algorithms/test_adam.py
```diff
@@ -0,0 +1,156 @@
...
+        # call adam
+        params = adam(
+            f,
+            params,
+            max_iter=max_iter,
+            lr=lr,
+        )
+
+        # minimizer of Rosenbrock function
+        sol = torch.tensor([a, a**2])
```
fzimmermann89: consider renaming it to analytical_solution or similiar.
I do not see a point in saving a couple of letters here in the variable name and needing a comment to describe what sol means

tests/algorithms/test_adam.py
```diff
@@ -0,0 +1,156 @@
...
+        (128, 256, 0),
+        (128, 256, 1),
+        (32, 64, 0),
+        (32, 64, 1),
+    ],
+)
+def test_adam_sat_recovery(ny, nx, bounds_flag):
+    """Test lbfgs functionality for a saturation recovery example."""
+
+    # image dimension
```
fzimmermann89: if you feel im_shape is not easy to understand, please consider remaining it to image_dimension instead of commenting its meaning

tests/phantoms/_SaturationRecoveryEllipsePhantom.py
```diff
@@ -0,0 +1,86 @@
...
+
+from mrpro.data import SpatialDimension
+from mrpro.operators import Operator
+from mrpro.operators.models import SaturationRecovery
+from mrpro.phantoms import EllipsePhantom
+from tests._RandomGenerator import RandomGenerator
+
+
+# TODO: maybe make it possible to use more and different ellipses?
+class SaturationRecoveryEllipsePhantom2D:
```
fzimmermann89: do we need to be this fancy here?

Do we need a special saturation recovery phantom?

fzimmermann89: consider maybe a T1Phantom?
Do we need to add a polynomial etc to the ellipses to have a simple test object?

tests/phantoms/_SaturationRecoveryEllipsePhantom.py
```diff
@@ -0,0 +1,86 @@
...
+    def __init__(self, im_shape: SpatialDimension[int], ti: torch.Tensor) -> None:
+        self.im_shape = im_shape
+        self.ti = ti
+
+    def get_sat_recovery_model(self) -> Operator:
```
fzimmermann89: I fail to see the purpose of this complexity.

tests/phantoms/_SaturationRecoveryEllipsePhantom.py
fzimmermann89: we could (after maybe removing some of the complexity here) make this a fixture that is used by the test, https://docs.pytest.org/en/6.2.x/fixture.html

tests/algorithms/test_adam.py
```diff
@@ -0,0 +1,156 @@
...
+        # check if they are close
+        tol = 1e-1
+        mse = F.mse_loss(sol, x)
+        assert mse < tol
+
+
+@pytest.mark.parametrize(
+    'ny, nx, bounds_flag',
+    [
+        (128, 128, 0),
```
fzimmermann89: Does this parameterization make any sense?

Why should adam fail with a different number of pixels?
this only increases runtime of the tests

src/mrpro/operators/_ConstraintsOp.py
```diff
@@ -0,0 +1,129 @@
...
+#   limitations under the License.
+
+
+import torch
+
+from mrpro.operators import Operator
+
+
+# TODO: inheriting from Operator throws "_backward_hooks"-error
+class ConstraintsOp(Operator):
```
fzimmermann89: this is missing a lot of tests, right?

fzimmermann89 (changes requested):
I believe there is some room for cleanup...

fzimmermann89:
"With pytest.raises" means we want this to cause an error.
We should only use this to check if we throw errors where we want to throw them, not to ignore errors..

tests/algorithms/test_lbfgs.py
koflera: I am not sure... when run locally (on e81330), I seem to need it, after having rebased to main. Before it seemed not necessary.

tests/algorithms/test_lbfgs.py
koflera: this test was removed

src/mrpro/algorithms/_adam.py
koflera: changed to `optim`

src/mrpro/operators/_ConstraintsOp.py
koflera: renamed to `sigmoid`

src/mrpro/operators/_ConstraintsOp.py
koflera: i changed that... the only disadvantage is that now it is necessary to use `int `for `beta` which could have been nice to be able to decrease it

src/mrpro/operators/functionals/_L2_data_discrepancy.py
koflera: good point. I changed it to mse_data_discrepancy, added remarks (also about the identification of C as R^2) and added some tests

tests/algorithms/test_adam.py
koflera: i deleted this test and replaced it by one where we only have the standard combination of a and b, i.e. (`a=1, b=100`) and test different non-linear optimizers

src/mrpro/algorithms/_adam.py
```diff
@@ -0,0 +1,99 @@
...
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+import torch
+from torch.optim import Adam
+
+
+def adam(
+    f,
```
fzimmermann89: Type hint for f?

tests/phantoms/_SaturationRecoveryEllipsePhantom.py
koflera: i removed the entire test as it does not test the functionality of the functions but it could rather be used as an example. I replace it with a one where `lbfgs `and `adam `are tested also using the `ConstraintsOp`

src/mrpro/algorithms/_adam.py
```diff
@@ -0,0 +1,99 @@
...
+    betas: tuple[float, float] = (0.9, 0.999),
+    eps: float = 1e-8,
+    weight_decay: float = 0,
+    amsgrad: bool = False,
+    foreach: bool | None = None,
+    maximize: bool = False,
+    capturable: bool = False,
+    differentiable: bool = False,
+    fused: bool | None = None,
+) -> list[torch.Tensor]:
```
fzimmermann89, line 35: Should this be an inplace function that modifies params or should it return params?

Currently is does both, which might be a bit confusing.

tests/algorithms/test_optimizers.py
```diff
@@ -0,0 +1,80 @@
...
+from mrpro.algorithms import lbfgs
+from mrpro.operators import ConstraintsOp
+from tests import RandomGenerator
+from tests.operators._OptimizationTestFunctions import Rosenbrock
+
+
+@pytest.mark.parametrize(
+    'optimizer_name, bounds_flag',
+    [
+        ('adam', 0),
```
fzimmermann89: Question: why is this not a bool?
Like contraint_value True or False?
Do bools not work there?

tests/algorithms/test_optimizers.py
```diff
@@ -0,0 +1,80 @@
...
+        cop = ConstraintsOp(bounds=((-1, 1),)) if bounds_flag else None
+
+        def f(x):  # TODO: Use @ later
+            return rosen_brock(x) if cop is None else rosen_brock(cop(x))
+
+        # hyperparams for optimizer
+        lr = 1e-2
+        max_iter = 250
+
+        if optimizer_name == 'adam':
```
fzimmermann89: Can we use a getattr here or even better, in the parameteize decorator already the function as a function instead of a string?

tests/operators/test_ConstraintsOp.py
```diff
@@ -0,0 +1,108 @@
...
+    random_generator = RandomGenerator(seed=0)
+
+    # random tensor with arbitrary values
+    x = random_generator.float32_tensor(size=(36,))
+
+    # define constraints operator using the bounds
+    Cop = ConstraintsOp(bounds)
+
+    # transform tensor to be component-wise in the range defined by bounds
+    (cx,) = Cop((x,))
```
fzimmermann89: This should be cop(x)

Operators take tensors as input

src/mrpro/operators/_ConstraintsOp.py
```diff
@@ -0,0 +1,145 @@
...
+            elif lb is None and ub is not None:
+                # case (None,b); corresponds to (-\infty, b)
+                xc.append(ub - self.softplus(-x[i], beta=self.beta_softplus))
+            elif lb is None and ub is None:
+                # case (None,None); corresponds to (-\infty, \infty), i.e. no transformation
+                xc.append(x[i])
+
+        return tuple(xc)
+
+    def inverse(self, xc: tuple[torch.Tensor, ...]) -> tuple[torch.Tensor, ...]:
```
fzimmermann89: This should be *xc
I. E. multiple tensors

src/mrpro/operators/_ConstraintsOp.py
```diff
@@ -0,0 +1,145 @@
...
+
+        return F.softplus(x, beta=beta)
+
+    @staticmethod
+    def softplus_inverse(x: torch.Tensor, beta: int = 1) -> torch.Tensor:
+        """Inverse of 'softplus_transformation."""
+
+        return beta * x + torch.log(-torch.expm1(-beta * x))
+
+    def forward(self, x: tuple[torch.Tensor, ...]) -> tuple[torch.Tensor, ...]:
```
fzimmermann89: This should be *x

fzimmermann89 (changes requested):
Operators take multiple tensors as input, not a tuple of tensors.

So forward(*x) to pack all of these into a tue you can then loop over in the function

src/mrpro/algorithms/_adam.py
koflera: true... maybe better to return params then? i don't have a strong opinion on that, though

tests/algorithms/test_optimizers.py
koflera: done

tests/algorithms/test_optimizers.py
koflera: done

src/mrpro/algorithms/_adam.py
fzimmermann89: neither do I.
Also slight preference for not modifying the input.
But then we need a 
`p.detach().clone().requires_grad_(True)`
in the optimisation functions

tests/algorithms/test_lbfgs.py
fzimmermann89: it depends on the pytorch version in the environment.
if you don't `pip install` again (or update the environment) you will continue to use the old pytorch version which is older then our requirements in the pyproject.toml

src/mrpro/operators/_ConstraintsOp.py
fzimmermann89: I believe this is bug in pytorch typehints /docstring.
It should work with beta being float.

fzimmermann89: Instead, you could also use:

-(1/beta) * torch.nn.functional.logsigmoid(-beta*x)

which is also softplus and numerically stable afaik

fzimmermann89:
mypy fails with:
```
src/mrpro/operators/_ConstraintsOp.py:22: error: Bad number of arguments, expected: at least 1, given: 0  [type-arg]
src/mrpro/operators/functionals/_mse_data_discrepancy.py:21: error: Bad number of arguments, expected: at least 1, given: 0  [type-arg]
```

the tests fail because of #174 (not our fault, also tracked on ismrmrd python repo)

src/mrpro/operators/_ConstraintsOp.py
koflera: done

src/mrpro/operators/_ConstraintsOp.py
koflera: done

tests/operators/test_ConstraintsOp.py
koflera: done

src/mrpro/algorithms/_adam.py
koflera: done

src/mrpro/algorithms/_adam.py
koflera: done

ckolbPTB:
> mypy fails with:
> 
> ```
> src/mrpro/operators/_ConstraintsOp.py:22: error: Bad number of arguments, expected: at least 1, given: 0  [type-arg]
> src/mrpro/operators/functionals/_mse_data_discrepancy.py:21: error: Bad number of arguments, expected: at least 1, given: 0  [type-arg]
> ```

Can we use something like "# type: ignore [type-arg]" here to satisfy mypy if this is due to a problem in 3rd party code?

src/mrpro/operators/_ConstraintsOp.py
koflera: done

src/mrpro/operators/_ConstraintsOp.py
koflera: done

src/mrpro/algorithms/_adam.py
```diff
@@ -0,0 +1,101 @@
+"""ADAM for solving non-linear minimization problems."""
+
+# Copyright 2024Physikalisch-Technische Bundesanstalt
```
fzimmermann89: missing space

src/mrpro/algorithms/_adam.py
```diff
@@ -0,0 +1,101 @@
...
+#   limitations under the License.
+
+import torch
+from torch.optim import Adam
+
+from mrpro.operators import Operator
+
+
+def adam(
+    f: Operator[*tuple[torch.Tensor, ...], tuple[torch.Tensor, ...]],
```
fzimmermann89: I think it should be
```suggestion
    f: Operator[*tuple[torch.Tensor, ...], tuple[torch.Tensor]],
```
i.e. f can only return a single tensor.

src/mrpro/algorithms/_adam.py
```diff
@@ -0,0 +1,101 @@
...
+    fused: bool | None = None,
+) -> list[torch.Tensor]:
+    """Adam for non-linear minimization problems.
+
+    Parameters
+    ----------
+    f
+        scalar-valued function to be optimized
+    params
+        list of parameters to be optimized
```
fzimmermann89: make it clear that these will not be touched.
so maybe `intial_points` ? or at least mention it in the docstring

src/mrpro/operators/_ConstraintsOp.py
```diff
@@ -0,0 +1,145 @@
...
+        return F.sigmoid(beta * x)
+
+    @staticmethod
+    def sigmoid_inverse(x: torch.Tensor, beta: float = 1.0) -> torch.Tensor:
+        """Constraint x to be in the range given by 'bounds'."""
+
+        return torch.logit(x) / beta
+
+    @staticmethod
+    def softplus(x: torch.Tensor, beta: int = 1) -> torch.Tensor:
```
fzimmermann89: beta: float
and silence the mypy error with a comment. This will soon be float in pytorch
https://github.com/pytorch/pytorch/issues/120175

src/mrpro/operators/_ConstraintsOp.py
```diff
@@ -0,0 +1,145 @@
...
+        """Constrain x to be in (bound,infty)."""
+
+        return -(1 / beta) * torch.nn.functional.logsigmoid(-beta * x)
+
+    @staticmethod
+    def softplus_inverse(x: torch.Tensor, beta: int = 1) -> torch.Tensor:
```
fzimmermann89: beta:float

src/mrpro/operators/functionals/_mse_data_discrepancy.py
```diff
@@ -0,0 +1,67 @@
...
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+import torch
+import torch.nn.functional as F
+
+from mrpro.operators import Operator
+
+
+class mse_data_discrepancy(Operator[torch.Tensor, tuple[torch.Tensor, ...]]):
```
fzimmermann89: `Operator[*tuple[torch.Tensor,...], tuple[torch.Tensor]]`

It can take in a not specified number of tensors, but will return always exactly one tensor.

fzimmermann89 (approved):

fzimmermann89:
- There are some typing issues I think. The operator types are not optimal, see my comments above.
The functionals can be typed as returning only a single tensor.
- softplus can soon take in a float in pytorch. I reported the bug in their typing (the code always worked with float) and it got fixed. maybe we can already type beta as float and set  `#type: ignore SOMETHING` to ignore the error ..

Otherwise, I  think the code makes sense. You also convinced my that it makes sense to have the `with pytest.raises` there as a reminder to remove it once this line does not raise an error any more.

src/mrpro/operators/functionals/_mse_data_discrepancy.py
koflera: should this really take an arbitrary number of tensors as input? I definetely agree with `tuple[torch.Tensor]` as output, but if this function implements `1/N * || . - data||_2^2`, then the input of the operator should be of the same as data, i.e only one tensor.

if you are thinking about something like `||Ax - y||_2^2`, then the user should make sure that `Ax ` is only one tensor in order to be used with this function

fzimmermann89: agree

src/mrpro/operators/functionals/_mse_data_discrepancy.py
```diff
@@ -0,0 +1,67 @@
...
+    """MSE loss function.
+
+        This class implements the function
+            1./N * || . - data ||_2^2,
+        where N equals to the number of elements of the tensor.
+
+        N.B. if one of data or input is complex-valued, we
+        identify the space C^N with R^2N and multiply the output
+        by 2. By this, we achieve that for example
+            MSE(1) = MES(1+1j*0) = 1.
```
ckolbPTB: typo: MES -> MSE

src/mrpro/operators/functionals/_mse_data_discrepancy.py
```diff
@@ -0,0 +1,67 @@
...
+        # observed data
+        self.data = data
+
+    def forward(self, x: torch.Tensor) -> tuple[torch.Tensor]:
+        """Calculate the MSE of the input.
+
+        Parameters
+        ----------
+        x
+            tensor whose mse with respect to data should be calculated
```
ckolbPTB: typo: mse -> MSE

src/mrpro/operators/functionals/_mse_data_discrepancy.py
```diff
@@ -0,0 +1,67 @@
...
+
+        Returns
+        -------
+            MSE of the different of the input and the given data
```
ckolbPTB: typo: different -> difference

tests/algorithms/test_optimizers.py
```diff
@@ -0,0 +1,75 @@
...
+            f,
+            params_init,
+            max_iter=max_iter,
+            lr=lr,
+        )
+
+        # test if the obtained solution is close to the analytical
+        torch.testing.assert_close(torch.tensor(params_result), analytical_sol)
+
+        # test if the optimizer didn't change the initialization but returned copies
```
ckolbPTB: from the text description it is not clear if the optimizer should return a copy or not. maybe: "ensure that the optimizer returned a copy and did not change the initialisation values

tests/operators/functionals/test_mse_functional.py
```diff
@@ -0,0 +1,45 @@
...
+        ((0.0 + 1j * 1, 0.0), (0.0 + 1j * 1, 0.0), (0.0)),  # same tensors; both complex-valued with real part=0
+    ],
+)
+def test_mse_functional(data, x, expected_mse):
+
+    # create mse functional based on data, i.e.
+    # 1/N*|| . - data||_2^2
+    mse_dc = mse_data_discrepancy(torch.tensor(data))
+
+    # calculate mse and compared to expected value
```
ckolbPTB: typo: compared -> compare, mse -> MSE

fzimmermann89:
I addressed some of the review comments regarding the comments ( ;-) ) and added tests for what I think should be illegal bounds.
```
        ((1.0, -1.0), (-1.0, 1.0)),  # first one is invalid
        ((-1.0, 1.0), (1.0, -1.0)),  # second one is invalid
        ((1.0, 1.0),),   ### QUESTION: IS THIS VALID OR INVALID? ####
        ((torch.nan, 1),),
        ((-1, torch.nan),),
        ((torch.inf, -torch.inf),),
```
            
Some of these seem not to raise an exception in the constraintOp. If they are actually *valid*, feel free to remove them from the test.
Also, do we want to allow inf and -inf for bounds? This feels (for me) more natural to write than None for an unbounded variable.

fzimmermann89:
@koflera :
Maybe clean up the commit message next time you merge a PR...

PR #133, opened 2023-12-14
update supported python version to 3.10 & 3.11

### LINKED ISSUE #100: pytorch not compatible with python3.12
Add a line suggesting the python version to use in README.
pytorch currently supports Python 3.8-3.11 (https://pytorch.org/get-started/locally/).

schuenke:
Pytorch does not support Python 3.12, yet.

Therefore, I limited the python version in pyproject.toml to ">3.10, <3.12" and adjusted the badge in the README

closes #100

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ MR image reconstruction and processing package specifically developed for PyTorc
 
 This package supports ismrmrd-format for MR raw data. All data containers utilize PyTorch-tensors to ensure easy integration in PyTorch-based network schemes.
 
-![Python](https://img.shields.io/badge/python-3.10+-blue)
+![Python](https://img.shields.io/badge/python-3.10%20|%203.11-blue)
 [![License](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](https://opensource.org/licenses/Apache-2.0)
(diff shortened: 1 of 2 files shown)


fzimmermann89:
I would NOT limit our compatibility because a dependency does not support it. Only if our code is not ready for a new version.

I'd limit the python compatibility to `>=3.10,<3.2`

I agree with the batch / not mentioning python 3.12 in the readme, not to confuse users,
 but veto the requirements change.

Furthermore, pytorch nightly supports python 3.12.

fzimmermann89 (changes requested):
I'd prefer not to limit our compatibility based on compartibility of dependencies. This is something the package manager should take care of.

schuenke:
> I would NOT limit our compatibility because a dependency does not support it. Only if our code is not ready for a new version.
> 
> I'd limit the python compatibility to `>=3.10,<3.2`
> 
> I agree with the batch / not mentioning python 3.12 in the readme, not to confuse users, but veto the requirements change.
> 
> Furthermore, pytorch nightly supports python 3.12.

I would argue that we currently cannot test our code in 3.12, which means that in general we cannot guarantee that our code is 3.12 compatible, which IMO means that we do not support it, no ?!

fzimmermann89:
Yes, but semantic versioning should mean all code running under 3.11 should run under 3.12, right?

Edit: But python is not using semver... 

Ok,....

fzimmermann89 (approved):

PR #136, opened 2023-12-20
Gpu support

### LINKED ISSUE #67: Move KData to device
Implement standard pytorch functionality to move data + ktraj + header between cpu and gpu (e.g.  .to(device)). Add flags to select of ktraj and header are moved.

ckolbPTB:
I tested a simple reconstruction pipeline on a GPU. Only needed a few fixes for it to run. 

I have implemented a very basic .cuda() and .cpu() functionality for KTrajectory and KData. If we are more or less happy with this approach I would extend this to IData, CsmData...

I also added tests which are marked "cuda". In order to run only the cuda-tests `pytest -m cuda` can be used.

I added the pytest flag `-m "not cuda"` to pytest.yml such that these tests are excluded from CI.

### DIFF
--- a/src/mrpro/data/_Data.py
+++ b/src/mrpro/data/_Data.py
@@ -1,0 +1,73 @@
+"""Base class for data objects."""
+
+# Copyright 2023 Physikalisch-Technische Bundesanstalt
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#       http://www.apache.org/licenses/LICENSE-2.0
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+from __future__ import annotations
+
+import dataclasses
+from abc import ABC
+from typing import Any
+
+import torch
+
+
+@dataclasses.dataclass(slots=True, frozen=True)
+class Data(ABC):
+    """A general data class with field data and header."""
+
+    data: torch.Tensor
+    header: Any
+
+    def to(self, *args, **kwargs) -> Data:
+        """Perform dtype and/or device conversion of data.
+
+        A torch.dtype and torch.device are inferred from the arguments
+        of self.to(*args, **kwargs). Please have a look at the
+        documentation of torch.Tensor.to() for more details.
+        """
+        return Data(header=self.header, data=self.data.to(*args, **kwargs))
+
+    def cuda(
+        self,
+        device: torch.device | None = None,
+        non_blocking: bool = False,
+        memory_format: torch.memory_format = torch.preserve_format,
+    ) -> Data:
+        """Create copy of object with data in CUDA memory.
+
+        Parameters
+        ----------
+        device
+            The destination GPU device. Defaults to the current CUDA device.
+        non_blocking
+            If True and the source is in pinned memory, the copy will be asynchronous with respect to the host.
+            Otherwise, the argument has no effect.
+        memory_format
+            The desired memory format of returned tensor.
+        """
+        return Data(
+            header=self.header,
+            data=self.data.cuda(
+                device=device, non_blocking=non_blocking, memory_format=memory_format
+            ),  # type: ignore [call-arg]
+        )
+
+    def cpu(self, memory_format: torch.memory_format = torch.preserve_format) -> Data:
+        """Create copy of object in CPU memory.
+
+        Parameters
+        ----------
+        memory_format
+            The desired memory format of returned tensor.
+        """
+        return Data(header=self.header, data=self.data.cpu(memory_format=memory_format))  # type: ignore [call-arg]
--- a/src/mrpro/data/_DcfData.py
+++ b/src/mrpro/data/_DcfData.py
@@ -98,9 +98,9 @@ class DcfData:
 
         # Calculate dcf only for unique k-space positions
         traj_dim = traj.shape
-        traj = np.round(traj.numpy(), decimals=UNIQUE_ROUNDING_DECIMALS)
-        traj = traj.reshape(dim, -1)
-        traj_unique, inverse, counts = np.unique(traj, return_inverse=True, return_counts=True, axis=1)
+        traj_numpy = np.round(traj.cpu().numpy(), decimals=UNIQUE_ROUNDING_DECIMALS)
+        traj_numpy = traj_numpy.reshape(dim, -1)
+        traj_unique, inverse, counts = np.unique(traj_numpy, return_inverse=True, return_counts=True, axis=1)
 
         # Especially in 3D, errors in the calculation of the convex hull can occur for edge points. To avoid this,
         # the corner points of a cube bounding box are added here. The bounding box is chosen very large to ensure these
@@ -135,7 +135,7 @@ class DcfData:
         # Sort dcf values back into the original order (i.e. before calling unique)
         dcf = np.reshape((dcf / counts)[inverse], traj_dim[1:])
 
-        return torch.tensor(dcf, dtype=torch.float32)
+        return torch.tensor(dcf, dtype=torch.float32, device=traj.device)
 
     @classmethod
     def from_traj_voronoi(cls, traj: KTrajectory) -> DcfData:
--- a/src/mrpro/data/_KData.py
+++ b/src/mrpro/data/_KData.py
@@ -171,3 +171,89 @@ class KData:
             )
 
         return cls(kheader, kdata, ktraj)
+
+    def to(
+        self,
+        device: torch.device | str | None = None,
+        dtype: None | torch.dtype = None,
+        non_blocking: bool = False,
+        copy: bool = False,
+        memory_format: torch.memory_format = torch.preserve_format,
+    ) -> KData:
+        """Perform dtype and/or device conversion of trajectory and data.
+
+        Parameters
+        ----------
+        device
+            The destination device. Defaults to the current device.
+        dtype
+            Dtype of the k-space data, can only be torch.complex64 or torch.complex128.
+            The dtype of the trajectory (torch.float32 or torch.float64) is then inferred from this.
+        non_blocking
+            If True and the source is in pinned memory, the copy will be asynchronous with respect to the host.
+            Otherwise, the argument has no effect.
+        copy
+            If True a new Tensor is created even when the Tensor already matches the desired conversion.
+        memory_format
+            The desired memory format of returned Tensor.
+        """
+        # Only complex64 and complex128 supported for kdata.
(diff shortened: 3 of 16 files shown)


fzimmermann89:
Some remarks:

I would try to match pytorch .cuda /.cpu / .to signature. What is the reasoning behind the custom handling of the default device? Can't we just pass on the arguments to the Tensor.cpu / .. function and let pytorch decide ?

We should test that for the operators Operator.cuda() ... works. This is inherited from nn.module. For it to work, all tensors should be registered as buffers or parameters, the commands will be passed tp submodules.

I would maybe also add .clone and .detach to the data classes.

ckolbPTB:
> I would try to match pytorch .cuda /.cpu / .to signature. 

You mean to also add `non_blocking` and `memory_format` to the signature `.cuda(device=None, non_blocking=False, memory_format=torch.preserve_format)`?

> Can't we just pass on the arguments to the Tensor.cpu / .. function and let pytorch decide ?

Yes, true.

ckolbPTB:
Issues/questions to be addressed:

- I get the following error from mypy: 
```
src/mrpro/data/_KData.py:195: error: Unexpected keyword argument "memory_format" for "cuda" of "_TensorBase"  [call-arg]
/home/runner/.cache/pre-commit/repo23gjt902/py_env-python3/lib/python3.10/site-packages/torch/_C/__init__.pyi:2009: note: "cuda" of "_TensorBase" defined here
```

- Not sure of the abstract class `Data` is really necessary

- I did not add a `to()` method to `KData` because I am not sure what it should do as in general `KData.data` and `KData.traj` have different dtypes.

src/mrpro/data/_Data.py
```diff
@@ -0,0 +1,73 @@
...
+        device: torch.device | None = None,
+        non_blocking: bool = False,
+        memory_format: torch.memory_format = torch.preserve_format,
+    ) -> Data:
+        """Create copy of object with data in CUDA memory.
+
+        Parameters
+        ----------
+        device
+            The destination GPU device. Defaults to the current CUDA device.
```
fzimmermann89, line 51:
```suggestion
            The destination GPU device. Defaults to the current device.
```
can also be CPU, mps, ...

src/mrpro/data/_DcfData.py
```diff
@@ -91,14 +91,17 @@ def _dcf_2d3d_voronoi(traj: torch.Tensor) -> torch.Tensor:
...
+
         # 2D and 3D trajectories supported
         dim = traj.shape[0]
         if dim not in (2, 3):
             raise ValueError(f'Only 2D or 3D trajectories supported, not {dim}D.')
 
         # Calculate dcf only for unique k-space positions
         traj_dim = traj.shape
-        traj = np.round(traj.numpy(), decimals=UNIQUE_ROUNDING_DECIMALS)
+        traj = np.round(traj.cpu().numpy(), decimals=UNIQUE_ROUNDING_DECIMALS)
```
fzimmermann89: Maybe we can use the oppertunity to give the input traj  and the numpy traj different names? 
This would allow mypy to consider them as array / tensor not array|tensor

this would also remove `traj_device = traj.device`

fzimmermann89:
reported that tensor.cpu has wrong typing information  at https://github.com/pytorch/pytorch/issues/118386

src/mrpro/data/_DcfData.py
ckolbPTB: I left the input traj as traj because this should be our standard traj tensor.

fzimmermann89:
Maybe we should use a decorator to mark tests that require CUDA? 
this is how its often done in other packages, but TBH I don't know how this is implemented then in pytest.

src/mrpro/data/_KData.py
```diff
@@ -171,3 +171,89 @@ def reshape_acq_data(data):
...
+        elif dtype == torch.complex128:
+            dtype_traj = torch.float64
+        else:
+            raise ValueError(f'dtype {dtype} not supported. Only torch.complex64 and torch.complex128 is supported.')
+
+        return KData(
+            header=self.header,
+            data=self.data.to(
+                device=device, dtype=dtype, non_blocking=non_blocking, copy=copy, memory_format=memory_format
+            ),  # type: ignore [call-overload]
```
fzimmermann89, line 215: is # type: ignore [call-overload] actually  requiered here?

src/mrpro/data/_KData.py
```diff
@@ -171,3 +171,89 @@ def reshape_acq_data(data):
...
+            Otherwise, the argument has no effect.
+        memory_format
+            The desired memory format of returned Tensor.
+        """
+        return KData(
+            header=self.header,
+            data=self.data.cuda(
+                device=device, non_blocking=non_blocking, memory_format=memory_format
+            ),  # type: ignore [call-arg]
+            traj=self.traj.cuda(device=device, non_blocking=non_blocking, memory_format=memory_format),
```
fzimmermann89, line 244: and why is # type: ignore [call-overload] not required here as well?

src/mrpro/data/_KData.py
```diff
@@ -171,3 +171,89 @@ def reshape_acq_data(data):
...
+
+        Parameters
+        ----------
+        memory_format
+            The desired memory format of returned Tensor.
+        """
+        return KData(
+            header=self.header,
+            data=self.data.cpu(memory_format=memory_format),  # type: ignore [call-arg]
+            traj=self.traj.cpu(memory_format=memory_format),
```
fzimmermann89, line 258: why is # type: ignore [call-overload] only required for one of the tensors?
Is mypy not checking the other one ?

fzimmermann89: ..this would mean that there is any issue with our typing somewhere else.
.data and .traj should behave the same wrt type hints.

tests/data/test_idata.py
```diff
@@ -44,3 +45,26 @@ def test_IData_from_kheader_and_tensor(random_kheader, random_test_data):
...
+    idat_cuda = idat.cuda()
+    assert idat_cuda.data.is_cuda
+
+
+@pytest.mark.cuda
+def test_IData_cpu(random_kheader, random_test_data):
+    """Move IData object to CUDA memory and back to CPU memory."""
+    idat = IData.from_tensor_and_kheader(data=random_test_data, kheader=random_kheader)
+    idat_cpu = idat.cuda().cpu()
+    assert idat_cpu.data.is_cpu
```
fzimmermann89, line 70: is_cpu() ?

tests/data/test_idata.py
```diff
@@ -44,3 +45,26 @@ def test_IData_from_kheader_and_tensor(random_kheader, random_test_data):
...
+    idat_complex128 = idat.to(dtype=torch.complex128)
+    assert idat_complex128.data.dtype == torch.complex128
+
+
+@pytest.mark.cuda
+def test_IData_cuda(random_kheader, random_test_data):
+    """Move IData object to CUDA memory."""
+    idat = IData.from_tensor_and_kheader(data=random_test_data, kheader=random_kheader)
+    idat_cuda = idat.cuda()
+    assert idat_cuda.data.is_cuda
```
fzimmermann89, line 62: can this ever fail?

shouldnt it be is_cuda() ?

tests/data/test_idata.py
fzimmermann89: () is missing

tests/data/test_kdata.py
```diff
@@ -80,3 +80,47 @@ def test_KData_modify_header(ismrmrd_cart, field, value):
...
+    assert kcuda.traj.ky.is_cuda
+    assert kcuda.traj.kx.is_cuda
+
+
+@pytest.mark.cuda
+def test_KData_cuda(ismrmrd_cart):
+    """Move KData object to CUDA memory."""
+    k = KData.from_file(ismrmrd_cart.filename, DummyTrajectory())
+    kcuda = k.cuda()
+    assert kcuda.data.is_cuda
```
fzimmermann89, line 113: see above...

src/mrpro/data/_KData.py
ckolbPTB: no .data is of type Tensor but .traj is of type Trajectory

src/mrpro/data/_KData.py
ckolbPTB: yes it is needed. mypy is also configured to throw an error if a type ignore is not needed

tests/data/test_idata.py
ckolbPTB: is_cpu and is_cuda is a bool not a function

src/mrpro/data/_KData.py
fzimmermann89: which argment is the issue? 
I try to get this fixed in pytorch, but to looked ok to me.
` def to(self, device: Optional[DeviceLikeType] = None, dtype: Optional[_dtype] = None, non_blocking: _bool = False, copy: _bool = False, *, memory_format: Optional[torch.memory_format] = None) -> Tensor:`
is the typehint in pytorch. 
what am I missing?

src/mrpro/data/_KData.py
fzimmermann89: ah, thanks, got it

src/mrpro/data/_KData.py
fzimmermann89: ah, it got recently fixed in https://github.com/pytorch/pytorch/pull/113647

src/mrpro/operators/_FourierOp.py
```diff
@@ -139,10 +139,10 @@ def __init__(
 
             self._omega = torch.stack(omega, dim=-2)
             self._fwd_nufft_op = KbNufft(
-                im_size=nufft_im_size, grid_size=grid_size, numpoints=numpoints, kbwidth=kbwidth
+                im_size=nufft_im_size, grid_size=grid_size, numpoints=numpoints, kbwidth=kbwidth, device=traj.kx.device
             )
             self._adj_nufft_op = KbNufftAdjoint(
-                im_size=nufft_im_size, grid_size=grid_size, numpoints=numpoints, kbwidth=kbwidth
+                im_size=nufft_im_size, grid_size=grid_size, numpoints=numpoints, kbwidth=kbwidth, device=traj.kx.device
```
fzimmermann89, line 143: do we want to add a Ktraj.device ? Or more general, a Data.device?

fzimmermann89:
lgtm.

three remarks: 

At some point in the near future for a new pytorch version some of the #ignore types will no longer be needed, but for older pytorch versions they will be.  this is something to keep in mind.

do we want to add a .device property to the data? instead of traj.kx.device it would allow traj.device

we should at some point set something up such that we don't have to write a _cuda test for all functions, but instead parameterize them to use either CPU (or even mixed devices)

fzimmermann89 (approved):

src/mrpro/operators/_FourierOp.py
ckolbPTB: A KData.device to me would only make sense if the entire object i.e. also the header is then on a certain device. For Ktraj it would make sense but I would postpone it because we still need something to verify that kx, ky, kz are actually on the same device, have some logic what happens if we assign mixed tensors to kx, ky, kz...

PR #137, opened 2023-12-21
clean up

### LINKED ISSUE #126: SensitivityOp interface
The output of the SensitivityOp should be defined via type hinting: https://github.com/PTB-MR/mrpro/blob/770376be712b94bf44e520767fed89cfeb48321c/src/mrpro/operators/_SensitivityOp.py#L26C23-L26C31

`forward(self, img_data: torch.Tensor) -> torch.Tensor`: 
and without `torch.Tensor` in `Returns`

Same for adjoint.

ckolbPTB:
This fixes #126 and hopefully makes the rest of the code a bit more consistent.

### DIFF
--- a/src/mrpro/operators/_FourierOp.py
+++ b/src/mrpro/operators/_FourierOp.py
@@ -25,8 +25,18 @@ from mrpro.operators import LinearOperator
 def is_uniform(x: torch.Tensor, atol: float = 1e-6) -> torch.Tensor:
     """Check if a tensor contains equidistant sampling points.
 
-    If the tensor has batch-dimension>1, this is checked for all
-    batches.
+    Parameters
+    ----------
+    x
+        tensor to be checked
+    atol
+        tolerance of how similar the values have to be
+
+    Returns
+    -------
+        boolean tensor if sampling points are equidistant.
+
+    If the tensor has batch-dimension>1, this is checked for all batches.
     """
     # TODO: currrently, this function is never used here.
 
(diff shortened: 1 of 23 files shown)


src/mrpro/operators/_FourierOp.py
```diff
@@ -173,7 +179,7 @@ def get_target_pattern(fft_dims: tuple[int, ...], nufft_dims: tuple[int, ...], i
             elif dim in ignore_dims:
                 ignore_dims_str += ' ' + dim_str
 
-        target_pattern = f'(other{fft_dims_str}{ignore_dims_str}) coil{nufft_dims_str}'
+        target_pattern = f'(other{fft_dims_str}{ignore_dims_str}) coils{nufft_dims_str}'
```
fzimmermann89, line 182: If you change coil -> coils here, you also have to change it when this pattern is used.
Also, maybe  make other capture multiple dimensions already via ... ?

src/mrpro/operators/_FourierOp.py
```diff
@@ -262,7 +268,7 @@ def adjoint(self, y: torch.Tensor) -> torch.Tensor:
 
         # move dim where FFT was already performed such nuFFT can be performed
         if len(self._nufft_dims) != 0:
-            init_pattern = 'other coil dim2 dim1 dim0'
+            init_pattern = 'other coils dim2 dim1 dim0'
```
fzimmermann89, line 271: same as above

src/mrpro/operators/_FourierOp.py
fzimmermann89: example would be line 230

fzimmermann89 (changes requested):

fzimmermann89:
Re the other dimension:

I would prefer if we allow in our code for this dimension to actually be more than 1 dimensnion.
So we dont split the KData into dynamics, ... by default, but our code works if it is split.

This would, as a start, mean replacing 'other' by '...' in all einops pattern, and counting dimensions from the back for swapaxes etc.

What do you think, Christoph?

fzimmermann89 (approved):

ckolbPTB:
I think any change which makes the code more general is a good idea. As this will probably require more changes than a few clean-up ones, I created an issue for this #138

PR #139, opened 2024-01-02
Cartesian FFT

ckolbPTB:
Currently the `FourierOp` simply calls FFT on `KData.data` which only works if the data is correctly sorted and if the data is fully sampled. 

Here I have started to implement the code to also allow reconstruction of partial Fourier/echo and regular undersampled Cartesian data. 

Encoded space and recon space are now also correctly taken into consideration for the reco:

encoded_recon_space

Open issues:

- Should trajectory be within [-pi, pi] or in encoded space dimensions?
- How to correctly detect FFT dimensions? Maybe detect if trajectory is on Cartesian grid...
- How to deal with non-regular Cartesian undersampling?
- Lots of optimisation and pythonisation of code needed. Please add your suggestions/pseudo code and I will happily implement it.

### DIFF
--- a/src/mrpro/data/traj_calculators/_KTrajectoryIsmrmrd.py
+++ b/src/mrpro/data/traj_calculators/_KTrajectoryIsmrmrd.py
@@ -45,7 +45,7 @@ class KTrajectoryIsmrmrd:
 
         Returns
         -------
-            trajectory in the shape of the original raw data within [-pi, pi].
+            trajectory in the shape of the original raw data
         """
 
         # Read out the trajectory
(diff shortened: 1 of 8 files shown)


fzimmermann89:
Maybe we can discuss this in the next strategy meeting as well.


In the last two hackathons, I believe, the idea was that the Fourier
Operator should only do an FFT if the trajectory is separable, sorted. and
fully sampled, otherwise it will call the NUFFT.
Instead, there is an Operator that Sorts and zero-fills the data given a
trajectory.


This would remove the rather complex sorting logic into another Operator
that can be chained with the Fourier Operator (we might even define this
chain  with a concise name, for convenience)
And, more importantly, in iterative algorithms, the rearaging and
zero-filling would only have to be done once,saving some copies in the
hot-path.





Am Di., 2. Jan. 2024 um 17:51 Uhr schrieb github-actions[bot] <
***@***.***>:

ckolbPTB:
This was split up into PR #143, #146, #142 and #140

PR #140, opened 2024-01-16
Define trajectories to fit encoding_matrix and encoding_limits

ckolbPTB:
@JoHa0811 Can you please double check the modifications of `KTrajectoryPulseq` - I don't have any test data from the scanner to verify that this still works.


JoHa0811 (changes requested):
In _KData, line 166 must read:

torch.broadcast_shapes(kdata[..., 0, :, :, :].shape, shape)

JoHa0811 (approved):
Calculations work

Issue #141, opened 2024-01-16
_KData shape broadcasting bug


JoHa0811:
In _KData, line 166 must read:

torch.broadcast_shapes(kdata[..., 0, :, :, :].shape, shape)

in order for these checks to run for data coming from pulseq calculators

ckolbPTB:
Fixed by PR #145

PR #143, opened 2024-01-17
Fftshift for FourierOp

ckolbPTB:


ckolbPTB:
@JoHa0811 and @schuenke: it would be great if you could test this with your data to make sure everything is now nicely centered.

schuenke:
> @JoHa0811 and @schuenke: it would be great if you could test this with your data to make sure everything is now nicely centered.

I tested the following cases that worked fine:
1. 2D dynamic cartesian pulseq data using KTrajectoryCartesian calculator
2. 2D dynamic radial pulseq data using KTrajectoryRadial2D calculator

I tried to test KTrajectoryPulseq as well, but it is not working at all currently even after fixing #141 and #144

tests/utils/test_fft.py
```diff
@@ -54,3 +55,14 @@ def test_image_to_kspace_as_inverse():
...
+
+def test_change_image_size():
+    """Test changing image size by cropping and padding."""
+    ishape_orig = (100, 200, 50)
+    ishape_new = (80, 100, 240)
+    iorig = torch.randn(*ishape_orig, dtype=torch.complex64)
+    inew = change_image_shape(iorig, ishape_new)
+
+    # Compare overlapping region
+    torch.testing.assert_close(iorig[10:90, 50:150, :], inew[:, :, 95:145])
```
koflera: although the shapes returned are correct; this test does not test whether cropping is the adjoint of zero-padding.
In fact, the test indeed fails as soon as for the original and the target shape there is one dimension for which the original and target dimensionality are not both odd or both even.
for example:
`ishape_orig = (100, 200, 50), ishape_new = (80, 100, 240)` and `ishape_orig = (101, 200, 50), ishape_new = (83, 100, 240)` are both fine; while `ishape_orig = (101, 200, 50), ishape_new = (84, 100, 240)` or `ishape_orig = (100, 200, 50), ishape_new = (83, 100, 240)` will not pass the adjoint test.

Not sure whether this is a relevant case at all, but anyway

tests/utils/test_fft.py
```diff
@@ -54,3 +55,14 @@ def test_image_to_kspace_as_inverse():
...
+
+
+def test_change_image_size():
+    """Test changing image size by cropping and padding."""
+    ishape_orig = (100, 200, 50)
+    ishape_new = (80, 100, 240)
+    iorig = torch.randn(*ishape_orig, dtype=torch.complex64)
+    inew = change_image_shape(iorig, ishape_new)
+
+    # Compare overlapping region
```
koflera: consider including something like

```
u_shape = (101, 201, 50)
v_shape = (13, 221, 64)
u = torch.randn(u_shape, dtype=torch.complex64)
v = torch.randn(v_shape, dtype=torch.complex64)
Au = change_image_shape(u, v_shape)
AHv = change_image_shape(v, u_shape)
torch.testing.assert_close(torch.vdot(Au.flatten(), v.flatten()), torch.vdot(u.flatten(), AHv.flatten()))
```

src/mrpro/operators/_FourierOp.py
```diff
@@ -85,38 +85,36 @@ def __init__(
         nufft_dims = []
         fft_dims = []
         ignore_dims = []
-        fft_s = []
+        fft_recon_shape = []
```
koflera, line 88: this already starts to look a lot like having a domain and a range for the operator ;)

src/mrpro/utils/fft.py
```diff
@@ -13,37 +13,93 @@
 #   limitations under the License.
 
 import torch
+import torch.nn.functional as F
 
 
-def kspace_to_image(kdat: torch.Tensor, dim: tuple[int, ...] = (-1, -2, -3)) -> torch.Tensor:
+def change_image_shape(idat: torch.Tensor, idat_shape_new: tuple[int, ...]) -> torch.Tensor:
```
koflera: the name "change_image_shape" somewhat suggests that this operation is for images, while we also intend to use it for k-space data.
We could also consider to make it a LinearOperator that uses `torch.nn.functional.pad`, which also works with negative paddings, making it easy to implement the adjoint as well.
the required paddings could then be inferred from the arguments `fft_recon_shape` and `fft_encoding_shape` when defining the FourierOperator

koflera (changes requested):

src/mrpro/operators/_PadOp.py
```diff
@@ -0,0 +1,103 @@
...
+from mrpro.operators import LinearOperator
+from mrpro.utils import change_data_shape
+
+
+class PadOp(LinearOperator):
+    """Pad operator class."""
+
+    def __init__(
+        self,
+        dim: tuple[int, ...],
```
koflera, line 26: the argument `dim` can be inferred from `orig_shape` and `padded_shape` by checking for the dimensions where `orig_shape` and `padded_shape` do not coincide... this maybe is also less prone to errors

src/mrpro/utils/_change_data_shape.py
```diff
@@ -0,0 +1,49 @@
+"""Wrapper for FFT and IFFT."""
```
koflera: comment does not fit content

src/mrpro/operators/_PadOp.py
koflera: the argument dim can be inferred from orig_shape and padded_shape by checking for the dimensions where orig_shape and padded_shape do not coincide... this maybe is also less prone to errors

src/mrpro/operators/_PadOp.py
```diff
@@ -0,0 +1,103 @@
...
+        if len(dim) != len(orig_shape) or len(dim) != len(padded_shape):
+            raise ValueError('Dim, orig_shape and padded_shape have to be of same length')
+
+        super().__init__()
+        self.dim: tuple[int, ...] = dim
+        self.orig_shape: tuple[int, ...] = orig_shape
+        self.padded_shape: tuple[int, ...] = padded_shape
+
+    @staticmethod
+    def _pad_data(x: torch.Tensor, dim: tuple[int, ...], padded_shape: tuple[int, ...]) -> torch.Tensor:
```
koflera: after getting rid of the argument `dim`, which seems to be unneccesary (see comment above), this function seems to more or less coincide with the `_change_data_shape.py` in utils.
Maybe consider extending `_pad_data` here by integrating the content of `_change_data_shape` but keeping the name `_pad_data`?

tests/utils/test_change_data_shape.py
```diff
@@ -0,0 +1,30 @@
...
+import torch
+
+from mrpro.utils._change_data_shape import change_data_shape
+
+
+def test_change_data_shape_content():
+    """Test changing data size by cropping and padding."""
+    dshape_orig = (100, 200, 50)
+    dshape_new = (80, 100, 240)
+    dorig = torch.randn(*dshape_orig, dtype=torch.complex64)
```
koflera: use `RandomGenerator`

tests/utils/test_change_data_shape.py
```diff
@@ -0,0 +1,30 @@
+"""Tests for image space - k-space transformations."""
```
koflera: comment does not fit content

tests/utils/test_change_data_shape.py
```diff
@@ -0,0 +1,30 @@
...
+
+def test_change_data_shape_content():
+    """Test changing data size by cropping and padding."""
+    dshape_orig = (100, 200, 50)
+    dshape_new = (80, 100, 240)
+    dorig = torch.randn(*dshape_orig, dtype=torch.complex64)
+    dnew = change_data_shape(dorig, dshape_new)
+
+    # Compare overlapping region
+    torch.testing.assert_close(dorig[10:90, 50:150, :], dnew[:, :, 95:145])
```
koflera: this test seems to be already covered by the one testing the padding operator; consider removing it

src/mrpro/utils/_change_data_shape.py
```diff
@@ -0,0 +1,49 @@
...
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+import numpy as np
+import torch
+import torch.nn.functional as F
+
+
+def change_data_shape(dat: torch.Tensor, dat_shape_new: tuple[int, ...]) -> torch.Tensor:
```
fzimmermann89: couple of possible issues:
1. name sounds like a reshape operation.
2. consider data instead of data
3. consider dat_shape_new -> new_shape (or I might call you yoda)
4. what if data has a different number of dimensions than the new shape?

```suggestion
def pad_or_crop(data: torch.Tensor, new_shape: tuple[int, ...]|torch.Size) -> torch.Tensor:
    """Change shape of data by cropping or zero-padding.

    Parameters
    ----------
    data
        data
    new_shape
        target shape of data

    Returns
    -------
        data padded or cropped to shape
    """

    if len(new_shape) != data.ndim:
        raise ValueError("length of shape should match dimensions of data")
    npad=[]
    for old, new in zip(data.shape, new_shape):
        diff = new-old
        #  ensures padding and cropping lead to the matching asymmetry for odd shape differences
        before = math.trunc(diff/2) 
        after = diff - before
        npad.append(before)
        npad.append(after)
    # npad has to be reversed because pad expects it in reversed order
    if any(npad):
        data = F.pad(data, npad[::-1])
    return data
```
Or, including the along_dimension functionality:

```suggestion
import torch
import torch.nn.functional as F
import numpy as np
import math

def normalize_index(ndim: int, index:int):
    """Normlize possibly negative indices

    Parameters
    ----------
    ndim
        number of dimensions
    index
        index to normalize. negative indices count from the end.
    Raises
    ------
    IndexError
        if index is outside [-ndim,ndim)
    """

    if 0<index<ndim:
        return index
    elif -ndim<=index<0:
        return ndim + index
    else:
        raise IndexError(f"Invalid index {index} for {ndim} data dimensions")
    
def pad_or_crop(data: torch.Tensor, new_shape: tuple[int, ...]|torch.Size, dim:None|tuple[int,...]=None) -> torch.Tensor:
    """Change shape of data by cropping or zero-padding.
    

    Parameters
    ----------
    data
        data
    new_shape
        desired shape of data
    dim:
        dimensions the new_shape corropsoends to. None (default) is interpreted as last len(new_shape) dimensions.

    Returns
    -------
        data padded or cropped to shape
    """
    
    if len(new_shape) > data.ndim:
            raise ValueError("length of new shape should not exceed dimensions of data")
    
    if dim is None: # Use last dimensions
        new_shape = data.shape[:-len(new_shape)] + new_shape

    else:
        if len(new_shape) != len(dim):
            raise ValueError("length of shape should match length of dim")
        dim = tuple(normalize_index(idx) for idx in dim) # raises if any not in [-data.ndim,data.ndim)
        if len(dim)!=len(set(dim)): # this is why we normalize
            raise ValueError("repeated values are not allowed in dims")
        (new_shape:=torch.tensor(data.shape))[dim]=new_shape
                
    npad=[]
    for old, new in zip(data.shape, new_shape):
        diff = new-old
        after = math.trunc(diff/2)
        before = diff - after
        npad.append(before)
        npad.append(after)
    
    if any(npad):
        # F.pad expects paddings in reversed order
        data = F.pad(data, npad[::-1])
    return data

File: src/mrpro/operators/_PadOp.py (hunk 5 above)
Comment (fzimmermann89) [reply src/mrpro/operators/_PadOp.py]: this is differnt, you might want to specify "pad along dimension 1" and not first calculate the final shape.
I would leave dim as an argument, but maybe allow None -> last len(orig_shape) dimensions>

Comment (ckolbPTB) [reply src/mrpro/operators/_PadOp.py]: Yes, that was the idea. If you want to pad along dimension k2 you should not have to care about all the other dimensions (which could be changed by other operators). 

I would not use None here, because setting `dim = ()` does not do anything and so I don't find it intuitive if `dim = None` that then the last `len(orig_shape)` are padded. I think this makes more sense for `change_data_shape/pad_or_crop` and I would add it there.

File: tests/utils/test_change_data_shape.py (hunk 10 above)
Comment (ckolbPTB) [reply tests/utils/test_change_data_shape.py]: As there are small differences between the two I would like to keep it.

File: src/mrpro/operators/_PadOp.py (hunk 7 above)
Comment (ckolbPTB) [reply src/mrpro/operators/_PadOp.py]: I removed _pad_data and changed change_data_shape to Felix 2nd suggestion

File: src/mrpro/utils/_change_data_shape.py (hunk 11 above)
Comment (ckolbPTB) [reply src/mrpro/utils/_change_data_shape.py]: I incorporated your second suggestion with the dimension option to be able to use it direclty in the PadOp

File: tests/operators/test_zero_pad_op.py [hunk 12]
```diff
@@ -0,0 +1,56 @@
...
+    [
+        ((101, 201, 50), (13, 221, 64)),
+        ((100, 200, 50), (14, 220, 64)),
+        ((101, 201, 50), (14, 220, 64)),
+        ((100, 200, 50), (13, 221, 64)),
+    ],
+)
+def test_zero_pad_op_ajoint(u_shape, v_shape):
+    """Test adjointness of pad operator."""
+    u = torch.randn(u_shape, dtype=torch.complex64)
```
koflera: RandomGenerator for consistency

tests/operators/test_zero_pad_op.py
```diff
@@ -0,0 +1,56 @@
...
+import torch
+
+from mrpro.operators import ZeroPadOp
+
+
+def test_zero_pad_op_content():
+    """Test correct padding."""
+    dshape_orig = (2, 100, 3, 200, 50, 2)
+    dshape_new = (2, 80, 3, 100, 240, 2)
+    dorig = torch.randn(*dshape_orig, dtype=torch.complex64)
```
koflera: RandomGenerator for consistency

tests/operators/test_fast_fourier_op.py
```diff
@@ -36,21 +35,34 @@ def test_kspace_to_image(npoints, a):
...
+    ],
+)
+def test_fast_fourier_op_adjoint(encoding_shape, recon_shape):
+    """Test adjointness of Fast Fourier Op."""
 
-    # Create random 3D data set
-    npoints = [200, 100, 50]
-    idat = torch.randn(*npoints, dtype=torch.complex64)
+    # Create test data
+    x = torch.randn(recon_shape, dtype=torch.complex64)
```
koflera: RandomGenerator for consistency

koflera (approved):

src/mrpro/utils/_zero_pad_or_crop.py
```diff
@@ -0,0 +1,89 @@
...
+            raise ValueError('length of shape should match length of dim')
+        dim = tuple(normalize_index(data.ndim, idx) for idx in dim)  # raises if any not in [-data.ndim,data.ndim)
+        if len(dim) != len(set(dim)):  # this is why we normalize
+            raise ValueError('repeated values are not allowed in dims')
+        new_shape_full = torch.tensor(data.shape)
+        for i, d in enumerate(dim):
+            new_shape_full[d] = new_shape[i]
+
+    npad = []
+    for old, new in zip(torch.tensor(data.shape), new_shape_full):
```
fzimmermann89, line 79: why is this here now?

src/mrpro/utils/_zero_pad_or_crop.py
```diff
@@ -0,0 +1,89 @@
...
+
+    if dim is None:  # Use last dimensions
+        new_shape = data.shape[: -len(new_shape)] + new_shape
+    else:
+        if len(new_shape) != len(dim):
+            raise ValueError('length of shape should match length of dim')
+        dim = tuple(normalize_index(data.ndim, idx) for idx in dim)  # raises if any not in [-data.ndim,data.ndim)
+        if len(dim) != len(set(dim)):  # this is why we normalize
+            raise ValueError('repeated values are not allowed in dims')
+        new_shape_full = torch.tensor(data.shape)
```
fzimmermann89, line 74: and this

fzimmermann89, line 74: why is this a tensor if you don't use tuple indexing in the next line?

src/mrpro/utils/_zero_pad_or_crop.py
```diff
@@ -0,0 +1,89 @@
...
+        for i, d in enumerate(dim):
```
fzimmermann89, line 75: is there a reason for this loop instead of just indexing with dims?

src/mrpro/utils/_zero_pad_or_crop.py
fzimmermann89, line 79: why are the shapes tensors anyway?

fzimmermann89:
the padding is broken for dims=None.

src/mrpro/utils/_zero_pad_or_crop.py
ckolbPTB: otherwise the difference in the line below does not work

Issue #19, opened 2023-07-08
Cartesian acquisition model


ckolbPTB:
Create Cartesian acquisition model

Check list:
- [ ]     Create branch
- [ ]     Implement code
- [ ]     Add comments
- [ ]     Implement tests
- [ ]     Squash commmits
- [ ]     Create PR
- [ ]     Assign reviewer(s) to PR

fzimmermann89:
https://github.com/ckolbPTB/mrpro/issues/9#issuecomment-1750410990

schuenke:
Is this resolved by #142 ?!

ckolbPTB:
Closed by #146

PR #146, opened 2024-01-19
Cartesian sampling operator

### LINKED ISSUE #123: Undersampling in Fourier Operator for FFT directions not yet available
Currently, the Fourier Operator does not yet support undersampling along directions for which the Fourier transform can be performed on a Cartesian grid.

With the current implementation, i.e. how to perform the FFTs/nuFFTs is fully inferred from solely the k-space trajectories, it probably makes sense to create binary masks in the `__init__` of the FourierOp and to use them in the forward/adjoint rather than passing a mask as an argument in the forward/adjoint.

ckolbPTB:


ckolbPTB:
closes #19

src/mrpro/data/_KTrajectory.py
```diff
@@ -41,11 +43,99 @@ class KTrajectory:
...
+        -------
+            type of trajectory along 'kz-ky-kx' or 'k2-k1-k0'
+        """
+        # Matrix describing trajectory-type [(kz, ky, kx), (k2, k1, k0)]
+        # Start with everything not on a grid (arbitrary k-space locations).
+        # We use the value of the enum-type to make it easier to do array operations.
+        traj_type_matrix = torch.tensor([[TrajType.NOTONGRID.value] * 3] * 3, dtype=torch.int8)
+
+        for ind, ks in enumerate((self.kz, self.ky, self.kx)):
+            are_values_on_grid = (ks - ks.to(dtype=torch.int64).to(dtype=ks.dtype)) <= tolerance
```
fzimmermann89: `torch.rint` ?

src/mrpro/data/_KTrajectory.py
```diff
@@ -41,11 +43,99 @@ class KTrajectory:
...
+        for dim in (-3, -2, -1):
+            if torch.all(traj_type_matrix[dim, :] == TrajType.SINGLEVALUE.value):
+                traj_type[dim] = TrajType.SINGLEVALUE.value
+            elif torch.any(traj_type_matrix[dim, :] == TrajType.NOTONGRID.value):
+                traj_type[dim] = TrajType.NOTONGRID.value
+
+        return [TrajType(value) for value in traj_type]
+
+    @property
+    def traj_type_along_kzyx(self, tolerance: float = 1e-2) -> list[TrajType]:
```
fzimmermann89: The name looks a bit ugly.
Maybe remove "traj"

src/mrpro/data/_KTrajectory.py
```diff
@@ -41,11 +43,99 @@ class KTrajectory:
...
 
-    def as_tensor(self, stack_dim=0):
+    def _type_of_traj(
+        self, tolerance: float = 1e-2, orientation: Literal['kz-ky-kx', 'k2-k1-k0'] = 'kz-ky-kx'
+    ) -> list[TrajType]:
+        """Type of trajectory.
+
+        This function checks if the entries of the trajectory along certain dimensions
+            - are of shape 1 -> TrajType.SINGLEVALUE
+            - lie on a Cartesian grid -> TrajType.ONGRID
```
fzimmermann89, line 67: does it check cartesian grid?
isnt it just regular spacing long the dimension?

src/mrpro/data/_KTrajectory.py
```diff
@@ -41,11 +43,99 @@ class KTrajectory:
...
+
+        Returns
+        -------
+            broadcasted shape of trajectory
+        """
         shape = np.broadcast_shapes(self.kx.shape, self.ky.shape, self.kz.shape)
         return tuple(shape)
 
-    def as_tensor(self, stack_dim=0):
+    def _type_of_traj(
```
fzimmermann89: maybe have this inner function always calcuate kx ky kz AND k0 k1 k2 and remove the orientation?

src/mrpro/operators/_CartesianSamplingOp.py
```diff
@@ -0,0 +1,172 @@
...
+                    + coil_idx * self._encoding_shape.z * self._encoding_shape.y * self._encoding_shape.x
+                )
+
+            ysort = torch.zeros(
+                *(y.shape[:2] + (self._encoding_shape.z, self._encoding_shape.y, self._encoding_shape.x)),
+                dtype=y.dtype,
+                device=y.device,
+            )
+
+            return ysort.put_(self._fft_idx_full, y, accumulate=True)
```
fzimmermann89: I would chainthis directly to the zeros call above.
As only after the put `ysorted `is y sorted, and to signify this is one of the exceptions where in place operations are ok: on newly created tensors that will never require gradients.

src/mrpro/operators/_CartesianSamplingOp.py
```diff
@@ -0,0 +1,172 @@
...
+        """
+
+        if self._kshape[1:] != y.shape[-3:]:
+            raise ValueError('k-space data shape missmatch')
+
+        if len(self._fft_dims) > 0 and len(self._fft_idx) > 0:
+            if len(self._fft_idx_full) == 0 or self._fft_idx_full.shape[1] != y.shape[1]:
+                coil_idx = torch.ones(
+                    *(y.shape[:2] + self._fft_idx.shape[2:]), dtype=torch.int64, device=y.device
+                ) * rearrange(
```
fzimmermann89: is this the same as 
`einops.repeat(torch.arange(y.shape[-4], device=y.device), " coils-> other coils k2 k1 k0", k0=y.shape[-1], ...  `
i.e. 
arrange instead of linspace, 
using einops repeat instead of reanage and multiplying by ones?

src/mrpro/operators/_CartesianSamplingOp.py
```diff
@@ -0,0 +1,172 @@
...
+                    torch.linspace(0, y.shape[1] - 1, y.shape[1], dtype=torch.int64, device=y.device),
+                    'coils->1 coils 1 1 1',
+                )
+                other_idx = torch.ones_like(self._fft_idx, dtype=torch.int64, device=y.device) * rearrange(
```
fzimmermann89: see above

fzimmermann89:
Regarding ypur question of caching the results:
We can use the tensor._version to check if the tensor has been modified.

a) invalidate the cache in the __setattr__ 
b)  in the property, check if the cache exists and if a _cache_version=[int,int,int] is the same as the versions 
in kx,ky,kz's ._version. Otherwise, invalidate chache. Calculate properties. And cache.

The inner _check_type function would then update the cache for both kxkykz as well as for k2k1k0 orientation in one go...

I think this could make for a nicer interface, we could have a type_kx, type_k2 .... property instead of a string specifying the order and a list as return values..

src/mrpro/data/_KTrajectory.py
ckolbPTB: I don't think torch.rint exists but torch.round does the job

src/mrpro/data/_KTrajectory.py
ckolbPTB: the values do not have to be regularly spaced but have to lie on a grid, i.e. [0 1 2 3 5 10] would also be ONGRID

src/mrpro/operators/_CartesianSamplingOp.py
ckolbPTB: nice

src/mrpro/data/_KTrajectory.py
fzimmermann89: My question is regarding the 'Cartesian' here.
kz ky kx are by Definition Cartiesian. 
but do we check for perpendicularity in k2 k1 k0?
Cartesian would imply yes, but I don't see where, maybe I am missing it

ckolbPTB: The way I see it is, that the points of the trajectory are always defined in the kz-ky-kx space. If we check the type along k2 we are checking if all the points along k2 lie on the kz-ky-kx grid.

ckolbPTB:
I added a first (incomplete) implementation of the cache idea where the function `_type_of_traj()` is only called if something has changed in kz, ky or kx. Otherwise the cached result is returned. @fzimmermann89 is this roughly what you had in mind?

Currently this only works for in-place changes of kx,ky and kz (see `test_traj_type_update()`). setters would have to be defined for kx, ky and kz to update the ._version for this too. Might also need a new variable to not miss with the standard way tensor._version works. 

Question is, is it worth it? `_type_of_traj()` is really quick and not sure if the overhead of caching is really justified or if we should simply call this function whenever the type of the trajectory is queried. Opinions?

fzimmermann89:
Yeah, I came to a different conclusion: Maybe you were right and we should make KTrajecteroy frozen (and throw an exception if the _version changes).
Otherwise yes, we would need a setter to invalidate the cache also on replacement of one of the tensors..

fzimmermann89:
I woul prefer to change the type logic to a flag that checks different properties.
For now, it willjust be ongrid and singlevalue.
Both can be set or not set in the flag.
This feels like a more flexible way to do this and this pattern can more easy be applied to different properties as well.

so, maybe take a look at some point at #150 , which is based on your branch just before te last commit

fzimmermann89:
@schuenke or @koflera 
Do you mind having a look, after starting the review, I made some extensive refactor changes, also to the FourierOperator, which I now rebased onto this PR.
One of the goals was to enable multiple 'other' dims in the Fourier Operator and some general streamlining of code.

So maybe we could use an independent review / sanity check. Would be kind of stupid for me to continue the review now :D

fzimmermann89:
Do we want to rebase it already to operatorsugar? or first merge #128 which makes all operators return tuples of tensors...

ckolbPTB:
I would suggest to first merge #128

src/mrpro/operators/_CartesianSamplingOp.py
```diff
@@ -0,0 +1,206 @@
...
+        Returns
+        -------
+            k-space data in original (i.e. acquired) dimensions
+        """
+        if self._encoding_shape != SpatialDimension(*x.shape[-3:]):
+            raise ValueError('k-space data shape missmatch')
+
+        if self._fft_dims and self._fft_idx is not None:
+            # Calculate the full index if it has not been calculated yet or if the other and coil dimension has changed
+            if (self._fft_idx_full is None) or (self._fft_idx_full.shape[:2] != x.shape[:2]):
```
fzimmermann89: make this work for more than one other dim

src/mrpro/operators/_CartesianSamplingOp.py
```diff
@@ -0,0 +1,206 @@
...
+        -------
+            k-space data sorted into encoding_space matrix
+        """
+
+        if self._kshape[1:] != y.shape[-3:]:
+            raise ValueError('k-space data shape missmatch')
+
+        if self._fft_dims and self._fft_idx is not None:
+            # Calculate the full index if it has not been calculated yet or if the other and coil dimension has changed
+            if self._fft_idx_full is None or self._fft_idx_full.shape[:2] != y.shape[:2]:
```
fzimmermann89: todo: make this work for more than one other dim

fzimmermann89: this should work now

fzimmermann89:
@ckolbPTB:
What do you think about this?

@schuenke 
I would like to get this merged soonish.... As it also fixes some other issues with the Fourier Operator

src/mrpro/operators/_CartesianSamplingOp.py
```diff
@@ -0,0 +1,144 @@
...
+
+        return (x_reshaped,)
+
+    def adjoint(self, y: torch.Tensor) -> tuple[torch.Tensor,]:
+        """Adjoint operator sorting data into the encoding_space matrix.
+
+        Parameters
+        ----------
+        y
+            k-space data in original (i.e. acquired shape)
```
ckolbPTB: k-space data in original **(i.e. acquired)** shape

src/mrpro/operators/_CartesianSamplingOp.py
```diff
@@ -0,0 +1,144 @@
...
+
+        if self._kshape[-3:] != y.shape[-3:]:
+            raise ValueError('k-space data shape missmatch')
+
+        if not self._needs_indexing:
+            return (y,)
+
+        y_2d = rearrange(y, '... k2 k1 k0 -> (...) (k2 k1 k0)')
+
+        # scatter does not broadcast, so we need to expand the fft_idx to the full dataset
```
ckolbPTB: Suggested additional comment: 
fft_idx is expanded by creating len(y_2d) copies of the indices. In the next step torch.scatter_ will then ensure that the index is correctly extended to the len(y_2d) dimensions.

src/mrpro/operators/_CartesianSamplingOp.py
```diff
@@ -0,0 +1,144 @@
...
+        idx_expanded = self._fft_idx.expand(len(y_2d), -1)
+
+        y_scattered = torch.zeros(
```
ckolbPTB, line 138: Suggested additional comment:
torch.zeros and torch.scatter_ joined to ensure [PLEASE ENTER REASON]

schuenke (approved):
I hab a look at the code yesterday. Didn't have the time to test it with real world data, but @ckolbPTB said he did and it worked for all the cases.

The comments are definitely clearer / more helpful now. Thanks 👍 

I would suggest that we delete the wasabi_example script as I did yesterday in main.

PR #149, opened 2024-01-26
Hotfix #143: zero_pad_or_crop broken for dim=None

fzimmermann89:


fzimmermann89:
This function is also not covered by any unittest., only  by some integration tests further via the Operators.

ckolbPTB:
> This function is also not covered by any unittest., only by some integration tests further via the Operators.

Doesn't this cover it: https://github.com/PTB-MR/mrpro/blob/main/tests/utils/test_zero_pad_or_crop.py

ckolbPTB:
Should we then modify the test to specifically test for dim=None? E.g.:

```
**import pytest**
import torch

from mrpro.utils._zero_pad_or_crop import zero_pad_or_crop
from tests import RandomGenerator

**@pytest.mark.parametrize('dim', [(-3, -2, -1), None])**
def test_zero_pad_or_crop_content(**dim**):
    """Test changing data by cropping and padding."""
    generator = RandomGenerator(seed=0)
    dshape_orig = (100, 200, 50)
    dshape_new = (80, 100, 240)
    dorig = generator.complex64_tensor(dshape_orig)
    dnew = zero_pad_or_crop(dorig, dshape_new, **dim=dim**)

    # Compare overlapping region
    torch.testing.assert_close(dorig[10:90, 50:150, :], dnew[:, :, 95:145])
```

ckolbPTB (approved):

fzimmermann89:
missed this test. but we should also check odd dimensions, odd differences, mixed orders, repeated dims.....

PR #150, opened 2024-01-28
Refactor KTrajectory type and FourierOP

fzimmermann89:
A suggestion for a refactor of some code.

- Trajectory Type is now a Flag with values singlevalue and ongrid.  'notongrid' is not 'ongrid'
- KTrajectory is now frozen, This causes issues in the wasabi example. We might reconsider making it unfrozen.
- Trajectory type is now cached (and inalidated if the tensor version cahnges because of some inplace operation)

- FourierOP uses reshape/permute etc in the nufft. The advantage of rearange gets lost of we have compicated code building the expression string.
- Steps toward allowing multiple other dimensions in the FourierOP
- I do not fully understand to NotImplemented Exception. IMHO this check could be wrong,

- Flake8 and Black disagree in E203. I disabled E203 in flake8

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -161,3 +161,7 @@ cython_debug/
 
 # Documentation
 **/_autosummary
+
+# anaconda related directories
+.conda
+.micromamba
--- a/src/mrpro/data/_KTrajectory.py
+++ b/src/mrpro/data/_KTrajectory.py
@@ -45,19 +45,8 @@ class KTrajectory:
     kx: torch.Tensor
     """(other,k2,k1,k0), frequency encoding direction k0 if Cartesian."""
 
-    _type_along_kzyx: tuple[TrajType, TrajType, TrajType]
-    _type_along_k210: tuple[TrajType, TrajType, TrajType]
-    _expected_tensor_versions: tuple[int, int, int]
-
-    # Type of trajectory
-    _type_kz_ky_kx: list[TrajType]  # for kz, ky and kx directions
-    _type_k2_k1_k0: list[TrajType]  # for k2, k1 and k0 directions
-
     # Tolerance for trajectory type estimation (i.e. how close do the values have to be to grid points)
     type_tolerance: float
-
-    # Tensor version of kz, ky and kx to detect any changes which require type update
-    _traj_version: list[int]
 
     @property
     def broadcasted_shape(self) -> tuple[int, ...]:
@@ -76,8 +65,6 @@ class KTrajectory:
         Checks if the entries of the trajectory along certain dimensions
             - are of shape 1 -> TrajType.SINGLEVALUE
             - lie on a Cartesian grid -> TrajType.ONGRID
-            - none of the above -> TrajType.NOTONGRID
-
 
         Parameters
         ----------
@@ -114,16 +101,12 @@ class KTrajectory:
     @property
     def type_along_kzyx(self) -> tuple[TrajType, TrajType, TrajType]:
         """Type of trajectory along kz-ky-kx."""
-        if self._expected_tensor_versions != (self.kz._version, self.ky._version, self.kx._version):
-            raise NotImplementedError('The trajectory has been modified inplace. This is not supported')
-        return self._type_along_kzyx
+        return self._traj_types(self.type_tolerance)[0]
 
     @property
     def type_along_k210(self) -> tuple[TrajType, TrajType, TrajType]:
         """Type of trajectory along k2-k1-k0."""
-        if self._expected_tensor_versions != (self.kz._version, self.ky._version, self.kx._version):
-            raise NotImplementedError('The trajectory has been modified inplace. This is not supported')
-        return self._type_along_k210
+        return self._traj_types(self.type_tolerance)[1]
 
     def as_tensor(self, stack_dim=0) -> torch.Tensor:
         """Tensor representation of the trajectory.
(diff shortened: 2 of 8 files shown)


fzimmermann89:
Precommit fails in mypy in the wasabi example because KTrajctory is now frozen and the example directly changes the Tensor.

Test fails in test_fourier_op.py::test_fourier_not_supported_traj, as I commented out this check for now. This needs a fix.

.gitignore
```diff
@@ -161,3 +161,7 @@ cython_debug/
 
 # Documentation
 **/_autosummary
+
+# anaconda related directories
```
ckolbPTB, line 165: Why is this needed?

src/mrpro/data/_KTrajectory.py
```diff
@@ -36,9 +36,18 @@ class KTrajectory:
...
+
+    ky: torch.Tensor
+    """(other,k2,k1,k0), phase encoding direction k1 if Cartesian."""
+
+    kx: torch.Tensor
+    """(other,k2,k1,k0), frequency encoding direction k0 if Cartesian."""
+
+    _type_along_kzyx: tuple[TrajType, TrajType, TrajType]
+    _type_along_k210: tuple[TrajType, TrajType, TrajType]
+    _expected_tensor_versions: tuple[int, int, int]
```
ckolbPTB: I would remove `_expected_tensor_versions: tuple[int, int, int]` and always run `_traj_types()` whenever traj_type is queried. not efficient but should still be ok. Otherwise errors about changed tensors would occur in random places.

src/mrpro/data/_KTrajectory.py
```diff
@@ -61,106 +70,60 @@ def broadcasted_shape(self) -> tuple[int, ...]:
...
-    def _type_of_traj(self) -> None:
-        """Calculate type of trajectory.
+    def _traj_types(self, tolerance) -> tuple[tuple[TrajType, TrajType, TrajType], tuple[TrajType, TrajType, TrajType]]:
+        """Calculate the trajectory type along kzkykx and k2k1k0.
 
-        This function checks if the entries of the trajectory along certain dimensions
+        Checks if the entries of the trajectory along certain dimensions
             - are of shape 1 -> TrajType.SINGLEVALUE
             - lie on a Cartesian grid -> TrajType.ONGRID
             - none of the above -> TrajType.NOTONGRID
```
ckolbPTB: TrajType.NOTONGRID does not exist anymore

src/mrpro/data/_KTrajectory.py
```diff
@@ -61,106 +70,60 @@ def broadcasted_shape(self) -> tuple[int, ...]:
...
+        return (type_zyx[0], type_zyx[1], type_zyx[2]), (type_210[0], type_210[1], type_210[2])
 
-        Parameters
-        ----------
-        tolerance
-            tolerance used to determine if trajectory points lie on a gird
+    @property
+    def type_along_kzyx(self) -> tuple[TrajType, TrajType, TrajType]:
+        """Type of trajectory along kz-ky-kx."""
+        if self._expected_tensor_versions != (self.kz._version, self.ky._version, self.kx._version):
```
ckolbPTB: remove - see comment above

src/mrpro/data/_KTrajectory.py
```diff
@@ -61,106 +70,60 @@ def broadcasted_shape(self) -> tuple[int, ...]:
...
 
-        Returns
-        -------
-            type of trajectory along k2-k1-k0
-        """
-        return self._type_k2_k1_k0
+    @property
+    def type_along_k210(self) -> tuple[TrajType, TrajType, TrajType]:
+        """Type of trajectory along k2-k1-k0."""
+        if self._expected_tensor_versions != (self.kz._version, self.ky._version, self.kx._version):
```
ckolbPTB: remove - see comment above

src/mrpro/data/_KTrajectory.py
```diff
@@ -212,9 +173,19 @@ def __init__(
         if len(shape) != 4:
             raise ValueError('The k-space trajectory tensors should each have 4 dimensions.')
 
+        type_along_kzyx, type_along_k210 = self._traj_types(grid_detection_tolerance)
```
ckolbPTB: not needed here

.gitignore
fzimmermann89: We can also remove it. I put it there because I use a vscode extension that creates these folders in the project folder, thus they always show up in git status etc.
I am pretty sure we will never want to add a .conda environment to git, so adding it in the .gitignore along all the other ignored stuff will not hurt.

src/mrpro/operators/_FourierOp.py
```diff
@@ -58,38 +56,46 @@ def __init__(
...
-            # E.g. kx is of shape (1,1,30,1,1): kx sampling along k2.
-            if self._fft_dims != [dim for dim in (-3, -2, -1) if traj.traj_type_along_k210[dim] == TrajType.ONGRID]:
+        if self._nufft_dims:
+            fft_dims_k210 = [
+                dim
+                for dim in (-3, -2, -1)
+                if (traj.type_along_k210[dim] & TrajType.ONGRID)
+                and not (traj.type_along_k210[dim] & TrajType.SINGLEVALUE)
+            ]
+            if self._fft_dims != fft_dims_k210:
```
ckolbPTB, line 90: This checked here because as far as I can see, there is only a problem if the FFT dimensions are not aligned and a combination of FFT and NUFFT is carried out. Maybe we can rephrase the error message to make this clearer.

.gitignore
ckolbPTB: Our .gitignore is really big anyway so probably doesn't matter. I prefer to use something like .*/ to e.g. ignore all folders starting with a period and then adding specific folders which are needed. Makes it easier to read and maintain. Maybe for the future...

ckolbPTB:
@fzimmermann89 let me know if I have missed anything otherwise feel free to merge

PR #152, opened 2024-01-31
Hotfix for PyTorch 2.2

ckolbPTB:
Pytorch 2.2 was released yesterday and it fixed some problems with mypy and `tensor.to()`. As our pre-commit framework always uses the latest pytorch version this now causes an error because our `KData.to()` implementation does not raise an error anymore and hence the exception `# type: ignore [call-overload]` is not needed anymore. 

This fixes this but now we also have to ensure that Pytorch version is at least 2.2.

@fzimmermann89 or @koflera can you see a problem with increasing the required pytorch version?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ classifiers = [
 ]
 dependencies = [
     "numpy>=1.23,<2.0",
-    "torch>=2.0,<3.0",
+    "torch>=2.2,<3.0",
     "torchio",
     "ismrmrd",
     "einops",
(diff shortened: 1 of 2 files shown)


fzimmermann89:
and soon the .cuda and .cpu will also be fixed :) 

Until we are at a somewhat stable release that is used in bigger projects, I would follow torch and python to current versions.
Once we have a Version 1.0 out, I would try to stay on versions for longer...

fzimmermann89 (approved):

schuenke (approved):
👍

Issue #119, opened 2023-11-28
Hardcoded ignore flags


ckolbPTB:
Currently we have the ignore flags which define which raw data we do not use for `KData` hardcoded here:
https://github.com/PTB-MR/mrpro/blob/467a3b387a7893289094b6e8d6dd806b869e86f8/src/mrpro/data/_KData.py#L87

The required ignore flags will depend on the type of scan we want to reconstruct so this should be adaptable by the user. 

Suggestion:
```
    def from_file(
        cls,
        filename: str | Path,
        ktrajectory: KTrajectoryCalculator | KTrajectory | KTrajectoryIsmrmrd,
        header_overwrites: dict[str, object] | None = None,
        dataset_idx: int = -1,
        ignore_flags: AcqFlags = AcqFlags.ACQ_IS_NOISE_MEASUREMENT | AcqFlags.ACQ_IS_SURFACECOILCORRECTIONSCAN_DATA
```

fzimmermann89:
I am pretty sure that AcqFlags.ACQ_IS_SURFACECOILCORRECTIONSCAN_DATA is a misnomer.
At least the siemens converter stores `MDH_RAWDATACORRECTION` in this flag, i.e. if the raw data should be corrected by a coilwise factor.

pymapvbvd uses to following to detect non-images:
``` noImaScan = (mask.MDH_ACQEND | mask.MDH_RTFEEDBACK | mask.MDH_HPFEEDBACK
                 | mask.MDH_PHASCOR | mask.MDH_NOISEADJSCAN | mask.MDH_PHASESTABSCAN
                 | mask.MDH_REFPHASESTABSCAN | mask.MDH_SYNCDATA
                 | (mask.MDH_PATREFSCAN & ~mask.MDH_PATREFANDIMASCAN))```
 ```     
       
  with indices into EvalInfoMask as follows:
  ```
   MDH_ACQEND: 0
        MDH_RTFEEDBACK: 1
        MDH_HPFEEDBACK: 2
        MDH_SYNCDATA: 5
        MDH_RAWDATACORRECTION: 10
        MDH_REFPHASESTABSCAN: 14
        MDH_PHASESTABSCAN: 15
        MDH_SIGNREV: 17
        MDH_PHASCOR: 21
        MDH_PATREFSCAN: 22
        MDH_PATREFANDIMASCAN: 23
        MDH_REFLECT: 24
        MDH_NOISEADJSCAN: 25
        MDH_VOP: 53 - 32
[...]

PR #153, opened 2024-02-01
Enable data which does not allow for a reshape into k2 and k1

ckolbPTB:
This allows to read in data which cannot be reshaped into a 2D matrix of k2 and k1. E.g. 2D Poisson-based Cartesian undersampling along the phase and slice encoding direction.


schuenke (approved):
Tested it with some additional examples and it worked fine.
At some point we might want to split the from_file() classmethod, but I would leave this for a separate PR, where we can also address #119

PR #157, opened 2024-02-09
Clean up code

### LINKED ISSUE #148: Spring clean
This is a collection of small issues which are not (yet) bugs but should be tiedied up at some point:

To make things consistent with the other tests:
- [x] tests/operators/models/ should be removed and [test_models.py](https://github.com/PTB-MR/mrpro/blob/main/tests/operators/models/test_models.py) moved to tests/operators/
- [x] WASABI and WASABITI are defined in operators/models so their tests should be in tests/operators/test_models.py
- [x] [test_FourierOp.py](https://github.com/PTB-MR/mrpro/blob/main/tests/operators/test_FourierOp.py) should be renamend to test_fourierop.py or (maybe easier to read) test_fourier_op.py. The latter would mean to rename [test_sensitivityop.py](https://github.com/PTB-MR/mrpro/blob/main/tests/operators/test_sensitivityop.py) to test_sensitivity_op.py

Further issues:
- [x] [_FourierOp.py](https://github.com/PTB-MR/mrpro/blob/main/src/mrpro/operators/_FourierOp.py) is missing a descriptive string as the first line. Can we somehow check for this in our CI framework?
- [ ] [wasabi_example.py](https://github.com/PTB-MR/mrpro/blob/main/examples/wasabi_example.py) is using absolute path. Data should be on zenodo/figshare... to allow for automatic running. Will be bug for #118

### LINKED ISSUE #156: Package name does not match PEP8
Our package name is currently "**MRpro**" with two capital letters at the beginning, but the official style guide for Python code (PEP8) recommends "short, all-lowercase names" for packages and modules. Find the corresponding section [HERE](https://peps.python.org/pep-0008/#package-and-module-names)

Thus, imo, we should change the python package name, but keep the name with leading capital letters in the Readme / docs etc...

schuenke:
This PR addresses #148 and #156

The following things should / need to be done:

- [x] reactivate docstring checks to enforece docstrings for all classes and methods
- [x] rename project from MRpro to mrpro to fulfill PEP8 recommendation (see #156)
- [x] move tests for WASABI and WASABITI to tests/operators/models
- [x] [test_FourierOp.py](https://github.com/PTB-MR/mrpro/blob/main/tests/operators/test_FourierOp.py) should be renamend to test_fourierop.py or (maybe easier to read) test_fourier_op.py. The latter would mean to rename [test_sensitivityop.py](https://github.com/PTB-MR/mrpro/blob/main/tests/operators/test_sensitivityop.py) to test_sensitivity_op.py
- [ ] [wasabi_example.py](https://github.com/PTB-MR/mrpro/blob/main/examples/wasabi_example.py) is using absolute path. Data should be on zenodo/figshare... to allow for automatic running. Will be bug for https://github.com/PTB-MR/mrpro/pull/118

closes #148 
closes #156


schuenke:
Maybe we want to merge this one as well before the hackathon?

Imo the example_script can be fixed later...

pyproject.toml
```diff
@@ -100,18 +100,7 @@ force_single_line = true
...
 convention = "numpy"
-add_ignore = [
-    "D100",
-    "D101",
-    "D102",
-    "D103",
-    "D104",
-    "D105",
-    "D106",
-    "D107",
```
fzimmermann89, line 110: just double checking, we want to have docstrings on classes AND their __init__ ?

src/mrpro/data/_KData.py
```diff
@@ -47,6 +47,18 @@
 
 @dataclasses.dataclass(slots=True, frozen=True)
 class KData:
+    """MR raw data / k-space data class.
```
fzimmermann89: Didnt we at some point decide to do 

```
header: KHeader
"""header containing meta data"""

data: torch.Tensor
"""fantastic data"""

traj: torch.Tensor
"""...and where to find it"""
```
to document attributes? do we then need a docstring in the class that does the same?
I would prefer to have a docstring on the attributes instead of on the class.

pyproject.toml
schuenke: I would say so, yes. In general, init arguments and class attributes can differ. So, having docstrings for both makes sense in my opinion.

src/mrpro/data/_TrajectoryDescription.py
```diff
@@ -32,6 +32,14 @@ class TrajectoryDescription:
 
     @classmethod
     def from_ismrmrd(cls, trajectoryDescription: trajectoryDescriptionType) -> TrajectoryDescription:
+        """Create TrajectoryDescription from ismrmrd trajectoryDescriptionType.
+
+        Parameters
+        ----------
+        trajectoryDescription
+            ismrmrd trajectoryDescriptionType object.
```
fzimmermann89: This also feels kind of useless, right?

src/mrpro/operators/models/_InversionRecovery.py
```diff
@@ -16,27 +16,33 @@
...
         ti
-            inversion times
+            inversion times [s]
         """
         super().__init__()
         self.ti = torch.nn.Parameter(ti, requires_grad=ti.requires_grad)
 
     def forward(self, m0: torch.Tensor, t1: torch.Tensor) -> torch.Tensor:
-        """Apply the forward model.
+        """Forward method of the Inversion Recovery signal model.
```
fzimmermann89:
```suggestion
        ""Apply Inversion Recovery signal model.
```

fzimmermann89: that it is the `forward `method is kinda obvious, right?

src/mrpro/operators/models/_MOLLI.py
```diff
@@ -16,27 +16,35 @@
...
+        b
+            Parameter b in MOLLI signal model
+        t1
+            longitudinal relaxation time T1 [s]
 
         Returns
         -------
-            Image data tensor (other, c, z, y, x)
+        torch.Tensor
+            data tensor with dimensions ((... inv_times), coils, z, y, x)
```
fzimmermann89: did we decide to use this style?
I thought we decided not to put the types here, as we already have this docu,eted by type hints
```suggestion
           signal with dimensions ((... inv_times), coils, z, y, x)
```

fzimmermann89 (approved):

src/mrpro/operators/models/_MOLLI.py
fzimmermann89: and similiar...

fzimmermann89:
Sorry for merging #128 before this one, now this needs to be rebased ... ** *innocent look* **

schuenke:
I can do this tomorrow morning in the train. No prob.

src/mrpro/operators/models/_MOLLI.py
schuenke: I didn't change this yet.
Will check the numpy docstring style documentation again.

fzimmermann89: We should try to aim to keep this consistent in our code base. 

I am pretty sure currently more function do not specify the return type in the docstring.

If you decide one way or the other, it would be great to change it everywhere.

schuenke: removed the return types ✔️

src/mrpro/data/_KData.py
schuenke: I removed the `Attributes` in the docstring.
The documentation in the style you pointed out will be in the future when we do this for all classes I would say.

PR #159, opened 2024-02-11
Fix ignore flags

### LINKED ISSUE #119: Hardcoded ignore flags

fzimmermann89:
Should now on default ignore all Siemens non-image acquisitions, and flags are configurable.
 
Needs some testing on real data.

Fixes #119

### DIFF
--- a/src/mrpro/data/_KData.py
+++ b/src/mrpro/data/_KData.py
@@ -44,6 +44,18 @@ KDIM_SORT_LABELS = (
     'set',
 )
 
+# Same criteria as https://github.com/wtclarke/pymapvbvd/blob/master/mapvbvd/mapVBVD.py uses
+DEFAULT_IGNORE_FLAGS = (
+    AcqFlags.ACQ_IS_NOISE_MEASUREMENT
+    | AcqFlags.ACQ_IS_DUMMYSCAN_DATA
+    | AcqFlags.ACQ_IS_HPFEEDBACK_DATA
+    | AcqFlags.ACQ_IS_NAVIGATION_DATA
+    | AcqFlags.ACQ_IS_PHASECORR_DATA
+    | AcqFlags.ACQ_IS_PHASE_STABILIZATION
+    | AcqFlags.ACQ_IS_PHASE_STABILIZATION_REFERENCE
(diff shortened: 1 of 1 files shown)


schuenke:
In general, it lgtm. 

Like you wrote, this needs some testing with real data (ideally with some vendor-specific and some pulseq data).

I was wondering if an option to add additional flags to the default list (when using `ignore_flags = None`) would be a handy feature? Assuming I would like to add a single additional flag, I would image that it would be nice if I wouldn't have to manually list all the defaults plus the additional flag, but just the additional one, no?

fzimmermann89:
To whom it may concern: Please let me know which of the two commits you prefer.
Or if you have suggestions on how to handle ACQ_IS_PARALLEL_CALIBRATION being set on images, but only if also ACQ_IS_PARALLEL_CALIBRATION_AND_IMAGING is set (otherwise it is pure calibration data and should by default be ignored)

fzimmermann89:
discussion with @schuenke @koflera  and @ckolbPTB : will go with variant 2 for now.

fzimmermann89:
The code should be correct, when we looked at the diff we only saw the diff between version 1 and version 2: The missing line was the same between both versions.
So for review, just look at the overall changes of this PR.

ckolbPTB (approved):

schuenke (approved):
lgtm

Issue #138, opened 2023-12-21
Make "other" dimension more flexible


ckolbPTB:
Suggestion by @fzimmermann89 

Re the other dimension:

I would prefer if we allow in our code for this dimension to actually be more than 1 dimensnion.
So we dont split the KData into dynamics, ... by default, but our code works if it is split.

This would, as a start, mean replacing 'other' by '...' in all einops pattern, and counting dimensions from the back for swapaxes etc.

ckolbPTB:
Make sure that always the correct terms (e.g. coils and not coil) are used.

Functions/Classes to double check

- [x] https://github.com/PTB-MR/mrpro/blob/0730e0608366b422c59c0c5ffa50a94fba671cf2/src/mrpro/algorithms/prewhiten_kspace.py#L30
- [x] https://github.com/PTB-MR/mrpro/blob/0730e0608366b422c59c0c5ffa50a94fba671cf2/src/mrpro/operators/CartesianSamplingOp.py#L26
- [x] https://github.com/PTB-MR/mrpro/blob/0730e0608366b422c59c0c5ffa50a94fba671cf2/src/mrpro/operators/DensityCompensationOp.py#L25
- [x] https://github.com/PTB-MR/mrpro/blob/0730e0608366b422c59c0c5ffa50a94fba671cf2/src/mrpro/operators/SensitivityOp.py#L25
- [x] https://github.com/PTB-MR/mrpro/blob/0730e0608366b422c59c0c5ffa50a94fba671cf2/src/mrpro/data/traj_calculators/KTrajectoryCartesian.py#L48
- [ ] https://github.com/PTB-MR/mrpro/blob/0730e0608366b422c59c0c5ffa50a94fba671cf2/src/mrpro/data/traj_calculators/KTrajectoryPulseq.py#L73
- [x] https://github.com/PTB-MR/mrpro/blob/0730e0608366b422c59c0c5ffa50a94fba671cf2/src/mrpro/data/IData.py#L63
- [x] https://github.com/PTB-MR/mrpro/blob/0730e0608366b422c59c0c5ffa50a94fba671cf2/src/mrpro/data/KNoise.py#L65
- [x] https://github.com/PTB-MR/mrpro/blob/0730e0608366b422c59c0c5ffa50a94fba671cf2/src/mrpro/data/KTrajectory.py#L30
[...]

PR #160, opened 2024-02-11
Allow more than one other dim in KTrajectory

fzimmermann89:
Just relaxes one of the checks: KTrajectory can now be created from tensors with more than 4 dimensions


schuenke:
Maybe we should include the refactoring of the rest of the code (as mentioned in #138) in this PR? It's adjusted in many functions already, but I think we didn't adjust all code yet.

fzimmermann89:
I am pretty sure there are lots of functions left that could use a fix - - but this particular one I need right know and would like to get merged ASAP..

schuenke (approved):

PR #166, opened 2024-02-14
variant 1: conditional import

fzimmermann89:


fzimmermann89:
so this does not work....

fzimmermann89:
Almost:
After a bit of thinking about it, we should not use KData inside the functions.

Only the object we got passed in.
So we call normal instance methods etc it.
To create a new object., we should use object.__class__() or type(object)() 

This will also work with inheritance: 
If somebody creates a KData child, it's reshape functions should return a child instance, not the parent KData instance. This is currently broken and should anyways be fixed.. 

Then all that remains are type hints. This can be solved nicely with type vars and/or a type checking conditional

Issue #169, opened 2024-02-16
Speed up dcf calculation


ckolbPTB:
Currently the dcf calculation is very slow, specifically this part: https://github.com/PTB-MR/mrpro/blob/025bc796b845aabdf255b2ed584f0ff26ae0cb28/src/mrpro/data/_DcfData.py#L118C18-L118C36

I tried to exchange the `ThreadPoolExecutor` with `multiprocessing.Pool` but this did not make any change. 

Possible ways to solve this: Find a pytorch compatible way of calculation Voronoi cells or at least replace `ConvexHull(v).volume` by a function running on GPU

fzimmermann89:
The issue is that the number of vertices in each cell is different. This makes vectorization difficult.
I will keep this as a todo until the next long train ride and would test if we can write a small separate utility that also wraps qhull with cython, but allows a list of arrays of vertices and internally parallelizes over the list.

fzimmermann89:
this would be a pure python version to calculate the volume/area of a convex polyhedron (vornoi cells are afik always convex)
```python
def calculate_volume(vertices, centroid=None):
    """
    Compute the volume (or area in 2D) of a Voronoi cell defined by a set of vertices and its centroid.
    
    Parameters:
        vertices
               An array of shape (N, 2) or (N, 3) representing the vertices of the Voronoi cell.
        centroid (optional): 
              An array of length 2 or 3 representing the centroid of the Voronoi cell.
              If None, the centroid will be calculated as the mean of the vertices.
    
    Returns:
        volume (float): The volume (or area in 2D) of the Voronoi cell.
    """
    num_vertices = vertices.shape[0]
    assert num_vertices >= 3, "Voronoi cell requires at least 3 vertices"

    # Determine the dimensionality of the input
    dim = vertices.shape[1]

    # If centroid is not provided, calculate the mean of the vertices
    if centroid is None:
        centroid = np.mean(vertices, axis=0)

    # Shift vertices by the centroid to avoid precision issues
    shifted_vertices = vertices - centroid

    # Form all combinations of vertices to create triangular faces (or faces in 2D)
    i, j, k = np.indices((num_vertices, num_vertices, num_vertices))
    valid_indices = (i < j) & (j < k)

    # Filter out the valid combinations of vertices
    i, j, k = i[valid_indices].flatten(), j[valid_indices].flatten(), k[valid_indices].flatten()

    # Calculate the volume (or area in 2D) of each simplex formed by the vertices and the centroid
[...]

PR #170, opened 2024-02-16
Switch to Processes for  voronoi cell volume calculation

fzimmermann89:
Ref: Speed up dcf calculation #169

Scipy.spatial Qhull does not seem to work in parallel in multiple threads, using processes makes it~4x faster on my notebook. Would still prefer a gpu version.


ckolbPTB (approved):
Tested on the cluster and confirmed speed up.

Issue #91, opened 2023-10-12
Consider making SpatialDimension frozen and/or new class


fzimmermann89:
It would be nice to have a common structure for x,y,z values.
This includes Nx,Ny,Nz for image sizes in the phantom functions.

Now there are three questions:
### 1. Are we fine with having the same structure for these
- real space distances (in meters) 
- image resolution (in meters) 
- delta_Ks (reciprocal meters)
- Image dimensions (pixel).

Or do we want to have seperate dataclasses by 'units', that all implement the (already existing xyz interface) and are siblings of spatialdimension.
A subquestions would be, if we decide to use one class for all of them: Is SpatialDimension the best name for it?

### 2. Should we make SpatialDimension frozen.
It could be nice to be able to use SpatialDimension also as Parameters for functions with default values,
For example  `def generate_phantom(shape=SpatialDimension(10,10,10)`
Without SpatialDimension being frozen, this is really error prone, as:

```python
def generate_phantom(shape=SpatialDimension(10,10,10), overwrite=True)
   if overwrite:
     shape.x=1
   print(shape.x)
   
generate_phantom(False) #10
generate_phantom(True) #1
generate_phantom(False) #1   
``` 
as the default argument will only be initialized once at load time, but we allow modifying it. 
(This is way you should use tuples as default arguments, not lists)

We could instead also come up with a DefaultSpatialDimension(SpatialDimension) the we use for defaults, and is just SpatialDimension made immutable. So all parameters can still by typed SpatialDimension but the defaults are immutable.

### 3. Should for convenience this functions also accept a tuple (z,y,x)
 to be able to write:
`generate_phantom(shape=(10,10,10)) ` instead of` generate_phantom(shape=SpatialDimension(10,10,10))`.
We could also say "yes, but we do always use the SpatialDimension argument in our internal code"
This way we have clarity what is going on, but for the reconstruction-pipeline userscripts the usage is a bit simpler/shorter.

schuenke:
Regarding point 3:
Especially in the userscripts it would definitely be nicer to accept tuples in addition to SpatialDimension objects.
This would mean that we allow `SpatialDimension[x] | tuple[x, x, x]` for all arguments of operators and functions that might be used in such scripts (e.g. sampling ops, fourier op, phantom and filter functions etc)

How would we like to ensure to _always use the SpatialDimension argument in our internal code_ than? Simply check for something like 

```python
if not isinstance(arg, SpatialDimension):
[...]

PR #171, opened 2024-02-16
Reco examples

JoHa0811:
Added two reconstruction examples; one or fully-sampled cartesian data and one for radial-interleafed
Both files have markdown comments in order to be visually appealling in their jupyter-notebook end-form


examples/radial_interleafed_example.py
```diff
@@ -0,0 +1,125 @@
...
+for i in image:
+    print(i.shape)
+    plt.matshow(torch.abs(i))
+    break
+# %% [markdown]
+# # Save reconstructed images
+# In case you would like to save the reconstructed images,
+# set the NIFTI flag at the top to True
+# %%
+if NIFTI:
```
fzimmermann89: Q: is there a reason for this flag? why not just always do it in the example?

examples/radial_interleafed_example.py
```diff
@@ -0,0 +1,125 @@
...
+    recon_shape=SpatialDimension(1, 256, 256),
+    encoding_shape=SpatialDimension(1, 256, 256),
+    traj=data.traj,
+    oversampling=SpatialDimension(1, 1, 1),
+)
+(xcoils,) = ft_op.H(data.data * dcf.data)
+idata = IData.from_tensor_and_kheader(xcoils, data.header)
+
+smoothing_width = SpatialDimension(z=1, y=5, x=5)
+csm = CsmData.from_idata_walsh(idata, smoothing_width)
```
fzimmermann89, line 73: maybe we could consider generating csms from KData or an image tensor directly?
This reads really complicated right now..

examples/radial_interleafed_example.py
```diff
@@ -0,0 +1,125 @@
...
+    zenodo_cmd = f'zenodo_get 10.5281/zenodo.10664974 -o {str(data_folder)}'
+
+    # Execute the Zenodo command
+    out = subprocess.call(zenodo_cmd, shell=True)
+
+    # Check if the download was successful
+    if out == 0:
+        print('Download successful.')
+        break
+    else:
```
fzimmermann89: Q: why do we need to retry this?

examples/cartesian_reconstruction_example.py
```diff
@@ -0,0 +1,125 @@
...
+# # Reconstruction
+# Given the loaded data, the Fourier Operator is initialized next,
+# with which the appropriate Fourier Transform is carried out.
+# Furthermore, coil sensitivity maps are created to control for varying
+# coil sensitivities throughout the scan.
+
+# %%
+# perform FT and CSM
+ft_op = FourierOp(
+    recon_shape=SpatialDimension(1, 256, 256),
```
fzimmermann89, line 92: we should really consider also allowing tuples here I think...

examples/cartesian_reconstruction_example.py
```diff
@@ -0,0 +1,125 @@
...
+
+    # Check if the download was successful
+    if out == 0:
+        print('Download successful.')
+        break
+    else:
+        print(f'Download failed! Attempt {attempt} of {max_attempts}')
+        attempt += 1
+
+if attempt > max_attempts:
```
fzimmermann89: just an else would suffce here

fzimmermann89: https://stackoverflow.com/questions/3295938/else-clause-on-python-while-statement

examples/radial_interleafed_example.py
JoHa0811: Because when executed in a notebook, this line fails. But when executed locally, it works.
We can also get rid of it, if we want all functionality to run online as well.

examples/radial_interleafed_example.py
JoHa0811: Because it fails relatively often. This is probably due to the Zenodo API.

examples/cartesian_reconstruction_example.py
JoHa0811: Thanks

examples/radial_interleafed_example.py
fzimmermann89: ah, thats why there is a retry option for zenodo_get...  :D  (`-r INT`)


also, what is the advantage of subprocess call over 
```import zenodo_get
from pathlib import Path
DATASET = "10669837" # use dataset instead of doi to avoid additional call to resolve doi...
outdir = Path("/tmp/bliblablu")
zenodo_get.zenodo_get([DATASET, "-r", 5, "-o", outdir]) # r: retries
```

examples/radial_interleafed_example.py
fzimmermann89: i see,
i wouldn't mind removing this completely from the examples, as this is not part of mrpro (yet)

JoHa0811: Sounds good, I will remove it

examples/radial_interleafed_example.py
JoHa0811: of course
Shall we do this as part of a seperate issue?

examples/radial_interleafed_example.py
JoHa0811: I will re-write the retry logic to what you posted
The advantge of subprocess I cannot tell you. This code was taken from PTBRecon

fzimmermann89: The I would suggest using the python call of the function.

examples/radial_interleafed_example.py
```diff
@@ -0,0 +1,125 @@
...
+        print(f'Download failed! Attempt {attempt} of {max_attempts}')
+        attempt += 1
+
+if attempt > max_attempts:
+    # Abort after reaching the maximum number of attempts
+    raise ConnectionError('Zenodo download failed after 3 attempts!')
+
+# %%
+filepath = data_folder
+seq_filename = '20240130_CEST_interleafed_radial_256px_fov256_8mm_50spokes_100ms.seq'
```
ckolbPTB: Seq file is not used

examples/radial_interleafed_example.py
```diff
@@ -0,0 +1,125 @@
...
+h5_filename = 'meas_MID00044_FID00370_20240130_CEST_interleafed_radial_256px_fov256_8mm_50spokes_100m.h5'
```
ckolbPTB, line 37: maybe we can think of filenames which are a bit nicer to read...

examples/radial_interleafed_example.py
```diff
@@ -0,0 +1,125 @@
...
+# %% [markdown]
+# # Read in data
+# Data is read in using the downloaded .h5 file and a corresponding
+# trajectory calculator (cartesian in this case). Given the trajectory
```
ckolbPTB: (2D Golden radial in this case)

examples/radial_interleafed_example.py
```diff
@@ -0,0 +1,125 @@
+# %% [markdown]
+# # Basic example for reconstruction of radial interleafed sampled data
```
ckolbPTB: interlea**v**ed

examples/radial_interleafed_example.py
```diff
@@ -0,0 +1,125 @@
...
+# This notebook is meant as a first introduction inot mrpro-based
```
ckolbPTB: into

examples/radial_interleafed_example.py
```diff
@@ -0,0 +1,125 @@
...
+# reconstruction of radial interleafed data.
```
ckolbPTB: What does the "interleaved" actually refer to?

examples/radial_interleafed_example.py
```diff
@@ -0,0 +1,125 @@
...
+# Data is read in using the downloaded .h5 file and a corresponding
+# trajectory calculator (cartesian in this case). Given the trajectory
+# calculator, mrpro will automatically read, sort and validate the data.
+# %%
+data = KData.from_file(
+    ktrajectory=KTrajectoryRadial2D(),
+    filename=f'{filepath}/{h5_filename}',
+)
+# %% [markdown]
+# # Reconstruction
```
ckolbPTB, line 49: Move density compensation to separate cell with explaination

examples/radial_interleafed_example.py
```diff
@@ -0,0 +1,125 @@
...
+# Given the loaded data, the Fourier Operator is initialized next,
+# with which the appropriate Fourier Transform is carried out.
+# Furthermore, density compensation and coil sensitivity maps are created to
+# control for varying coil sensitivities ad densities throughout the scan.
```
ckolbPTB: remove "to control for varying coil sensitivities ad densities throughout the scan."

examples/radial_interleafed_example.py
```diff
@@ -0,0 +1,125 @@
...
+# %%
+# Densitiy compensation
```
ckolbPTB: Density

examples/radial_interleafed_example.py
```diff
@@ -0,0 +1,125 @@
...
+dcf = DcfData.from_traj_voronoi(traj=data.traj)
+
+# %%
+# perform FT and CSM
```
ckolbPTB, line 55: I would also here make separate steps: 1) Fourier transform yielding images of shape (other, coils, z, y, x) and then 2) calculate coil maps of shape (other, coils, z, y, x) and 3) coil combination to yield images of shape (other, 1, z, y, x)

examples/radial_interleafed_example.py
```diff
@@ -0,0 +1,125 @@
...
+
+smoothing_width = SpatialDimension(z=1, y=5, x=5)
+csm = CsmData.from_idata_walsh(idata, smoothing_width)
+sensitivity_op = SensitivityOp(csm)
+(x,) = sensitivity_op.H(xcoils)
+
+# %% [markdown]
+# # See the results
+# %%
+image = x.abs().square().sum(1).sqrt()
```
ckolbPTB: Why summation? Coils have already been combined... Abs is called again during plotting...

examples/radial_interleafed_example.py
```diff
@@ -0,0 +1,125 @@
...
+image = image.squeeze()
```
ckolbPTB: I would prefer something like `plt.matshow(i[0,0,0,:,:]) # plotting single slice`

examples/radial_interleafed_example.py
```diff
@@ -0,0 +1,125 @@
...
+for i in image:
+    print(i.shape)
+    plt.matshow(torch.abs(i))
+    break
```
ckolbPTB: why break?

examples/cartesian_reconstruction_example.py
schuenke: You mentioned this in #91 already. I like the idea, but maybe we should continue the discussion there.

examples/radial_interleafed_example.py
JoHa0811: Done, thanks :)

examples/radial_interleafed_example.py
JoHa0811: removed the line, thank you

examples/radial_interleafed_example.py
JoHa0811: I simplified the entire code-block and made it more comprehensible (hopefully)
Same goes for the comment below

examples/radial_interleafed_example.py
JoHa0811: You're right, this was unneccessary. I simplified the expression

examples/radial_interleafed_example.py
JoHa0811: As far as I understand the densitiy compensation needs to be calculated before the coil sensitvitiy maps, since idata receives xcoils, which in turn is generated via "ft_op.H(data.data * dcf.data)"
I have disentangled the logic as far as I could. Please let me know if it can be refactored more :)

ckolbPTB:
superseded by other PRs on radial (#235) and Cartesian reconstruction (#233)

PR #161, opened 2024-02-12
Add separable nd-filters and sliding window

fzimmermann89:
some utilities: 
uniform filter,  gaussian filter, sliding windows views

PR #173, opened 2024-02-16
Inati

### LINKED ISSUE #64: Csm inati method

JoHa0811:
Implements the Inati CSM method and integrates the new folder style 
closes #64 and #161


src/mrpro/data/_CsmData/_CsmData.py
```diff
@@ -93,3 +93,21 @@ def from_idata_walsh(
         csm_data = csm_fun(idata.data)
 
         return cls(header=idata.header, data=csm_data)
+
+    @classmethod
+    def from_idata_inati(cls, data: torch.Tensor, ks: int, power: int, padding_mode='circular') -> CsmData:
```
schuenke: if the method is called `from_idata...` I would expect that data is of type IData

src/mrpro/data/_CsmData/_CsmData.py
```diff
@@ -93,3 +93,21 @@ def from_idata_walsh(
...
+        """Coil sensitivity maps using the Inati method.
+
+        Details of the method can be found in Inati et al. 2004.
+
+        Parameters
+        ----------
+        data: Images of shape (coil, E1, E0)
+        ks: kernel size
+        power: number of iterations
+        padding_mode: padding mode for the sliding window
```
schuenke, line 108: would be nice to use the same style as in all other methods...

```python
"""
...
Parameters
----------
data
    images of shape...
ks
   kernel size
...
"""

```

src/mrpro/data/_CsmData/_inati.py
```diff
@@ -0,0 +1,36 @@
...
+
+def coil_map_study_2d_Inati(data: torch.Tensor, ks: int, power: int, padding_mode='circular'):
+    """Coil sensitivity maps using the method described in Inati et al. 2004.
+
+    Parameters
+    ----------
+    data: Images of shape (coil, E1, E0)
+    ks: kernel size
+    power: number of iterations
+    padding_mode: padding mode for the sliding window
```
schuenke, line 14: see comment about style from above

src/mrpro/data/_CsmData/_inati.py
```diff
@@ -0,0 +1,36 @@
+import torch
+
+from mrpro.utils.window import sliding_window
+
+
+def coil_map_study_2d_Inati(data: torch.Tensor, ks: int, power: int, padding_mode='circular'):
+    """Coil sensitivity maps using the method described in Inati et al. 2004.
```
schuenke, line 7: Maybe use a `References` section as recommended in the [official style guide](https://numpydoc.readthedocs.io/en/latest/format.html#references)

or at least list the complete information including journal, doi etc

JoHa0811: Yes, good call!
Only problem is that i don't know which paper this is from, as this was the only reference in @fzimmermann89 implementation. I will happily beautify this, if someone could give me a link to the paper :)

ckolbPTB: 1. Inati S J, Hansen, M. S. & Peter, K. A solution to the phase problem in adaptvie coil combination. in Proceedings of the 21st Annual Meeting of ISMRM, Salt Lake City, USA, 2672 (2013).
2. Inati, S. & Hansen, M. A Fast Optimal Method for Coil Sensitivity Estimation and Adaptive Coil Combination for Complex Images. in Proceedings of Joint Annual Meeting ISMRM-ESMRMB, Milan, Italy, 7115 (2014).

fzimmermann89:
The biggest issues with this code are that we need to 
a) check if it reproduces ptbpyrecon
b) extend it to 3d
c) proper tests
I am not to convinced by having an example for this.

As this is still just a draft that only copied over my code, I would abstain from any further comments and wait until we have #161 merged (this was on purpose a separate PR first)

I do not agree that this suppresseds #161 as all issues mypy complained about were silenced here by removing the typehints.

Please don't do that!

Instead, you are invited to continue working on #161 first.

ckolbPTB:
Superseded by #345

PR #178, opened 2024-02-20
KData functions refactored into Mixin classes

ckolbPTB:
The only way I could make it work is by using static methods. 

For something like 
```
class KDataSplitMixin():
     def split_data(self, idx):
```
`self` will be of type `KDataSplitMixin` and hence `self.header` and so on will lead to errors because `header` is not defined for `KDataSplitMixin`

The solution I found was to define protocols: https://mypy.readthedocs.io/en/latest/more_types.html#mixin-classes but this does not seem viable for our classes.

Another point for discussion is the naming. Having `KData` in front of every mixin class seems to be overkill but on the other hand makes it very clear what it does without checking in which folder this class is defined. 

`prewhiten_kspace` and `remove_readout_os` also need to be modified like these methods but I would do this then in a separate PR.


schuenke:
hmm... I dont really like the staticmethods and the syntax that comes along with it... 

```python
kdata = KData.rearrange_k2_k1_into_k1(kdata_before)
```

is much more unpleasant than

```python
kdata = kdata.rearrange_k2_k1_into_k1()
```

imo.

I will have a look again and check if I find a better way to solve it.

schuenke:
The only solution I found so far is using `Protocol` from `typing` as also mentioned in the [link](https://mypy.readthedocs.io/en/latest/more_types.html#mixin-classes) @ckolbPTB posted.

We would need to define a protocol like

```python
from typing import Protocol

class KDataProtocol(Protocol):
    @property
    def header(self) -> KHeader: ...
    @property
    def data(self) -> torch.Tensor: ...
    @property
    def traj(self) -> KTrajectory: ...

    def __init__(self, header: KHeader, data: torch.Tensor, traj: KTrajectory): ...
```

and than use it for type hints in the MixIn classes... for example:

```python
class KDataRearrangeMixin:
    """Rearrange KData."""

    def rearrange_k2_k1_into_k1(self: KDataProtocol) -> KDataProtocol:
        ...
```

This works for me. Passes mypy and pytest. 

However, I first thought it would work without defining the `__init__` in the protocol, which is unfortunately not the case. 
Without this, it would be a nice solution imo. With this required definition I am not 100% convinced tbh.

What do you think @ckolbPTB @fzimmermann89

fzimmermann89:
Can you remove the init and type hint the return as Self?
Maybe you need to do self.__cls__(..) for the return inside the function.

schuenke:
> Can you remove the init and type hint the return as Self? Maybe you need to do self.**cls**(..) for the return inside the function.

Not sure if I get what you suggested... currently this MixIn class would look like this:

```python
class KDataRearrangeMixin:
    """Rearrange KData."""

    def rearrange_k2_k1_into_k1(self: KDataProtocol) -> KDataProtocol:
        """Rearrange kdata from (... k2 k1 ...) to (... 1 (k2 k1) ...).

        Parameters
        ----------
        kdata
            K-space data (other coils k2 k1 k0)

        Returns
        -------
            K-space data (other coils 1 (k2 k1) k0)
        """

        # Rearrange data
        kdat = rearrange(self.data, '... coils k2 k1 k0->... coils 1 (k2 k1) k0')

        # Rearrange trajectory
        ktraj = rearrange(self.traj.as_tensor(), 'dim ... k2 k1 k0-> dim ... 1 (k2 k1) k0')

        # Create new header with correct shape
        kheader = copy.deepcopy(self.header)

        # Update shape of acquisition info index
        def reshape_acq_info(info):
            return rearrange(info, 'other k2 k1 ... -> other 1 (k2 k1) ...')

        kheader.acq_info = modify_acq_info(reshape_acq_info, kheader.acq_info)

        return type(self)(kheader, kdat, type(self.traj).from_tensor(ktraj))

```

If I remove the `__init__` from `KDataProtocol`, mypy complains:
`error: Too many arguments for "KDataProtocol"  [call-arg]`

Which return do you suggest to type hint as `self`? And how?

ckolbPTB:
I am not sure the Protocol-approach scales very well. I haven't tested it but I guess if we define another Mixin which uses `rearrange_k2_k1_into_k1` then we would have to define a second Protocol class which has this function and so in each Mixin we might have a dedicated Protocol class. Not sure we want that...

schuenke:
The `KDataProtocol(Protocol)` is defined once for KData. For every MixIn class (e.g. KDataRearrangeMixin) we just import the KDataProtocol and use it for type hints. So if we add more MixIns for KData, we don't have to define another Protocol.

Or did you mean another Class that might import `rearrange_k2_k1_into_k1` as a method? In this case, we would need to define another Protocol for this class as well, sure. But currently we assume `rearrange_k2_k1_into_k1` is only used by KData, no? This is why it's placed in the `_kdata` folder?!

schuenke:
The error you described in your first post can btw be solved using KData as type hint for self:

```python
class KDataSplitMixin():
     def split_data(self: KData, idx):
```

However, in our case this leads to the error
`"The erased type of self '...*' is not a supertype of its class"`
which I didn't manage so resolve yet...

ckolbPTB:
What I meant was that if we add another Mixin, e.g. `KDataArrangeMixin` which uses the function of another Mixin

```
class KDataArrangeMixin:
    def arrange(self: KDataProtocol) -> KDataProtocol:
        x = self.rearrange_k2_k1_into_k1(self)
        return x
```

Then this would not work because `KDataProtocol `does not have a function `rearrange_k2_k1_into_k1`. Can you simply add functions to a Protocol class? Would we then have to create a Protocol-class which includes all properties and functions of `KData`?

schuenke:
> What I meant was that if we add another Mixin, e.g. `KDataArrangeMixin` which uses the function of another Mixin
> 
> ```
> class KDataArrangeMixin:
>     def arrange(self: KDataProtocol) -> KDataProtocol:
>         x = self.rearrange_k2_k1_into_k1(self)
>         return x
> ```
> 
> Then this would not work because `KDataProtocol `does not have a function `rearrange_k2_k1_into_k1`. Can you simply add functions to a Protocol class? Would we then have to create a Protocol-class which includes all properties and functions of `KData`?

Ah, got it... I will test it and report here, but I guess we have to add all methods to the protocol, yeah.

schuenke:
I just pushed the implementation using Protocol. It's not meant to be the final version, but I guess it's easier to discuss this option having the code available here... 

In summary:

- we use KDataProtocol for type hinting in the MixIn classes
- all methods, that are used within the MixIn classes, must be in the Protocol
- currently there is no option to disable pydocstyle checks for a single file/class. This is why I had to put "useless docstrings" to all KDataProtocol methods as well. There are some discussions to introduce an ignore-decorator for pydocstyle, but it's not available atm

ckolbPTB:
> all methods, that are used within the MixIn classes, must be in the Protocol

Potentially we would also need to add **all** the KData methods to Protocol should we need them in an additional function.

schuenke (approved):
In the current form we have KData and _KDataProtocol implemented in `_KData.py` and all mixin classes defined in separate files in the same `_kdata` subdirectory.

`_KDataProtocol ` is required for type hinting in the mixin classes and all properties and methods that are used in the mixin classes must be defined in `_KDataProtocol`.

Properties/methods that are not used in the mixin classes are not defined in `_KDataProtocol` !

Using the leading underscore marks `_KDataProtocol` as a private class so that its properties and methods do not require a docstring in our current setup, which prevents duplication of documentation and thus potential sources of errors/mismatches. 

IMO this is the cleanest solution and we should adapt CsmData, DcfData etc. in the same way.

fzimmermann89:
Can we maybe split this PR into two parts:

I agree with the protocol etc but am not really happy with the reshaping functions we used to test it out..

Maybe we can do one PR with the protocols only moving existing functions into new files and a second PR with the respaong logic?

PR #179, opened 2024-02-21
Use shoelace equation for 2d dcf area calculation

fzimmermann89:
Rel: #169


fzimmermann89:
what happend here??

fzimmermann89:
Ops. I forced pushed to this PR after pushing the only changed contained to main.
I thought a branch protection would prohibit me from doing a push to main.
This explains why this PR is closed now...

Issue #180, opened 2024-02-23
RFC: Coil Compression


fzimmermann89:
I would like to use a simple SVD coil compression.

Questions is how and where to implement it.

## Version 1 ##
First I thought for me it might be a function 
`coil_compress(target:Kdata|Idata, basis:KDataIData|None,  virutual_coils:int, compressionmatrix=None,separate_dims:tuple[int]=None, return_Compressionmatrix:bool) -> Kdata | tuple[Kdata, tensor]`

Explanation:
`target`: that will be compressed
if `basis` is None: Use target to figure out the compression matrix. if IData or KData: use IData.data to figure out the compression matrix
`virtual_coils`: number of virtual coils to compres into.

if `compressionmatrix` is supplied, use it instead of calculating a new one.
`return_Compressionmatrix`: wether the compression matrix be returned as well for reuse on a different dataset.

not sure about how to specify broadcasting:
Like should the same compression be applied for all "other" axis or separate?
I consider `separate_dims` which defaults to the same compression for all other dimensions, but you can give a tuple of dimensions numbers the operations will be looped over. So specifiny sparate_dims=0 would mean: Make a different compression matrix for each 'other'.
Or should this be done by first selecting only the part of KData that should be used and looping outside of the function?

## Version 2 ##
Even though initially I thought having another operator for the compression matrix might be overkill. 
But it would allow for writing
```
coil_compressinion=Compression.from_kdata(k)
compressed=coil_compression(other_k)
```
so it would a bit cleaner to apply it to different data..

## Discussion ##
Now after writing this, I think I prefer  version 2... ....
What do you think? Maybe @ckolbPTB  @schuenke @JoHa0811  (and of course anybody else who has an opinion on that!)

ckolbPTB:
I also prefer version 2, because the compression matrix should usually be calculated for a subset of the data (e.g. only for one echo and then applied to all other echos). With the above split that can be easily done by the user, otherwise we would have to put this logic into the coil_compression function.

ckolbPTB:
closed by #470

PR #181, opened 2024-02-23
Add PCA-based compression operator

fzimmermann89:
Rel #180


fzimmermann89:
this looks ugly...

fzimmermann89:
So, coil compression as an operator? Or some function?

ckolbPTB:
Would it make sense to have a PCA operator which is then used by a coil-compression function?

fzimmermann89:
One of the goals is to be able to first construct the compression matrix, then apply it to different data.

So we could move the application to kdata into a function

fzimmermann89:
We should consider how to deal with Data object in operators where it makes sense to apply them to kdata or Idata as well as to tensors.

Either we use some dispatch library, stdlib single dispatch, manual overloads and isinstance or mixing/traits, some optimal subclass manipulation in linear operator (similar to the autograd stuff)

fzimmermann89:
pre-commi.ci autofix

src/mrpro/operators/_CoilCompression.py
```diff
@@ -0,0 +1,89 @@
...
+    def __init__(self, data: Data | torch.Tensor, n_components, separate_dims: tuple[int, ...] | None = None):
+        """Construct a PCA based soft coil compression operator.
+
+        Parameters
+        ----------
+        data
+            Data to used to find the principal components.
+        n_components
+            Number of principal components to keep.
+        separate_dim, optional
```
fzimmermann89:
```suggestion
        separate_dims
```

src/mrpro/operators/_DataOperator.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+
+    def __init__(self, operator: Operator[torch.Tensor, tuple[torch.Tensor,]], return_type: type[Data] | None = None):
+        """Initialize the DataOperator.
+
+        Parameters
+        ----------
+        Operator
+            The operator to apply
+        returntype
```
fzimmermann89:
```suggestion
        operator
            The operator to wrap
        return_type
```

src/mrpro/operators/_DataOperator.py
```diff
@@ -0,0 +1,49 @@
+import dataclasses
+from copy import deepcopy
+
+import torch
+
+from mrpro.data import Data
+from mrpro.operators import Operator
+
+
+class DataOperator(torch.nn.Module):
```
fzimmermann89: Spit out of this pr

src/mrpro/operators/_CoilCompression.py
```diff
@@ -0,0 +1,89 @@
...
+import torch
+
+from mrpro.data import Data
+from mrpro.operators import LinearOperator
+
+
+class CoilCompression(LinearOperator):
+    """PCA based soft coil compression operator."""
+
+    def __init__(self, data: Data | torch.Tensor, n_components, separate_dims: tuple[int, ...] | None = None):
```
fzimmermann89:
```suggestion
    def __init__(self, data: Data | torch.Tensor, n_components:int, separate_dims: tuple[int, ...] | None = None):
```

src/mrpro/operators/_CoilCompression.py
```diff
@@ -0,0 +1,89 @@
+import einops
+import numpy as np
+import torch
+
+from mrpro.data import Data
+from mrpro.operators import LinearOperator
+
+
+class CoilCompression(LinearOperator):
```
fzimmermann89: Rename to PCACompressionOp

fzimmermann89:
Does anybody want to take this PR?

The signature should be changed, instead of separate_dims as an argument and only working on dim -4, there should be a dim argument which dimension to compress, defaulting to -4.
And a shared_dims (tuple of int or ellipsis) with the opposite meaning of separate dims. A single compression matrix should be constructed for these dimensions, with ellipsis (default) meaning all dimensions. For dimensions not specified as shared, normal natxhing shall be used.

The data operator part should just be removed here.

fzimmermann89:
@ckolbPTB are you interested?

ckolbPTB:
To me separate dims seems to more intuitive parameter and easier to define if the shape of data used for calculating the compression matrix is different to the input data to forward.

ckolbPTB:
I would then add a `compress_coils` algorithm in a separate PR

fzimmermann89:
> To me separate dims seems to more intuitive parameter and easier to define if the shape of data used for calculating the compression matrix is different to the input data to forward.

yes, the implementation would be easier, but using it would be a bit inconsistent, as we assume all leading dimensions are always batch dimensions. 
otherwise adding another leading dimensions needs a code change to batch along that dimensions. this is not what we want.

This is why I deliberately suggested having the inverse, a joint_dim argument -- and why I tried to find somebody else to tackle this ;)

ckolbPTB:
> yes, the implementation would be easier, but using it would be a bit inconsistent, as we assume all leading dimensions are always batch dimensions.

I did not mean the implementation but the usage. One example woud be PCA-based compression for qMRI. Let's assume we have one dimension a which is supposed to be treated separately. We would calculate N signal curves in the shape (a,t,N) and then calculate the compression matrix for this. This compression matrix would then be applied to the data of shape (a,t,z,y,x). If I used joint_dim then I would have to specific joint_dim=-1 for the init but then change it when applying to the data. If I used separate_dims it would be 0 for both. 

Also for coil compression I would like to define the separate_dim as the slice dimension and not care if I have additional leading dimensions due to dynamics or echos or something like this.

fzimmermann89:
If I am not mistaken we have to specify four dimensions:
- the dimension in which to compress (coils or time)  when creating the compression 
- dimension to compress when applying the data
-  dimensions to batch over vs dimensions to consider as samples on the data matrix when creating the compression 
- how to broadcast the result at application 

 
Either we allow to specify all of them or certain applications require either unsqueezes or reaaranges.

We could just keep the separate&dim logic for know, although it is kind of a foot gun that adding a batch dimension in front of the data will make this by default a shared dimension.

Maybe at some point er can come up with better definitions here what do to.

fzimmermann89:
>
> Also for coil compression I would like to define the separate_dim as the slice dimension and not care if I have additional leading dimensions due to dynamics or echos or something like this. 

This breaks both pytorch and somewhat mrpro conventions: leading dimensions are considered batch dimensions, thus should be treated as separate.
 If you want to share compression along dynamics, for example, you should specify that. If you then use batches of different subjects, it should treat the new 0tj dim as a separate batch dim. So the logic is exactly the opposite.

fzimmermann89:
If I am not mistaken we have to specify three dimensions:
- the dimension in which to compress (coils or time)  when creating the compression 
- dimension to compress when applying the data
-  dimensions to batch over vs dimensions to consider as samples on the data matrix when creating the compression (this is where we disagree) - do you want to use the same compression for different slices, subjects etc or the same?
 
Either we allow to specify all three, or certain applications require either unsqueezes or reaaranges.

I used the code to construct one coil compression operator based on one sample and applied it to differently shaped inputs with different batch dimensions
So it would be creation_dim=coil, joint_dim=z y x, application_dim=coil. At application this should now broadcast in the leading dimensions of the data and the same compression is applied for rotation,slice,....

In a qmri example, you would want to create the matrix with (t,N)-shaped samples, and compress in time. You want to apply it to data 
( t, slice, rotation, coil,zyx)
Here the creation dim is 0, the application dim is 0 and the there is no joint_dim.

If the data were 
( Subject, t, slice, rotation, coil,zyx) you would have to specify an application dimension or create the matrix with (subject or 1,t,N) samples.

Specifying separate dimensions instead of joint dimensions would not help here.

fzimmermann89:
Also, thinking about it, maybe the differences between use in qmri and coil compression are greater than the common parts. And itigjt ne easier to just have to things that do a bit of (different) reshaping before applying a PCA ..?

ckolbPTB:
> We could just keep the separate&dim logic for know, although it is kind of a foot gun that adding a batch dimension in front of the data will make this by default a shared dimension.

But isn't this the definition of a batch dimension?

fzimmermann89:
> > We could just keep the separate&dim logic for know, although it is kind of a foot gun that adding a batch dimension in front of the data will make this by default a shared dimension.
> 
> But isn't this the definition of a batch dimension? 

It would usually be separate dimension, i.e. having a batch dimension is equivalent to having a for loop over the different batch samples. So no joint operations connecting the batch samples (beside batch norm as THE special case)
L

fzimmermann89:
Example:  having a batch of three subjects should do separate coil compressions for each of the subjects.

ckolbPTB:
@fzimmermann89 could you have a look at this please?

src/mrpro/operators/PCACompressionOp.py
```diff
@@ -0,0 +1,84 @@
...
+        The operator carries out an SVD followed by a threshold of the n_components largest values along the last
+        dimension of a data with shape (*other, joint_dim, compression_dim). A single SVD is carried out for everything
+        along joint_dim. Other are batch dimensions.
+
+        Consider combining this operator with RearrangeOp to make sure the data is in the correct shape before applying.
+
+        Parameters
+        ----------
+        data
+            Data to be used to find the principal components of shape (*other, joint_dim, compression_dim)
```
schuenke: Maybe change to:
`Data of shape (*other, joint_dim, compression_dim) to be used to find the principal components. `
Else sounds like the components have that shape...

src/mrpro/operators/PCACompressionOp.py
schuenke, line 1: Maybe we want to add the option to determine n_components automatically from the data? For an old project I implemented this using one approach and looked up two other ones, that I didn't implement yet. Not sure if we want this option (n_components could then maybe be of type int or Literal(implemented methods)), but I will paste my old code/references here just in case:

```python
def det_malinowski_factor(eigenvalues: np.ndarray) -> int:
    """
    Determines minimum number of principal components according to: Malinowski, E. R. Determination of the number of
    factors and the experimental error in a data matrix. Anal. Chem. 49, 612–617 (1977).
    """
    ind = np.zeros([eigenvalues.shape[0]-1, 1])
    for n in range(ind.shape[0]):
        ind[n] = np.sqrt(np.sum(eigenvalues[n:]/(eigenvalues.shape[0]-(n+1)))) / (eigenvalues.shape[0]-(n+1))**2

    return int(np.argmin(ind))


def det_nelson_factor(eigenvalues: np.ndarray) -> int:
    """
    Determines minimum number of principal components according to: Nelson, L.R. (2005). Some observations on the
    scree test, and on coefficient alpha. Thai Journal of Educational Research and Measurement: 3(1), 1-17.
    """
    # Todo: implement function :)
    pass


def det_median_noise_factor(eigenvalues: np.ndarray) -> int:
    """
    Determines minimum number of principal components according to: Manjón, J.V. et al. (2015). MRI noise estimation and
    denoising using non-local PCA. Medical Image Analysis 22 (2015) 35–47
    """
    # Todo: implement function :)
    pass

```

schuenke (approved):

src/mrpro/operators/PCACompressionOp.py
ckolbPTB: new issue #479

ckolbPTB:
@fzimmermann89 could you merge this or is anything still missing?

fzimmermann89:
I will check later and merge it.

fzimmermann89:
pre-commit.ci autofix

fzimmermann89:
Does anybody know if it is possible to make RearrangeOp a link in the documentation?

fzimmermann89:
> Does anybody know if it is possible to make RearrangeOp a link in the documentation?

? @lrlunin or @ckolbPTB

ckolbPTB:
> Does anybody know if it is possible to make RearrangeOp a link in the documentation?

I don't think I understand the question....

fzimmermann89:
In function signatures etc, class names are linked in the documentation to the respective page.

I was trying to make the reference to `RearrangeOp` in the docstring a clickable link in the documentation -- but failed.

fzimmermann89:
image
vs 
image

PR #182, opened 2024-02-23
Cart sampling index int64

ckolbPTB:
trajectory is float32 which is not sufficient for accurate indexing of large datasets. Needs to be int64


fzimmermann89:
good catch!

schuenke (approved):
lgtm

PR #183, opened 2024-02-24
Direct Reconstruction

fzimmermann89:
a helper for kdata -> idata using simple direct (adjoint fourier) reconstruction

This is not ready for review in the sense of 'ready for merge', but I would like to get some input if we want to have sth like this and how we want to do it


fzimmermann89:
This is more an RFC, should we at some point add algorithms like that?

Should these be Classes or just Functions?
Not sure if it is useful to construct the adjoint reconstruction and apply it to different data, 
or if it should just be a function that does the steps currently in` __init__` and `forward` in one go.


(Also, it contains some fixes for general operator logic, some of them included in #189 and the Data-As-Buffer #188 and should be rebased maybe split into 2-3 commits to main.)

fzimmermann89:
This is at least a handy  short cut to get some images on screen:
```python
from mrpro.data import KData, KNoise
from mrpro.algorithms.reconstruction import AdjointReconstruction
from mrpro.data.traj_calculators import KTrajectoryRadial2D
filename = "test.h5"

k = KData.from_file(filename, KTrajectoryRadial2D())
# for now required 
k.header.encoding_matrix.z = 1
k.header.recon_matrix.x = 240
k.header.recon_matrix.y = 240

noise = KNoise.from_file(filename, 0)

reco = AdjointReconstruction.from_kdata(k, noise)
image = reco(k)

plt.matshow(image.rss().squeeze())
````

**Comment by fzimmermann89** [comment]
I removed some stuff to make this PR a bit easier to discuss.

**Comment by fzimmermann89** [comment]
@ckolbPTB
Whats the status here?
Should I polish it up (add tests) and rename it to direct reconstruction?

Are there any issues with to general structure and idea?

**Comment by ckolbPTB** [comment]
> Should I polish it up (add tests) and rename it to direct reconstruction?

yes please

> Are there any issues with to general structure and idea?

no

**Comment by fzimmermann89** [comment]
Some of the changes:
- Move EinsumOp to einops.einsum, this allows multi-character axis names, similar to rearrange (I want to use k1, coil,...)
- Add image.rss helper function, returning the coil combined image. if is is already coil combined, it can still be used to get a magnitude image. 
- add example notebook
-  change docstrings in existing radial example notebook to mention the new one as a more high level approach
~~-  allow dcfdata@operator ( this is not yet tested outside of the notebook)~~
- add .cpu / .to /. cuda to KNoise. 
- moved some imports into functions to avoid circular imports in the data classes.

missing:
- do we need another tests for the direct reconstruction or is the notebook test good enough for now? -> Good enough
~~- any suggestions for a simple test?~~
- tests for knoise.cpu/to/cuda
~~- test for dcf matmul~~

**Comment by fzimmermann89** [comment]
- added as_tensor functions to dcfdata and csmdata
- fixed the imports in __init__'s: vscode's pyright did not resolve `mrpro.data` if we use `import mrpro.data` in the `__init__`
- Fix issues with typing in the operator class: At least in the direct_reconstruction example now all types are correct. Before, a lot of the compositions did not have a return type.
Here I needed to ignore two typing errors: first, the typing overlap with torch.Tensor for rmul (reported to pytorch) and the other one is that mypy complains in the matmul of LinearOperator, that the arguments are wider than those of the Operator-Baseclass. This is IMHO wrong, mypy just does not 'get' there that a LinearOperator is also an Operator, thus accepting Operator and LinearOperator should be the same as accepting Operator....
- Undo using "@" to apply an Operator to a tensor. If we want to have this, we can re-add this in the future, I was getting typing issues and did not felt this is worth the hassle to add it in this PR..
-undo the @ application of dcf. instead, I use the as_tensor function
- add 2 different levels of manual reconstruction and asserts that the results are the same.

**Comment by fzimmermann89** [comment]
Are we ok with this PR touching different parts of the code and fixing the typing issues etc as well?
Or should I split it into to PRs?

File: examples/pulseq_2d_radial_golden_angle.py [hunk 1]
```diff
@@ -1,5 +1,9 @@
 # %% [markdown]
 # # Reconstruction of 2D golden angle radial data from pulseq sequence
+# Here we manually do all steps of a direction reconstruction, i.e.
+# CSM estimation, density compensation, coil combination and inverse transform.
```
ckolbPTB: maybe: inverse transform -> pseudo-inversion transform

src/mrpro/algorithms/reconstruction/_direct_reconstruction.py
```diff
@@ -0,0 +1,122 @@
...
+from mrpro.data._DcfData import DcfData
+from mrpro.data._IData import IData
+from mrpro.data._kdata._KData import KData
+from mrpro.data._KNoise import KNoise
+from mrpro.operators._FourierOp import FourierOp
+from mrpro.operators._LinearOperator import LinearOperator
+
+
+class DirectReconstruction(Reconstruction):
+    """Direct Reconstruction by Inverse Fourier Transform."""
```
ckolbPTB: maybe also: Pseudo-inverse

src/mrpro/algorithms/reconstruction/_direct_reconstruction.py
```diff
@@ -0,0 +1,122 @@
...
+        """Initialize DirectReconstruction.
+
+        Parameters
+        ----------
+        fourier_operator
+            Instance of the FourierOperator which adjoint is used for reconstruction.
+        csm
+            Sensitivity maps for coil combination. If None, no coil combination will be performed.
+        noise
+            Used for Prewhitening
```
ckolbPTB: prewhitening

src/mrpro/algorithms/reconstruction/_direct_reconstruction.py
```diff
@@ -0,0 +1,122 @@
...
+        dcf
+            Density compensation. If None, no dcf will be performed.
+            Also set to None, if the FourierOperator is already density compensated.
+        """
+        super().__init__()
+        self.FourierOperator = fourier_operator
```
ckolbPTB: why is this not self.fourier_operator? based on our naming convention I think it should actually be self.fourier_op

src/mrpro/algorithms/reconstruction/_direct_reconstruction.py
```diff
@@ -0,0 +1,122 @@
...
+from mrpro.operators._LinearOperator import LinearOperator
+
+
+class DirectReconstruction(Reconstruction):
+    """Direct Reconstruction by Inverse Fourier Transform."""
+
+    def __init__(
+        self,
+        fourier_operator: LinearOperator,
+        csm: None | CsmData = None,
```
ckolbPTB, line 35: I would use csmdata, dcfdata and knoise to higlight that these are dataclasses

src/mrpro/algorithms/reconstruction/_direct_reconstruction.py
```diff
@@ -0,0 +1,122 @@
...
+        """
+        super().__init__()
+        self.FourierOperator = fourier_operator
+        # TODO: Make this buffers once DataBufferMixin is merged
+        self.dcf = dcf
+        self.csm = csm
+        self.noise = noise
+
+    @classmethod
+    def from_kdata(cls, kdata: KData, noise: KNoise | None = None, coil_combine: bool = True) -> Self:
```
ckolbPTB, line 61: maybe "do_coil_combine" to highlight that this is a bool?

src/mrpro/algorithms/reconstruction/_direct_reconstruction.py
```diff
@@ -0,0 +1,122 @@
...
+    def recalculate_csm_walsh(self, kdata: KData, noise: KNoise | None = None):
+        """Update the CSM from KData using Walsh.
+
+        Parameters
+        ----------
+        kdata
+            KData used for adjoint reconstruction, which is then used for
+            Walsh CSM estimation.
+        noise
+            Noise measurement for prewhitening (optional)
```
ckolbPTB: remove "optional"

src/mrpro/algorithms/reconstruction/_direct_reconstruction.py
```diff
@@ -0,0 +1,122 @@
...
+        """Apply the reconstruction.
+
+        Parameters
+        ----------
+        kdata
+            k-space data to reconstruct.
+
+        Returns
+        -------
+        image
```
ckolbPTB: should be idata

src/mrpro/algorithms/reconstruction/_direct_reconstruction.py
```diff
@@ -0,0 +1,122 @@
...
+        device = kdata.data.device
+        if self.noise is not None:
+            kdata = prewhiten_kspace(kdata, self.noise.to(device))
+        operator = self.FourierOperator
+        if self.csm is not None:
+            operator = operator @ self.csm.as_operator()
+        if self.dcf is not None:
+            operator = self.dcf.as_operator() @ operator
+        operator = operator.to(device)
+        (image_data,) = operator.H(kdata.data)
```
ckolbPTB: should be img

src/mrpro/data/_IData.py
```diff
@@ -65,6 +65,11 @@ class IData(Data):
 
     header: IHeader
 
+    def rss(self, keepdim: bool = False) -> torch.Tensor:
```
ckolbPTB, line 68: doc string incomplete

ckolbPTB:
Based on our discussion when we switched to ruff and the infor here: https://ptb-mr.github.io/mrpro/contributor_guide.html

The naming in the example script should be adapted _operator -> _op, and image_tensor -> img and so on

examples/ruff.toml
```diff
@@ -1,3 +1,9 @@
 extend = "../pyproject.toml"
 
-lint.extend-ignore = ["D100", "T20"]
+lint.extend-ignore = [
+    "D100",   #undocumented-public-module
+    "T20",    #print
+    "E402",   #module-import-not-at-top-of-file
+    "ERA001", #commented-out-code
```
ckolbPTB: Do we really need this? I would prefer to not ignore that

src/mrpro/operators/_DensityCompensationOp.py
```diff
@@ -0,0 +1,39 @@
...
+
+from __future__ import annotations
+
+import torch
+
+from mrpro.data._DcfData import DcfData
+from mrpro.operators._EinsumOp import EinsumOp
+
+
+class DensityCompensationOp(EinsumOp):
```
ckolbPTB, line 25: needs  a test

src/mrpro/algorithms/reconstruction/_direct_reconstruction.py
fzimmermann89: but this is already given by the type?

src/mrpro/algorithms/reconstruction/_direct_reconstruction.py
fzimmermann89: this is already defined by the type?

src/mrpro/algorithms/reconstruction/_direct_reconstruction.py
fzimmermann89: mhm.
```
        (img_data,) = operator.H(kdata.data)
        img = IData.from_tensor_and_kheader(img_data, kdata.header)
```
```
        (img,) = operator.H(kdata.data)
        image= IData.from_tensor_and_kheader(img_data, kdata.header)
```
```
        (img,) = operator.H(kdata.data)
        idata= IData.from_tensor_and_kheader(img, kdata.header)
```

the last one looks really strange to me, because the data of the image is named image, and the container is named data....

src/mrpro/algorithms/reconstruction/_direct_reconstruction.py
ckolbPTB: yes but later in the code or if you call DirectReconstruction.csm you don't know if csm is now csmdata or csm_tensor.

src/mrpro/algorithms/reconstruction/_direct_reconstruction.py
ckolbPTB: an image-data-container contains an image and header information. to me this seems fine.

fzimmermann89:
> image_tensor

I agree with _op, 
but not with img vs img_tensor.
I think in the examples it makes it way clearer that this is the tensor... 

we only agreed to use 'img' as the one abbreviation for image, afaik...

src/mrpro/algorithms/reconstruction/_direct_reconstruction.py
fzimmermann89: but to call the variable idata does not really show what it is, that it is an image.. (except if you know about the idata class)

so img in the name would be better
but img_data to me should be the actual image data, not the container...


Here, I would really prefer
```
  (img_data,) = operator.H(kdata.data)
  img = IData.from_tensor_and_kheader(img_data, kdata.header)
```
As easier to understand...

ckolbPTB:
> but not with img vs img_tensor.

but then let's use `img_tensor` instead of `img_data`. then at least there is no confusion between `idata` and `img_data`

fzimmermann89:
moved to #261

Issue #187, opened 2024-02-25
Add register_data or similar to Operators?


fzimmermann89:
For tensors, registering them as buffers makes sure they are moved to GPU if torch.nn.Module.cuda is called.

We currently do not have a way to achieve something similar for Data.

So either we register only the data.data as buffers and ignore the header etc, or we would need
- a similiar function to register data (or overload register_buffer)
-overload operator. _apply to also apply the function to all data.data (see pytorch nn. Module source code)

PR #188, opened 2024-02-25
Data as register_buffer usable buffers

fzimmermann89:
Related #187

### DIFF
--- a/src/mrpro/operators/_SensitivityOp.py
+++ b/src/mrpro/operators/_SensitivityOp.py
@@ -27,7 +27,7 @@ class SensitivityOp(LinearOperator):
     ) -> None:
 
         super().__init__()
-        self.register_buffer('CSM', csm)
+        self.register_buffer('csm', csm)
 
     def forward(self, img_data: torch.Tensor) -> tuple[torch.Tensor,]:
         """Apply the forward operator, thus expand the coils dimension.
@@ -41,7 +41,7 @@ class SensitivityOp(LinearOperator):
         -------
             image data tensor with dimensions (other coils z y x).
         """
-        return (self.CSM.data * img_data,)
+        return (self.csm.data * img_data,)
 
     def adjoint(self, img_data: torch.Tensor) -> tuple[torch.Tensor,]:
         """Apply the adjoint operator, thus reduce the coils dimension.
@@ -56,4 +56,4 @@ class SensitivityOp(LinearOperator):
             image data tensor with dimensions (other 1 z y x).
         """
 
-        return ((self.CSM.data.conj() * img_data).sum(-4, keepdim=True),)
+        return ((self.csm.data.conj() * img_data).sum(-4, keepdim=True),)
--- a/src/mrpro/operators/_SensitivityOp.py
+++ b/src/mrpro/operators/_SensitivityOp.py
@@ -25,7 +25,9 @@ class SensitivityOp(LinearOperator):
         self,
         csm: CsmData,
     ) -> None:
-        self.C = csm.data
+
+        super().__init__()
+        self.register_buffer('CSM', csm)
 
     def forward(self, img_data: torch.Tensor) -> tuple[torch.Tensor,]:
         """Apply the forward operator, thus expand the coils dimension.
@@ -39,7 +41,7 @@ class SensitivityOp(LinearOperator):
         -------
             image data tensor with dimensions (other coils z y x).
         """
-        return (self.C * img_data,)
+        return (self.CSM.data * img_data,)
 
     def adjoint(self, img_data: torch.Tensor) -> tuple[torch.Tensor,]:
         """Apply the adjoint operator, thus reduce the coils dimension.
@@ -54,4 +56,4 @@ class SensitivityOp(LinearOperator):
             image data tensor with dimensions (other 1 z y x).
         """
 
-        return ((self.C.conj() * img_data).sum(-4, keepdim=True),)
+        return ((self.CSM.data.conj() * img_data).sum(-4, keepdim=True),)
--- a/src/mrpro/utils/_databuffermixin.py
+++ b/src/mrpro/utils/_databuffermixin.py
@@ -84,13 +84,11 @@ class DataBufferMixin(torch.nn.Module):
                 return data[name]
         return super().__getattr__(name)
 
-    def __setattr__(self, name: str, value: torch.Tensor | torch.nn.Module | Data) -> None:
+    def __setattr__(self, name: str, value) -> None:
         """Set Attribute."""
-        if isinstance(value, Data):
-            data = self.__dict__.get('_data')
-            if data is not None and name in data:
-                # TODO: hooks ?
-                data[name] = value
+        if isinstance(value, Data) and (data := self.__dict__.get('_data')) is not None and name in data:
+            # TODO: hooks ?
+            data[name] = value
         else:
             return super().__setattr__(name, value)
 
--- a/src/mrpro/utils/_databuffermixin.py
+++ b/src/mrpro/utils/_databuffermixin.py
@@ -1,16 +1,19 @@
 from collections import OrderedDict
+
 import torch
-from mrpro.data import Data
+
+from mrpro.data._Data import Data
 
 
 class DataBufferMixin(torch.nn.Module):
-    """A Mixin that allow to set a mrpro.data.Data as a Buffer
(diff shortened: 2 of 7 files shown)


fzimmermann89:
I used the sensitivity operator as a test, even though it really does not matter there.

The goal is to be able to also "register_buffer" a Data object, similar to a tensor, on Operators (and other custom modules that use the DataBufferMixin.)

For example, a custom model might want to save some knoise in the init, and it should be moved to the GPU if one called custommodel.cpu() (same logic as for other buffer)

Not sure if reusing the register_buffer name is smart, or if there should be a separate register_data (open for comments!)

src/mrpro/operators/_SensitivityOp.py
```diff
@@ -25,7 +25,9 @@ def __init__(
         self,
         csm: CsmData,
     ) -> None:
-        self.C = csm.data
+
+        super().__init__()
+        self.register_buffer('CSM', csm)
```
ckolbPTB: To be somewhat consistent the buffer should be 'csm' and not capitalised. Maybe even '_csm' as we want to discourage users from overwriting it after the _init_

src/mrpro/utils/_databuffermixin.py
```diff
@@ -0,0 +1,104 @@
+from collections import OrderedDict
+
+import torch
+
+from mrpro.data._Data import Data
+
+
+class DataBufferMixin(torch.nn.Module):
```
ckolbPTB: If this is only one class, should the file then be called _DataBufferMixin.py?

src/mrpro/utils/_databuffermixin.py
```diff
@@ -0,0 +1,104 @@
...
+    """A Mixin that allow to set a mrpro.data.Data as a Buffer.
```
ckolbPTB: mrpro.Data.data

src/mrpro/utils/_databuffermixin.py
```diff
@@ -0,0 +1,104 @@
...
+
+    This implements the same logic for mrpro.Data as torch.nn.Module
+    uses for Buffers, but stores them in a separate dictionary called
+    _data
+
+    The main use is to allow Data objects to be automatically moved to a
```
ckolbPTB: ...to allow Data.data to be automatically moved... (we are not moving Data.header at the moment)

src/mrpro/utils/_databuffermixin.py
```diff
@@ -0,0 +1,104 @@
...
+    Used in Operators, for example.
+    """
+
+    _data: dict[str, Data]
+    # could maybe be requiered at some point,
+    # see torch.nn.Module
+    # call_super_init = True
+
+    def register_buffer(self, name: str, data: torch.Tensor | None | Data, persistent: bool = True) -> None:
+        r"""Add a buffer to the module.
```
ckolbPTB: remove r

src/mrpro/utils/_databuffermixin.py
```diff
@@ -0,0 +1,104 @@
...
+        """
+        if not isinstance(data, Data):
+            return super().register_buffer(name, data, persistent)
+
+        if '_data' not in self.__dict__:
+            raise AttributeError('cannot assign buffer before super().__init__() call')
+        elif not isinstance(name, str):
+            raise TypeError(f'buffer name should be a string. Got {torch.typename(name)}')
+        elif '.' in name:
+            raise KeyError("buffer name can't contain \".\"")
```
ckolbPTB: can't -> cannot otherwise formatting of error message is not nice

src/mrpro/utils/_databuffermixin.py
```diff
@@ -0,0 +1,104 @@
...
+        elif name == '':
+            raise KeyError("buffer name can't be empty string \"\"")
```
ckolbPTB: same as above

ckolbPTB (changes requested):

ckolbPTB:
Would be good to add a simple test, e.g. something like:

```
dat1 = Data(data=torch.randn(2,3,4), header='dat1')
dat2 = Data(data=torch.randn(2,3,4), header='dat2')
buf = DataBufferMixin() 

buf.register_buffer(name='dat1', data=dat1)
buf.register_buffer(name='dat2', data=dat2)

buf.cuda()
assert buf.dat1.data.is_cuda
assert buf.dat2.data.is_cuda
```

ckolbPTB:
We would then also have to have a look at all the other Operators and check what needs to be added in the init via this buffer scheme.

src/mrpro/operators/_SensitivityOp.py
fzimmermann89: I would not underscore it, as it is ok to change the cms later on.

src/mrpro/utils/_databuffermixin.py
fzimmermann89: This is the same message pytorch prints. I'd prefer to keep it as is

fzimmermann89:
> We would then also have to have a look at all the other Operators and check what needs to be added in the init via this buffer scheme.

Yes, this is something we should do, independent of this PR. Most Tensors should be registered as buffers if they are not parameters. I did not pay to much attention to this so far..

src/mrpro/utils/_databuffermixin.py
fzimmermann89: same as above indeed ;)

src/mrpro/utils/_databuffermixin.py
fzimmermann89: Isnt the base object called 'Data'  and stored in  mrpro.data?
So we can register` mpro.data.Data` as a Buffer.

(mrpro.data.Data.data would be a tensor, which we could register as a buffer without this code)

Or what am I missing?

src/mrpro/operators/_SensitivityOp.py
ckolbPTB: I was thinking of something like 
```
S = SensitivityOp(csm_cpu)
S.cuda()
S.csm = csm_cpu
```
Will csm then still be a registered_buffer? Will csm then still be on the cuda device?

src/mrpro/utils/_databuffermixin.py
ckolbPTB: You are right, I forgot about one ".data"

src/mrpro/utils/_databuffermixin.py
ckolbPTB: Just because pytorch does it does not mean it is correct ;). Keep it if you want, I don't feel strongly about it, just noticed it when I tested the errors.

ckolbPTB:
> Yes, this is something we should do, independent of this PR.

This would mean that we implement something which is only supported by parts of the code. I don't think this is a good idea...

fzimmermann89:
For torch.nn.modules, Tensors save in self should be either buffers or parameters. 

If we use our mrpro.data.Data in operators and depends on this PR. Otherwise we would save the data aus a buffer and disregard the header.

src/mrpro/operators/_SensitivityOp.py
fzimmermann89: Will csm then still be a registered_buffer? -> Yes. It would replace the value of the existing buffer, but it would still be a buffer.
Will csm then still be on the cuda device? -> Only if you moved it there. The .cuda move moves all buffers, it does not enforce that they are kept there

fzimmermann89:
Still not 100 percent convinced this is the best way to do it.

I had to add is_coplex and is_floatingpoint to Data because pytorch will call it in module.to

The alternative would be to only overwrite
To Cuda and cpu
To work on ALL attributes that are of type Data..

fzimmermann89:
What is the status here? Do we want to have this functionally?

ckolbPTB:
I think we want this functionality but because of

> Still not 100 percent convinced this is the best way to do it.

I thought we need more discussion on how to do it.

fzimmermann89:
As far as I can see there are 4 possible approaches:

- overwrite operator.to/cuda/cpu etc. These are currently implemented by pytorch.

- overwrite nn.module.apply to always move all of your data types.

- never use one of our data classes as an attribute but only raw tensors registered as buffers. I.e. the init we save only the data attribute and the parameters of the header we need. We either add setters to the operators or don't allow changing or viewing the data off an existing operstor. 

- register_buffer the data classes (this PR).

ckolbPTB:
requires better approach to handle moving tensors to gpu for operators

PR #189, opened 2024-02-25
Fix operator chaining

fzimmermann89:
Two spots I missed when we switched to operators returning tuples:
In the @ chain, we need to unpack the return  of the inner operator when calling the outer operator.

Mypy did not catch that as the types on the general operator are not bound to Tensor. 

Edit: A couple more issues....
We should maybe consider running mypy on the tests as well...

(Maybe https://stackoverflow.com/questions/78013162/typehint-function-args-tupleargs-with-constrain-on-the-args could at some point help here)


fzimmermann89:
TODO: Check if mypy likes the tests

schuenke (approved):
The need to use typing.cast() does not necessarily make the code easier to read, but unfortunately it seems to be unavoidable. 

The rest looks fine and I checked already that everything is compatible with the ruff settings from #193

PR #191, opened 2024-02-26
Same ndim for all AcqInfo fields

### LINKED ISSUE #158: Shape of acq_info.idx.user -> switch to separate user_0, user_1, etc?
Almost all `acq_info.idx` have shape `[other, k2, k1]`, only user is of shape `[other, k2, k1,8]`.
Do we maybe instead want split it into 8 different idx and have all acq_info idx have the same shape?
Otherwise, `user` would always either need a special case (or use up the ellipses in a rearrange..)

(I think I just decided at some point while implementing acqinfo to have the user_idx as one tensor and now regret the decision -- what are your opinions?)

ckolbPTB:
closes #158

### DIFF
--- a/src/mrpro/algorithms/__init__.py
+++ b/src/mrpro/algorithms/__init__.py
@@ -1,2 +1,4 @@
 from mrpro.algorithms._prewhiten_kspace import prewhiten_kspace
 from mrpro.algorithms._remove_readout_os import remove_readout_os
+from mrpro.algorithms._lbfgs import lbfgs
+from mrpro.algorithms._adam import adam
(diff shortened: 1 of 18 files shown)


fzimmermann89:
..or idx all one dimension less?
and only the other acqinfo (where we can have multiple values in one field) shape [...,1]/[...,N]
Because now when have many useless [0] in our code for the idx... 

Not sure if I still prefer this... What do you think?

ckolbPTB:
I agree that for the sake of readability and ease of use, we should keep acq-idx without the trailing 1. Might lead to some "if-clauses" later on to distinguish between acq-info and acq-idx but that's probably a small price to pay.

schuenke (approved):
lgtm

removing the trailing dimension for idx was a good suggestion 👍

PR #193, opened 2024-02-28
Switch to Ruff for linting and formatting

### LINKED ISSUE #185: pydocstyle is no longer maintened and has isssues with overload
We are currently using pydocstyle in our precommit.
This is deprecated https://github.com/PyCQA/pydocstyle/pull/658 and, more importantly, has a number of issues with 
`@overload`. https://github.com/PyCQA/pydocstyle/issues/613, https://github.com/PyCQA/pydocstyle/issues/525, https://github.com/PyCQA/pydocstyle/issues/635

For example, I fail to make something like this pass our precommit

```
@overload __add__(self,other:float)-> float: ...

@overload __add__(self,other:int)-> int: ...

def __add__(self,other:int|float)->int|float
''' Add an int or float. '''
```

which I would like to use..

schuenke:
[Ruff](https://docs.astral.sh/ruff/) is _an extremely fast Python linter and code formatter, written in Rust._
It can replace several tools in our current linting workflow like flake8, black, isort, pydocstyle etc. 

# SUMMARY of changes

## linting/formating configuration

- removed the following pre-commi hooks: double-quote-string-fixer, end-of-file-fixer, trailing-whitespace, flake8, docformatter, pydocstyle, pyupgrade, codespell, black
- added typos hook (replacement for codespell)
- added ruff (replacement for all other)
- added separate ruff configuration files (`ruff.toml`) for `examples` and `tests` folder
- upgraded to latest version of all hooks

## general code style / type-hinting changes

- remove all mutable default arguments (see [Ruff rule B006](https://docs.astral.sh/ruff/rules/mutable-argument-default/)): e.g., replace lists with sequences, tensors with single value with float, 1D-tensors with tuples (see also #198) 
- remove all function calls in default arguments (see [Ruff rule B008](https://docs.astral.sh/ruff/rules/function-call-in-default-argument/), [explanatory example ](https://www.valentinog.com/blog/tirl-python-default-arguments/))
- moved pytest fixture that are used in several files into conftest.py (see #211 for further suggestion)
- remove `subprocess.call()` in example scripts

## fixes

- return types of optimizers (adam and lbfgs)
- optimizer tests

## Renaming of  public classes / properties / attributes (no guarantee for completeness):

- `mse_data_discrepancy` (class) -> `MSEDataDiscrepancy` (class)
- `KHeader.num_coils` -> `KHeader.n_coils`
- `KHeader.calib_mode` -> `KHeader.calibration_mode`

## mrpro naming convention

- follow [pep8 naming convention](https://peps.python.org/pep-0008/#naming-conventions) (e.g. all lowercase variable names, CapWords class names)
- use more descriptive variable names when applicable (e.g., `result` instead of `res`, `tolerance_squared` instead of `sqtol`)
- use variable names with starting with `n_` for all variables decribing a number of ... (e.g., `n_coils` instead of `ncoils` or `num_coils`)
- use variable names ending with `_op` for operators (e.g., `fourier_op`)
- use `img` as variable name for images


fzimmermann89:
"B006", # mutable data structures for argument defaults
        "B008", # do not perform calls in argument defaults

I would keep this enabled..
I was looking for some way to get notified if I do this error!

If we want to do this somewhere, we should disable the check there and add a comment why it is ok to do it like this...

also, I do not fully get your remark regarding the test folder..
black etc are currently also running on tests, why add tests to the ignore list?

schuenke:
> ```
> "B006", # mutable data structures for argument defaults
>     "B008", # do not perform calls in argument defaults
> ```
> 
> I would keep this enabled.. I was looking for some way to get notified if I do this error!
> 
> If we want to do this somewhere, we should disable the check there and add a comment why it is ok to do it like this...
> 
> also, I do not fully get your remark regarding the test folder.. black etc are currently also running on tests, why add tests to the ignore list?

I would vote to enable both as well, but didn't want to change to much before knowing that you agree 😄 

C901 is something that we probably cannot avoid easily
B904 doesn't look to crucial to me...

Should we enable ruff in `tests`?

fzimmermann89:
how many errors do we get if you enable it ? :D

src/mrpro/operators/_FourierOp.py
```diff
@@ -59,13 +59,17 @@ def __init__(
...
+                if i in dims
+            ]
 
         def get_traj(traj, dims):
-            return [k for k, i in zip((traj.kz, traj.ky, traj.kx), (-3, -2, -1)) if i in dims]
+            return [k for k, i in zip((traj.kz, traj.ky, traj.kx), (-3, -2, -1), strict=False) if i in dims]
 
         self._ignore_dims, self._fft_dims, self._nufft_dims = [], [], []
-        for dim, type in zip((-3, -2, -1), traj.type_along_kzyx):
+        for dim, type in zip((-3, -2, -1), traj.type_along_kzyx, strict=False):
```
fzimmermann89: why strict=False?

either strict=True or disabling the check....

schuenke:
> how many errors do we get if you enable it ? :D

Many 😄 
This is why I disabled it for the first commit...

src/mrpro/operators/_FourierOp.py
```diff
@@ -97,11 +101,14 @@ def get_traj(traj, dims):
 
             self._nufft_im_size = get_spatial_dims(recon_shape, self._nufft_dims)
             grid_size = [
-                int(s * os) for s, os in zip(self._nufft_im_size, get_spatial_dims(oversampling, self._nufft_dims))
+                int(s * os)
+                for s, os in zip(self._nufft_im_size, get_spatial_dims(oversampling, self._nufft_dims), strict=False)
```
fzimmermann89: this should be strict=True imho

src/mrpro/utils/_zero_pad_or_crop.py
```diff
@@ -75,7 +75,7 @@ def zero_pad_or_crop(
         new_shape = tuple(new_shape[dim.index(i)] if i in dim else s for i, s in enumerate(data.shape))
 
     npad = []
-    for old, new in zip(data.shape, new_shape):
+    for old, new in zip(data.shape, new_shape, strict=False):
```
fzimmermann89: thi sshould also be strict=true imho

src/mrpro/operators/_FourierOp.py
schuenke: Yeah, saw this just after I pushed the commit. It was an auto adjustment by ruff. Will change to true.

src/mrpro/data/_SpatialDimension.py
```diff
@@ -72,7 +72,7 @@ def from_array_xyz(
         conversion, optional
             will be called for each value to convert it, by default None
         """
-        if not isinstance(data, (np.ndarray, torch.Tensor)):
+        if not isinstance(data, np.ndarray | torch.Tensor):
```
fzimmermann89, line 75: whaaaat? this is allowed. nice. didnt know this was valid. TIL.

src/mrpro/utils/_zero_pad_or_crop.py
schuenke: changed in [c0e2d68](https://github.com/PTB-MR/mrpro/pull/193/commits/c0e2d68acdd11289bc3819520efef2bef172fa2c)

src/mrpro/operators/_FourierOp.py
schuenke: changed in [c0e2d68](https://github.com/PTB-MR/mrpro/pull/193/commits/c0e2d68acdd11289bc3819520efef2bef172fa2c)

src/mrpro/operators/_FourierOp.py
schuenke: changed in [c0e2d68](https://github.com/PTB-MR/mrpro/pull/193/commits/c0e2d68acdd11289bc3819520efef2bef172fa2c)

tests/data/_Dicom2DTestImage.py
```diff
@@ -44,8 +44,11 @@ def __init__(
         matrix_size_y: int = 128,
         matrix_size_x: int = 256,
         te: float = 3.7,
-        phantom: EllipsePhantom = EllipsePhantom(),
+        phantom: EllipsePhantom | None = None,
```
fzimmermann89, line 46: can we use ellipsephantom as a fixture here?

schuenke:
if I enable `warn_return_any` for mypy, I currently still get the following errors:
```
src\mrpro\operators\_Operator.py:61: error: Returning Any from function declared to return "Tout"  [no-any-return]
src\mrpro\data\_CsmData.py:65: error: Returning Any from function declared to return "Tensor"  [no-any-return]
src\mrpro\phantoms\coils.py:66: error: Returning Any from function declared to return "Tensor"  [no-any-return]
src\mrpro\phantoms\_EllipsePhantom.py:101: error: Returning Any from function declared to return "Tensor"  [no-any-return]
src\mrpro\operators\models\_WASABITI.py:100: error: Returning Any from function declared to return "tuple[Tensor]"  [no-any-return]
src\mrpro\operators\models\_WASABI.py:98: error: Returning Any from function declared to return "tuple[Tensor]"  [no-any-return]
src\mrpro\operators\_LinearOperator.py:69: error: Returning Any from function declared to return "tuple[Tensor]"  [no-any-return]
```

I will have a look at this again.


@fzimmermann89 some of the rules you enabled (to see the world burn 😜) are conflicting if we enable them in general (e.g. D203 & D211, D212 & D213) or with our own naming convention  (e.g. N802, N999, ...) and some do not make sense as we discussed already (e.g. ANN101, ANN102, ANN204). Should we just enable individual rules from D, N, ANN? Which ones would that be?

pyproject.toml
```diff
@@ -89,9 +89,38 @@ include = ["src/**/*.py"]
 exclude = ["docs/**"]
 
 [tool.ruff.lint]
-extend-select = ["B", "C", "SIM", "UP", "Q", "COM", "A", "S", "I", "D"]
+select = [
```
fzimmermann89, line 91: This is a great idea!

src/mrpro/algorithms/_remove_readout_os.py
```diff
@@ -51,14 +51,14 @@ def remove_readout_os(kdata: KData) -> KData:
...
         end_cropped_readout = start_cropped_readout + kdata.header.recon_matrix.x
 
-        def crop_readout(input_):
+        def crop_readout(input_: torch.Tensor):
             return input_[..., start_cropped_readout:end_cropped_readout]
 
         # Transform to image space, crop to reconstruction matrix size and transform back
-        FFOp = FastFourierOp(dim=(-1,))
-        (dat,) = FFOp.adjoint(kdata.data)
+        ff_op = FastFourierOp(dim=(-1,))
```
fzimmermann89: maybe we can allow upper-case variable names inside functions?
An operator looks nicer if it is a capital letter.. (personal preference)

src/mrpro/data/_AcqInfo.py
```diff
@@ -118,15 +118,15 @@ def tensor(data):
             # Remove any uncessary dimensions
             return torch.tensor(np.squeeze(data))
 
-        def tensor_2d(data):
+        def tensor_2d(input_: np.ndarray):
```
fzimmermann89: rename to data or x?

src/mrpro/data/_TrajectoryDescription.py
```diff
@@ -25,18 +25,18 @@ class TrajectoryDescription:
     """TrajectoryDescription dataclass."""
 
     identifier: str = ''
-    userParameterLong: dict[str, int] = dataclasses.field(default_factory=dict)
-    userParameterDouble: dict[str, float] = dataclasses.field(default_factory=dict)
-    userParameterString: dict[str, str] = dataclasses.field(default_factory=dict)
+    user_parameter_long: dict[str, int] = dataclasses.field(default_factory=dict)
```
fzimmermann89, line 28: userParameterLong is the name used by ismrmrd...
should we consider keeping it and marking this with a` #noqa`  comment to disable this rele here?

fzimmermann89: or do we always follow pep8?
not sure which one I prefer.

src/mrpro/operators/_ConstraintsOp.py
```diff
@@ -14,7 +14,7 @@
 
 
 import torch
-import torch.nn.functional as F
+import torch.nn.functional as F  # noqa: N812
```
fzimmermann89, line 18: 👍

src/mrpro/operators/_FourierOp.py
```diff
@@ -59,14 +59,14 @@ def __init__(
...
-        def get_spatial_dims(spatial_dims, dims):
+        def get_spatial_dims(spatial_dims: SpatialDimension, dims: list[int]):
             return [
                 s
                 for s, i in zip((spatial_dims.z, spatial_dims.y, spatial_dims.x), (-3, -2, -1), strict=True)
                 if i in dims
             ]
 
-        def get_traj(traj, dims):
+        def get_traj(traj: KTrajectory, dims: list[int]):
```
fzimmermann89: Sequence[int], #198

src/mrpro/operators/functionals/_mse_data_discrepancy.py
```diff
@@ -13,12 +13,12 @@
...
 
 import torch
-import torch.nn.functional as F
+import torch.nn.functional as F  # noqa: N812
 
 from mrpro.operators import Operator
 
 
-class mse_data_discrepancy(Operator[*tuple[torch.Tensor], tuple[torch.Tensor]]):
+class MseDataDiscrepancy(Operator[*tuple[torch.Tensor], tuple[torch.Tensor]]):
```
fzimmermann89: MSEDataDiscrepancy vs MseDataDiscrepancy ?
(no preference)

src/mrpro/utils/_rgetattr.py
```diff
@@ -15,6 +15,6 @@
 import functools
 
 
-def rgetattr(obj, attr, *args):
+def rgetattr(obj, attr, *args):  # noqa: ANN001
     """Recursive getattr for nested attributes."""
-    return functools.reduce(lambda obj, attr: getattr(obj, attr, *args), [obj] + attr.split('.'))
+    return functools.reduce(lambda obj, attr: getattr(obj, attr, *args), [obj] + attr.split('.'))  # noqa: RUF005
```
fzimmermann89: general question:
We we prefe`r [1, *foo, 5, 6] `over` [1] + foo + [5, 6]` ?
Why is this use here different than the others?
Should we disable the rule in general or change it here as well?

fzimmermann89 (approved):
wow, much work, gut job, few questions.

fzimmermann89:
```
==================================== ERRORS ====================================
__________________ ERROR collecting tests/data/test_ktraj.py ___________________
In test_ktype_along_kzyx: function uses no argument 'im_shape'
__________________ ERROR collecting tests/data/test_ktraj.py ___________________
In test_ktype_along_kzyx: function uses no argument 'im_shape'
__________________ ERROR collecting tests/data/test_ktraj.py ___________________
In test_ktype_along_kzyx: function uses no argument 'im_shape'
__________________ ERROR collecting tests/data/test_ktraj.py ___________________
In test_ktype_along_kzyx: function uses no argument 'im_shape'
____________ ERROR collecting tests/operators/models/test_wasabi.py ____________
tests/operators/models/test_wasabi.py::test_WASABI_signal_model_shape: in "parametrize" the number of names (12):
  ('offset_max', 'offset_nr', 'b0_shift', 'rb1', 'c', 'd', 'p', 'other', 'coils', 'z', 'y', 'x')
must be equal to the number of values (10):
  (250, 101, 0, 1.0, 1.0, 2.0, 4, 1, 1, 1)
____________ ERROR collecting tests/operators/models/test_wasabi.py ____________
tests/operators/models/test_wasabi.py::test_WASABI_signal_model_shape: in "parametrize" the number of names (12):
  ('offset_max', 'offset_nr', 'b0_shift', 'rb1', 'c', 'd', 'p', 'other', 'coils', 'z', 'y', 'x')
must be equal to the number of values (10):
  (250, 101, 0, 1.0, 1.0, 2.0, 4, 1, 1, 1)
____________ ERROR collecting tests/operators/models/test_wasabi.py ____________
tests/operators/models/test_wasabi.py::test_WASABI_signal_model_shape: in "parametrize" the number of names (12):
  ('offset_max', 'offset_nr', 'b0_shift', 'rb1', 'c', 'd', 'p', 'other', 'coils', 'z', 'y', 'x')
must be equal to the number of values (10):
  (250, 101, 0, 1.0, 1.0, 2.0, 4, 1, 1, 1)
____________ ERROR collecting tests/operators/models/test_wasabi.py ____________
tests/operators/models/test_wasabi.py::test_WASABI_signal_model_shape: in "parametrize" the number of names (12):
  ('offset_max', 'offset_nr', 'b0_shift', 'rb1', 'c', 'd', 'p', 'other', 'coils', 'z', 'y', 'x')
must be equal to the number of values (10):
  (250, 101, 0, 1.0, 1.0, 2.0, 4, 1, 1, 1)
___________ ERROR collecting tests/operators/models/test_wasabiti.py ___________
tests/operators/models/test_wasabiti.py::test_WASABITI_signal_model_shape: in "parametrize" the number of names (11):
  ('offset_max', 'offset_nr', 'b0_shift', 'rb1', 't1', 'p', 'other', 'coils', 'z', 'y', 'x')
must be equal to the number of values (9):
  (250, 101, 0, 1.0, 1.0, 4, 1, 1, 1)
___________ ERROR collecting tests/operators/models/test_wasabiti.py ___________
tests/operators/models/test_wasabiti.py::test_WASABITI_signal_model_shape: in "parametrize" the number of names (11):
  ('offset_max', 'offset_nr', 'b0_shift', 'rb1', 't1', 'p', 'other', 'coils', 'z', 'y', 'x')
must be equal to the number of values (9):
  (250, 101, 0, 1.0, 1.0, 4, 1, 1, 1)
___________ ERROR collecting tests/operators/models/test_wasabiti.py ___________
tests/operators/models/test_wasabiti.py::test_WASABITI_signal_model_shape: in "parametrize" the number of names (11):
  ('offset_max', 'offset_nr', 'b0_shift', 'rb1', 't1', 'p', 'other', 'coils', 'z', 'y', 'x')
must be equal to the number of values (9):
  (250, 101, 0, 1.0, 1.0, 4, 1, 1, 1)
___________ ERROR collecting tests/operators/models/test_wasabiti.py ___________
tests/operators/models/test_wasabiti.py::test_WASABITI_signal_model_shape: in "parametrize" the number of names (11):
  ('offset_max', 'offset_nr', 'b0_shift', 'rb1', 't1', 'p', 'other', 'coils', 'z', 'y', 'x')
must be equal to the number of values (9):
  (250, 101, 0, 1.0, 1.0, 4, 1, 1, 1)
_____________ ERROR collecting tests/operators/test_fourier_op.py ______________
In test_fourier_fwd_adj_property: function uses no argument 's0'
_____________ ERROR collecting tests/operators/test_fourier_op.py ______________
In test_fourier_fwd_adj_property: function uses no argument 's0'
_____________ ERROR collecting tests/operators/test_fourier_op.py ______________
In test_fourier_fwd_adj_property: function uses no argument 's0'
_____________ ERROR collecting tests/operators/test_fourier_op.py ______________
In test_fourier_fwd_adj_property: function uses no argument 's0'
--------- generated xml file: /home/runner/work/mrpro/mrpro/pytest.xml ---------

```

src/mrpro/utils/_rgetattr.py
schuenke: Didn't have the time to write a comment, when I pushed the commit, but I didn't change anything for _rgetattr because we don't use it anyway, do we? I wanted to discuss if we wanna remove it....

schuenke:
> ```
> ==================================== ERRORS ====================================
> __________________ ERROR collecting tests/data/test_ktraj.py ___________________
> In test_ktype_along_kzyx: function uses no argument 'im_shape'
> __________________ ERROR collecting tests/data/test_ktraj.py ___________________
> In test_ktype_along_kzyx: function uses no argument 'im_shape'
> __________________ ERROR collecting tests/data/test_ktraj.py ___________________
> In test_ktype_along_kzyx: function uses no argument 'im_shape'
> __________________ ERROR collecting tests/data/test_ktraj.py ___________________
> In test_ktype_along_kzyx: function uses no argument 'im_shape'
> ____________ ERROR collecting tests/operators/models/test_wasabi.py ____________
> tests/operators/models/test_wasabi.py::test_WASABI_signal_model_shape: in "parametrize" the number of names (12):
>   ('offset_max', 'offset_nr', 'b0_shift', 'rb1', 'c', 'd', 'p', 'other', 'coils', 'z', 'y', 'x')
> must be equal to the number of values (10):
>   (250, 101, 0, 1.0, 1.0, 2.0, 4, 1, 1, 1)
> ____________ ERROR collecting tests/operators/models/test_wasabi.py ____________
> tests/operators/models/test_wasabi.py::test_WASABI_signal_model_shape: in "parametrize" the number of names (12):
>   ('offset_max', 'offset_nr', 'b0_shift', 'rb1', 'c', 'd', 'p', 'other', 'coils', 'z', 'y', 'x')
> must be equal to the number of values (10):
>   (250, 101, 0, 1.0, 1.0, 2.0, 4, 1, 1, 1)
> ____________ ERROR collecting tests/operators/models/test_wasabi.py ____________
> tests/operators/models/test_wasabi.py::test_WASABI_signal_model_shape: in "parametrize" the number of names (12):
>   ('offset_max', 'offset_nr', 'b0_shift', 'rb1', 'c', 'd', 'p', 'other', 'coils', 'z', 'y', 'x')
> must be equal to the number of values (10):
>   (250, 101, 0, 1.0, 1.0, 2.0, 4, 1, 1, 1)
> ____________ ERROR collecting tests/operators/models/test_wasabi.py ____________
> tests/operators/models/test_wasabi.py::test_WASABI_signal_model_shape: in "parametrize" the number of names (12):
>   ('offset_max', 'offset_nr', 'b0_shift', 'rb1', 'c', 'd', 'p', 'other', 'coils', 'z', 'y', 'x')
> must be equal to the number of values (10):
>   (250, 101, 0, 1.0, 1.0, 2.0, 4, 1, 1, 1)
> ___________ ERROR collecting tests/operators/models/test_wasabiti.py ___________
> tests/operators/models/test_wasabiti.py::test_WASABITI_signal_model_shape: in "parametrize" the number of names (11):
>   ('offset_max', 'offset_nr', 'b0_shift', 'rb1', 't1', 'p', 'other', 'coils', 'z', 'y', 'x')
> must be equal to the number of values (9):
>   (250, 101, 0, 1.0, 1.0, 4, 1, 1, 1)
> ___________ ERROR collecting tests/operators/models/test_wasabiti.py ___________
> tests/operators/models/test_wasabiti.py::test_WASABITI_signal_model_shape: in "parametrize" the number of names (11):
>   ('offset_max', 'offset_nr', 'b0_shift', 'rb1', 't1', 'p', 'other', 'coils', 'z', 'y', 'x')
> must be equal to the number of values (9):
>   (250, 101, 0, 1.0, 1.0, 4, 1, 1, 1)
> ___________ ERROR collecting tests/operators/models/test_wasabiti.py ___________
> tests/operators/models/test_wasabiti.py::test_WASABITI_signal_model_shape: in "parametrize" the number of names (11):
>   ('offset_max', 'offset_nr', 'b0_shift', 'rb1', 't1', 'p', 'other', 'coils', 'z', 'y', 'x')
> must be equal to the number of values (9):
>   (250, 101, 0, 1.0, 1.0, 4, 1, 1, 1)
> ___________ ERROR collecting tests/operators/models/test_wasabiti.py ___________
> tests/operators/models/test_wasabiti.py::test_WASABITI_signal_model_shape: in "parametrize" the number of names (11):
>   ('offset_max', 'offset_nr', 'b0_shift', 'rb1', 't1', 'p', 'other', 'coils', 'z', 'y', 'x')
> must be equal to the number of values (9):
>   (250, 101, 0, 1.0, 1.0, 4, 1, 1, 1)
> _____________ ERROR collecting tests/operators/test_fourier_op.py ______________
> In test_fourier_fwd_adj_property: function uses no argument 's0'
> _____________ ERROR collecting tests/operators/test_fourier_op.py ______________
> In test_fourier_fwd_adj_property: function uses no argument 's0'
> _____________ ERROR collecting tests/operators/test_fourier_op.py ______________
> In test_fourier_fwd_adj_property: function uses no argument 's0'
> _____________ ERROR collecting tests/operators/test_fourier_op.py ______________
> In test_fourier_fwd_adj_property: function uses no argument 's0'
> --------- generated xml file: /home/runner/work/mrpro/mrpro/pytest.xml ---------
> ```

I removed multiple unused arguments in several functions/tests, but looks like I forgot some changes. Will have a look at it tomorrow.

src/mrpro/operators/functionals/_mse_data_discrepancy.py
schuenke: same for me... @ckolbPTB: what's your favorite? 😄

tests/data/_Dicom2DTestImage.py
schuenke: Currently the tests and test data are a bit messy. We already have a `ph_ellipse` fixture, which is just a `EllipsePhantomTestData`, which is just an `EllipsePhantom` (this is part of src/mrpro/phantoms) with different default ellipses...

do we want to be able to use `_Dicom2DTestImage` with other EllipsePhantoms? @fzimmermann89 @ckolbPTB 
If not, using the `ph_ellipse` fixture would make more sense indeed.

src/mrpro/data/_TrajectoryDescription.py
schuenke: IMO we should always follow pep8 except for a few exceptions such as LinearOperator.H, which we mark with `# noqa: N812`

src/mrpro/operators/functionals/_mse_data_discrepancy.py
ckolbPTB: I find **MseDataDiscrepancy** easier to read and prefer that.

src/mrpro/data/_TrajectoryDescription.py
ckolbPTB: I would go for user_parameter_long. I think it is easy enough to figure out that this corresponds with userParameterLong in ISMRMRD.

tests/data/_Dicom2DTestImage.py
ckolbPTB: I think fixture should be fine.

src/mrpro/operators/_FourierOp.py
schuenke: I will leave it as it is for the moment to avoid inconsistencies in our code... should be handled in a separate PR adressing #198

src/mrpro/utils/_rgetattr.py
schuenke: I will change it to be consistent with the rest of our code. For an explanation why we should prefer the unpacking op for concatination, check the Ruff doc for rule [RUF005](https://docs.astral.sh/ruff/rules/collection-literal-concatenation/).

Still, if we don't use the function anymore, we should probably remove it...

src/mrpro/algorithms/_remove_readout_os.py
schuenke: I think we now agreed to always follow pep8 for variable names and only allow exceptions in special methods and/or property names like `LinearOperator.H`

fzimmermann89: did we? I thought  we cut this discussion due to time constraints. ... ... 🙈

tests/operators/functionals/test_mse_functional.py
```diff
@@ -41,6 +39,6 @@ def test_mse_functional(data, x, expected_mse):
     1/N*|| . - data||_2^2
     """
 
-    mse_dc = mse_data_discrepancy(torch.tensor(data))
+    mse_dc = MseDataDiscrepancy(torch.tensor(data))
```
fzimmermann89: mse_op

fzimmermann89:
Why is https://github.com/PTB-MR/mrpro/blob/b31ae8eaa4b3a61e4c78844b62f2b390966b6f81/tests/operators/test_fast_fourier_op.py#L66 allowed? Or am I missing the changes somehow?

fzimmermann89:
Can we maybe include the list and tuple -> Sequence typehint in this PR?
This is also something that 
a) should not break anything
b) is a kind of style choice
c) is in some random places all around the code, making it annoying to merge (similar to this PR)

...would you mind if I add it to this PR.... ?

src/mrpro/algorithms/_lbfgs.py
```diff
@@ -55,19 +54,18 @@ def lbfgs(
     history_size, optional
         update history size, by default 10
     line_search_fn, optional
-        line search algorithm, either ‘strong_wolfe’ or None,
+        line search algorithm, either 'strong_wolfe' or None,
```
fzimmermann89: can we do this with a proper type hint ?

tests/_RandomGenerator.py
```diff
@@ -144,13 +144,13 @@ def int64(self, low: int = -1 << 63, high: int = 1 << 63 - 1) -> int:
     def uint64(self, low: int = 0, high: int = 1 << 64) -> int:
         # pytorch does not support uint64, so we use int64 instead
         # and then convert to uint64
-        range = high - low
+        range_ = high - low
```
fzimmermann89, line 149: span?

fzimmermann89: value_range?

schuenke:
> Why is
> 
> https://github.com/PTB-MR/mrpro/blob/b31ae8eaa4b3a61e4c78844b62f2b390966b6f81/tests/operators/test_fast_fourier_op.py#L66
> allowed? Or am I missing the changes somehow?

It is "allowed" because we currently ignore the following rules in our tests:

```
lint.extend-ignore = [
    "ARG001", # unused argument
    "ANN",    # flake8-annotations
    "D",      # pydocstyle
    "N",      # pep8-naming
    "S101",   # use of assert detected
    "PT011",  # pytest-raises-too-broad
]
```

Everything except S101 could be enabled if we want, but would result in a lot of manual refactoring... didn't do it yet because I didn't want to waste time on stuff that is maybe not desired.

schuenke:
> Can we maybe include the list and tuple -> Sequence typehint in this PR? This is also something that a) should not break anything b) is a kind of style choice c) is in some random places all around the code, making it annoying to merge (similar to this PR)
> 
> ...would you mind if I add it to this PR.... ?

It's fine for me. We touch almost all files anyway 😁

src/mrpro/algorithms/_remove_readout_os.py
ckolbPTB: I remember to either use "ff_op" or "FF" but not "FFOp". I think I am in favour or variable names which provide information and would actually go for "fast_fourier_op".

src/mrpro/operators/_FourierOp.py
```diff
@@ -32,9 +32,7 @@ def __init__(
         recon_shape: SpatialDimension[int],
         encoding_shape: SpatialDimension[int],
         traj: KTrajectory,
-        oversampling: SpatialDimension[float] = SpatialDimension(
-            2.0, 2.0, 2.0
-        ),  # TODO: maybe not the nicest, since only used for nuFFT
+        oversampling: float | SpatialDimension[float] = 2.0,  # only used for nuFFT
```
ckolbPTB: Could we also have this just as a float. I cannot think of any application where we would need different oversampling factor along different directions...

src/mrpro/operators/_SensitivityOp.py
```diff
@@ -21,10 +21,7 @@
 class SensitivityOp(LinearOperator):
     """Sensitivity operator class."""
 
-    def __init__(
-        self,
-        csm: CsmData,
-    ) -> None:
+    def __init__(self, csm: CsmData) -> None:
         self.C = csm.data
```
ckolbPTB: should this then be `self.csm`?

src/mrpro/operators/functionals/_mse_data_discrepancy.py
fzimmermann89: Just for reference: PEP8 actually would suggest MSEDataDiscrepancy

> When using acronyms in CapWords, capitalize all the letters of the acronym. Thus HTTPServerError is better than HttpServerError

src/mrpro/algorithms/_remove_readout_os.py
fzimmermann89: ok, überstimmt.

src/mrpro/utils/_rgetattr.py
fzimmermann89: TIL.

tests/algorithms/test_optimizers.py
```diff
@@ -50,7 +38,6 @@ def test_optimizers_rosenbrock(optimizer, enforce_bounds_on_x1):
 
         # save to compare with later as optimization should not change the initial points
         params_init_before = [i.detach().clone for i in params_init]
-        params_init_grad_before = [i.grad.clone() for i in params_init]
```
fzimmermann89, line 53: why is this removed?

tests/algorithms/test_optimizers.py
```diff
@@ -67,16 +54,11 @@ def test_optimizers_rosenbrock(optimizer, enforce_bounds_on_x1):
...
+        params_result = optimizer(functional, params_init, max_iter=max_iter, lr=lr)
 
         # obtained solution should match analytical
         torch.testing.assert_close(torch.tensor(params_result), analytical_solution)
 
-        for p, before, grad_before in zip(params_init, params_init_before, params_init_grad_before):
+        for p, before in zip(params_init, params_init_before, strict=True):
             assert p == before, 'the initial parameter should not have changed during optimization'
-            assert p.grad == before, 'the inital paramters gradient should not have changed during optimization'
+            assert p.grad == before, 'the initial parameters gradient should not have changed during optimization'
```
fzimmermann89: it should test vs grad_before

tests/algorithms/test_optimizers.py
fzimmermann89: remark: maybe would be even better to also randomly set  i.grad before to some values that should not be touched by the optimizer.

schuenke: > why is this removed?

good question... probably happened in one of the semi-manual corrections of the initial ruff formats adding strict=False everywhere...

src/mrpro/operators/functionals/_mse_data_discrepancy.py
schuenke: than MSEDataDiscrepancy it is 👍

schuenke:
Thanks for the latest fixes @fzimmermann89, but I think we have to stop adding more and more changes to this PR so that maybe @ckolbPTB or @koflera have a chance to review the current changes related to our linting/formatting workflow and #199.

I also realized that we still have several tests that check for more than one property, e.g. correct shape and adjointness test in a single test function, but IMO this should be handled in a separate PR.

Would be nice, if we could get this one merged as soon as possible, because it will effect all other open PRs as well.

tests/algorithms/test_optimizers.py
fzimmermann89: I fixed it and made sure the test actually runs and doesnt always pass no matter what the optimizer does.

.pre-commit-config.yaml
```diff
@@ -2,7 +2,7 @@ default_language_version:
   python: python3
 
 repos:
-  - repo: 'https://github.com/pre-commit/pre-commit-hooks'
+  - repo: "https://github.com/pre-commit/pre-commit-hooks"
```
ckolbPTB: Sometimes we use "" sometimes we don't for the repo names. Any reason for this?

src/mrpro/operators/_LinearOperator.py
```diff
@@ -26,7 +26,7 @@
 
 
 # LinearOperators have exactly one input and one output,
-# and are fullfill f(a*x + b*y) = a*f(x) + b*f(y)
+# and are fulfill f(a*x + b*y) = a*f(x) + b*f(y)
```
ckolbPTB: `and fulfill` without the `are`

src/mrpro/algorithms/_adam.py
```diff
@@ -20,7 +22,7 @@
 
 def adam(
     f: Operator[*tuple[torch.Tensor, ...], tuple[torch.Tensor]],
```
ckolbPTB, line 24: I am not sure I fully understand the current way of tuple, list and Sequence as an input. In this example we say that f should expect a tuple but allow Sequence (i.e tuple, list and range) for initial_parameters. We also return the output parameters as list. Seems to me like we are switching all the time. Why do we not stick with tuple for the parameters? This would then also match the fact that we are not modifying them but return a copy.

src/mrpro/data/_CsmData.py
```diff
@@ -27,7 +27,9 @@ class CsmData(QData):
 
     @staticmethod
     def _iterative_walsh_csm(
-        coil_images: torch.Tensor, smoothing_width: SpatialDimension[int], niter: int
+        coil_images: torch.Tensor,
+        smoothing_width: SpatialDimension[int],
```
fzimmermann89, line 31: maybe mention in docstring that int means same smoothing along all non-singleton k dimensions?

fzimmermann89: there should be a test calling it with int and one calling it with spatialdimension

src/mrpro/data/traj_calculators/_KTrajectoryPulseq.py
```diff
@@ -77,7 +73,7 @@ def __call__(self, kheader: KHeader) -> KTrajectoryRawShape:
 
         sample_size = num_samples.shape[0]
 
-        def reshape_pulseq_traj(k_traj, encoding_matrix):
+        def reshape_pulseq_traj(k_traj: torch.Tensor, encoding_matrix: int):
```
ckolbPTB: `encoding_matrix` is a bit misleading here because everywhere else encoding_matrix refers to three dimensions but here we are talking of a single dimension. Maybe `encoding_size` or `encoding_length` or `kdim` or `k_extend`

src/mrpro/data/_KData.py
```diff
@@ -149,18 +139,25 @@ def from_file(
         num_total_unique = torch.as_tensor([len(unique_idxs[label]) for label in KDIM_SORT_LABELS]).prod()
 
         # Define function to find index label combinations. This is used to determine the dimensions of k1 and k2.
```
fzimmermann89: can we move this into a docstring of the idx_label_combination?

tests/data/_PulseqRadialTestSeq.py
```diff
@@ -28,25 +27,25 @@ def __init__(self, seq_filename: str, Nx=256, Nspokes=10):
...
             number of frequency encoding points
-        Nspokes, optional
+        n_spokes, optional
             number of spokes
         """
         seq = pypulseq.Sequence()
         fov = 200e-3
 
-        delta = torch.pi / Nspokes
+        delta = torch.pi / n_spokes
```
ckolbPTB: delta_angle

tests/data/_PulseqRadialTestSeq.py
```diff
@@ -28,25 +27,25 @@ def __init__(self, seq_filename: str, Nx=256, Nspokes=10):
...
         deltak = 1 / fov
```
ckolbPTB: delta_k

tests/data/_PulseqRadialTestSeq.py
```diff
@@ -28,25 +27,25 @@ def __init__(self, seq_filename: str, Nx=256, Nspokes=10):
...
 
         system = pypulseq.Opts()
         rf, gz, _ = pypulseq.make_sinc_pulse(flip_angle=0.1, slice_thickness=1e-3, system=system, return_gz=True)
-        gx = pypulseq.make_trapezoid(channel='x', flat_area=Nx * deltak, flat_time=2e-3, system=system)
-        adc = pypulseq.make_adc(num_samples=Nx, duration=gx.flat_time, delay=gx.rise_time, system=system)
+        gx = pypulseq.make_trapezoid(channel='x', flat_area=n_x * deltak, flat_time=2e-3, system=system)
```
ckolbPTB: delta_k

tests/data/_PulseqRadialTestSeq.py
```diff
@@ -28,25 +27,25 @@ def __init__(self, seq_filename: str, Nx=256, Nspokes=10):
...
+        adc = pypulseq.make_adc(num_samples=n_x, duration=gx.flat_time, delay=gx.rise_time, system=system)
         gx_pre = pypulseq.make_trapezoid(channel='x', area=-gx.area / 2 - deltak / 2, duration=2e-3, system=system)
         gz_reph = pypulseq.make_trapezoid(channel='z', area=-gz.area / 2, duration=2e-3, system=system)
 
-        for spoke in range(Nspokes):
+        for spoke in range(n_spokes):
             phi = delta * spoke
```
ckolbPTB: delta_angle

tests/data/_PulseqRadialTestSeq.py
```diff
@@ -57,14 +56,14 @@ def __init__(self, seq_filename: str, Nx=256, Nspokes=10):
...
+        self.n_spokes = n_spokes
         self.seq = seq
         self.seq_filename = seq_filename
 
-        kz = torch.zeros(1, 1, Nspokes, Nx)
-        phi = torch.pi / Nspokes * torch.arange(Nspokes)
-        k0 = deltak * torch.linspace(-Nx / 2, Nx / 2 - 1, Nx)
+        kz = torch.zeros(1, 1, n_spokes, n_x)
+        phi = torch.pi / n_spokes * torch.arange(n_spokes)
+        k0 = deltak * torch.linspace(-n_x / 2, n_x / 2 - 1, n_x)
```
ckolbPTB: delta_k

tests/data/_PulseqRadialTestSeq.py
ckolbPTB: We could also replace phi with angle

tests/data/test_kheader.py
ckolbPTB, line 1: ln 30: num_coils -> n_coils

src/mrpro/data/_KData.py
```diff
@@ -149,18 +139,25 @@ def from_file(
...
-        def idx_label_combination(average_idx, slice_idx, contrast_idx, phase_idx, repetition_idx, set_idx):
+        def idx_label_combination(
+            average_idx: list[int],
```
fzimmermann89: these typhintes are wrong!
why doenst mypy complain in line 160?
this function is called there with ints as paramters.

tests/data/test_traj_calculators.py
```diff
@@ -15,219 +15,228 @@
...
+    assert trajectory.kx.shape == valid_shape[2]
 
 
 def test_KTrajectoryRpe_uniform(valid_rpe_kheader):
     """Calculate RPE trajectory with uniform angle."""
-    num_rpe_lines = valid_rpe_kheader.acq_info.idx.k1.shape[1]
-    ktrajectory1 = KTrajectoryRpe(angle=torch.pi / num_rpe_lines, shift_between_rpe_lines=torch.tensor([0]))
-    ktraj1 = ktrajectory1(valid_rpe_kheader)
+    n_rpe_lines = valid_rpe_kheader.acq_info.idx.k1.shape[1]
+    trajectory_calculator_1 = KTrajectoryRpe(angle=torch.pi / n_rpe_lines, shift_between_rpe_lines=torch.tensor([0]))
```
ckolbPTB: either `trajectory_calculator1` or `trajectory_1`

tests/helper.py
```diff
@@ -15,7 +15,7 @@
 import torch
 
 
-def rel_image_diff(im1, im2):
+def relative_image_difference(im1, im2):
```
ckolbPTB: img1, img2

tests/operators/functionals/test_mse_functional.py
```diff
@@ -12,15 +12,13 @@
...
 import pytest
 import torch
-
-from mrpro.operators.functionals._mse_data_discrepancy import mse_data_discrepancy
+from mrpro.operators.functionals._mse_data_discrepancy import MSEDataDiscrepancy
 
 
 @pytest.mark.parametrize(
-    'data, x, expected_mse',
+    ('data', 'x', 'expected_mse'),
```
ckolbPTB, line 21: If we use `MSEDataDiscrepancy` should we then not also use `expected_MSE` (which in my opinion looks ugly...)

tests/operators/test_cartesian_sampling_op.py
```diff
@@ -31,35 +32,40 @@ def test_cart_sampling_op_data_match():
...
+    trajectory_sub = KTrajectory(
+        kz=trajectory.kz[:, ::2, :, :],
+        ky=trajectory.ky[:, :, ::4, :],
+        kx=trajectory.kx[:, :, :, ::3],
+    )
+    sampling_op_sub = CartesianSamplingOp(encoding_shape=encoding_shape, traj=trajectory_sub)
 
-    # Verify that the fully-sampled sampling operator does not do anything because the data is already sorted
-    assert not SOp._needs_indexing
+    # Verify that the fully-sampled sampling operator does not do an_ything because the data is already sorted
```
ckolbPTB: an_ything?

tests/operators/test_cartesian_sampling_op.py
```diff
@@ -31,35 +32,40 @@ def test_cart_sampling_op_data_match():
...
+    trajectory = create_traj(k_shape, nkx, nky, nkz, sx, sy, sz)
 
     # Create matching data
     random_generator = RandomGenerator(seed=0)
     kdata = random_generator.complex64_tensor(size=k_shape)
 
     # Create sampling operator
     encoding_shape = SpatialDimension(k_shape[-3], k_shape[-2], k_shape[-1])
-    SOp = CartesianSamplingOp(encoding_shape=encoding_shape, traj=ktraj)
+    sampling_op = CartesianSamplingOp(encoding_shape=encoding_shape, traj=trajectory)
```
ckolbPTB, line 43: maybe `cart_sampling_op`?

tests/operators/test_sensitivity_op.py
```diff
@@ -14,88 +14,78 @@
...
     random_generator = RandomGenerator(seed=0)
-    Nz, Ny, Nx = 100, 100, 100
-    num_coils = 4
+    n_z, n_y, n_x = 100, 100, 100
+    n_coils = 4
 
     # Generate sensitivity operator
-    random_tensor = random_generator.complex64_tensor(size=(1, num_coils, Nz, Ny, Nx))
+    random_tensor = random_generator.complex64_tensor(size=(1, n_coils, n_z, n_y, n_x))
     random_csm = CsmData(data=random_tensor, header=QHeader(fov=SpatialDimension(1.0, 1.0, 1.0)))
```
ckolbPTB: `random_csmdata`?

tests/phantoms/test_ellipse_phantom.py
```diff
@@ -14,50 +14,44 @@
...
-    im_dim = SpatialDimension(z=1, y=ph_ellipse.ny, x=ph_ellipse.nx)
-    im = ph_ellipse.phantom.image_space(im_dim)
-    kdat = ph_ellipse.phantom.kspace(ph_ellipse.ky, ph_ellipse.kx)
-    FFOp = FastFourierOp(dim=(-1, -2))
-    (irec,) = FFOp.adjoint(kdat)
-    # Due to discretisation artifacts the reconstructed image will be different to the reference image. Using standard
+    img_dimension = SpatialDimension(z=1, y=ellipse_phantom.n_y, x=ellipse_phantom.n_x)
+    img = ellipse_phantom.phantom.image_space(img_dimension)
+    kdata = ellipse_phantom.phantom.kspace(ellipse_phantom.ky, ellipse_phantom.kx)
+    fourier_op = FastFourierOp(dim=(-1, -2))
```
ckolbPTB, line 52: `ff_op` ?

tests/utils/test_filters.py
ckolbPTB, line 1: replace res with filtered_data

Ln 43 unifirm -> uniform

tests/operators/test_zero_pad_op.py
```diff
@@ -12,34 +12,34 @@
...
 import torch
-
 from mrpro.operators import ZeroPadOp
+
 from tests import RandomGenerator
+from tests.helper import dotproduct_adjointness_test
 
 
 def test_zero_pad_op_content():
     """Test correct padding."""
```
ckolbPTB: actually also tests cropping (i.e. negative padding). Suggestion for docstring: Test correct padding and cropping (i.e. negative padding size)

ckolbPTB (changes requested):

ckolbPTB:
Currently we are using the class names in lowercase for variables, .e.g `kdata` or `idata`. The only exception I found was `KTrajectory`. Here we use `trajectory` rather than `ktrajectory`. Shall we adapt that to make it more consistent?

tests/operators/functionals/test_mse_functional.py
fzimmermann89: Not if we want to follow PEP8 naming guide: MSE only where we use CamelCase, i.e. class names, not for variable names

src/mrpro/algorithms/_adam.py
fzimmermann89: We used to mix tuple and int for parameters. This causes issues if we call it with the other one.
So 
```
def f(a:tuple[int,...])-> None:
    ...
f([1])
```
would make mypy complain. 

Sequence is anything that has getitem and has a length, so list and tuple both work.
So we should always use Sequence for inputs. This also means we cannot modify it (as it doesnt specify that it has a setitem). So its a tuple or a list treated like a tuple -> good for inputs. Otherwise the caller has to convert to a tuple.


Our return type is more narrow, we actually write what we return. We return tuples for operators.

The `*tuple[torch.Tensor, ...]` means that it is a function with more than one input. Nowhere in the inputs is a tuple involved, it is just a way to write `Operator: tensor,tensor,tensor,... -> (tensor,)`

This is only required if we want to write "arbittrary many inputs", otherwise it is, for example:
`Operator[torch.Tensor, tuple[torch.Tensor]], Operator[torch.Tensor, torch.Tensor, tuple[torch.Tensor]],`
This is the same style of generic typing as used for `Callable`.

So, long story short:
Sequence for inputs.
unpacked tuple in the typhints for operator (and callable) if we want to use an ellipsis.
tuple as the return type.

tests/phantoms/test_ellipse_phantom.py
fzimmermann89: I changed this to fourier_op as it more closely resembles its use here. 

that is a `fastfourier `does not matter the understand the rest of the code, but that it does a Fourier-transform helps.
the exact type will be shown on hover in an IDE anyways. and ff does in general not mean fouriertransform for me, only fft has this meaning... so either `fft_op` or `fourier_op`

fzimmermann89:
I would stick with trajectory, as there will ever only be a single trajectory. The idata/kdata is less related to the classes, but the the kind of data in my mind. 

Maybe related is the question: how to call k-data and image-data and trajectory-data that is not in an IData/KData class. There are a couple of places where we have data first as a tensor, then as a dataclass. this also causes some mypy ambiguity. 
img vs idata? both img?
k vs kdata? kdata_tensor?
trajectory, ktrajectory  ?

src/mrpro/data/_CsmData.py
```diff
@@ -68,7 +70,7 @@ def _iterative_walsh_csm(
     def from_idata_walsh(
         cls,
         idata: IData,
-        smoothing_width: SpatialDimension[int] = SpatialDimension(5, 5, 5),
+        smoothing_width: int | SpatialDimension[int] = 5,
         niter: int = 3,
```
fzimmermann89: niter-> power_iterations

docstring: number of iterations used to determine dominand eigenvector

src/mrpro/data/traj_calculators/_KTrajectoryPulseq.py
```diff
@@ -77,7 +73,7 @@ def __call__(self, kheader: KHeader) -> KTrajectoryRawShape:
...
             with torch.no_grad():
```
fzimmermann89: why is here a no_grad?
this is unnecessry.

src/mrpro/algorithms/_adam.py
ckolbPTB: > tuple as the return type.

Why does adam then return a list?

ckolbPTB:
> I would stick with trajectory, as there will ever only be a single trajectory. The idata/kdata is less related to the classes, but the the kind of data in my mind.

If we want other people to use a similar coding style it would be nice to have some simply guidelines and "use dataclasses as variable names" seems easier then saying use "idata for IData" but use "trajectory for KTrajectory and csm for CsmData" seems more complicated. If we think the "K" in "KTrajectory" is redundant then we should rename it.

src/mrpro/algorithms/_adam.py
fzimmermann89: good point.

src/mrpro/data/_CsmData.py
schuenke: added `test_CsmData_interactive_Walsh_smoothing_width` testing equality of csm data using calling iterative walsh with SpatialDimension and int

.pre-commit-config.yaml
schuenke: I removed all unnecessary quotes in the config

src/mrpro/algorithms/_adam.py
schuenke: fixed return type(s) for adam and lbfgs

src/mrpro/data/traj_calculators/_KTrajectoryPulseq.py
schuenke: changed to `encoding_size`

tests/operators/test_cartesian_sampling_op.py
schuenke: if we wanna rename it, I would suggest `carteasian_sampling_op`, but since we are in `test_cartesian_sampling_op.py` and only deal with cartesian data, I think we can stick with just `sampling_op` here

tests/phantoms/test_ellipse_phantom.py
schuenke: IMO `fourier_op` is fine

src/mrpro/data/_CsmData.py
schuenke: changed to `power_iterations` for now but not sure if I would prefer `n_iterations`...

fzimmermann89: I *think* the iterations are mostly useless in Walsh, it should just be big enough. There is never a need to reduce it to get better results.
n_iterations give the impression as if the number of iterations were a parameter of walsh's method, like "how many smoothing does it do". But I think it is not, it is just an implementation detail of our code.  In the paper, they just determine the dominant eigenvector. If we were smarter, we could most likely remove it and use a stopping different criterion.

so if you want `n_...` I would suggest:
 `n_power_iterations` 

(This will also appear in some other algorithms later on,with the same meaning (for example, espirit). For some time I thought it would make sense to adjust this parameter for espirit, but instead, it makes more sense to adjust the threshold....)

tests/data/_Dicom2DTestImage.py
schuenke: as discussed, we will maybe refactor this in a future PR

src/mrpro/data/_CsmData.py
schuenke: I will leave it as `power_iterations`

src/mrpro/data/_KData.py
schuenke: will be refactored in #207 anyway

schuenke:
I added a summary of (the most important) changes in the first post. 

IMO the PR is ready to be merged @ckolbPTB

fzimmermann89:
Can you remove the rgetattr and acceleration factor?

Both files have not been used for a long time.

fzimmermann89 (approved):

ckolbPTB:
One last thing: Can we put what you wrote under "mrpro naming convention" in the readme. alternative: paste and copy into an issue and asign it to me.

ckolbPTB (approved):

fzimmermann89:
I copied it over to the readme, feel free to expand on that.

Also, I removed rgetattr and AccelerationFactor. Both were implemented by me in the beginning but we soon decided NOT to use them and they have been dead code for months.
@schuenke I assume you agree as you suggested it in the first place..

Issue #194, opened 2024-03-01
Progress bars / verbose flag / callback functions / ... etc


koflera:
**Current Status**: We will add a callback to optimizer. Maybe to some other long running functions with defined progress as well.

## TODO ##
Add callback to:
- [x] CG (#195) 
- [x] ADAM
- [x] LBFGS

Update: we decided that it is not ot needed for 
- [ ] Direct Reconstruction of many 'others' ?  (#183)

## Original Issue ##

> For algorithms, it might be nice to have an option to be able to track some information when running the routine. 
> For example, for CG, it migh bi interesting to check the current residual, for adam/lbfgs/... the value of the loss function, etc.
> 
> Should we maybe start to add something like a `verbose` flag for algorithms?

fzimmermann89:
For low-level algorithms, I would not add a progressbar or output, but instead a callback, similar to what scipy does (https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.minimize.html):
You can specify a function that will be called after each iteration with the current parameters.

For  higher level algorithms, such as 'CG - Sense Reconstruction', 'Superresolution-Reconstruction' etc I would agree that a verbose flag with progress bars is helpful.

fzimmermann89:
The advantage of a callback over a verbose setting is, that most of the time, I would assume, we either
a) run in vscode with a debugger and can use its logging/debugging functions or
b) we run something that takes longer from the terminal, where a general verbose would just print difficult to parse information to stdout. 

Either way, what would a good 'verbose' be for a low-level algorithm?
A callback could instead use a problem dependent logging which is set at the higher level,  for example:
```
def callback_neptune(x):
   neptune_run['current_residual'].append(f(x))
```

An open question for me is what the signature of the callback should be:
Is it aways just a single tensor with the current estimate, or something more involved such as a 
(current_x, other_algorithm_dependend_stuff_as_dict) ?  or always a dict, with guaranteed key "x" for the current solution and all other keys algorithm dependent?

Pro single tensor:
 - consistent interface for all optimizers
 
Pro More information:
 - simple debugging by (for example), setting 
 `callback=lambda infos:print(infos['beta'])`

 Disadvantage:
 - different algorithms will return different information, so to fully allow mypy in user code, we need a `CG_CallbackDict` type (or are there smarted ways to do the typehinting than https://mypy-play.net/?mypy=latest&python=3.12&gist=e0ac8a4cf0cd8fe00d0fdde9be582eb3

[...]

PR #195, opened 2024-03-01
Conjugate Gradient

koflera:
This contains the implementation of the conjugate gradient method for linear systems of the form `Hx=b`, where `H` is a self-adjoint `LinearOperator`.


koflera:
> This contains the implementation of the conjugate gradient method for linear systems of the form `Hx=b`, where `H` is a self-adjoint `LinearOperator`.

A point for discussion would already be the following: Should this method be intened to be used with leading batch-dimension>1? If yes, we have to decide on how to act regarding the stopping criterion... Should the CG only stop if for all considered problems the stopping criterion is met?
I think this could in general be problematic as one could have differently-conditioned problems which require different stopping criterions/number of iterations...

I think I would leave the batch-version of the CG maybe as a future implementation as "layer" to be used within a NN. What do you think?

koflera:
Also, is this considered to be a "basic" algorithm for which we would like to include a callback? see also #194

fzimmermann89:
Without thinking too long about it:

- regarding batch: I would consider the operator as (other...) - > (other...) so the norm will be after flattening all dimensions and we consider a batched problem always as one big problem, not as batches.. I think this is what you are doing, right? 

- can we optimize the inner loop a bit? 

- if you want some verbosity on the optimizer, I'd add a callback to cg.

tests/algorithms/test_cg.py
```diff
@@ -0,0 +1,87 @@
...
+
+    # generate random matrices (avoid zeros by using ab offset of 1.)
+    if complex_valued_flag:
+        A = 1 + random_generator.complex64_tensor(size=(N, N))
+        H = A.conj().T @ A
+    else:
+        A = 1 + random_generator.float32_tensor(size=(N, N))
+        H = A.T @ A
+
+    class MatMult(LinearOperator):
```
fzimmermann89: Just wondering, maybe we could put this into the main of mrpro as a MatrixOperator ?
Thi might be useful for other stuff as well

tests/algorithms/test_cg.py
```diff
@@ -0,0 +1,87 @@
...
+
+        def __init__(self, Mat: torch.Tensor) -> None:
+            super().__init__()
+            self.Mat = Mat
+
+        def forward(self, x) -> tuple[torch.Tensor]:
+            return (self.Mat @ x,)
+
+        def adjoint(self, x: torch.Tensor) -> tuple[torch.Tensor]:
+            return (self.Mat.conj().T @ x,) if self.Mat.is_complex() else (self.Mat.T @ x,)
```
fzimmermann89: you can use Tensor.mH for real and complex here afaik

tests/algorithms/test_cg.py
```diff
@@ -0,0 +1,87 @@
...
+
+def create_self_adjoint_op(N, complex_valued_flag):
+    """Create self-adjoint operator H and wrap as LinearOperator."""
+    random_generator = RandomGenerator(seed=0)
+    N = 32
+
+    # generate random matrices (avoid zeros by using ab offset of 1.)
+    if complex_valued_flag:
+        A = 1 + random_generator.complex64_tensor(size=(N, N))
+        H = A.conj().T @ A
```
fzimmermann89: this can be A.mH for both cases

tests/algorithms/test_cg.py
```diff
@@ -0,0 +1,87 @@
...
+
+import pytest
+import torch
+
+from mrpro.algorithms import cg
+from mrpro.operators import LinearOperator
+from tests import RandomGenerator
+
+
+def create_self_adjoint_op(N, complex_valued_flag):
```
fzimmermann89: complex_valued_flag -> complex_valued:bool

or maybe consider making this a parameterized fixture:
https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#fixture-parametrize

tests/algorithms/test_cg.py
```diff
@@ -0,0 +1,87 @@
...
+    """Create self-adjoint operator H and wrap as LinearOperator."""
+    random_generator = RandomGenerator(seed=0)
+    N = 32
```
fzimmermann89: N is overwritten

tests/algorithms/test_cg.py
fzimmermann89, line 23: this could be a fixture

fzimmermann89: with the parameterization here

tests/algorithms/test_cg.py
```diff
@@ -0,0 +1,87 @@
...
+
+        def adjoint(self, x: torch.Tensor) -> tuple[torch.Tensor]:
+            return (self.Mat.conj().T @ x,) if self.Mat.is_complex() else (self.Mat.T @ x,)
+
+    HOp = MatMult(H)
+    return HOp
+
+
+@pytest.mark.parametrize('complex_valued_flag', [True, False])
+def test_cg_convergence(complex_valued_flag):
```
fzimmermann89: this could be `test_cg_convergence(selfadjointoperator)`
with the operator a parameterized fixture

tests/algorithms/test_cg.py
```diff
@@ -0,0 +1,87 @@
...
+    xtrue = (
+        1 + random_generator.complex64_tensor(size=(N,))
+        if complex_valued_flag
+        else 5 * random_generator.float32_tensor(size=(N,))
+    )
+    (b,) = HOp(xtrue)
+    x0 = torch.ones_like(xtrue)
+    xcg = cg(HOp, b, x0=x0, max_iter=N)
+    rtol = 1e-3 * torch.vdot(b.flatten(), b.flatten())
+    atol = 1e-3
```
fzimmermann89: why atol as a variable?
why is rtol dependent on |b|^2?

src/mrpro/algorithms/_cg.py
```diff
@@ -0,0 +1,81 @@
...
+        x = x0.clone() if x0 is not None else b.clone()
+
+    for i in range(max_iter):
+        rTr = torch.vdot(r.flatten(), r.flatten())
+        (Hp,) = H(p)
+        alpha = rTr / (torch.vdot(p.flatten(), Hp.flatten()))
+        x = x + alpha * p
+        r = r - alpha * Hp
+
+        sqnorm_r = torch.vdot(r.flatten(), r.flatten()).real
```
fzimmermann89: can we avoid half of the torch.vdot(r.flatten(), r.flatten())'s?

src/mrpro/algorithms/_cg.py
```diff
@@ -0,0 +1,81 @@
...
+        if sqnorm_r <= tol * sqnorm_b:
```
fzimmermann89: only check this if tol=!0
each check will synchronize cuda afaik.

src/mrpro/algorithms/_cg.py
```diff
@@ -0,0 +1,81 @@
...
+            return x
+
+        if i < max_iter - 1:
```
fzimmermann89: Maybe there is a nicer way to write the loop without this if check?

src/mrpro/algorithms/_cg.py
```diff
@@ -0,0 +1,81 @@
...
+
+    if x0 is not None:
+        if x0.shape != b.shape:
+            raise ValueError(
+                'CG solves linear systems Hx=b with self-adjoint operator;'
+                'this implies that x0 has to have the same shape as b'
+                f'\nshapes {x0.shape, b.shape} are incompatible'
+            )
+
+    # initial guess of the residual
```
fzimmermann89: ? initial guess of the residual ?

src/mrpro/algorithms/_cg.py
```diff
@@ -0,0 +1,81 @@
...
+
+import torch
+
+from mrpro.operators import LinearOperator
+
+
+def cg(
+    H: LinearOperator, b: torch.Tensor, x0: torch.Tensor | None = None, max_iter: int = 128, tol: float = 1e-4
+) -> torch.Tensor:
+    """CG for solving a linear system Hx=b, where H isself-adjoint.
```
ckolbPTB: isself-adjoint -> is self-adjoint

src/mrpro/algorithms/_cg.py
```diff
@@ -0,0 +1,81 @@
...
+
+    N.B. if the condition of H is very large, a small residual does not necessarily
+    imply that the solution is accurate.
+
+    Parameters
+    ----------
+    H
+        Self-adjoint operator
```
ckolbPTB: self-adjoint

src/mrpro/algorithms/_cg.py
```diff
@@ -0,0 +1,81 @@
...
+    b
+        right-hand-side of the system
+    x0, optional
+        starting value of the iteration; if not chosen, x0:=b
```
ckolbPTB: maybe: starting value, if None it will be set to b

src/mrpro/algorithms/_cg.py
```diff
@@ -0,0 +1,81 @@
...
+    max_iter, optional
+        number of maximal iterations
```
ckolbPTB: maximum number of iterations

tests/algorithms/test_cg.py
```diff
@@ -0,0 +1,87 @@
...
+from mrpro.operators import LinearOperator
+from tests import RandomGenerator
+
+
+def create_self_adjoint_op(N, complex_valued_flag):
+    """Create self-adjoint operator H and wrap as LinearOperator."""
+    random_generator = RandomGenerator(seed=0)
+    N = 32
+
+    # generate random matrices (avoid zeros by using ab offset of 1.)
```
ckolbPTB: using **an** offset

ckolbPTB: actually maybe better to use 
`random_generator.complex64_tensor(size=(N, N), low=1)`

tests/algorithms/test_cg.py
```diff
@@ -0,0 +1,87 @@
...
+    (b,) = HOp(xtrue)
+    x0 = torch.ones_like(xtrue)
+    xcg = cg(HOp, b, x0=x0, max_iter=N)
+    rtol = 1e-3 * torch.vdot(b.flatten(), b.flatten())
+    atol = 1e-3
+
+    # test if solution is accurate
+    torch.testing.assert_close(xcg, xtrue, rtol=rtol, atol=atol)
+
+    # test if cg stops if the ground-truth is the initial guess
```
ckolbPTB: This tests if one iteration of CG is the ground truth but not if the algorithm stops.

tests/algorithms/test_cg.py
```diff
@@ -0,0 +1,87 @@
...
+    xcg_one_iter = cg(HOp, b, x0=xtrue, max_iter=1)
+    assert (xtrue == xcg_one_iter).all()
+
+
+def test_invalid_shapes():
+
+    # check if wrongly set-up linear system throws error
```
ckolbPTB: check that incompatible shapes of x0 and b lead to error

src/mrpro/algorithms/_cg.py
koflera: i refactored the code a little bit (actually, there was also a mistake) and quite closely followed the structure of scipy... they also have a similar check

koflera:
> Without thinking too long about it:
> 
>     * regarding batch: I would consider the operator as (other...) - > (other...) so the norm will be after flattening all dimensions and we consider a batched problem always as one big problem, not as batches.. I think this is what you are doing, right?

exactly, I added a comment in the documentation that explains how this case is handled and also included some tests to check that this works indeed.

>     * can we optimize the inner loop a bit?
I now closely followed the scipy version, but maybe there is still room for improvments?

>     * if you want some verbosity on the optimizer, I'd add a callback to cg.
done

src/mrpro/operators/_MatrixOp.py
```diff
@@ -0,0 +1,34 @@
...
+
+from mrpro.operators import LinearOperator
+
+
+class MatrixOp(LinearOperator):
+    """Simple Linear Operator that implements matrix multiplication."""
+
+    def __init__(self, matrix: torch.Tensor) -> None:
+        super().__init__()
+        self.matrix = matrix
```
fzimmermann89:
```suggestion
        self.register_buffer('matrix',matrix)
```

tests/algorithms/test_cg.py
```diff
@@ -0,0 +1,154 @@
...
+
+    # distinguish cases with mb>1 or mb=1 to appropriately construct the operator
+    # for scipy's cg
+    batchsize = right_hand_side.shape[0] if len(right_hand_side.shape) == 3 else 1
+
+    # if batchsize>1, construct H = diag(H1,...,H_mb) and b=[b1,...,b_mb]^T, otherwise just take the matrix
+    h_matrix_np = scipy.linalg.block_diag(*h_operator.matrix.numpy()) if batchsize > 1 else h_operator.matrix.numpy()
+
+    # choose small tolerance to avoid exiting the for loop in the cg
+    tolerance = 1e-12
```
fzimmermann89: can we set tolerance=0 ?

tests/algorithms/test_cg.py
```diff
@@ -0,0 +1,154 @@
...
+    """Test if our implementation is close to the one of scipy."""
+    # create operator, right-hand side and ground-truth data
+    h_operator, right_hand_side, solution = system
+
+    # generate invalid initial value
+    starting_value = torch.zeros_like(right_hand_side)
+
+    # distinguish cases with mb>1 or mb=1 to appropriately construct the operator
+    # for scipy's cg
+    batchsize = right_hand_side.shape[0] if len(right_hand_side.shape) == 3 else 1
```
fzimmermann89, line 115: I think we could remove this

tests/algorithms/test_cg.py
```diff
@@ -0,0 +1,154 @@
...
+)
+def system(request):
+    """Generate data for creating a system Hx=b with linear and self-adjoint
+    H."""
+    random_generator = RandomGenerator(seed=0)
+
+    # get parameters for the test
+    batchsize, vectorsize, complex_valued = request.param
+
+    # if mb=1, it corresponds to one linear system; for mb>1, multiple systems
```
fzimmermann89: why mb in the comment?

tests/algorithms/test_cg.py
```diff
@@ -0,0 +1,154 @@
...
+    )
+
+    # make sure H is self-adjoint
+    h_matrix = a_matrix.mH @ a_matrix
+
+    # construct matrix multiplication as LinearOperator
+    h_operator = MatrixOp(h_matrix)
+
+    # create ground-truth data and right-hand side of the system
+    vector = (
```
fzimmermann89: consider writing as 
```
if complex_valued:
    vector = random_generator.complex64_tensor(size=vector_shape, high=1.0)
else:
    vector = random_generator.float32_tensor(size=vector_shape, low=-1.0, high=1.0)

```
as this Is mire redable and even shorter ...

tests/algorithms/test_cg.py
```diff
@@ -0,0 +1,154 @@
...
+    (right_hand_side,) = h_operator(vector)
+
+    return h_operator, right_hand_side, vector
+
+
+def test_cg_convergence(system):
+    """Test if CG delivers accurate solution."""
+
+    # create operator, right-hand side and ground-truth data
+    h_operator, right_hand_side, solution = system
```
fzimmermann89: question: why is h_operator and  right_hand_side? As in, why is the operator named like in the equation Hx=b, b and x are not..

h_operator and b
or linear_op and right_hand_side?

tests/algorithms/test_cg.py
```diff
@@ -0,0 +1,154 @@
...
+
+def test_cg_stopping_after_one_iteration(system):
+    """Test if cg stops after one iteration if the ground-truth is the initial
+    guess."""
+    # create operator, right-hand side and ground-truth data
+    h_operator, right_hand_side, solution = system
+
+    # callback function; should not be called since cg should exit for loop
+    def callback(solution):
+        assert False
```
fzimmermann89: consider adding text to assert

tests/algorithms/test_cg.py
fzimmermann89: can we use less teneries?

src/mrpro/operators/_MatrixOp.py
```diff
@@ -0,0 +1,34 @@
+"""Fourier Operator."""
```
ckolbPTB: Matrix Operator

src/mrpro/algorithms/_cg.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+from collections.abc import Callable
+
+import torch
+
+from mrpro.operators import LinearOperator
+
+
+def conjugate_gradient(
```
ckolbPTB: as we have adam and lbfgs I would suggest to call this also cg

src/mrpro/algorithms/_cg.py
```diff
@@ -0,0 +1,124 @@
...
+    imply that the solution is accurate.
+
+    Parameters
+    ----------
+    operator
+        self-adjoint operator (named H above)
+    right_hand_side
+        right-hand-side of the system (named b above)
+    starting_value, optional
+        starting value of the iteration; if None, it will be set to b
```
ckolbPTB: will be set to right_hand_side

src/mrpro/algorithms/_cg.py
```diff
@@ -0,0 +1,124 @@
...
+    # assign starting value to the solution
+    solution = starting_value.clone() if starting_value is not None else right_hand_side.clone()
+
+    # for the case where the residual is exactly zero
+    if torch.vdot(residual.flatten(), residual.flatten()) == 0:
+        return solution
+
+    # squared tolerance;
+    # (we will check ||residual||^2 < tolerance^2 instead of ||residual|| < tol
+    # #to avoid the computation of the root for the norm)
```
ckolbPTB: #to -> to

src/mrpro/algorithms/_cg.py
```diff
@@ -0,0 +1,124 @@
...
+    tolerance_squared = tolerance**2
+
+    # dummy value the old squared norm of the residual;
```
ckolbPTB: initialise the squared norm of the residual with a dummy value

src/mrpro/algorithms/_cg.py
```diff
@@ -0,0 +1,124 @@
...
+    norm_squared_residual = None
+
+    for i in range(max_iterations):
+
+        # calculate the square norm of the residual
+        residual_flat = residual.flatten()
+        norm_squared_residual_new = torch.vdot(residual_flat, residual_flat).real
+
+        # check if the solution is already accurate enough
+        if tolerance != 0:
```
ckolbPTB: Does this mean that for tolerance = 0 the algorithm will always run for max_iterations even if norm_squared_residual_new == 0? Is this what we want?

src/mrpro/algorithms/_cg.py
```diff
@@ -0,0 +1,124 @@
...
+    # squared tolerance;
+    # (we will check ||residual||^2 < tolerance^2 instead of ||residual|| < tol
+    # #to avoid the computation of the root for the norm)
+    tolerance_squared = tolerance**2
+
+    # dummy value the old squared norm of the residual;
+    # #only required for initialization
+    norm_squared_residual = None
+
+    for i in range(max_iterations):
```
ckolbPTB: i -> iteration

src/mrpro/operators/_MatrixOp.py
```diff
@@ -0,0 +1,56 @@
...
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+
+import torch
+
+from mrpro.operators import LinearOperator
+
+
+class MatrixOp(LinearOperator):
```
ckolbPTB: This would also require a test

tests/algorithms/test_cg.py
```diff
@@ -0,0 +1,151 @@
...
+        assert False, 'CG did not exit'
+
+    # the test should fail if we reach the callback
+    xcg_one_iteration = conjugate_gradient(
+        operator, right_hand_side, starting_value=solution, max_iterations=10, tolerance=1e-4, callback=callback
+    )
+    assert (xcg_one_iteration == solution).all()
+
+
+def test_implementation(system):
```
ckolbPTB: test_compare_cg_to_scipy

tests/algorithms/test_cg.py
```diff
@@ -0,0 +1,151 @@
...
+    """Test if our implementation is close to the one of scipy."""
+    # create operator, right-hand side and ground-truth data
+    operator, right_hand_side, _ = system
+
+    # generate invalid initial value
```
ckolbPTB: why "invalid"?

tests/algorithms/test_cg.py
```diff
@@ -0,0 +1,151 @@
...
+    # if batchsize>1, construct H = diag(H1,...,H_mb) and b=[b1,...,b_mb]^T, otherwise just take the matrix
+    matrix_np = scipy.linalg.block_diag(*operator.matrix.numpy()) if batchsize > 1 else operator.matrix.numpy()
+
+    # choose zero tolerance to avoid exiting the for loop in the cg
+    tolerance = 0.0
+
+    # test for different maximal number of iterations
+    for max_iterations in [1, 2, 4, 8]:
+        # run our cg and scipy's cg and compare the results
+        (xcg_scp, _) = cg_scp(
```
ckolbPTB, line 115: cg_solution_scipy

tests/algorithms/test_cg.py
```diff
@@ -0,0 +1,151 @@
...
+    """Test if CG throws error in case of shape-mismatch."""
+    # create operator, right-hand side and ground-truth data
+    h_operator, right_hand_side, _ = system
+
+    # generate invalid initial starting point
+    starting_value = torch.zeros(
+        h_operator.matrix.shape[-1] + 1,
+    )
+    with pytest.raises(ValueError, match='incompatible'):
+        _ = conjugate_gradient(h_operator, right_hand_side, starting_value=starting_value, max_iterations=10)
```
ckolbPTB: I don't think "_ =" is needed.

ckolbPTB (changes requested):

src/mrpro/operators/_MatrixOp.py
koflera: I am not sure whether a test is really necessary here, since this is really not more than a wrapper for torch's @ operator. What do we expect could go wrong?

src/mrpro/algorithms/_cg.py
fzimmermann89: `# dummy value. new value will be set in loop before first usage
`
"initialize he squared norm of the residual" is redundant.

src/mrpro/algorithms/_cg.py
```diff
@@ -0,0 +1,124 @@
...
+
+        # update estimates of the solution and the residual
+        (operator_conjugate_vector,) = operator(conjugate_vector)
+        alpha = norm_squared_residual_new / (
+            torch.vdot(conjugate_vector.flatten(), operator_conjugate_vector.flatten())
+        )
+        solution += alpha * conjugate_vector
+        residual -= alpha * operator_conjugate_vector
+
+        norm_squared_residual = norm_squared_residual_new
```
fzimmermann89: residual_squared = residual_squared_new

src/mrpro/operators/_MatrixOp.py
```diff
@@ -0,0 +1,56 @@
...
+        ----------
+        x
+            tensor to be multiplied with matrix
+
+        Returns
+        -------
+            result of matrix-vector multiplication
+        """
+
+        return (self.matrix @ x,)
```
fzimmermann89: Can we make this a bit more general with regards to batching?
As is, this will be of of limited real world use due to reshaping and batching required.
let my try out something....

src/mrpro/algorithms/_cg.py
koflera: is the suggestion just to leave out the "norm"? or to have `norm_residual_squared` instead of `residual_norm_squared`?
for the first, I would say it doesn't make sense as it is really the norm and a not a poit-wise squared operation we mean here.

src/mrpro/algorithms/_cg.py
koflera: this is related to @fzimmermann89's comment above where he mentioned that cuda will synchronize after each check. 

i think it makes sense to keep this logic. The tolerance and the maximal number of iterations are two stopping criterions and if either of the two is met, the iteration stops.

Setting tolerance=0 therefore corresponds to only having the maximal number of iterations as a stopping criterion. 
I would say this is the desired behaviour of the algorithm regardless of whether or not this corresponds to what one would do in practice.

having a exactly zero residual is very unlikely, I would say. Even in the absence of noise, rounding errors will proably avoid having ||r||==0. The check in the beginning before the iterations is to prevent the iterations to start if we start with the exact solution (also something that would probably not occur in practice).

fzimmermann89: tolarance == 0 meaning: dont stop on a tolerance criterion feels quite logical to me.
maybe we can also explicitly mention this in the docstring

src/mrpro/algorithms/_cg.py
fzimmermann89: both-ish :D  suggestions would be
1. leave out the norm 
(imho implied by residual squared -> for an physicist a squared vector usually means the norm implied by the dot product, vector^H vector.)
2. to put "residual" first, as this is the important part of the name.

src/mrpro/operators/_MatrixOp.py
fzimmermann89: see the other PR where I requested your review

fzimmermann89:
For lbfgs we use 'initial parameters', here it is 'starting value', can we use similar names for the optimizers?
-> initial value ?

src/mrpro/algorithms/_cg.py
ckolbPTB: > maybe we can also explicitly mention this in the docstring

yes, please

src/mrpro/operators/_MatrixOp.py
ckolbPTB: > What do we expect could go wrong?

famous last words. 

In general our tests should cover as much of our code as possible. We might change some interfaces of the Operators and then we might need to adapt this interface too. If there is no test it is very likely that we would miss this. Nevertheless, probably @fzimmermann89 PR will add a test for this.

src/mrpro/algorithms/_cg.py
koflera: ok, I think I prefer to keep the word "norm" in then and go for option 2).

fzimmermann89:
can we add a test that checks if the callback is called?

ckolbPTB (approved):

PR #197, opened 2024-03-03
Continuous ir model

ckolbPTB:
Signal model for continuous acquisition with and without inversion pulses.


fzimmermann89:
Some basic questions / comments:

- I do not really understand the parameters of the model.
Why is it required to specify tau and tacq_start?
Shouldn't it be enough to specify when acquisitions happen and when inversions happen? Or maybe I misunderstood the parameters, in which case this is more a comment regarding the docstring ;)

- In general, can we maybe come up with better names than `tacq` or `tau`  ? 
 I would say spending a couple more characters to write out the meaning of these could be beneficial. Same  for `ind` and `idx`.

-  Is there a reason for the indexing instead of directly handling the times?

  For an extreme example, you could consider looping in the forward over the inversion times, for each inversion time selecting the acquisitions after it and before the next one, then figuring out each tau, and calculating the part of the signal. Finally, stacking the signals. 

  Even though this might on GPU cause additional Synchronisation, would this be significantly slower, justifying the increased complexity?

  
If yes, you might consider underscoring the attributes that cannot be changed after creating the model as they would require recalculation of the indices. 

-  Is there an advantage of using torch.div over normal "/" (or, if you prefer, t1.reciprocal()) ?

- can we write the last reshaping in a way that works even if t1 is just a scalar-tensor?

- We might consider making the tensors in operators  either buffers or parameters to allow move to gpu.

src/mrpro/operators/models/_TransientInversionRecovery.py
```diff
@@ -0,0 +1,190 @@
...
+    multiple inversion pulses. All the timings (signal_time_points, inversion_time_points, first_adc_time_point)
+    have to be given relative to the same clock. Values can be given in s, ms... but have to have all the same units.
+
+    Overview of the sequence
+    M0 equilibrium magnetisation
+    Mz measured magnetisation
+    M* steady-state magnetisation
+
+
+    (1) Start of data acquisition (first_adc_time_point) (optional)
```
fzimmermann89: I forgot, was there a reason why you prefer to call it the start of acquisition instead of the start of RF pulses?
As there are not necessarily adcs there, right?

src/mrpro/operators/models/_TransientInversionRecovery.py
```diff
@@ -0,0 +1,190 @@
...
+    (2) Data acquisition before first inversion pulse (optional)
+        Mz(t) = M* + (M0 - M*)e^(-t * R1*)
+    (3) First inversion pulse (inversion_time_points[0])
+    (4) Time between inversion pulse and start of data acquisition, (time_inversion_adc), e.g. due to spoiler gradient
```
fzimmermann89: delay_inversion_adc?

src/mrpro/operators/models/_TransientInversionRecovery.py
```diff
@@ -0,0 +1,190 @@
...
+            repetition time
+        inversion_time_points
+            time stamp of each inversion
+        time_inversion_adc
+            time between inversion pulse and start of data acquisition
+        first_adc_time_point
+            time stamp of first acquisition
+        """
+        super().__init__()
+        tr = torch.Tensor(tr)
```
fzimmermann89: torch.as_tensor

fzimmermann89: otherwise there will be a warning if the parameter is a tensor.

src/mrpro/operators/models/_TransientInversionRecovery.py
```diff
@@ -0,0 +1,190 @@
...
+        time_inversion_adc = torch.Tensor(time_inversion_adc)
+
+        self.signal_time_points = torch.nn.Parameter(signal_time_points, requires_grad=signal_time_points.requires_grad)
+        self.tr = torch.nn.Parameter(tr, requires_grad=tr.requires_grad)
+        self.time_inversion_adc = torch.nn.Parameter(time_inversion_adc, requires_grad=time_inversion_adc.requires_grad)
+
+        self.first_adc_time_point = torch.Tensor(first_adc_time_point) if first_adc_time_point is not None else None
+        self.index_before_first_inversion = torch.where(signal_time_points < inversion_time_points[0])[0]
```
fzimmermann89: this should be register_buffer

src/mrpro/operators/models/_TransientInversionRecovery.py
```diff
@@ -0,0 +1,190 @@
...
+                'the start of the acquisitions first_adc_time_point has to be defined.',
+            )
+
+        # Verify that no acquisition occurred before start of acquisition
+        if self.first_adc_time_point is not None and any(signal_time_points < first_adc_time_point):
+            raise ValueError('Acquisitions detected before start of acquisition.')
+
+        # Add final "virtual" inversion pulse to allow for easier looping later
+        inversion_time_points = torch.concat((inversion_time_points, (signal_time_points.max() + 1).reshape(1)), dim=0)
+        self.inversion_time_points = torch.nn.Parameter(
```
fzimmermann89: is the function a) differentiable wrt to the inversion time points? If not, this should not be a Parameter but a buffer.
b) can you later on change the inversion time points? If not, it should be _inversion_time_points with an underscore, right?

Sorry, I did not remember the content of this function when you asked the question earlier..

src/mrpro/operators/models/_TransientInversionRecovery.py
```diff
@@ -0,0 +1,190 @@
...
+            longitudinal relaxation time T1
+        alpha
+            flip angle
+
+        Returns
+        -------
+            signal with dimensions ((... times), coils, z, y, x)
+        """
+        t1 = torch.where(t1 == 0, 1e-10, t1)
+        t1_star = 1 / (1 / t1 - torch.log(torch.cos(torch.deg2rad(alpha))) / self.tr)
```
fzimmermann89: wasnt the decision at some point to have angles in radians by default?

fzimmermann89: @schuenke

src/mrpro/operators/models/_TransientInversionRecovery.py
```diff
@@ -0,0 +1,190 @@
...
+
+            # Signal at the beginning of the next inversion pulse
+            time = (
+                self.inversion_time_points[inversion_index + 1]
+                - self.inversion_time_points[inversion_index]
+                - self.time_inversion_adc
+            )
+            m0_plus = m0_star - (m0_tau + m0_star) * torch.exp(-time / t1_star)
+
+        return (rearrange(signal, 't ... c z y x -> (... t) c z y x'),)
```
fzimmermann89: this will flatten all other dims, I think this is not what we wan here, right?
If my m0, t1, etc is split by slices, it should return the signal split by slcies

tests/operators/models/test_transient_inversion_recovery.py
```diff
@@ -0,0 +1,93 @@
...
+@pytest.mark.parametrize(
+    ('t', 'result'),
+    [
+        # short ti
+        (0, '-m0'),
+        # long ti
+        (20, 'm0_star'),
+    ],
+)
+def test_signal_boundaries_single_inversion(t, result):
```
fzimmermann89: can we add a check for the signal at t1=0, t1=inf, m0=0, m0=inf.
it should not return nans.

Also, it would already be nice to add tests for the gradients.
Or, instead, I can add the tests for each signal model during the hackathon.
If the later, I would only add tests. If one of the tests fails, I will then open PRs for each failing signal function for somebody else to fix it...

src/mrpro/operators/models/_TransientInversionRecovery.py
ckolbPTB: An inversion pulse is also an RF pulse. I guess it could be first_excitation_pulse_time_point but this a bit long. As TE << T1 acquisition and RF_ex can be considered at the same time point and yes in theory you could have a RF_ex without ADC but in practice there is no reason why you would do that.

src/mrpro/operators/models/_TransientInversionRecovery.py
ckolbPTB: not sure what we want. I would prefer to have one other dimension with clear labels what is what. Otherwise I have to remember that I am applying first a SignalModel and then a MotionOperator and hence my tensor looks like (motion state, inversion_times, other,...) rather than (inversion_times, motion_states, other,...).

src/mrpro/operators/models/_TransientInversionRecovery.py
fzimmermann89: At least to my, this made it more difficult to understand to signal model. As I did not get the distingion between adcs and acquisitions...
Excitation and acquisition would have been easier to understand.

src/mrpro/operators/models/_TransientInversionRecovery.py
ckolbPTB: Currently all the other models flatten along other so I would suggest to keep it like this at the moment and if want to change it then change it for all models in a separate PR

tests/operators/models/test_transient_inversion_recovery.py
ckolbPTB: > If the later, I would only add tests. If one of the tests fails, I will then open PRs for each failing signal function for somebody else to fix it...

sounds like a good plan!

src/mrpro/operators/models/_TransientInversionRecovery.py
ckolbPTB: out of curiosity, what needs to be on the gpu? Do we need to be able to pu every variable in self on the gpu?

src/mrpro/operators/models/_TransientInversionRecovery.py
fzimmermann89: Do we really want to introduce this bug just because no one has ifxex it yet for the other signal models?
The signal  models  should just keep the shape of the other dimensions unchanged...

src/mrpro/operators/models/_TransientInversionRecovery.py
ckolbPTB: first_acquisition_time_point?

src/mrpro/operators/models/_TransientInversionRecovery.py
ckolbPTB: > The signal models should just keep the shape of the other dimensions unchanged...

I am not so sure. Eventually we want to map from parameter_map -> kdata and kdata is of shape (other, coils, k2, k1, k0) and therefore we need to somehow flatten the other dimension. Simply appending new dimensions at the beginning doesn't sound like a great idea to me.

src/mrpro/operators/models/_TransientInversionRecovery.py
ckolbPTB: I don't think that the signal is differentiable wrt to any of the timings because they are dependent on each other. 

If I simply do `self.register_buffer('_inversion_time_points', inversion_time_points)` then at least in my VSCode `self._inversion_time_points` is of type Any and I lose all my type hinting. How is `register_buffer` usually used?

src/mrpro/operators/models/_TransientInversionRecovery.py
fzimmermann89: But matching the order will be an issue. 
So is it (other T) or (t other) and how to convert from one to the other. 
The signal models and optimizers work with tensors, so you would need to do this reordering anyways. 
I thin it is easier to split in the reconstruction the kdata into a suitable shape using a rearrange, then do the fitting. It is easier to remember that the last other dimension is time than which values of the tensor.. 



Compromise:
Always flatten the time dimension into the last other dimension.

So for the other dimensions
-  (slice, 1) would result in (slice, t) < almost- how I am using it in my code, I would just need to add a singleton dimension

- (other) in (other*T) <- current behavior

So it is 't ...  other ->... (other t)' or similar for the reshape

src/mrpro/operators/models/_TransientInversionRecovery.py
ckolbPTB: as_tensor

ckolbPTB:
I have not managed to get this model to work with vmap.

ckolbPTB:
Making this model work nicely with vmap would require restricting it to a special use case. We will revisit this if somebody needs this model.

PR #201, opened 2024-03-06
docker build separated from tests

ckolbPTB:


ckolbPTB:
Summary:
- container images are build in separate docker.yml
- docker images are build for changes of               
   - 'docker/*'
   - 'pyproject.toml'
   - '/.github/docker.yml'
- docker images are first build with tag test and pushed to the github registry
- tests are run in the images (does not update any test badges and coverage reports etc)
- once all tests are passed images are tagged as latest and pushed to github registry
- ptytest is now run on all images found on the github registry with tag latest

schuenke (approved):
Sounds and looks good to me. If there are any problems, we will probably only realize when it's merged and in action...

fzimmermann89:
Just two final questions (I havent looked at the actual code):
- Would this break if two PRs are modifying the pyproject.toml with both starting the action that pushes a test image? Will the images get mixed up or overwritten?
- Will old images remain in the github container registry? If yes, is there a limit how much space we can take up or how many versions we can have?

ckolbPTB:
> Will old images remain in the github container registry? If yes, is there a limit how much space we can take up or how many versions we can have?

Yes, an old image will remain in the registry, but I can't think of any way to figure out from a current PR which images can be automatically deleted. We will have to delete them manually. Nevertheless, I think most of the time we will overwrite images and only once we upgrade to a new python will this be necessary.

> Would this break if two PRs are modifying the pyproject.toml with both starting the action that pushes a test image? Will the images get mixed up or overwritten?

There will be a conflict if two PRs push a test image at the same time in the sense that they will be overwritten or gcr will return an error that the same image is already being pushed. Not sure which one. This requires pushes to the PR within 20-30s which I believe will happen very rarely. To avoid that we would have to use PRs dependent tags but then we really have to clean up the tags after successful testing and at the moment I don't see the need for this. If this error occurs simply reruning the failing job would fix this. 

One problem I realised is that if a new Dockerfile is added then there is a short time frame in which only image:test but not image:latest exists. This would currently lead to a failing test because we only check for the available images and then assume it has a tag latest. I will try to add a check that verifies that the tag latest exists here: 

https://github.com/PTB-MR/mrpro/blob/81b7ee0102d7f4fe991881c698640063618fa9c9/.github/workflows/pytest.yml#L21

UPDATE: test are only run for containers with tag "latest"

PR #204, opened 2024-03-07
Return copy in remove readout os

fzimmermann89:
``Should this return a view, a copy or modify in place?

Before, it modified the original header and returned a new KData object with a view into the data.
This combination IMHO the worst option and somewhat surpring.

I think this should better return a copy of everything. 
Although it results in a copy operation on the actual data, it should make it easier for python to clean up memory.
As long as we hold a reference to a view, the original array will not be garbage collected afaik, 
```
bigarray = np.array(BIGNUMBER)
bigarray = bigarray[0]
```
would still occupy O(BIGNUMBER) memory.

So it might be better to pay the cost of creating a copy and temporarily up to double the memory usage, but allowing python to clean up the memory afterwards.

### DIFF
--- a/src/mrpro/algorithms/_remove_readout_os.py
+++ b/src/mrpro/algorithms/_remove_readout_os.py
@@ -34,7 +34,7 @@ def remove_readout_os(kdata: KData) -> KData:
 
     Returns
     -------
-        (A new copy of) K-space data with oversampling removed.
+        Copy of K-space data with oversampling removed.
 
     Raises
     ------
--- a/src/mrpro/algorithms/_remove_readout_os.py
+++ b/src/mrpro/algorithms/_remove_readout_os.py
@@ -12,6 +12,8 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
+from copy import deepcopy
+
 import torch
 
 from mrpro.data import KData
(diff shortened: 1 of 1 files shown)


ckolbPTB:
> So it might be better to pay the cost of creating a copy and temporarily up to double the memory usage, but allowing python to clean up the memory afterwards.

Yes and I thought that was what I was doing, but I guess I forgot the deepcopy of the header

src/mrpro/algorithms/_remove_readout_os.py
```diff
@@ -31,49 +34,46 @@ def remove_readout_os(kdata: KData) -> KData:
 
     Returns
     -------
-        K-space data with oversampling removed.
+        (A new copy of) K-space data with oversampling removed.
```
ckolbPTB: Copy of k-space data

The brackets might suggest that this only happens sometimes and a copy is usually something new.

ckolbPTB (approved):

fzimmermann89:
> > So it might be better to pay the cost of creating a copy and temporarily up to double the memory usage, but allowing python to clean up the memory afterwards.
> 
> Yes and I thought that was what I was doing, but I guess I forgot the deepcopy of the header

and of the trajectory I think....

PR #207, opened 2024-03-09
Allow different total number of phase encode steps in KData

fzimmermann89:
This simplifies the logic to determine `n_k2 `and `n_k1 `in `KData`. A lot of the complexity was due to the error message for the `NotImplementedError`.

Before, there were some edgecases that were not fully covered.  Now we also have a fallback for different numbers of phase encodes for different "other" labels, with shape (total_aquisitions, 1, 1, k0) with a warning instead of an exception.

(I would also be fine with removing the warning. I added it as I think it could  for us equally likely be a data/user error than the expected resullt.)

I also renamed` KTrajectoryRawShape.reshape` as this looked to close to `Tensor.reshape`: I was wondering why we would reshape with "sort_idx" as one of the sizes. Now it is `sort_and_reshape`

Note: This is currently vs the ruff PR, but will be rebased onto main and will be merged seperatly, either before or after #193

### DIFF
--- a/src/mrpro/data/_KTrajectoryRawShape.py
+++ b/src/mrpro/data/_KTrajectoryRawShape.py
@@ -47,7 +47,7 @@ class KTrajectoryRawShape:
     repeat_detection_tolerance: None | float = 1e-3
     """tolerance for repeat detection, by default 1e-3, None to disable."""
 
-    def reshape(
+    def sort_and_reshape(
         self,
         sort_idx: np.ndarray,
         n_k2: int,
--- a/src/mrpro/data/_kdata/_KData.py
+++ b/src/mrpro/data/_kdata/_KData.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 
 import dataclasses
 import datetime
+import warnings
 from pathlib import Path
 from typing import Protocol
 
@@ -26,6 +27,7 @@ import ismrmrd
 import numpy as np
 import torch
 from einops import rearrange
+from mrpro.data._AcqInfo import AcqIdx
 from mrpro.data._AcqInfo import AcqInfo
 from mrpro.data._EncodingLimits import Limits
 from mrpro.data._kdata._KDataRearrangeMixin import KDataRearrangeMixin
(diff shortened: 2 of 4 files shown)


fzimmermann89:
This will require some testing on actual data which I have not done yet.

ckolbPTB:
> Before, there were some edgecases that were not fully covered. Now we also have a fallback for different numbers of phase encodes for different "other" labels, with shape (total_aquisitions, 1, 1, k0) with a warning instead of an exception.

I am not sure I understand what the reason for this is. Yes, we can do this, but we don't have any code which allows for the reconstruction of such data.

ckolbPTB:
How about having another `KDataRaw` class which has a `from_file` method which does everything up to this point: https://github.com/PTB-MR/mrpro/blob/caa89eba74c65c64d498ffb996f94054393cb8ad/src/mrpro/data/_kdata/_KData.py#L132 

Then we have a method in `KData` called `from_raw` which does the rest (and of course keep from_file which would do both steps in one call). `from_raw` would then throw an error if `KDataRaw` is not valid, i.e. cannot be reshaped into our matrix. Then we would have an option to fix certain parts of the data but make it clear that `KDataRaw` cannot yet be used for image reconstruction.

fzimmermann89:
Moving the reshaping in a different function could simplify the from_file method a bit and allow us to call it again at a later point after changing some indices.


I would prefer NOT to add a new class for the different shapes. A warning is IMHO enough.
Most functions should really work the same irregardless of the shape. If you want to do an nmr spectrum you should be able to do it. The gain of the new class does not outweigh the added complexity wrt to mixins and protocols..

ckolbPTB:
> If you want to do an nmr spectrum you should be able to do it. 

We can do this already because the different spectra are not encoded as different phase encoding points. Hence everything will work as expected.

> Most functions should really work the same irregardless of the shape.

True, but it is not about the shape but the fact that we split kdata into dimensions which require Fourier transform (k0, k1, k2) and are combined in a single image and dimensions which lead to different images (coils, other). If we now put phase encoding dimensions into other, our code will "work" but the result will be wrong. 

> The gain of the new class does not outweigh the added complexity wrt to mixins and protocols.

KDataRaw would not require any Mixins or Protocols because it would really only be a data class with a single method.

fzimmermann89:
Then it doesnt really help. If we cannot at least reshape, rearrange, and select subsets KDataRaw use useless.
If we do this, it should be the other way around and everything should by default be implemented for KDataRaw except if need to know the phase encoding steps.

I still vote to return the Data and print a warning. We can give the warning a 
We can explicitly mention that reconstruction will most likely fail and use a bit stronger language in the warning

If you really want to have it fail in your code, you can use (or similar, depending on the actual message we use for the warning) to treat it as an exception
```
import warnings
warnings.filterwarnings("error","Data Shape Issue" ) 
```

schuenke (approved):
I like the new simplified n_k1 and n_k2 logic/implementation! Much cleaner and easier to understand imo. 

Regarding the n_acqs_per_other > 1 case: I don't have a very strong opinion, but I think it's nice to be able to read the data despite potential wrong labels/flags. Imo, raising a (strong) warning in these cases is sufficient. I would love to have the option (without modifying the code 😄 ) to read in such data in case I messed up the labels in a pulseq sequence etc... If we don't wanna allow it by default, it should at least be optional. 

Moving the sort and reshape function of acq_info into AcqInfo itself etc. would be nice to keep the KData functions short and clear and would actually fit much better to the rest of our implementation style. However, I think this would be part of #243 , no ?!

ckolbPTB:
I still don't think it is a good idea to return KData which will break the rest of the code but luckily we are three admins so we can have a majority vote. Maybe extend the warning message for n_acqs_per_other > 1 by mentioning that the rest of the reconstruction will most likely yield unexpected results.

Issue #208, opened 2024-03-09
Feature Request: Rotation


fzimmermann89:
I consider adding a Rotation property to the AcqInfo.
The getter would return a [scipy.spatial Rotation ](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.html) from slice_dir, phase_dir and read_dir, the setter set these from a Rotation. 

Even though it is just a one-liner, it could be useful for other to avoid many different ad-hoc implementations dealing with the rotation angles etc.. 

Or is this unnecessary and the directions are deemed enough as an interface?

fzimmermann89:
I started to implement a pytorch version of Rotation that also works with our axis name order and tensors. 

Some of the conversions will not be differentible and happen on cpu. For example, the conversion from a rotation to Euler angles. Reimlementing this in pytorch is IMHO currently not worth the effort and I would call into scipy. I will issue a warning if the conversation is called with a tensor that requires gradients.

Once this is done, we can decide if we want to have it in mrpro.utils, and if yes, we we want to use it in AcqInfo.

PR #209, opened 2024-03-09
Rotation  class

fzimmermann89:
This is an implementation of a rotation class, implementing most of the interface of scipy.spatial.transform.Rotation but for torch.Tensors and using zhe zyx order of axes.

(See #208 )

Notable changes are:
- `.apply `is replaced by forward / call, as the class is derived from a torch.nn.Module. This might change in the future.
- arbitrary number of batch dimensions. scipy only supports a list of rotations, i.e. a single batch dimension.
-  no rotation group functions.
- align_vectors calls scipy version
- mean has a dim and keep dim parameter, similar to Tensor.mean.
- forward can be applied to SpatialDimension 

maybe Todo's for later:
- add dim to concatenate, extending scipy.Rotation
- add .rearrange

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,6 +36,7 @@ dependencies = [
     "pydicom",
     "pypulseq@git+https://github.com/imr-framework/pypulseq",
     "torchkbnufft>=1.4.0",
+    "scipy>=1.12"
 ]
 
 [project.optional-dependencies]
--- a/src/mrpro/utils/_Rotation.py
+++ b/src/mrpro/utils/_Rotation.py
@@ -46,8 +46,8 @@ from scipy.spatial.transform import Rotation as Rotation_scipy
 
 from mrpro.data import SpatialDimension
 
-AXIS_ORDER = ('x', 'y', 'z')  # This can be modified
-QUAT_AXIS_ORDER = (*AXIS_ORDER, 'w')  # Do not modify
+AXIS_ORDER = 'zyx'  # This can be modified
+QUAT_AXIS_ORDER = AXIS_ORDER + 'w'  # Do not modify
 assert QUAT_AXIS_ORDER[:3] == AXIS_ORDER, 'Quaternion axis order has to match axis order'  # noqa: S101
 
 
@@ -67,24 +67,29 @@ def _compose_quaternions_single(p: torch.Tensor, q: torch.Tensor) -> torch.Tenso
 
 
 def _compose_quaternions(p: torch.Tensor, q: torch.Tensor) -> torch.Tensor:
+    """Calculate p * q, with p and q batched quaternions."""
     p, q = torch.broadcast_tensors(p, q)
     product = torch.vmap(_compose_quaternions_single)(p.reshape(-1, 4), q.reshape(-1, 4)).reshape(p.shape)
     return product
 
 
-def _canonical_quaternion(q: torch.Tensor) -> torch.Tensor:
-    x, y, z, w = (q[..., QUAT_AXIS_ORDER.index(axis)] for axis in 'xyzw')
+def _canonical_quaternion(quaternion: torch.Tensor) -> torch.Tensor:
+    """Convert to canonical form, i.e. positive w."""
+    x, y, z, w = (quaternion[..., QUAT_AXIS_ORDER.index(axis)] for axis in 'xyzw')
     needs_inversion = (w < 0) | ((w == 0) & ((x < 0) | ((x == 0) & ((y < 0) | ((y == 0) & (z < 0))))))
-    q = torch.where(needs_inversion.unsqueeze(-1), -q, q)
-    return q
+    canonical_quaternion = torch.where(needs_inversion.unsqueeze(-1), -quaternion, quaternion)
+    return canonical_quaternion
 
 
 def _matrix_to_quaternion(matrix: torch.Tensor) -> torch.Tensor:
+    """Convert matrix to quaternion."""
     if matrix.shape[-2:] != (3, 3):
         raise ValueError(f'Invalid rotation matrix shape {matrix.shape}.')
-    batch_shape = matrix.shape[:-2]
-    m00, m01, m02, m10, m11, m12, m20, m21, m22 = torch.unbind(matrix.flatten(start_dim=-2), -1)
 
+    batch_shape = matrix.shape[:-2]
+    # matrix elements
+    m00, m01, m02, m10, m11, m12, m20, m21, m22 = torch.unbind(matrix.flatten(start_dim=-2), -1)
+    # q,r,s are some permutation of x,y,z
     qrsw = torch.nn.functional.relu(
         torch.stack(
             [
@@ -97,7 +102,9 @@ def _matrix_to_quaternion(matrix: torch.Tensor) -> torch.Tensor:
         )
     )
     q, r, s, w = qrsw.unbind(-1)
-
+    # all these are the same except in edge cases.
+    # we will choose the one that is most numerically stable.
+    # we calculate all choices as this is faster
     candidates = torch.stack(
         (
             *(q, m10 + m01, m02 + m20, m21 - m12),
@@ -107,7 +114,7 @@ def _matrix_to_quaternion(matrix: torch.Tensor) -> torch.Tensor:
         ),
         dim=-1,
     ).reshape(*batch_shape, 4, 4)
-
+    # now we make the choice.
     # the choice will not influence the gradients.
     choice = qrsw.argmax(dim=-1)
     quaternion = candidates.take_along_dim(choice[..., None, None], -2).squeeze(-2) / (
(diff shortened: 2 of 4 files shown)


fzimmermann89:
@schuenke regarding#193 

The pre-commit fails without a helpful error message... I have no idea what the issue is. Running it twice does not help..

```
ruff-format..............................................................Failed
- hook id: ruff-format
- files were modified by this hook

1 file reformatted, 89 files left unchanged

docformatter.............................................................Failed
- hook id: docformatter
- exit code: 3
- files were modified by this hook
```

fzimmermann89:
I would like to have the integration of this into the rest of our code base as a seperate PR, to a llow for an easier review and to avoid having this one open for weeks.

So please review this as a stand alone rotation code.

fzimmermann89:
maybe Todo's:
- add dim to concatenate, extending scipy.Rotation
- add .rearrange
- add property test for single rotation

guastinimara:
I have many questions, and still not sure how to use the reviews in github, so I'll write everything here okay? And then understand how all of this works.

General question: 
if I understood it correctly, Rotation is a class of rotations stored in their quaternions from, which can come from rotation matrices, euler representation and vectors representation. I've seen that there is the choice of computing the "canonical" version of a quaternion. But if I understood it correctly, in general elements of his class are not necessarily in the canonical form, or not? 
If so my question is why? Couldn't it be clearer to always store the quaternions in their canonical form or it doesn't make sense?

guastinimara:
Short questions:

- def mean: I don't understand what dim could be? I get it if it's the batch dimension, but can we mean over other dimensions? What would this mean in practice for rotations?
- def __getitem__: does this work also if indexer is a list of integers or only single integer?
- def align_vectors: I'm still understanding this, I think I'll have questions about it later
- def magnitude: is there a reason why angles are given in radians?

guastinimara:
- **def from_matrix**:
I don't really understand what you're doing in matrix_to_quaternion (I'm not familiar with this), maybe can you add a reference to the method as a comment there or send it here so that I can get it? I don't know whether this is a standard straightforward method.
Also, since a rotation matrix corresponds to two quaternions, is there here a line where that choice between the two is taken?
I couldn't find the source [2] F. Landis Markley, "Unit Quaternion from Rotation Matrix" to be publicly available, and I cant see in matrix_to_quaternion if this case where the the matrix is not orthogonal is already considered. Is this something that is there already and I didn't see that? or that needs to be added?

guastinimara:
- **def _make_elementary_quat(axis: str, angles: torch.Tensor):** 
Stupid question: why angles plural? Isn't it one angle if we are making the rotation around one axes?

guastinimara:
- **def _quaternion_to_matrix** :  
in the comment: we use q, r, s for same permutation of "xyz" ??? --> what does this mean??
Maybe I would add https://en.wikipedia.org/wiki/Rotation_matrix#Conversions (Quaternions) here, or put the formula of 
the matrix from there.
In general, do we prefer to add links to the descriptions or to be descriptive and put formulas already in the descriptions?

guastinimara:
- **def from_euler**: 
If I understood it correctly, we need angles.shape[-1] == n_axes? if so I would add it in the comment description as well, it would make it more clear for me at a first glance.

- **def _quaternion_to_euler**: where did you find the formula for this? Maybe I would add it either as a link or as a comment, it would be useful. but maybe you already want to add it as as_euler is not commented yet as as_quat is.

fzimmermann89:
> .... in general elements of his class are not necessarily in the canonical form, or not? If so my question is why? Couldn't it be clearer to always store the quaternions in their canonical form or it doesn't make sense?
> Also, since a rotation matrix corresponds to two quaternions, is there here a line where that choice between the two is taken?

In most cases, it does not matter which of the two quaternions is used. Only if you want to print it out, you might want to choose. Always transforming into the canonical form internally significantly increase the time it takes to multiply to rotations, for example. As _quaternions is not publicly accessible, but only the internal representation, it does not really matter that it is not always  in canonical form.

fzimmermann89:
> def mean: I don't understand what dim could be? I get it if it's the batch dimension, but can we mean over other dimensions? What would this mean in practice for rotations?

There are multiple batch dimensions. You can only mean over these dimensions. 

> def getitem: does this work also if indexer is a list of integers or only single integer?

Yes, the type hint is wrong. I will try to fix it.

> def magnitude: is there a reason why angles are given in radians?

1. Because all the scipy functions that this is based on use radians as default
2. because we decided to use radians as default #26 

> def _make_elementary_quat(axis: str, angles: torch.Tensor):

yes, this is inconsistent, although it can be a batch of angles, it should be named angle to be consistent with the other names in the class.

> def from_euler: If I understood it correctly, we need angles.shape[-1] == n_axes? 

Yes, agree.

> def _quaternion_to_matrix :
>in the comment: we use q, r, s for same permutation of "xyz" ??? --> what does this mean??

should be "some permutation", i.e. (q=x r=y s=z), or (q=y,r=z,s=x) or ... Do you have a suggestion how to describe this better?
This is used in a few places, as the rotation class can work with different orders of axes names... so it is not always: First axes is x, second is y and so on... So to avoid having a variable named 'x' that is actually 'z', I decided to name name q,r,s instead..
I am open to suggestions for comment or naming that would make it easier to understand

fzimmermann89:
> * **def from_matrix**:
> * F. Landis Markley, "Unit Quaternion from Rotation Matrix"

This (and the scipy.rotation.Rotation class, as mentioned in the beginning of the file) is the reference.
See also https://github.com/scipy/scipy/blob/main/scipy/spatial/transform/_rotation.pyx

schuenke:
IMO we need to add a scipy >= 1.12 dependency, because: 

with scipy 1.11.4 and before [c38fc76](https://github.com/PTB-MR/mrpro/pull/209/commits/c38fc763bc33d8fb4b09bc22386a8566cf9c562c) the following tests failed on my machine:

```
=========================== short test summary info ===========================
FAILED tests/utils/test_rotation.py::test_align_vectors_improper_rotation - a...
FAILED tests/utils/test_rotation.py::test_align_vectors_invalid_input - Asser...
FAILED tests/utils/test_rotation.py::test_align_vectors_align_constrain - Run...
FAILED tests/utils/test_rotation.py::test_align_vectors_near_inf - numpy.lina...
FAILED tests/utils/test_rotation.py::test_align_vectors_parallel - RuntimeWar...
FAILED tests/utils/test_rotation.py::test_weighted_mean_dims[shape0-True-0-expected_shape0]
FAILED tests/utils/test_rotation.py::test_weighted_mean_dims[shape2-True-dim2-expected_shape2]
FAILED tests/utils/test_rotation.py::test_weighted_mean_dims[shape4-False-None-expected_shape4]
======================== 8 failed, 157 passed in 1.55s ========================
```

with scipy 1.12.0 and before [c38fc76](https://github.com/PTB-MR/mrpro/pull/209/commits/c38fc763bc33d8fb4b09bc22386a8566cf9c562c) some additional tests passed:

```
=========================== short test summary info ===========================
FAILED tests/utils/test_rotation.py::test_weighted_mean_dims[shape0-True-0-expected_shape0]
FAILED tests/utils/test_rotation.py::test_weighted_mean_dims[shape2-True-dim2-expected_shape2]
FAILED tests/utils/test_rotation.py::test_weighted_mean_dims[shape4-False-None-expected_shape4]
======================== 3 failed, 162 passed in 1.60s ========================
```

with scipy 1.12.0 and after [c38fc76](https://github.com/PTB-MR/mrpro/pull/209/commits/c38fc763bc33d8fb4b09bc22386a8566cf9c562c) all tests pass 🎉 

UPDATE:
with scipy 1.11.4 and after [c38fc76](https://github.com/PTB-MR/mrpro/pull/209/commits/c38fc763bc33d8fb4b09bc22386a8566cf9c562c) we still got failing tests:

```
=========================== short test summary info ===========================
FAILED tests/utils/test_rotation.py::test_align_vectors_improper_rotation - a...
FAILED tests/utils/test_rotation.py::test_align_vectors_invalid_input - Asser...
FAILED tests/utils/test_rotation.py::test_align_vectors_align_constrain - Run...
FAILED tests/utils/test_rotation.py::test_align_vectors_near_inf - numpy.lina...
FAILED tests/utils/test_rotation.py::test_align_vectors_parallel - RuntimeWar...
======================== 5 failed, 160 passed in 2.05s ========================
```

==> we need scipy >= 1.12.0

fzimmermann89:
ups, forgot to commit the cahnge to the pyproject.toml.
we need scipy 1.12 due to:

> The method Rotation.align_vectors was extended to solve a constrained alignment problem where two vectors are required to be aligned precisely. 

https://docs.scipy.org/doc/scipy/release/1.12.0-notes.html

guastinimara:
> should be "some permutation", i.e. (q=x r=y s=z), or (q=y,r=z,s=x) or ... Do you have a suggestion how to describe this >better? This is used in a few places, as the rotation class can work with different orders of axes names... so it is not always: First >axes is x, second is y and so on... So to avoid having a variable named 'x' that is actually 'z', I decided to name name q,r,s >instead.. I am open to suggestions for comment or naming that would make it easier to understand

Now that you explained it is clearer, but I couldn't think of a better way to describe it that doesn't make it too confusing (also, it wouldn't make too much sense to trying to reconnect their names to x,y,z if they are not them respectively). 

I've just seen that you now added the commas from 'xyz' to 'x, y, z' in this comment and I was going to suggested that as well, so that for me is clear now.

fzimmermann89:
soooo... any news here?

schuenke (approved):

fzimmermann89:
I will merge this now, if tehre is anything missing or if we want to move it somewhere else, this is for a future PR..

PR #215, opened 2024-03-13
Restructure algorithms & Move to direct imports in src

### LINKED ISSUE #210: Move Iterative Walsh to algorithms?
I would suggest to move `iterative_walsh` 

https://github.com/PTB-MR/mrpro/blob/9abccff816a5a55d657cfb592af69e08f96dac56/src/mrpro/data/_CsmData.py#L29

to algorithms... having it in `CsmData` feels like suggesting that is the standard algorithm of choice, but we might also want to use something else, like espirit etc.

Also, given that iterative walsh seems to mainly work using a power-iteration, I would suggest to implement a general class for a power-iteration method (we will need this anyway for other algorithms as well) and to use that for iterative-walsh.

what do you think?

schuenke:
## Moving to direct imports
We now use direct imports in all of our src/. 
So, for example, instead of `from mrpro.data import KData`  we now use `from mrpro.data._kdata._KData import KData`
In tests and examples we use to convenient imports, that is `from mrpro.data import KData`.
## Restructuring of algorithms

- introduce submodules in `algorithms` for `optimizers` (currently adam and lbfgs) and `csm` methods (currently only walsh, but espirit etc. in the future)
- make `remove_readout_os` an instance method of KData instead of a function in `algorithms`

closes #210


fzimmermann89:
Either we remove the CSMData class and replace the sensitivity operator by a normal multiplication OR the sensitivity Operator should take CSMData as input. 

It does not make sense to first create a CSMData instance only to then use a raw tensor for an operator that only does a multiplication, which can equally well be expressed as a sumproduct.

Thus, please keep the input argument a CSMData instance. If you want to, you can optionally make it a tensor.

Which circular import is the issue? Maybe there is either an __future__.annotations missing our you can import from _Module directly.

schuenke:
> Either we remove the CSMData class and replace the sensitivity operator by a normal multiplication OR the sensitivity Operator should take CSMData as input.
> 
> It does not make sense to first create a CSMData instance only to then use a raw tensor for an operator that only does a multiplication, which can equally well be expressed as a sumproduct.
> 
> Thus, please keep the input argument a CSMData instance. If you want to, you can optionally make it a tensor.
> 
> Which circular import is the issue? Maybe there is either an **future**.annotations missing our you can import from _Module directly.

I commited the changes without touching the SensitivityOp so that the error can be checked in the logs:

```
ImportError while loading conftest '/home/runner/work/mrpro/mrpro/tests/conftest.py'.
tests/conftest.py:7: in <module>
    from mrpro.data import AcqInfo
/opt/hostedtoolcache/Python/3.11.8/x64/lib/python3.11/site-packages/mrpro/data/__init__.py:12: in <module>
    from mrpro.data._kdata._KData import KData
/opt/hostedtoolcache/Python/3.11.8/x64/lib/python3.11/site-packages/mrpro/data/_kdata/_KData.py:33: in <module>
    from mrpro.data._kdata._KDataRemoveOsMixin import KDataRemoveOsMixin
/opt/hostedtoolcache/Python/3.11.8/x64/lib/python3.11/site-packages/mrpro/data/_kdata/_KDataRemoveOsMixin.py:21: in <module>
    from mrpro.operators import FastFourierOp
/opt/hostedtoolcache/Python/3.11.8/x64/lib/python3.11/site-packages/mrpro/operators/__init__.py:5: in <module>
    from mrpro.operators._SensitivityOp import SensitivityOp
/opt/hostedtoolcache/Python/3.11.8/x64/lib/python3.11/site-packages/mrpro/operators/_SensitivityOp.py:17: in <module>
    from mrpro.data import CsmData
E   ImportError: cannot import name 'CsmData' from partially initialized module 'mrpro.data' (most likely due to a circular import) (/opt/hostedtoolcache/Python/3.11.8/x64/lib/python3.11/site-packages/mrpro/data/__init__.py)
```

I am ofc trying to find another fix, but wasn't successful until now... this is why it's still a draft PR

schuenke:
And I think it actually is a real circular import because `mrpro.data` is importing `SensitivityOp` from `mrpro.operators`, and `SensitivityOp` is importing `CsmData` from `mrpro.data`

Another working fix is to use a local import of CsmData in `SensitivityOp.__init__()` and only import CsmData at the top of the file when type checking, but imo we should not start using local imports...

schuenke:
@fzimmermann89 for some reason you modified the SensitivityOp tests in this PR. Can you have a look why they are failing now?

fzimmermann89:
Ah sorry, I will revert these and do a separate PR.

I wanted to move the sensitivity test over to the adjointness test function.  But the order of u and should be swapped.

schuenke:
Do we merge this as it is? Almost forgot about it to be honest... 
I merged current main and adjusted the cg implementation to fit the new structure.

schuenke:
> Do we merge this as it is? Almost forgot about it to be honest... I merged current main and adjusted the cg implementation to fit the new structure.

any further requests?

fzimmermann89 (approved):
Can you add some imports to the top-level init as discussed before merging?

fzimmermann89:
@schuenke: I added 2 missing imports to operators.init such that mrpro.operators.models.... works

This still needs a rebase.. Please check after rebase / merge that nothing "reappears" again :D 
I will not push the button.

PR #218, opened 2024-03-14
Clean up

### LINKED ISSUE #213: docstring 'optional' inconsistency
We sometimes have 'optional' in the docstrings if arguments have default values, but most of the time not.

Should we add 'optional', keep it as is or remove it?

I would prefer to remove it, as
1. it is already set by having a default argument 
2. easy to forget to add or change if the signature changes.
3. the convention comes from a time before type hints and to show, for example that something set to None means default value.

For me , all the arguments as for NOT putting the types in the docstring hold: We don't have to manually add the same information twice.

ckolbPTB:
- rename recon_/encoding_shape to recon_/encoding_matrix
- remove optional closes #213 
- some import fixes
- fix indentation in copy right statements

### DIFF
--- a/src/mrpro/operators/_CartesianSamplingOp.py
+++ b/src/mrpro/operators/_CartesianSamplingOp.py
@@ -98,7 +98,7 @@ class CartesianSamplingOp(LinearOperator):
         -------
             selected k-space data in acquired shape (as described by the trajectory)
         """
-        if self._encoding_matrix != SpatialDimension(*x.shape[-3:]):
+        if self._sorted_grid_shape != SpatialDimension(*x.shape[-3:]):
             raise ValueError('k-space data shape mismatch')
 
         if not self._needs_indexing:
@@ -140,7 +140,7 @@ class CartesianSamplingOp(LinearOperator):
         # is always constant zero and gradients w.r.t. src work as expected.
         y_scattered = torch.zeros(
             *broadcast_shape,
-            self._encoding_matrix.z * self._encoding_matrix.y * self._encoding_matrix.x,
+            self._sorted_grid_shape.z * self._sorted_grid_shape.y * self._sorted_grid_shape.x,
             dtype=y.dtype,
             device=y.device,
         ).scatter_(dim=-1, index=idx_expanded, src=y_kflat)
@@ -148,9 +148,9 @@ class CartesianSamplingOp(LinearOperator):
         # reshape to  ..., other, coil, k2_enc, k1_enc, k0_enc
         y_reshaped = y_scattered.reshape(
             *y.shape[:-3],
-            self._encoding_matrix.z,
-            self._encoding_matrix.y,
-            self._encoding_matrix.x,
+            self._sorted_grid_shape.z,
+            self._sorted_grid_shape.y,
+            self._sorted_grid_shape.x,
         )
 
         return (y_reshaped,)
--- a/src/mrpro/operators/_CartesianSamplingOp.py
+++ b/src/mrpro/operators/_CartesianSamplingOp.py
@@ -3,14 +3,16 @@
 # Copyright 2023 Physikalisch-Technische Bundesanstalt
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
-#   you may not use this file except in compliance with the License.
-#   You may obtain a copy of the License at
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at:
+#
 #       http://www.apache.org/licenses/LICENSE-2.0
-#   Unless required by applicable law or agreed to in writing, software
-#   distributed under the License is distributed on an "AS IS" BASIS,
-#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-#   See the License for the specific language governing permissions and
-#   limitations under the License.
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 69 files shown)


src/mrpro/operators/_CartesianSamplingOp.py
```diff
@@ -28,18 +28,18 @@ class CartesianSamplingOp(LinearOperator):
...
     """
 
-    def __init__(self, encoding_shape: SpatialDimension[int], traj: KTrajectory) -> None:
+    def __init__(self, encoding_matrix: SpatialDimension[int], traj: KTrajectory) -> None:
         """Initialize Sampling Operator class.
 
         Parameters
         ----------
-        encoding_shape
+        encoding_matrix
```
fzimmermann89, line 39: 3 years ago I would have completely misunderstood this docstring and expected the Matrix to actually be a matrix, I. E. some linear operator. 
I know it is common lab speak in MR, but maybe
the word shape should appear somewhere?

schuenke (approved):

schuenke:
Maybe we leave it open for a bit in case we find more things to clean up before the hackathon? :)

fzimmermann89:
I made FastFourierOp actually take an encoding and/or a recon matrix, now matching the name again.

also, I added comments and docstring on CartesianSamplingOp and FastFourierOp.

fzimmermann89:
I vote against keeping this open.

I want to get this merged as fast as possible, as a fix to #221 would otherwise need merging and rebasing.

Please let's try to do not use this as another 'touches all files' PR :D

schuenke:
Sure, we can also merge it now. Well, when all tests pass ;)

fzimmermann89:
Not sure if I like the rename to recon_matrix and encoding_matrix.

For the fastfourierop, it is now additional logic which can only handle some cases.
Can there be cases were the shape in the fourierop should not be the recon_matrix/encoding_matrix?

In #183 I instead added a utility class method to the fourierop that uses a kdata object to create a suitable operator, with `recon_shape=kdata.header.recon_matrix, encoding_shape=kdata.header.encoding_matrix, trajectory=kdata.traj`
maybe this is also a way to do it if we do not couple the recon_shape by name with recon_matrix...

src/mrpro/operators/_FastFourierOp.py
```diff
@@ -3,57 +3,112 @@
...
-            # No padding
+            padded_shape = encoding_matrix
+
+        if original_shape is not None and padded_shape is not None:
+            # perform padding / cropping
+            self._pad_op = ZeroPadOp(dim=dim, original_shape=original_shape, padded_shape=padded_shape)
+        elif encoding_matrix is None and recon_matrix is None:
+            # No padding no padding / cropping
             self._pad_op = ZeroPadOp(dim=(), original_shape=(), padded_shape=())
+        else:
```
fzimmermann89, line 107: this is new and afaik required. there is currently no good way to do it with only one of the matrices set.
before, it failed in zeropadop

src/mrpro/operators/_FastFourierOp.py
```diff
@@ -3,57 +3,112 @@
...
         super().__init__()
         self._dim = tuple(dim)
         self._pad_op: ZeroPadOp
-        if encoding_shape is not None and recon_shape is not None:
-            self._pad_op = ZeroPadOp(dim=dim, original_shape=recon_shape, padded_shape=encoding_shape)
+
+        if isinstance(recon_matrix, SpatialDimension):
+            original_shape: Sequence[int] | None = [int(astuple(recon_matrix)[d]) for d in dim]
+            if not all(d in (-1, -2, -3) for d in dim):
+                raise NotImplementedError(
```
fzimmermann89, line 84: I dont see a ay to do it for other indices..

src/mrpro/operators/_FastFourierOp.py
```diff
@@ -3,57 +3,112 @@
...
+    Therefore for forward and ajoint first ifftshift needs to be applied, then fftn or ifftn and then ifftshift.
+
+    """
 
     def __init__(
         self,
         dim: Sequence[int] = (-3, -2, -1),
-        recon_shape: Sequence[int] | None = None,
-        encoding_shape: Sequence[int] | None = None,
+        recon_matrix: SpatialDimension[int] | Sequence[int] | None = None,
```
fzimmermann89, line 47: if it is named recon_matrix, it should work with a recon_matrix

src/mrpro/operators/_CartesianSamplingOp.py
```diff
@@ -127,16 +138,16 @@ def adjoint(self, y: torch.Tensor) -> tuple[torch.Tensor,]:
...
             dtype=y.dtype,
             device=y.device,
         ).scatter_(dim=-1, index=idx_expanded, src=y_2d)
         # reshape to  ..., other, coil, k2_enc, k1_enc, k0_enc
         y_reshaped = y_scattered.reshape(
             *y.shape[:-3],
-            self._encoding_shape.z,
-            self._encoding_shape.y,
-            self._encoding_shape.x,
+            self._sorted_grid_shape.z,
```
fzimmermann89, line 151: @ckolbPTB : decided to call it _sorted_grid_shape, not tensor_shape....
not sure what the best name if, but this makes it at least to me a bit clearer that this is the shape of the data _after sorting_ it..

fzimmermann89 (commented):
@ckolbPTB maybe you can have another look and merge?

maybe you can check if the new expanded docstrings are correct or if I mixed something up..

src/mrpro/operators/_CartesianSamplingOp.py
fzimmermann89: I added it...

PR #220, opened 2024-03-14
move sensitivity operator tests to using  dotproduct_adjointness_test 

fzimmermann89:
We forgot to move this one over in #193 
Also, it now tests for multiple other dimensions and if the result is the same if created with a tensor and with CSMData.

I added an explicit shape check to dotproduct_adjointness_test, as the flatten before the dotproduct made the test insensitive to an operator returning the wrong shape in the adjoint.

### DIFF
--- a/tests/helper.py
+++ b/tests/helper.py
@@ -36,12 +36,49 @@ def relative_image_difference(img1, img2):
     return image_difference / image_mean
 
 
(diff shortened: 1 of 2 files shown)


schuenke (approved):
`csm_n_other `-> `n_other_csm`
`img_n_other` -> `n_other_img`
to match our naming convention. rest lgtm 👍

Issue #221, opened 2024-03-14
Bug: CartesianSamplingOp for different traj along other


ckolbPTB:
I think this code breaks if we have a different trajectory for other dimensions, because we broadcast along (... other coil):
https://github.com/PTB-MR/mrpro/blob/14b83b2e0f7014406b2b1895d80166c5c07be216/src/mrpro/operators/_CartesianSamplingOp.py#L121

Not sure if this can be easily fixed or if it requires a diffferent approach to how we handle the sorting here.

Issue #217, opened 2024-03-13
Mismatch encoding limits - encoding matrix


ckolbPTB:
The encoding matrix describes the size of the k-space which needs to be encoded based on the resolution and FOV settings. Commonly data acquisition is then carried out in this matrix (or smaller due to partial fourier or undersampling). This means the encoding limits are smaller or equal to the matrix. 

There are cases where this is not the case. If the encoding matrix along PE is 188 but the number of samples per shot (e.g. PE points acquired after a prep-pulse or a trigger signal) is 20, then actually 200 PE samples will be obtained leading to the encoding limits of [0, 199]. 

For Cartesian sampling this currently breaks then e.g. here https://github.com/PTB-MR/mrpro/blob/9abbd8c90563d5b922f42757314d2902de615627/src/mrpro/operators/_CartesianSamplingOp.py#L128.

Easy solution would be to ignore points outside of encoding_matrix but of course it would be nicer to extend the encoding_matrix to include these points. This would require image-based interpolation to achieve to correct resolution.

fzimmermann89:
But wouldn't doing an interpolation kind-of defeat the some of the purpose of a Cartesian FFT?
To get from, say, 0.9mm resolution to 1mm resolution correctly would most likely involve an interpolation at least as expensive as the FFT... (Fastest might actually be to crop in fourierspace....)

If an encoding_matrix smaller than what is actually acquired is specified, I would assume that the k-data is cropped to the matrix size.
(similarly, if the matrix if bigger I might assume zero padding to match the matrix...)

If I want to use all available data, I would have to increase the matrix in the header.
I would vote for doing exactly what is specified as best as possible and printing a warning: "There is unused k-space data  that could be used is the matrix size is increased by setting kdata.header.encoding_matrix = _increased matrix size_ " 


Or am I misunderstanding the issue?

ckolbPTB:
Yes, that makes sense. I was thinking we could have the 0.9mm during the entire iterative recon and then have an interpolation step at the end, but I agree that in the end this wouldn't be much different from reconstructing the 1mm data in the first place.

We could
(a) ignore them when we read in the data and not put them into kdata.data. Advantage: kdata.data does not contain anything which is later ignored.
(b) ignore them in the CartesianSamplingOp. Advantage: The encoding_matrix can be easily adapted after reading in the data to include the additional samples

Any preferences?

fzimmermann89:
Let me think about it... 
[...]

PR #222, opened 2024-03-14
Support bipolar readout

### LINKED ISSUE #50: K0 limit calculation
I think the calculation of the k0 limits here are wrong:
https://github.com/ckolbPTB/mrpro/blob/52d469d7ebecc7fe5fd8cf352da8a3a0ad17241c/src/mrpro/data/_KData.py#L102

They should be calculated based on `AcqInfo.number_of_samples` and `AcqInfo.center_sample`

### LINKED ISSUE #184: Bipolar gradient
Currently we only support a single type of readout: https://github.com/PTB-MR/mrpro/blob/9b62daeb99a0bb5bb364088b4aedba37fbfdf0c8/src/mrpro/data/traj_calculators/_KTrajectoryCalculator.py#L69

For a multi-echo acquisition with bipolar readouts the readout direction (and probably centre of the readout) is different between odd and even echos.

ckolbPTB:
Correct trajectory calculation for bipolar (i.e. reversed) readout.

closes #184 
closes #50 

requires #221
Update: actually does not require #217

ToDo:
- [x] Fix encoding limits for readout in KData
- [x] Add tests for bipolar readout

### DIFF
--- a/src/mrpro/data/_kdata/_KData.py
+++ b/src/mrpro/data/_kdata/_KData.py
@@ -135,7 +135,13 @@ class KData(KDataSplitMixin, KDataRearrangeMixin, KDataSelectMixin):
 
         # Fill k0 limits if they were set to zero / not set in the header
         if kheader.encoding_limits.k0.length == 1:
-            kheader.encoding_limits.k0 = Limits(0, kdata.shape[-1] - 1, kdata.shape[-1] // 2)
+            # The encodig limits should describe the encoded space of all readout lines. If we have two readouts with
+            # (number_of_samples, center_sample) of (100, 20) (e.g. partial Fourier in the negative k0 direction) and
+            # (100, 80) (e.g. partial Fourier in the positive k0 direction) then this should lead to encoding limits of
+            # [min=0, max=159, center=80]
+            max_center_sample = int(torch.max(kheader.acq_info.center_sample))
+            max_pos_k0_extend = int(torch.max(kheader.acq_info.number_of_samples - kheader.acq_info.center_sample))
+            kheader.encoding_limits.k0 = Limits(0, max_center_sample + max_pos_k0_extend - 1, max_center_sample)
 
         # Sort kdata and acq_info into ("all other dim", coils, k2, k1, k0) / ("all other dim", k2, k1, acq_info_dims)
         # Fist, ensure each the non k1/k2 dimensions covers the same number of k1 and k2 points
--- a/src/mrpro/data/traj_calculators/_KTrajectoryCalculator.py
+++ b/src/mrpro/data/traj_calculators/_KTrajectoryCalculator.py
@@ -24,6 +24,7 @@ import torch
 from mrpro.data import KHeader
 from mrpro.data import KTrajectory
 from mrpro.data import KTrajectoryRawShape
+from mrpro.data.enums import AcqFlags
 
 
 class KTrajectoryCalculator(ABC):
@@ -57,19 +58,21 @@ class KTrajectoryCalculator(ABC):
         ValueError
             Center sample has to be the same for each readout
         """
-        n_samples = kheader.acq_info.number_of_samples
-        center_sample = kheader.acq_info.center_sample
-
-        # Verify that each readout has the same number of samples and same center sample
+        n_samples = kheader.acq_info.number_of_samples.flatten()
+        center_sample = kheader.acq_info.center_sample.flatten()
         if len(torch.unique(n_samples)) > 1:
             raise ValueError('Trajectory can only be calculated if each acquisition has the same number of samples')
-        if len(torch.unique(center_sample)) > 1:
-            raise ValueError('Trajectory can only be calculated if each acquisition has the same center sample')
+        n_k0 = int(n_samples[0])
 
-        # Calculate points along readout
-        n_k0 = int(n_samples[0, 0, 0])
-        k0 = torch.linspace(0, n_k0 - 1, n_k0, dtype=torch.float32) - center_sample[0, 0, 0]
-        return k0
+        # Data can be obtained with standard or reversed readout (e.g. bipolar readout).
+        k0 = torch.linspace(0, n_k0 - 1, n_k0, dtype=torch.float32)[None, :] - center_sample[:, None]
+        reversed_readout_idx = torch.where(kheader.acq_info.flags.flatten() & AcqFlags.ACQ_IS_REVERSE.value)[0]
+        # pytorch does not yet allow for negative indexing. torch.flip makes a copy of the data. providing a reversed
+        # index should be fastest
+        reversed_index = torch.linspace(n_k0 - 1, 0, n_k0, dtype=torch.int64)
+        k0[reversed_readout_idx, :] = k0[reversed_readout_idx[:, None], reversed_index[None, :]]
+
+        return k0.reshape(kheader.acq_info.center_sample.shape[:-1] + (n_k0,))
 
 
 class DummyTrajectory(KTrajectoryCalculator):
--- a/src/mrpro/data/traj_calculators/_KTrajectoryCalculator.py
+++ b/src/mrpro/data/traj_calculators/_KTrajectoryCalculator.py
@@ -58,21 +58,18 @@ class KTrajectoryCalculator(ABC):
         ValueError
             Center sample has to be the same for each readout
         """
-        n_samples = kheader.acq_info.number_of_samples.flatten()
-        center_sample = kheader.acq_info.center_sample.flatten()
-        if len(torch.unique(n_samples)) > 1:
+        n_samples = torch.unique(kheader.acq_info.number_of_samples)
+        center_sample = kheader.acq_info.center_sample
+        if len(n_samples) > 1:
             raise ValueError('Trajectory can only be calculated if each acquisition has the same number of samples')
-        n_k0 = int(n_samples[0])
+        n_k0 = int(n_samples.item())
 
         # Data can be obtained with standard or reversed readout (e.g. bipolar readout).
-        k0 = torch.linspace(0, n_k0 - 1, n_k0, dtype=torch.float32)[None, :] - center_sample[:, None]
-        reversed_readout_idx = torch.where(kheader.acq_info.flags.flatten() & AcqFlags.ACQ_IS_REVERSE.value)[0]
-        # pytorch does not yet allow for negative indexing. torch.flip makes a copy of the data. providing a reversed
-        # index should be fastest
-        reversed_index = torch.linspace(n_k0 - 1, 0, n_k0, dtype=torch.int64)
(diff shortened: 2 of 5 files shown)


fzimmermann89:
Do you have a test case that fails because of #221 ?

src/mrpro/data/traj_calculators/_KTrajectoryCalculator.py
```diff
@@ -78,8 +81,8 @@ class DummyTrajectory(KTrajectoryCalculator):
     Shape will fit to all data. Only used as dummy for testing.
     """
 
-    def __call__(self, header: KHeader) -> KTrajectory:
-        """Calculate dummy trajectory for given KHeader."""
-        kx = torch.zeros(1, 1, 1, header.encoding_limits.k0.length)
+    def __call__(self, header: KHeader) -> KTrajectory:  # noqa: ARG002
+        """Calculate dummy trajectory."""
+        kx = torch.zeros(1, 1, 1, 1)
```
fzimmermann89, line 81: why is this change required?

src/mrpro/data/traj_calculators/_KTrajectoryCalculator.py
```diff
@@ -57,19 +58,21 @@ def _kfreq(self, kheader: KHeader) -> torch.Tensor:
...
             raise ValueError('Trajectory can only be calculated if each acquisition has the same number of samples')
-        if len(torch.unique(center_sample)) > 1:
-            raise ValueError('Trajectory can only be calculated if each acquisition has the same center sample')
+        n_k0 = int(n_samples[0])
+
+        # Data can be obtained with standard or reversed readout (e.g. bipolar readout).
+        k0 = torch.linspace(0, n_k0 - 1, n_k0, dtype=torch.float32)[None, :] - center_sample[:, None]
+        reversed_readout_idx = torch.where(kheader.acq_info.flags.flatten() & AcqFlags.ACQ_IS_REVERSE.value)[0]
+        # pytorch does not yet allow for negative indexing. torch.flip makes a copy of the data. providing a reversed
+        # index should be fastest
```
fzimmermann89: pretty sure flip is faster and easier to read...

fzimmermann89: reversed_index is similiar to   reversed_readout_idx

src/mrpro/data/traj_calculators/_KTrajectoryCalculator.py
```diff
@@ -57,19 +58,21 @@ def _kfreq(self, kheader: KHeader) -> torch.Tensor:
...
+        center_sample = kheader.acq_info.center_sample.flatten()
         if len(torch.unique(n_samples)) > 1:
             raise ValueError('Trajectory can only be calculated if each acquisition has the same number of samples')
-        if len(torch.unique(center_sample)) > 1:
-            raise ValueError('Trajectory can only be calculated if each acquisition has the same center sample')
+        n_k0 = int(n_samples[0])
+
+        # Data can be obtained with standard or reversed readout (e.g. bipolar readout).
+        k0 = torch.linspace(0, n_k0 - 1, n_k0, dtype=torch.float32)[None, :] - center_sample[:, None]
+        reversed_readout_idx = torch.where(kheader.acq_info.flags.flatten() & AcqFlags.ACQ_IS_REVERSE.value)[0]
```
fzimmermann89: consider using a  boolean mask instead of torch.where:
```
reversed_readout_mask = kheader.acq_info.flags.flatten() & AcqFlags.ACQ_IS_REVERSE.value
 k0[reversed_readout_mask] = torch.flip(k0[reversed_readout_mask ])
```
or similiar should do the same and is easier to read

src/mrpro/data/traj_calculators/_KTrajectoryCalculator.py
```diff
@@ -57,19 +58,21 @@ def _kfreq(self, kheader: KHeader) -> torch.Tensor:
...
+        # pytorch does not yet allow for negative indexing. torch.flip makes a copy of the data. providing a reversed
+        # index should be fastest
+        reversed_index = torch.linspace(n_k0 - 1, 0, n_k0, dtype=torch.int64)
+        k0[reversed_readout_idx, :] = k0[reversed_readout_idx[:, None], reversed_index[None, :]]
 
-        # Calculate points along readout
-        n_k0 = int(n_samples[0, 0, 0])
-        k0 = torch.linspace(0, n_k0 - 1, n_k0, dtype=torch.float32) - center_sample[0, 0, 0]
-        return k0
+        return k0.reshape(kheader.acq_info.center_sample.shape[:-1] + (n_k0,))
```
fzimmermann89: https://docs.astral.sh/ruff/rules/collection-literal-concatenation/ does not yet catch this, as shape and the tuple are of differnt type, but you can even consider instead doing
` k0.reshape(*kheader.acq_info.center_sample.shape[:-1], n_k0)`
as reshape takes in integers

fzimmermann89:
consider the commit a suggestion for the k0 calculation that saves a few steps

src/mrpro/data/traj_calculators/_KTrajectoryCalculator.py
fzimmermann89: I checked how the synax should acutally be, see commit.

src/mrpro/data/traj_calculators/_KTrajectoryCalculator.py
fzimmermann89: see commit

fzimmermann89:
Could you also write a sentence in #50 how you solve it?

src/mrpro/data/traj_calculators/_KTrajectoryCalculator.py
ckolbPTB: before we set encoding_limits.k0 based on kdata.data.shape, therefore encoding_limits.k0 always matched the data. For a random header this is not the case anymore and therefore the only valid trajectory is one of size 1.

ckolbPTB:
> Could you also write a sentence in https://github.com/PTB-MR/mrpro/issues/50 how you solve it?

done

fzimmermann89 (approved):

src/mrpro/data/_kdata/_KData.py
```diff
@@ -135,7 +135,13 @@ def from_file(
...
         # Fill k0 limits if they were set to zero / not set in the header
         if kheader.encoding_limits.k0.length == 1:
-            kheader.encoding_limits.k0 = Limits(0, kdata.shape[-1] - 1, kdata.shape[-1] // 2)
+            # The encodig limits should describe the encoded space of all readout lines. If we have two readouts with
+            # (number_of_samples, center_sample) of (100, 20) (e.g. partial Fourier in the negative k0 direction) and
+            # (100, 80) (e.g. partial Fourier in the positive k0 direction) then this should lead to encoding limits of
+            # [min=0, max=159, center=80]
+            max_center_sample = int(torch.max(kheader.acq_info.center_sample))
+            max_pos_k0_extend = int(torch.max(kheader.acq_info.number_of_samples - kheader.acq_info.center_sample))
+            kheader.encoding_limits.k0 = Limits(0, max_center_sample + max_pos_k0_extend - 1, max_center_sample)
```
schuenke, line 144: nice example of descriptive variable names helping to understand the code 👍

src/mrpro/data/traj_calculators/_KTrajectoryCalculator.py
```diff
@@ -57,18 +58,17 @@ def _kfreq(self, kheader: KHeader) -> torch.Tensor:
         ValueError
             Center sample has to be the same for each readout
```
schuenke: This should be removed

src/mrpro/data/traj_calculators/_KTrajectoryCalculator.py
```diff
@@ -57,18 +58,17 @@ def _kfreq(self, kheader: KHeader) -> torch.Tensor:
...
-            raise ValueError('Trajectory can only be calculated if each acquisition has the same center sample')
+        n_k0 = int(n_samples.item())
 
-        # Calculate points along readout
-        n_k0 = int(n_samples[0, 0, 0])
-        k0 = torch.linspace(0, n_k0 - 1, n_k0, dtype=torch.float32) - center_sample[0, 0, 0]
+        # Data can be obtained with standard or reversed readout (e.g. bipolar readout).
+        k0 = torch.linspace(0, n_k0 - 1, n_k0, dtype=torch.float32) - center_sample
+        # Data can be obtained with standard or reversed readout (e.g. bipolar readout).
+        reversed_readout_mask = (kheader.acq_info.flags[..., 0] & AcqFlags.ACQ_IS_REVERSE.value).bool()
```
schuenke, line 68: Just a question unrelated to the code: can `ACQ_IS_REVERSE ` be set using pulseq labels? Or would this only work with an additional meta ismrmrd file with pulseq sequences?

src/mrpro/data/traj_calculators/_KTrajectoryCalculator.py
```diff
@@ -78,8 +78,8 @@ class DummyTrajectory(KTrajectoryCalculator):
     Shape will fit to all data. Only used as dummy for testing.
     """
 
-    def __call__(self, header: KHeader) -> KTrajectory:
-        """Calculate dummy trajectory for given KHeader."""
-        kx = torch.zeros(1, 1, 1, header.encoding_limits.k0.length)
+    def __call__(self, header: KHeader) -> KTrajectory:  # noqa: ARG002
```
schuenke, line 79: why do we keep the header argument if we don't need it anymore?

schuenke (approved):

src/mrpro/data/traj_calculators/_KTrajectoryCalculator.py
ckolbPTB: Not sure what pulseq can set but this is based on a standard Siemens raw data flag, similiar to calibration data and so on.

src/mrpro/data/traj_calculators/_KTrajectoryCalculator.py
ckolbPTB: because otherwise the signature of the function does not match anymore

src/mrpro/data/traj_calculators/_KTrajectoryCalculator.py
fzimmermann89: Do we need this if we use a pulseq trajectory calculator?

Otherwise, I think it will not be set...

fzimmermann89: @schuenke
In the pulseq reference manual, label extension section, the binary flag REV is mentioned

Can you try if setting that flag in pulseq to 1 sets the reversed acquisition flag in the ismrmrd?

schuenke: Can try it later today, yeah

PR #223, opened 2024-03-14
Fix 221: Cartesian sampling op for random undersampling with varying pattern along other

### LINKED ISSUE #221: Bug: CartesianSamplingOp for different traj along other

fzimmermann89:
First commit adds a test that fails, showing the issue.

Second commit solves the issue.

@ckolbPTB : If you have actual data, please consider checking this.

...at some point we might want to refactor the trajectory creation in the tests ....

Closes #221

### DIFF
--- a/src/mrpro/operators/_CartesianSamplingOp.py
+++ b/src/mrpro/operators/_CartesianSamplingOp.py
@@ -66,8 +66,9 @@ class CartesianSamplingOp(LinearOperator):
             kz_idx = rearrange(torch.arange(ktraj_tensor.shape[-3]), 'kz->1 kz 1 1')
 
         # 1D indices into a flattened tensor.
-        kidx = (kz_idx * encoding_shape.y * encoding_shape.x + ky_idx * encoding_shape.x + kx_idx).flatten()
-        self.register_buffer('_fft_idx', kidx.unsqueeze(0))
+        kidx = kz_idx * encoding_shape.y * encoding_shape.x + ky_idx * encoding_shape.x + kx_idx
+        kidx = rearrange(kidx, '... kz ky kx -> ... 1 (kz ky kx)')
+        self.register_buffer('_fft_idx', kidx)
         # we can skip the indexing if the data is already sorted
         self._needs_indexing = not torch.all(torch.diff(kidx) == 1)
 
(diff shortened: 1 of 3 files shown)


fzimmermann89:
argh. I should have pushed twice to show that the test actually failed...

The tests failed with
```python
    def forward(self, x: torch.Tensor) -> tuple[torch.Tensor,]:
        """Forward operator which selects acquired k-space data from k-space.
    
        Parameters
        ----------
        x
            k-space with dimensions given by encoding_shape
    
        Returns
        -------
            k-space data in original (i.e. acquired) dimensions
        """
        if self._encoding_shape != SpatialDimension(*x.shape[-3:]):
            raise ValueError('k-space data shape mismatch')
    
        if not self._needs_indexing:
            return (x,)
    
        x_2d = rearrange(x, '... k2_enc k1_enc k0_enc -> (...) (k2_enc k1_enc k0_enc)')
        # take_along_dim does broadcast, so no need for extending here
        x_indexed = torch.take_along_dim(x_2d, self._fft_idx, dim=-1)
        # reshape to (... other coil, k2, k1, k0)
>       x_reshaped = x_indexed.reshape(x.shape[:-3] + self._kshape[-3:])
E       RuntimeError: shape '[2, 5, 20, 20, 60]' is invalid for input of size 480000

src/mrpro/operators/_CartesianSamplingOp.py:99: RuntimeError
```

src/mrpro/operators/_CartesianSamplingOp.py
```diff
@@ -92,9 +93,9 @@ def forward(self, x: torch.Tensor) -> tuple[torch.Tensor,]:
         if not self._needs_indexing:
             return (x,)
 
-        x_2d = rearrange(x, '... k2_enc k1_enc k0_enc -> (...) (k2_enc k1_enc k0_enc)')
+        x_kflat = rearrange(x, '... coil k2_enc k1_enc k0_enc -> ... coil (k2_enc k1_enc k0_enc)')
         # take_along_dim does broadcast, so no need for extending here
-        x_indexed = torch.take_along_dim(x_2d, self._fft_idx, dim=-1)
+        x_indexed = torch.take_along_dim(x_kflat, self._fft_idx, dim=-1)
         # reshape to (... other coil, k2, k1, k0)
         x_reshaped = x_indexed.reshape(x.shape[:-3] + self._kshape[-3:])
```
ckolbPTB, line 100: self._kshape is not really the shape of the k-space data but the shape of the trajectory. Should we rename it?

ckolbPTB (approved):

ckolbPTB:
> If you have actual data, please consider checking this.

I verified it with bipolar gradient data

src/mrpro/operators/_CartesianSamplingOp.py
fzimmermann89: Funny. I had the same idea initially but undid the change. 
 it _should_ be the shape of the k space, using the trajectory is just how we estimate the shape... 
So in the forward and adjoint it reads better with kspape...

Issue #97, opened 2023-10-14
move largest EW/EV estimation into function


fzimmermann89:
We could move the estimation of the largest eigenwert and eigenvector by poweriteration into an `mrpro.algorithm`-function.
This function will be used toestimate the condition of Operators as well as in the espirit and walsh csm calculation.

The signature could maybe be  (pseudocode)
```
def poweriteration(operator:Callable[[torch.tensor],torch.tensor]|, v0:torch.tensor, norm:Callable[[torch.Tensor,torch.Tensor]]|Sequence[int], n_iterations:int=10)->tuple[torch.Tensor,torch.Tensor]:
   
   if isinstance(norm,Sequence):
       norm = lambda x: l2 norm along norm dimensions
   for i in range(n_iterations):
      ...
   return eigenvector, eigenvalue
```

ckolbPTB:
operator norm added in #224 EW/EV calculation only a few lines of code when optimised for a specific problem. general form too much overhead

PR #224, opened 2024-03-15
power iteration and operator-norm

koflera:
This implements a power iteration method to estimate the largest eigenvalue and largest eigenvector if a function. This can be further used to estimate the operator-norm of a LinearOperator.

Examples where this can be used are `iterative_walsh_inati` (see #210 and #97), the calculation of the operator norm for a simple Landweber method, the calculation of the upper bound for the step sizes to be used in pdhg, etc

### DIFF
--- a/src/mrpro/operators/LinearOperator.py
+++ b/src/mrpro/operators/LinearOperator.py
@@ -17,6 +17,7 @@
 from __future__ import annotations
 
 from abc import abstractmethod
+from collections.abc import Callable
 from collections.abc import Sequence
 from typing import overload
 
@@ -52,7 +53,10 @@ class LinearOperator(Operator[torch.Tensor, tuple[torch.Tensor]]):
         self,
         initial_value: torch.Tensor,
         dim: Sequence[int] | None,
-        n_max_iterations: int = 64,
+        max_iterations: int = 20,
+        relative_tolerance: float = 1e-4,
+        absolute_tolerance: float = 1e-5,
+        callback: Callable | None = None,
     ) -> torch.Tensor:
         """Power iteration for computing the operator norm of the linear operator.
 
@@ -60,7 +64,7 @@ class LinearOperator(Operator[torch.Tensor, tuple[torch.Tensor]]):
         ----------
         initial_value
             initial value to start the iteration; if the initial value contains a zero-vector for
-            one of the considered problems, the function throws and exception.
+            one of the considered problems, the function throws an value error.
         dim
             the dimensions of the tensors on which the operator operates.
             For example, for a matrix-vector multiplication example, a batched matrix tensor with shape (4,30,80,160),
@@ -74,42 +78,67 @@ class LinearOperator(Operator[torch.Tensor, tuple[torch.Tensor]]):
             domain of the considered operator (whose dimensionality is implicitly defined by choosing dim), such that
             the pointwise multiplication of the operator norm and elements of the domain (to be for example used
             in a Landweber iteration) is well-defined.
-        n_max_iterations , optional
-            maximal number of iterations
+        max_iterations
+            maximum number of iterations
+        relative_tolerance
+            absolute tolerance for the change of the operator-norm at each iteration;
+            if set to zero, the maximal number of iterations is the only stopping criterion used to stop
+            the power iteration
+        absolute_tolerance
+            absolute tolerance for the change of the operator-norm at each iteration;
+            if set to zero, the maximal number of iterations is the only stopping criterion used to stop
+            the power iteration
+        callback
+            user-provided function to be called at each iteration
 
         Returns
         -------
             an estimaton of the operator norm
         """
-        if n_max_iterations < 1:
-            raise Exception('The number of iterations should be larger than zero.')
+        if max_iterations < 1:
+            raise ValueError('The number of iterations should be larger than zero.')
 
         norm_initial_value = torch.linalg.vector_norm(initial_value, dim=dim, keepdim=True)
         if not (norm_initial_value > 0).all():
             if dim is None:
-                raise Exception('The initial value for the iteration should be different from the zero-vector.')
+                raise ValueError('The initial value for the iteration should be different from the zero-vector.')
             else:
-                raise Exception(
+                raise ValueError(
                     'Found at least one zero-vector as starting point. \
                     For each of the considered operators, the initial value for the iteration \
                     should be different from the zero-vector.'
                 )
 
         vector = initial_value
-        for _ in range(n_max_iterations):
+
+        # creaty dummy operator norm value that cannot be correct because by definition, the
+        # operator norm is a strictly positive number. This ensures that the first time the
+        # change between the old and the new estimate of the operator norm is non-zero and
+        # thus prevents the loop from exiting despite a non-correct estimate.
+        op_norm_old = torch.zeros(*tuple([1 for _ in range(vector.ndim)]))
+
+        for _ in range(max_iterations):
             # apply the operator to the vector
-            (vector,) = self.adjoint(*self(vector))
+            (vector_new,) = self.adjoint(*self(vector))
+
+            product = vector.real * vector_new.real
+            if vector.is_complex() and vector_new.is_complex():
+                product += vector.imag * vector_new.imag
+            dim = tuple(dim) if dim is not None else dim
+            op_norm = product.sum(dim, keepdim=True).sqrt()
+
+            if (absolute_tolerance > 0 or relative_tolerance > 0) and torch.isclose(
+                op_norm, op_norm_old, atol=absolute_tolerance, rtol=relative_tolerance
+            ).all():
+                break
 
             # normalize vector
-            vector /= torch.linalg.vector_norm(vector, dim=dim, keepdim=True)
+            vector_new /= torch.linalg.vector_norm(vector_new, dim=dim, keepdim=True)
+            vector = vector_new
+            op_norm_old = op_norm
 
-        (operator_vector,) = self.adjoint(*self(vector))
-
-        product = vector.real * operator_vector.real
-        if vector.is_complex() and operator_vector.is_complex():
-            product += vector.imag * operator_vector.imag
-        dim = tuple(dim) if dim is not None else dim
-        op_norm = product.sum(dim, keepdim=True).sqrt()
+            if callback is not None:
+                callback(op_norm)
 
         return op_norm
 
--- a/src/mrpro/operators/LinearOperator.py
+++ b/src/mrpro/operators/LinearOperator.py
@@ -98,6 +98,7 @@ class LinearOperator(Operator[torch.Tensor, tuple[torch.Tensor]]):
         if max_iterations < 1:
             raise ValueError('The number of iterations should be larger than zero.')
 
+        # check that the norm of the starting value is not zero
         norm_initial_value = torch.linalg.vector_norm(initial_value, dim=dim, keepdim=True)
         if not (norm_initial_value > 0).all():
             if dim is None:
@@ -109,6 +110,7 @@ class LinearOperator(Operator[torch.Tensor, tuple[torch.Tensor]]):
                     should be different from the zero-vector.'
                 )
 
+        # set initial value
         vector = initial_value
 
         # creaty dummy operator norm value that cannot be correct because by definition, the
@@ -122,11 +124,13 @@ class LinearOperator(Operator[torch.Tensor, tuple[torch.Tensor]]):
             # apply the operator to the vector
             (vector_new,) = self.adjoint(*self(vector))
 
+            # compute estimate of the operator norm
             product = vector.real * vector_new.real
             if vector.is_complex() and vector_new.is_complex():
                 product += vector.imag * vector_new.imag
             op_norm = product.sum(dim, keepdim=True).sqrt()
 
+            # check if stopping criterion is fulfillfed; if not continue the iteration
             if (absolute_tolerance > 0 or relative_tolerance > 0) and torch.isclose(
                 op_norm, op_norm_old, atol=absolute_tolerance, rtol=relative_tolerance
             ).all():
--- a/src/mrpro/operators/LinearOperator.py
+++ b/src/mrpro/operators/LinearOperator.py
@@ -81,9 +81,21 @@ class LinearOperator(Operator[torch.Tensor, tuple[torch.Tensor]]):
         -------
             an estimaton of the operator norm
         """
-        # check if initial value is exactly zero; if yes, set it to a random value
-        vector = torch.randn_like(initial_value) if (initial_value == 0.0).all() else initial_value
+        norm_initial_value = torch.linalg.vector_norm(initial_value, dim=dim, keepdim=True)
+        if dim is None:
+            if norm_initial_value == 0:
+                raise Exception('The initial value for the iteration should be different from the zero-vector.')
+        else:
+            if not (norm_initial_value > 0).all():
+                raise Exception(
+                    'Found at least one zero-vector as starting point. \
+                    For each of the considered operators, the initial value for the iteration \
+                    should be different from the zero-vector.'
+                )
+        if max_iterations < 1:
+            raise Exception('The number of iterations should be larger than zero.')
 
+        vector = initial_value
         for _ in range(max_iterations):
             # apply the operator to the vector
             (vector,) = self.adjoint(*self(vector))
--- a/src/mrpro/operators/LinearOperator.py
+++ b/src/mrpro/operators/LinearOperator.py
@@ -52,15 +52,15 @@ class LinearOperator(Operator[torch.Tensor, tuple[torch.Tensor]]):
         self,
         initial_value: torch.Tensor,
         dim: Sequence[int] | None,
-        max_iterations: int = 64,
+        n_max_iterations: int = 64,
     ) -> torch.Tensor:
         """Power iteration for computing the operator norm of the linear operator.
 
         Parameters
         ----------
         initial_value
-            initial value to start the iteration; if chosen exactly as zero-tensor,
-            we randomly generate an initial value.
+            initial value to start the iteration; if the initial value contains a zero-vector for
+            one of the considered problems, the function throws and exception.
         dim
             the dimensions of the tensors on which the operator operates.
             For example, for a matrix-vector multiplication example, a batched matrix tensor with shape (4,30,80,160),
@@ -74,29 +74,29 @@ class LinearOperator(Operator[torch.Tensor, tuple[torch.Tensor]]):
             domain of the considered operator (whose dimensionality is implicitly defined by choosing dim), such that
             the pointwise multiplication of the operator norm and elements of the domain (to be for example used
             in a Landweber iteration) is well-defined.
-        max_iterations, optional
+        n_max_iterations , optional
             maximal number of iterations
 
         Returns
         -------
             an estimaton of the operator norm
         """
+        if n_max_iterations < 1:
+            raise Exception('The number of iterations should be larger than zero.')
+
         norm_initial_value = torch.linalg.vector_norm(initial_value, dim=dim, keepdim=True)
-        if dim is None:
-            if norm_initial_value == 0:
+        if not (norm_initial_value > 0).all():
+            if dim is None:
                 raise Exception('The initial value for the iteration should be different from the zero-vector.')
-        else:
-            if not (norm_initial_value > 0).all():
+            else:
                 raise Exception(
                     'Found at least one zero-vector as starting point. \
                     For each of the considered operators, the initial value for the iteration \
                     should be different from the zero-vector.'
                 )
-        if max_iterations < 1:
-            raise Exception('The number of iterations should be larger than zero.')
 
         vector = initial_value
-        for _ in range(max_iterations):
+        for _ in range(n_max_iterations):
             # apply the operator to the vector
             (vector,) = self.adjoint(*self(vector))
 
--- a/src/mrpro/operators/LinearOperator.py
+++ b/src/mrpro/operators/LinearOperator.py
@@ -47,60 +47,52 @@ class LinearOperator(Operator[torch.Tensor, tuple[torch.Tensor]]):
         """Adjoint operator."""
         return AdjointLinearOperator(self)
 
-    def operator_norm(self,
-                        initial_value: torch.Tensor,
-                        dim: tuple | None = None,
-                        max_iterations: int = 4
-            ) -> torch.Tensor:
-            """Power iteration for computing the operator norm of the linear operator.
+    def operator_norm(
+        self, initial_value: torch.Tensor, dim: tuple | None = None, max_iterations: int = 64
+    ) -> torch.Tensor:
+        """Power iteration for computing the operator norm of the linear operator.
 
-            Parameters
-            ----------
-            initial_value
-                initial value to start the iteration; if chosen exactly as zero-tensor,
-                we randomly generate an initial value.
-            dim
-                dimensions referring to the dimensions of the tensors
-                on which the operator operates. For example, for a tensor with shape
-                (4,30,160,160) and dim = None, it is understood that
-                the the matrix representation of the considered operator
-                corresponds to a block diagonal operator and thus the algorithm returns
-                one sinle value. In contrast, if for example, dim=(-2,-1),
-                the algorithm returns computes a batched operator norm such that the
-                output can be
-            max_iterations, optional
-                maximal number of iterations, by default 256
-            tolerance, optional (TODO)
-                tolerance used to determine when to stop the iteration, by default 1e-6;
-                if set to zero, the maximal number of iterations is the only employed
-                stopping criterion
+        Parameters
+        ----------
+        initial_value
+            initial value to start the iteration; if chosen exactly as zero-tensor,
+            we randomly generate an initial value.
+        dim
+            dimensions referring to the dimensions of the tensors on which the operator operates.
+            For example, for a matrix-vector multiplication example,
+            with a batched matrix tensor with shape (4,30,80,160), tensors of shape (4,30,160) to be multiplied,
+            and dim = None, it is understood that the the matrix representation of the considered operator
+            corresponds to a block diagonal operator (with 4*30 matrices) and thus the algorithm returns
+            one single value. In contrast, if for example, dim=(-1), the algorithm computes a batched operator
+            norm and returns a tensor of shape (4,30,1) corresponding to the operator norms of the respective
+            matrices in the diagonal of the block-diagonal operator (if considered in matrix representation).
+            In any case, the output of the algorithm has the same shape as the elements of the domain
+            of the considered operator (whose dimensionality is implicitly defined by choosing dim)
+        max_iterations, optional
+            maximal number of iterations, by default 64
 
-            Returns
-            -------
-                an estimaton of the operator norm
-            """
+        Returns
+        -------
+            an estimaton of the operator norm
+        """
+        # check if initial value is exactly zero; if yes, set it to a random value
+        vector = torch.randn_like(initial_value) if (initial_value == 0.0).all() else initial_value
 
-            # check if initial value is exactly zero; if yes, set it to a random value
-            vector = torch.randn_like(initial_value) if (initial_value == 0.0).all() else initial_value
+        for _ in range(max_iterations):
+            # apply the operator to the vector
+            (vector,) = self.adjoint(*self(vector))
 
-            for _ in range(max_iterations):
+            # normalize vector
+            vector /= torch.linalg.vector_norm(vector, dim=dim, keepdim=True)
 
-                #apply the operator to the vector
-                (vector,) = self.adjoint(*self(vector))
+        (operator_vector,) = self.adjoint(*self(vector))
 
-                #normalize vector
-                vector /= vector.norm(dim=dim, keepdim=True)
+        product = vector.real * operator_vector.real
+        if vector.is_complex() and operator_vector.is_complex():
+            product += vector.imag * operator_vector.imag
+        op_norm = product.sum(dim, keepdim=True).sqrt()
 
-            (operator_vector,) = self.adjoint(*self(vector))
-
-            product =vector.real * operator_vector.real
-            if vector.is_complex() and operator_vector.is_complex():
-                product +=  vector.imag*operator_vector.imag
-            op_norm = product.sum(dim, keepdim=True).sqrt()
-
-            return op_norm
-
-
+        return op_norm
 
     @overload  # type: ignore[override]
     def __matmul__(self, other: LinearOperator) -> LinearOperator: ...
--- a/src/mrpro/operators/LinearOperator.py
+++ b/src/mrpro/operators/LinearOperator.py
@@ -17,6 +17,7 @@
 from __future__ import annotations
 
 from abc import abstractmethod
+from collections.abc import Sequence
 from typing import overload
 
 import torch
@@ -48,7 +49,10 @@ class LinearOperator(Operator[torch.Tensor, tuple[torch.Tensor]]):
         return AdjointLinearOperator(self)
 
     def operator_norm(
-        self, initial_value: torch.Tensor, dim: tuple | None = None, max_iterations: int = 64
+        self,
+        initial_value: torch.Tensor,
+        dim: Sequence[int] | None,
+        max_iterations: int = 64,
     ) -> torch.Tensor:
         """Power iteration for computing the operator norm of the linear operator.
 
@@ -58,18 +62,20 @@ class LinearOperator(Operator[torch.Tensor, tuple[torch.Tensor]]):
             initial value to start the iteration; if chosen exactly as zero-tensor,
             we randomly generate an initial value.
         dim
-            dimensions referring to the dimensions of the tensors on which the operator operates.
-            For example, for a matrix-vector multiplication example,
-            with a batched matrix tensor with shape (4,30,80,160), tensors of shape (4,30,160) to be multiplied,
-            and dim = None, it is understood that the the matrix representation of the considered operator
-            corresponds to a block diagonal operator (with 4*30 matrices) and thus the algorithm returns
-            one single value. In contrast, if for example, dim=(-1), the algorithm computes a batched operator
+            the dimensions of the tensors on which the operator operates.
+            For example, for a matrix-vector multiplication example, a batched matrix tensor with shape (4,30,80,160),
+            input tensors of shape (4,30,160) to be multiplied, and dim = None, it is understood that the the
+            matrix representation of the operator corresponds to a block diagonal operator (with 4*30 matrices)
+            and thus the algorithm returns a tensor of shape (1,1,1) containing one single value.
+            In contrast, if for example, dim=(-1,), the algorithm computes a batched operator
             norm and returns a tensor of shape (4,30,1) corresponding to the operator norms of the respective
             matrices in the diagonal of the block-diagonal operator (if considered in matrix representation).
-            In any case, the output of the algorithm has the same shape as the elements of the domain
-            of the considered operator (whose dimensionality is implicitly defined by choosing dim)
+            In any case, the output of the algorithm has the same number of dimensions as the elements of the
+            domain of the considered operator (whose dimensionality is implicitly defined by choosing dim), such that
+            the pointwise multiplication of the operator norm and elements of the domain (to be for example used
+            in a Landweber iteration) is well-defined.
         max_iterations, optional
-            maximal number of iterations, by default 64
+            maximal number of iterations
 
         Returns
         -------
@@ -90,6 +96,7 @@ class LinearOperator(Operator[torch.Tensor, tuple[torch.Tensor]]):
         product = vector.real * operator_vector.real
         if vector.is_complex() and operator_vector.is_complex():
             product += vector.imag * operator_vector.imag
+        dim = tuple(dim) if dim is not None else dim
         op_norm = product.sum(dim, keepdim=True).sqrt()
 
         return op_norm
--- a/src/mrpro/operators/LinearOperator.py
+++ b/src/mrpro/operators/LinearOperator.py
@@ -117,6 +117,7 @@ class LinearOperator(Operator[torch.Tensor, tuple[torch.Tensor]]):
         # thus prevents the loop from exiting despite a non-correct estimate.
         op_norm_old = torch.zeros(*tuple([1 for _ in range(vector.ndim)]))
 
+        dim = tuple(dim) if dim is not None else dim
         for _ in range(max_iterations):
             # apply the operator to the vector
             (vector_new,) = self.adjoint(*self(vector))
@@ -124,7 +125,6 @@ class LinearOperator(Operator[torch.Tensor, tuple[torch.Tensor]]):
             product = vector.real * vector_new.real
             if vector.is_complex() and vector_new.is_complex():
                 product += vector.imag * vector_new.imag
-            dim = tuple(dim) if dim is not None else dim
             op_norm = product.sum(dim, keepdim=True).sqrt()
 
             if (absolute_tolerance > 0 or relative_tolerance > 0) and torch.isclose(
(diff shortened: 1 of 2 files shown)


src/mrpro/algorithms/_power_iteration.py
```diff
@@ -0,0 +1,64 @@
...
+
+    for _ in range(max_iterations):
+        eigenvector_largest_old = eigenvector_largest
+
+        eigenvector_largest = operator(eigenvector_largest)[0]
+        if isinstance(operator, LinearOperator):
+            eigenvector_largest = eigenvector_largest[0]
+
+        eigenvalue_largest = torch.vdot(eigenvector_largest_old.flatten(), eigenvector_largest.flatten())
+        eigenvector_largest /= torch.linalg.norm(eigenvector_largest)
```
fzimmermann89: todo: norm should be configurable

src/mrpro/algorithms/_power_iteration.py
```diff
@@ -0,0 +1,64 @@
...
+
+        if torch.linalg.norm(eigenvector_largest - eigenvector_largest_old) < tolerance:
```
fzimmermann89: restructure, remove _old?

src/mrpro/algorithms/_power_iteration.py
```diff
@@ -0,0 +1,64 @@
...
+            break
+
+    return eigenvalue_largest, eigenvector_largest
```
fzimmermann89: remove _largest

src/mrpro/algorithms/_power_iteration.py
```diff
@@ -0,0 +1,64 @@
...
+        tolerance used to determine when to stop the iteration, by default 1e-6;
+        if set to zero, the maximal number of iterations is the only employed
+        stopping criterion
+
+    Returns
+    -------
+        the largest eigenvalue and eigenvector of the operator
+    """
+    # check if initial value is exactly zero; if yes, set it to a random value
+    if (initial_value == 0.0).all():
```
fzimmermann89: why?

fzimmermann89: this can be done outside
this will not match device and dtype

src/mrpro/algorithms/_power_iteration.py
```diff
@@ -0,0 +1,64 @@
...
+        eigenvector_largest = torch.randn_like(initial_value)
+    else:
+        eigenvector_largest = initial_value / torch.linalg.norm(initial_value)
+
+    for _ in range(max_iterations):
+        eigenvector_largest_old = eigenvector_largest
+
+        eigenvector_largest = operator(eigenvector_largest)[0]
```
fzimmermann89: (operator_vector,) =

src/mrpro/algorithms/_power_iteration.py
```diff
@@ -0,0 +1,64 @@
...
+        if isinstance(operator, LinearOperator):
```
fzimmermann89: move this outside the loop

src/mrpro/algorithms/_power_iteration.py
```diff
@@ -0,0 +1,64 @@
...
+            eigenvector_largest = eigenvector_largest[0]
+
+        eigenvalue_largest = torch.vdot(eigenvector_largest_old.flatten(), eigenvector_largest.flatten())
```
fzimmermann89: move outside loop

fzimmermann89: eigenvalue = norm(operator_vector) ?

fzimmermann89:
maybe we should talk about what a good power iteration function for our code is.

the eigenvector estimation that we need in most of the algorithms (walsh, espirit, etc) is just 3 lines. to use a separate function, we need to define the batched application of the matrix as a function, and the norm as a function, than call the power iteration function. 
To further calculate the eigenvalue, we need not only a norm, but also a scalar product.

Maybe we only have an operator norm on linear operators and continue to  write the power iteration for eigenvector estimation whenever we need it?

For the operator norm, there is still the questions how to define the norm/scalar product with regards to batching

so either we add a complicated logic to describe batching dimensions, or we only provide a operator_norm function outside the operator class and we have to provide the appropriate scalar product whenever you want to use it to calculate the norm of an operator.

koflera:
> maybe we should talk about what a good power iteration function for our code is.
> 
> the eigenvector estimation that we need in most of the algorithms (walsh, espirit, etc) is just 3 lines. to use a separate function, we need to define the batched application of the matrix as a function, and the norm as a function, than call the power iteration function. To further calculate the eigenvalue, we need not only a norm, but also a scalar product.
> 
> Maybe we only have an operator norm on linear operators and continue to write the power iteration for eigenvector estimation whenever we need it?
> 
> For the operator norm, there is still the questions how to define the norm/scalar product with regards to batching
> 
> so either we add a complicated logic to describe batching dimensions, or we only provide a operator_norm function outside the operator class and we have to provide the appropriate scalar product whenever you want to use it to calculate the norm of an operator.

What do you mean by "defining the norm with regards to batching dimensions"? I think this notion is not needed for computing the operator-norm. Why would one want to use a batch-size greater than one for computing it?

For other use cases, I would say that the appropriate way to compute the norm of a batch of tensors is to compute it for all the batched tensors and to return as many different values as there were tensors in the batch.

fzimmermann89:
We can have an arbitrary number of leading dimensions. Which of these is a batching dimension and which is a dimension the operator operates on is not defined. 
So either we need to define this for each operator, and pass it as a tuple of dimensions to the power iteration. OR we define the norm outside of the power iteration as a callable for each operator, and use it in the power iteration.

koflera:
ok, i see. However, I am still not sure whether this problem is really related to the computation of the operator norm or actually more general. Actually, for the computation of the operator norm, it is sufficient to work with a single element of the domain.

However, I see the point. Your second suggestion to me seems to go in the direction of how ODL handles things like that. 
There, operators have a domain and a range and one could associate an inner product to the respective spaces. 

But probably this requires a longer discussion.

koflera:
But have we actually ever used another space rather than R^N or C^N equipped with the Euclidean norms induced by the "standard" inner product?

fzimmermann89:
It is mostly which dimensions are batch dimensions. I would prefer not to add odl-like domain and range to all operators, this makes using them really painful and unpythonic.
Also, we currently use the power iterations on coverance matrices etc. So if we want to have one single function to use for these cases, we should have the norm as a callable parameter.

src/mrpro/algorithms/_power_iteration.py
koflera: to avoid the division by the zero-norm in the iteration

tests/operators/test_operator_norm.py
```diff
@@ -0,0 +1,96 @@
...
+from math import prod
+
+import torch
+from mrpro.operators import EinsumOp
+from mrpro.operators import FastFourierOp
+
+from tests import RandomGenerator
+
+
+def test_correct_result():
```
fzimmermann89: test_operatornorm_result?

tests/operators/test_operator_norm.py
```diff
@@ -0,0 +1,96 @@
...
+    matrix1 = torch.tensor([[2.0, 1, 0], [1.0, 2.0, 1.0], [0.0, 1.0, 2.0]])
+    operator1 = EinsumOp(matrix1, 'y x, x-> y')
+    random_vector1 = random_generator.float32_tensor(matrix1.shape[1])
+    operator1_norm_est = operator1.operator_norm(random_vector1, dim=None, max_iterations=32)
+    operator1_norm_true = 2 * (1 + cos(pi / 4))  # approximately 3.41421...
+    torch.testing.assert_close(operator1_norm_est.item(), operator1_norm_true, atol=1e-4, rtol=1e-4)
+
+    # test with Fast Fourier Operator (has norm 1 since norm="ortho" is used)
+    dim = (-3, -2, -1)
+    fouirer_op = FastFourierOp(dim=dim)
```
fzimmermann89: typo

tests/operators/test_operator_norm.py
fzimmermann89: split this into two tests

tests/operators/test_operator_norm.py
```diff
@@ -0,0 +1,96 @@
...
+    """Test if the implementation yields the correct result for different choices
+    of operators with known operator-norm."""
+    random_generator = RandomGenerator(seed=0)
+
+    # test with a 3x3 matrix with known largest eigenvalue
+    matrix1 = torch.tensor([[2.0, 1, 0], [1.0, 2.0, 1.0], [0.0, 1.0, 2.0]])
+    operator1 = EinsumOp(matrix1, 'y x, x-> y')
+    random_vector1 = random_generator.float32_tensor(matrix1.shape[1])
+    operator1_norm_est = operator1.operator_norm(random_vector1, dim=None, max_iterations=32)
+    operator1_norm_true = 2 * (1 + cos(pi / 4))  # approximately 3.41421...
```
fzimmermann89: 2+sqrt(2) ?

fzimmermann89: (why is there a cos in the operator norm)

tests/operators/test_operator_norm.py
```diff
@@ -0,0 +1,96 @@
...
+    # test with Fast Fourier Operator (has norm 1 since norm="ortho" is used)
+    dim = (-3, -2, -1)
+    fouirer_op = FastFourierOp(dim=dim)
+
+    random_image = torch.zeros(
+        4, 4, 8, 16, dtype=torch.complex64
+    )  # also tests that the initial value is set to a random value
+    fouirer_op_norm_batched = fouirer_op.operator_norm(random_image, dim=dim, max_iterations=64)
+    fourier_op_norm_non_batched = fouirer_op.operator_norm(random_image, dim=None, max_iterations=64)
+    fourier_op_norm_true = 1.0
```
fzimmermann89, line 117: move this under the comment that says that the norm is 1.

tests/operators/test_operator_norm.py
```diff
@@ -0,0 +1,96 @@
...
+    """Test if the batched calculation of the operator-norm works on a simple
+    matrix-vector multiplication example."""
+    random_generator = RandomGenerator(seed=0)
+    input_shape = (2, 4, 8, 8)
+
+    # dimensions which define the dimensionality of the considered vector space
+    dim = (-1,)
+
+    # create a tensor to be identified as 2 * 4 (=8) 8x8 square matrices
+    matrix1 = torch.arange(prod(input_shape)).reshape(input_shape).to(torch.float32)
```
fzimmermann89, line 164: why not random?

tests/operators/test_operator_norm.py
```diff
@@ -0,0 +1,96 @@
...
+    # the batched matrix-vector multiplication
+    operator1 = EinsumOp(matrix1, 'other1 other2 y x, other1 other2 x-> other1 other2 y')
+
+    random_vector1 = random_generator.float32_tensor((input_shape[0], input_shape[1], input_shape[3]))
+
+    # compute batched and non batched operator norms
+    operator1_norm_batched = operator1.operator_norm(random_vector1, dim=dim, max_iterations=32)
+    operator1_norm_non_batched = operator1.operator_norm(random_vector1, dim=None, max_iterations=32)
+
+    # using the fact that for a block-diagonal matrix, the eigenvalues are the list of
```
fzimmermann89: this is basis of the test, correct? maybe move this up to the test description

fzimmermann89 (commented):
see inline comments

tests/operators/test_operator_norm.py
koflera: just out of curiosity, why is it better to split it?

fzimmermann89: these are two completely different tests, nothing is shared between the two blocks. no reason to have them as one test.

tests/operators/test_operator_norm.py
koflera: good point. i had found this example in a numerics script and just took it from there as it was ;)

tests/operators/test_operator_norm.py
koflera: if the block-diagonal matrices are entirely randomly chosen, the largest eigenvalues of the respective matrices are pretty similar. by this construction, we make sure that it is clear which of the matrices is the "dominant" one and will have the largest of all largest eigenvalues. This eliminates the possibility that the test fails due to numerical issues instead of the  function not doing what it should do.

src/mrpro/operators/LinearOperator.py
```diff
@@ -47,6 +47,53 @@ def H(self) -> LinearOperator:  # noqa: N802
         """Adjoint operator."""
         return AdjointLinearOperator(self)
 
+    def operator_norm(
+        self, initial_value: torch.Tensor, dim: tuple | None = None, max_iterations: int = 64
```
ckolbPTB: dim should probably be Sequence[int]

src/mrpro/operators/LinearOperator.py
```diff
@@ -47,6 +47,53 @@ def H(self) -> LinearOperator:  # noqa: N802
...
+        dim
+            dimensions referring to the dimensions of the tensors on which the operator operates.
+            For example, for a matrix-vector multiplication example,
+            with a batched matrix tensor with shape (4,30,80,160), tensors of shape (4,30,160) to be multiplied,
+            and dim = None, it is understood that the the matrix representation of the considered operator
+            corresponds to a block diagonal operator (with 4*30 matrices) and thus the algorithm returns
+            one single value. In contrast, if for example, dim=(-1), the algorithm computes a batched operator
+            norm and returns a tensor of shape (4,30,1) corresponding to the operator norms of the respective
+            matrices in the diagonal of the block-diagonal operator (if considered in matrix representation).
+            In any case, the output of the algorithm has the same shape as the elements of the domain
```
ckolbPTB: Is this true? or do you mean the "**number of** elements of the domain"?

src/mrpro/operators/LinearOperator.py
```diff
@@ -47,6 +47,53 @@ def H(self) -> LinearOperator:  # noqa: N802
...
+            of the considered operator (whose dimensionality is implicitly defined by choosing dim)
+        max_iterations, optional
+            maximal number of iterations, by default 64
```
ckolbPTB: no default parameters

ckolbPTB: and no optional

src/mrpro/operators/LinearOperator.py
```diff
@@ -47,6 +47,53 @@ def H(self) -> LinearOperator:  # noqa: N802
...
+        initial_value
+            initial value to start the iteration; if chosen exactly as zero-tensor,
+            we randomly generate an initial value.
+        dim
+            dimensions referring to the dimensions of the tensors on which the operator operates.
+            For example, for a matrix-vector multiplication example,
+            with a batched matrix tensor with shape (4,30,80,160), tensors of shape (4,30,160) to be multiplied,
+            and dim = None, it is understood that the the matrix representation of the considered operator
+            corresponds to a block diagonal operator (with 4*30 matrices) and thus the algorithm returns
+            one single value. In contrast, if for example, dim=(-1), the algorithm computes a batched operator
```
ckolbPTB: a single value or a tensor of shape (1,1,1,1)?

tests/operators/test_operator_norm.py
```diff
@@ -0,0 +1,98 @@
...
+from tests import RandomGenerator
+
+
+def test_operator_norm_result():
+    """Test if the implementation yields the correct result for different choices
+    of operators with known operator-norm."""
+    random_generator = RandomGenerator(seed=0)
+
+    # test with a 3x3 matrix with known largest eigenvalue
+    matrix1 = torch.tensor([[2.0, 1, 0], [1.0, 2.0, 1.0], [0.0, 1.0, 2.0]])
```
ckolbPTB: why matrix1 and not just matrix?

tests/operators/test_operator_norm.py
```diff
@@ -0,0 +1,98 @@
...
+    operator1 = EinsumOp(matrix1, 'y x, x-> y')
+    random_vector1 = random_generator.float32_tensor(matrix1.shape[1])
+    operator1_norm_est = operator1.operator_norm(random_vector1, dim=None, max_iterations=32)
+    operator1_norm_true = 2 + sqrt(2)  # approximately 3.41421...
+    torch.testing.assert_close(operator1_norm_est.item(), operator1_norm_true, atol=1e-4, rtol=1e-4)
+
+
+def test_fourier_operator_norm():
+    """Test with Fast Fourier Operator (has norm 1 since norm="ortho" is used);
+    # also tests that the initial value is set to a random value if chosen as zero."""
```
ckolbPTB: Would it make sense to have a test which ensures that the same norm value is calculated independent if you give a random value oor chose the input as 0?

src/mrpro/operators/LinearOperator.py
koflera: actually, of shape (1,1,1)... I have further elaborated the comments.

tests/operators/test_operator_norm.py
koflera: i think it's so trivial that probably not worthy to be tested. Also, the fact that the test passes although we explicitly set zero as the initial value covers the case imho.

src/mrpro/operators/LinearOperator.py
koflera: mypy doesn't seem to like it

fzimmermann89: then you are doing something wrong ;)

src/mrpro/operators/LinearOperator.py
```diff
@@ -47,6 +48,59 @@ def H(self) -> LinearOperator:  # noqa: N802
...
+        initial_value: torch.Tensor,
+        dim: Sequence[int] | None,
+        max_iterations: int = 64,
+    ) -> torch.Tensor:
+        """Power iteration for computing the operator norm of the linear operator.
+
+        Parameters
+        ----------
+        initial_value
+            initial value to start the iteration; if chosen exactly as zero-tensor,
```
fzimmermann89: I really dont like this.

If I want to use a random tensor, I can create it on my own.
Maybe just throw an exception if in l.92 torch.linalg.vector_norm(vector, dim=dim, keepdim=True) is zero?

src/mrpro/operators/LinearOperator.py
```diff
@@ -47,6 +48,59 @@ def H(self) -> LinearOperator:  # noqa: N802
...
+            maximal number of iterations
+
+        Returns
+        -------
+            an estimaton of the operator norm
+        """
+        # check if initial value is exactly zero; if yes, set it to a random value
+        vector = torch.randn_like(initial_value) if (initial_value == 0.0).all() else initial_value
+
+        for _ in range(max_iterations):
```
fzimmermann89, line 123: should we raise an exception if max_iterations is smaller then 1?

tests/operators/test_operator_norm.py
```diff
@@ -0,0 +1,98 @@
...
+    torch.testing.assert_close(operator_norm_est.item(), operator_norm_true, atol=1e-4, rtol=1e-4)
+
+
+def test_fourier_operator_norm():
+    """Test with Fast Fourier Operator (has norm 1 since norm="ortho" is used);
+    # also tests that the initial value is set to a random value if chosen as zero."""
+    dim = (-3, -2, -1)
+    fourier_op = FastFourierOp(dim=dim)
+
+    random_image = torch.zeros(4, 4, 8, 16, dtype=torch.complex64)
```
fzimmermann89: this is not random

tests/operators/test_operator_norm.py
```diff
@@ -0,0 +1,98 @@
...
+    matrix2 = torch.block_diag(*[matrix1[kb, kt, ...] for kb in range(input_shape[0]) for kt in range(input_shape[1])])
+
+    # construct a linear operator from the second matrix; the linear operator implements
+    # the multiplication of the block-diagonal matrix with a 2*4*8*8 = 512-dimensional vector
+    operator2 = EinsumOp(matrix2, 'y x, x-> y')
+    random_vector2 = random_generator.float32_tensor(matrix2.shape[1])
+    operator2_norm_non_batched = operator2.operator_norm(random_vector2, dim=None, max_iterations=32)
+
+    # test whether the operator-norm calculated from the first operator and from the second
+    # one coincide
```
fzimmermann89: why coincide?
this should equal.

fzimmermann89: https://en.wikipedia.org/wiki/Mathematical_coincidence

tests/operators/test_operator_norm.py
```diff
@@ -0,0 +1,98 @@
...
+
+    # dimensions which define the dimensionality of the considered vector space
+    dim = (-1,)
+
+    # create a tensor to be identified as 2 * 4 (=8) 8x8 square matrices
+    matrix1 = torch.arange(prod(input_shape)).reshape(input_shape).to(torch.float32)
+
+    # construct a linear operator from the first matrix; the linear operator implements
+    # the batched matrix-vector multiplication
+    operator1 = EinsumOp(matrix1, 'other1 other2 y x, other1 other2 x-> other1 other2 y')
```
fzimmermann89: just a hint: you can use '... y x, ... x->... y'

fzimmermann89 (changes requested):

src/mrpro/operators/LinearOperator.py
```diff
@@ -47,6 +48,71 @@ def H(self) -> LinearOperator:  # noqa: N802
...
+        initial_value: torch.Tensor,
+        dim: Sequence[int] | None,
+        max_iterations: int = 64,
+    ) -> torch.Tensor:
+        """Power iteration for computing the operator norm of the linear operator.
+
+        Parameters
+        ----------
+        initial_value
+            initial value to start the iteration; if chosen exactly as zero-tensor,
```
ckolbPTB: adapt to new error message

src/mrpro/operators/LinearOperator.py
```diff
@@ -47,6 +48,71 @@ def H(self) -> LinearOperator:  # noqa: N802
...
+            in a Landweber iteration) is well-defined.
+        max_iterations, optional
+            maximal number of iterations
+
+        Returns
+        -------
+            an estimaton of the operator norm
+        """
+        norm_initial_value = torch.linalg.vector_norm(initial_value, dim=dim, keepdim=True)
+        if dim is None:
```
ckolbPTB: why is `if dim is None` treated separately? Should i`f not (norm_initial_value > 0).all()` also work for dim = None?

src/mrpro/operators/LinearOperator.py
```diff
@@ -47,6 +48,71 @@ def H(self) -> LinearOperator:  # noqa: N802
...
+
+            # normalize vector
+            vector /= torch.linalg.vector_norm(vector, dim=dim, keepdim=True)
+
+        (operator_vector,) = self.adjoint(*self(vector))
+
+        product = vector.real * operator_vector.real
+        if vector.is_complex() and operator_vector.is_complex():
+            product += vector.imag * operator_vector.imag
+        dim = tuple(dim) if dim is not None else dim
```
ckolbPTB: Is tuple(dim) really needed here? does .sum() not accept any Sequence element?

tests/operators/test_operator_norm.py
```diff
@@ -0,0 +1,142 @@
...
+
+import pytest
+import torch
+from mrpro.operators import EinsumOp
+from mrpro.operators import FastFourierOp
+
+from tests import RandomGenerator
+
+
+def test_operator_norm_illegal_max_iterations():
```
ckolbPTB: maybe "illegal" -> "invalid"

src/mrpro/operators/LinearOperator.py
```diff
@@ -47,6 +48,71 @@ def H(self) -> LinearOperator:  # noqa: N802
...
+            if norm_initial_value == 0:
+                raise Exception('The initial value for the iteration should be different from the zero-vector.')
+        else:
+            if not (norm_initial_value > 0).all():
+                raise Exception(
+                    'Found at least one zero-vector as starting point. \
+                    For each of the considered operators, the initial value for the iteration \
+                    should be different from the zero-vector.'
+                )
+        if max_iterations < 1:
```
ckolbPTB: I would move this up to the very beginning, as this is a simple test and should be done first before calculating any vector norms.

tests/operators/test_operator_norm.py
```diff
@@ -0,0 +1,142 @@
...
+    # test with a 3x3 matrix with known largest eigenvalue
+    matrix = torch.tensor([[2.0, 1, 0], [1.0, 2.0, 1.0], [0.0, 1.0, 2.0]])
+    operator = EinsumOp(matrix, 'y x, x-> y')
+    random_vector = random_generator.float32_tensor(matrix.shape[1])
+
+    with pytest.raises(Exception, match='zero'):
+        operator.operator_norm(random_vector, dim=None, max_iterations=0)
+
+
+def test_operator_norm_illegal_initial_value():
```
ckolbPTB: maybe "illegal" -> "invalid"

tests/operators/test_operator_norm.py
ckolbPTB: Some of your tests need doc strings

ckolbPTB (approved):

src/mrpro/operators/LinearOperator.py
koflera: in principle yes. I just thought that separating the two cases and printing different messages would be helpful for the user. For the case where one specfiies the dim along to which compute the operator-norm, it could be the case that the initial value is only exactly zero for one of the considered operators that are implicitly defined by the choice of dim. So I found that having different error messages would be more informative.

ckolbPTB: could we then have a more informative error message or maybe even to different error messages depending on `dim is None` or not?

src/mrpro/operators/LinearOperator.py
```diff
@@ -47,6 +48,71 @@ def H(self) -> LinearOperator:  # noqa: N802
         """Adjoint operator."""
         return AdjointLinearOperator(self)
 
+    def operator_norm(
+        self,
+        initial_value: torch.Tensor,
+        dim: Sequence[int] | None,
+        max_iterations: int = 64,
```
ckolbPTB: should actually be n_max_iterations but it also different in cg and adams/lbfgs...

src/mrpro/operators/LinearOperator.py
koflera: but currently we DO have two different error messages, no?

ckolbPTB: yes but the trigger for the error should be the same, independent if it is None or not. Just the info to the user should be different. Currently the trigger is also different, which I find confusing.

src/mrpro/operators/LinearOperator.py
koflera: I had tried and got this error:

`src/mrpro/operators/LinearOperator.py:111: error: Argument 1 to "sum" of "TensorBase" has incompatible type "Sequence[int] | None"; expected "int | Size | list[int] | tuple[int, ...] | None"  [arg-type]`

koflera: I will leave it as it is for the moment, but I have addressed the rest before my leave for the next few days.

src/mrpro/operators/LinearOperator.py
```diff
@@ -47,6 +48,71 @@ def H(self) -> LinearOperator:  # noqa: N802
...
+            matrix representation of the operator corresponds to a block diagonal operator (with 4*30 matrices)
+            and thus the algorithm returns a tensor of shape (1,1,1) containing one single value.
+            In contrast, if for example, dim=(-1,), the algorithm computes a batched operator
+            norm and returns a tensor of shape (4,30,1) corresponding to the operator norms of the respective
+            matrices in the diagonal of the block-diagonal operator (if considered in matrix representation).
+            In any case, the output of the algorithm has the same number of dimensions as the elements of the
+            domain of the considered operator (whose dimensionality is implicitly defined by choosing dim), such that
+            the pointwise multiplication of the operator norm and elements of the domain (to be for example used
+            in a Landweber iteration) is well-defined.
+        n_max_iterations , optional
```
fzimmermann89: remove optional

src/mrpro/operators/LinearOperator.py
```diff
@@ -47,6 +48,71 @@ def H(self) -> LinearOperator:  # noqa: N802
         """Adjoint operator."""
         return AdjointLinearOperator(self)
 
+    def operator_norm(
+        self,
+        initial_value: torch.Tensor,
+        dim: Sequence[int] | None,
+        n_max_iterations: int = 64,
```
fzimmermann89: i prefered the previous max_iterations.

src/mrpro/operators/LinearOperator.py
```diff
@@ -47,6 +48,71 @@ def H(self) -> LinearOperator:  # noqa: N802
...
+        """
+        if n_max_iterations < 1:
+            raise Exception('The number of iterations should be larger than zero.')
+
+        norm_initial_value = torch.linalg.vector_norm(initial_value, dim=dim, keepdim=True)
+        if not (norm_initial_value > 0).all():
+            if dim is None:
+                raise Exception('The initial value for the iteration should be different from the zero-vector.')
+            else:
+                raise Exception(
```
fzimmermann89: Runtime? ValueError?

tests/operators/test_operator_norm.py
```diff
@@ -0,0 +1,144 @@
...
+    random_generator = RandomGenerator(seed=0)
+    input_shape = (2, 4, 8, 8)
+    vector_shape = (input_shape[0], input_shape[1], input_shape[3])
+
+    # create a tensor to be identified as 2 * 4 (=8) 8x8 square matrices
+    matrix = random_generator.float32_tensor(size=input_shape)
+
+    # construct a linear operator from the first matrix; the linear operator implements
+    # the batched matrix-vector multiplication
+    operator = EinsumOp(matrix, 'other1 other2 y x, other1 other2 x-> other1 other2 y')
```
fzimmermann89: ... y x,

tests/operators/test_operator_norm.py
```diff
@@ -0,0 +1,144 @@
...
+
+    # dimensions which define the dimensionality of the considered vector space
+    dim = (-1,)
+
+    # create a tensor to be identified as 2 * 4 (=8) 8x8 square matrices
+    matrix1 = torch.arange(prod(input_shape)).reshape(input_shape).to(torch.float32)
+
+    # construct a linear operator from the first matrix; the linear operator implements
+    # the batched matrix-vector multiplication
+    operator1 = EinsumOp(matrix1, 'other1 other2 y x, other1 other2 x-> other1 other2 y')
```
fzimmermann89: "... y x, ... x-> ... y"

can be used for all the einsumops in the tests.

src/mrpro/operators/LinearOperator.py
fzimmermann89: this seems like a really high default, is this usually required?
I would assume we get a good estimate after 20 iterations or so

fzimmermann89:
Two questions:
 Is there a simple way to specify a relative difference between iterates as a stopping criterion  - without having to add much computation in each iteration, i.e. can this be written to only check for changes of the vector ?
 If yes, it would be good to include this (and rename n_max_iterations back to max_iterations)
 
 Otherwise, I would rename _n_max_iterations_ to _n_iterations_ -- as it is the fixed number of iteratinos.

@ckolbPTB : n_max_iterations for cg et.c seems quiteredundant. I would prefer _max_iterations_ if it is a bound and _n_iterations_ if it is the actually number of iterations.

ckolbPTB:
> @ckolbPTB : n_max_iterations for cg et.c seems quiteredundant. I would prefer _max_iterations_ if it is a bound and _n_iterations_ if it is the actually number of iterations.

Fine by me as long is we use it consistenly for cg (currently `max_iterations`), lbfgs/adam (currently `max_iter`). Also in CsmData/iterative_walsh we use `power_iterations` which should then probably be `n_power_iterations`

koflera:
> > @ckolbPTB : n_max_iterations for cg et.c seems quiteredundant. I would prefer _max_iterations_ if it is a bound and _n_iterations_ if it is the actually number of iterations.
> 
> Fine by me as long is we use it consistenly for cg (currently `max_iterations`), lbfgs/adam (currently `max_iter`). Also in CsmData/iterative_walsh we use `power_iterations` which should then probably be `n_power_iterations`

ok, so let's maybe agree that in general `n` stands for number, and we use the word `max` only if there an additional check that potentially allows the iteration to stop?

I.e. for CG we should either change to `n_iterations `or include a stopping criterion like checking the norm of the residual and use `n_max_iterations`.
For ADAM, we should use `n_iterations`, for LBFGS we should check what is exactly done.

fzimmermann89:
CG: max_iterations (tolerance criterion)
Lbfgs: max_iterations (tolerance criterion)
Adam: n_iterations
Here: n_iterstions

Never n_max_iterstions. :)

ckolbPTB:
iterative_walsh/CsmData: n_iterations

fzimmermann89:
> iterative_walsh/CsmData: n_iterations

For Walsh, I would even consider removing that option and either defining a constant  in the file and using that ,  checking for convergence, or using an existing pytorch funxtion. This only finds the largest ev of the matrix. 

As described in the paper, Walsh is not an iterative algoritm. So having this option there is either useless or confusing, is it is more of a technical detail within part of the algorithm..

koflera:
> CG: max_iterations (tolerance criterion) Lbfgs: max_iterations (tolerance criterion) Adam: n_iterations Here: n_iterstions
> 
> Never n_max_iterstions. :)

Fine for me.

src/mrpro/operators/LinearOperator.py
koflera: fine for me. I guess, however, that the convergence depends quite a lot on the considered operator. I tried with the nabla operator for `dim=(-4,-3,-2,-1)` and `dim=(-2,-1)` and the accuracy seems to be fine after 20 iterations.

koflera:
> Two questions: Is there a simple way to specify a relative difference between iterates as a stopping criterion - without having to add much computation in each iteration, i.e. can this be written to only check for changes of the vector ? If yes, it would be good to include this (and rename n_max_iterations back to max_iterations)
> 
> Otherwise, I would rename _n_max_iterations_ to _n_iterations_ -- as it is the fixed number of iteratinos.
> 
> @ckolbPTB : n_max_iterations for cg et.c seems quiteredundant. I would prefer _max_iterations_ if it is a bound and _n_iterations_ if it is the actually number of iterations.

I guess then we need something like a `vector_old` and `vector_new`, which, as far as I remember we did not want to have in the beginning.

fzimmermann89:
> > Two questions: Is there a simple way to specify a relative difference between iterates as a stopping criterion - without having to add much computation in each iteration, i.e. can this be written to only check for changes of the vector ? If yes, it would be good to include this (and rename n_max_iterations back to max_iterations)
> > Otherwise, I would rename _n_max_iterations_ to _n_iterations_ -- as it is the fixed number of iteratinos.
> > @ckolbPTB : n_max_iterations for cg et.c seems quiteredundant. I would prefer _max_iterations_ if it is a bound and _n_iterations_ if it is the actually number of iterations.
> 
> I guess then we need something like a `vector_old` and `vector_new`, which, as far as I remember we did not want to have in the beginning.

This is not the issue here.

I wondered if it is mathematically possible to say something about the difference between iterates of the norm, without actually calculating these iterates, and instead only use some criterion on vector and vector_old.
Or put differently, how does is the relative difference of operator norm iterates related to vector_old and vector? Is there a shortcut?

src/mrpro/operators/LinearOperator.py
koflera: Re: depending what one considers to be "fine" :D

I now included a test using the finite differences operator as well, and for dim corresponding to a 1D, 2D, 3D and 4D kernels, 50 iterations are not enough to pass the test for absolute and relative tolerance 1e-2.

fzimmermann89: Ok, so 64 it is..

koflera:
> > > Two questions: Is there a simple way to specify a relative difference between iterates as a stopping criterion - without having to add much computation in each iteration, i.e. can this be written to only check for changes of the vector ? If yes, it would be good to include this (and rename n_max_iterations back to max_iterations)
> > > Otherwise, I would rename _n_max_iterations_ to _n_iterations_ -- as it is the fixed number of iteratinos.
> > > @ckolbPTB : n_max_iterations for cg et.c seems quiteredundant. I would prefer _max_iterations_ if it is a bound and _n_iterations_ if it is the actually number of iterations.
> > 
> > 
> > I guess then we need something like a `vector_old` and `vector_new`, which, as far as I remember we did not want to have in the beginning.
> 
> This is not the issue here.
> 
> I wondered if it is mathematically possible to say something about the difference between iterates of the norm, without actually calculating these iterates, and instead only use some criterion on vector and vector_old. Or put differently, how does is the relative difference of operator norm iterates related to vector_old and vector? Is there a shortcut?

I now included a check similar as in ODL, see https://odlgroup.github.io/odl/_modules/odl/operator/oputils.html#power_method_opnorm and additional tests to test whether it is possible to exit the loop before exhausting the for-loop. Obviously, the cost somewhat increases because we calculate the inner product between the old vector and the new one, but I guess this should be ok.

src/mrpro/operators/LinearOperator.py
```diff
@@ -47,6 +49,99 @@ def H(self) -> LinearOperator:  # noqa: N802
...
+        op_norm_old = torch.zeros(*tuple([1 for _ in range(vector.ndim)]))
+
+        for _ in range(max_iterations):
+            # apply the operator to the vector
+            (vector_new,) = self.adjoint(*self(vector))
+
+            product = vector.real * vector_new.real
+            if vector.is_complex() and vector_new.is_complex():
+                product += vector.imag * vector_new.imag
+            dim = tuple(dim) if dim is not None else dim
```
fzimmermann89: move this out of the loop

src/mrpro/operators/LinearOperator.py
```diff
@@ -47,6 +49,99 @@ def H(self) -> LinearOperator:  # noqa: N802
...
+            op_norm = product.sum(dim, keepdim=True).sqrt()
+
+            if (absolute_tolerance > 0 or relative_tolerance > 0) and torch.isclose(
+                op_norm, op_norm_old, atol=absolute_tolerance, rtol=relative_tolerance
+            ).all():
+                break
+
+            # normalize vector
+            vector_new /= torch.linalg.vector_norm(vector_new, dim=dim, keepdim=True)
+            vector = vector_new
```
fzimmermann89: vector= vector_new/torch.linalg.vector_norm(vector_new, dim=dim, keepdim=True)
?

fzimmermann89:
btw this is not what i meant: i specifically asked if there is a way to avoid the additional computation of the norm in each loop-- > so the answer is no. **shrugs** nevermind

fzimmermann89 (commented):
¯\_(ツ)_/¯

koflera:
> btwm this is not what i meant: i specifically asked if there is a way to avoid the additional computation of the norm in each loop-- > so the answer is no. **shrugs** nevermind

I thought this is what you meant with "some criterion on vector and vector_old". What other criterion less expansive than the inner product between the two were you thinking of? 

In terms of application of the adjoint and the forward, which might be the potentially computationally intensive part depending on the operator, the code is the same as before.

src/mrpro/operators/LinearOperator.py
koflera: what do you mean by that?  if we move it out of the loop, we only compute it once, which does not allow to track the operator-norm. 
do you mean to move out the code and define some function `def inner_product(vector1, vector2)` and to call that inside the loop?

koflera: or do you mean only the part `dim = tuple(dim) if dim is not None else dim` ?

fzimmermann89: the latter :)

fzimmermann89 (approved):

PR #225, opened 2024-03-18
Fix #216

### LINKED ISSUE #216: Failing tests pass pipeline
If I am not mistaken, currently the pytest github action passes with a green checkmark if the tests fail to even execute, for example, due to an syntax error.

fzimmermann89:
Closes #216

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -41,7 +41,6 @@ jobs:
         run: |
           echo "final list of images with tag latest: ${{ steps.set-matrix.outputs.imagenames }}"
 
-
   test:
     name: Run tests and get coverage report
     needs: get_dockerfiles
@@ -69,6 +68,7 @@ jobs:
 
       - name: Run PyTest
         run: |
+          set -o pipefail
           pytest -n 4 -m "not cuda" --junitxml=pytest.xml --cov-report=term-missing:skip-covered --cov=mrpro | tee pytest-coverage.txt
 
       - name: Pytest coverage comment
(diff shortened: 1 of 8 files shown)


fzimmermann89:
@ckolbPTB @schuenke

ckolbPTB:
If the tests are not run there are missing files but thy are only treated as warnings rather than errors:
File "/__w/mrpro/mrpro/./pytest.xml" doesn't exist
No content found in file "/__w/mrpro/mrpro/./pytest-coverage.txt"

schuenke:
> If the tests are not run there are missing files but thy are only treated as warnings rather than errors: File "/__w/mrpro/mrpro/./pytest.xml" doesn't exist No content found in file "/__w/mrpro/mrpro/./pytest-coverage.txt"

IMO the step before ("Run PyTest) should already fail because of the `ModuleNotFoundError`. First google search wasn't helpful, but I will have a look.

schuenke:
I think I got it... the problem was that we piped together the `pytest... ` and `tee pytest-coverage.txt` command.

In general, when piping together several commands, the exit status of the pipeline is the exit status of the last command. In our case `tee pytest-coverage.txt`. If we want the pipeline to fail if any command fails, we need to use `set -o pipefail`.

I don't know why I had to specify that this step uses bash as shell, but might be required due to the docker container?! I dunno...

Anyway, I think it works as expected now.

fzimmermann89:
can we try a 'normal' test error?
just making sure we still get a report if one of the tests fail..

schuenke:
> can we try a 'normal' test error? just making sure we still get a report if one of the tests fail..

Good point. Actually it doesn't work as expected. 

Maybe useful to fix this:

- a failing test results in exit code 1
- the import-error results in exit code 4

Maybe we can specify that we want the pipeline to fail only if exit code is > 1 ?

schuenke:
I now check for the existence of the pytest.xml file in an additional step after running pytest. This should do the job!

Wanna have a last look @fzimmermann89 ?

fzimmermann89:
Looks reasonable. Let's see how it behaves in practice.

schuenke (approved):

PR #226, opened 2024-03-20
allow batch dimension in signal models

### LINKED ISSUE #202: WASABI/WASABITI shape tests not covering their purpose
Currently the `test_WASABI_signal_model_shape` and `test_WASABITI_signal_model_shape` tests don't fully cover their purpose because all spatial dimensions are 1 in all 4 (or rather 3, because one case is duplicated) tested cases. We should change this. Maybe the [test for the t1 models](https://github.com/PTB-MR/mrpro/blob/main/tests/operators/models/test_t1_models.py) can serve as an example. 

Furthermore, after merging #193, the signal models should accept `float` instead of `torch.Tensor` for all single-value arguments. Could adjust this as well.

In addition, missing docstrings and some explaining comments should be added.

Maybe @JoHa0811 can take care of this? Best would be after #193 is merged.

### LINKED ISSUE #227: Wrong shapes in wasabi test
https://github.com/PTB-MR/mrpro/blob/b5640d07ee25eff7952c05851d7fdf5595bc8112/tests/operators/models/test_wasabi.py#L6

Is this on purpose?
only one of the parameter maps is of the expected number of dimensions for us..

fzimmermann89:
Modifies the signal models for T1 mapping and wasabi, such that leading dimensions before the 'other' dimension are considered as batch dimension and not flattened.
Wasabiti is a todo.
Also removes some edgecases from molli resulting in nan-gradients, should result in same signal (needs to be checked)

### DIFF
--- a/src/mrpro/operators/models/_MOLLI.py
+++ b/src/mrpro/operators/models/_MOLLI.py
@@ -19,7 +19,7 @@ from mrpro.operators import SignalModel
 class MOLLI(SignalModel[torch.Tensor, torch.Tensor, torch.Tensor]):
     """Signal model for Modified Look-Locker inversion recovery (MOLLI)."""
 
-    def __init__(self, ti: torch.Tensor):
+    def __init__(self, ti: float | torch.Tensor):
         """Initialize MOLLI signal model for T1 mapping.
 
         Parameters
--- a/src/mrpro/operators/models/_MOLLI.py
+++ b/src/mrpro/operators/models/_MOLLI.py
@@ -29,6 +29,7 @@ class MOLLI(SignalModel[torch.Tensor, torch.Tensor, torch.Tensor]):
             with shape (time, ...)
         """
         super().__init__()
+        ti = torch.as_tensor(ti)
         self.ti = torch.nn.Parameter(ti, requires_grad=ti.requires_grad)
 
     def forward(self, a: torch.Tensor, b: torch.Tensor, t1: torch.Tensor) -> tuple[torch.Tensor,]:
--- a/src/mrpro/operators/models/_MOLLI.py
+++ b/src/mrpro/operators/models/_MOLLI.py
@@ -47,10 +47,12 @@ class MOLLI(SignalModel[torch.Tensor, torch.Tensor, torch.Tensor]):
         -------
             signal with dimensions ((... inv_times), coils, z, y, x)
         """
-        t1 = torch.where(t1 == 0, 1e-10, t1)
-        a = torch.where(a == 0 | torch.equal((b / a), torch.ones_like(a)), a + 1e-10, a)
-        ti = self.ti[(...,) + (None,) * (a.ndim)]
-        t1_star = t1 / ((b / a) - 1)
-        y = a - b * torch.exp(-(ti / (t1_star)))
-        res = rearrange(y, 't ... c z y x -> (... t) c z y x')
-        return (res,)
+        ti = self.ti[..., None, :, None, None, None, None]  # *other t=1, c, z, y, x
+        a = a.unsqueeze(-5)  # *other t=1, c, z, y, x
+        b = b.unsqueeze(-5)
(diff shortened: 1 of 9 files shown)


fzimmermann89:
The wasabi test fails.
This is because it tests with testdata that does not have an x,y,z dimension.
We should discuss what dataassumptions the signal models shall make and in which axes they should broadcast.
I.e. where  to put the time axes. Currently, it will put it before the coil dimension, counting from the back. 
So the input always needs to have coil z y x.
we could consider making the position of the new offset/timepoint axes a parameter of the init function.

ckolbPTB:
Discussion summary:
Signal dimension (e.g. inversion time) is added as additional axis as the first other dimension before coils. 
E.g. ti = (7,5) and t1 = (7,1,11,17,13) leads to signal = (7,5,1,11,17,13) [OUTDATED]
Handling of the other indices defined by AcqInfo (i.e. assigning the inversion times to the AcqIdx contrast) will be handled one level higher.

UPDATE:
Signal dimension (e.g. inversion time) is added as additional axis as the first dimension. 
E.g. ti = (5,7) and t1 = (7,1,11,17,13) leads to signal = (5,7,1,11,17,13). This also allows for signals which are not (...other, coils, z, y, x). E.g. t1 = (4, 3, 1) and ti = (5, 4, 3) will lead to signal (5, 4, 3, 1)

fzimmermann89:
@ckolbPTB 
Do we want to change the wasabiti and the wasabi/wasabiti tests as well?

src/mrpro/operators/models/_MOLLI.py
```diff
@@ -37,20 +38,23 @@ def forward(self, a: torch.Tensor, b: torch.Tensor, t1: torch.Tensor) -> tuple[t
...
+            with shape (time ... other, coils, z, y, x)
         """
-        t1 = torch.where(t1 == 0, 1e-10, t1)
-        a = torch.where(a == 0 | torch.equal((b / a), torch.ones_like(a)), a + 1e-10, a)
-        ti = self.ti[(...,) + (None,) * (a.ndim)]
-        t1_star = t1 / ((b / a) - 1)
-        y = a - b * torch.exp(-(ti / (t1_star)))
-        res = rearrange(y, 't ... c z y x -> (... t) c z y x')
-        return (res,)
+        delta_ndim = a.ndim - (self.ti.ndim - 1)  # -1 for time
```
fzimmermann89, line 55: what if a and b have different dimensions...?

fzimmermann89: This is not supported, correct? 
..and not worth supporting I would say..

src/mrpro/operators/models/_WASABI.py
```diff
@@ -74,26 +74,33 @@ def forward(
...
+        b1 = self.b1_nom * relative_b1
 
         signal = (
             c
             - d
             * (torch.pi * b1 * self.gamma * self.tp) ** 2
             * torch.sinc(self.tp * torch.sqrt((b1 * self.gamma) ** 2 + delta_x**2)) ** 2
         )
-        signal = rearrange(signal, 'offsets ... c z y x -> (... offsets) c z y x')
+        signal = rearrange(signal, '... other offsets c z y x ->.... (other offsets) c z y x')
```
fzimmermann89: todo

ckolbPTB:
> Do we want to change the wasabiti and the wasabi/wasabiti tests as well?

Yes, I will grab Patrick today and try to fix the WASABI/WASABITI stuff

src/mrpro/operators/models/_MOLLI.py
ckolbPTB: > ..and not worth supporting I would say..

Agreed.

ckolbPTB:
@schuenke could you have a final look at this and then I think it is ready to merge

schuenke (approved):

PR #229, opened 2024-03-21
variable shapes for parameters and ti

ckolbPTB:
Time (ti) of signal models is always added as first dimension. Allows to also provide single voxel parameters.

### DIFF
--- a/src/mrpro/operators/models/_InversionRecovery.py
+++ b/src/mrpro/operators/models/_InversionRecovery.py
@@ -29,6 +29,7 @@ class InversionRecovery(SignalModel[torch.Tensor, torch.Tensor]):
             with shape (time, ...)
         """
         super().__init__()
+        ti = torch.as_tensor(ti)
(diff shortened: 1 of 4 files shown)


ckolbPTB:
I would also suggest to have an abstract SignalModel class "MonoexponentialSignalModel" which does the init and reshaping of ti for InversionRecovery, SatRecovery, MOLLI and then later also for the standard multi-echo T2-model.

fzimmermann89:
> I would also suggest to have an abstract SignalModel class "MonoexponentialSignalModel" which does the init and reshaping of ti for InversionRecovery, SatRecovery, MOLLI and then later also for the standard multi-echo T2-model. 

Mhm.. I somewhat like that the signal models are quite easy to grasp..
We could save a few lines of code, with some added complexity.
We still need an init for each to have a matching docstring. So we would save one line in the init and maybe one or two lines in forward at the custom of inheritance..



I will have a look at  the code this weekend, first impression on mobile phone seems great.

fzimmermann89 (approved):

PR #232, opened 2024-03-22
Detection tolerance for KTrajectoryPulseq

schuenke:
Before, it was not possible to set the `repeat_detection_tolerance`, which is used when creating a `KTrajectory` from a `KTrajectoryRawShape` if using the `KTrajectoryPulseq` trajectory calculator. 

The old default would result in kz not beeing added as a ignore_dimension, leading to an error when applying the fourier operator. 

We decided that `repeat_detection_tolerance` is (for now) only required for `KTrajectoryPulseq` as all other should use correct analytical trajectory values anyway.

### DIFF
--- a/src/mrpro/data/traj_calculators/_KTrajectoryPulseq.py
+++ b/src/mrpro/data/traj_calculators/_KTrajectoryPulseq.py
@@ -58,24 +58,31 @@ class KTrajectoryPulseq(KTrajectoryCalculator):
 
         # calculate k-space trajectory using PyPulseq
         k_traj_adc_numpy, _, _, _, _ = seq.calculate_kspace()
-        k_traj_adc = torch.from_numpy(k_traj_adc_numpy)
+        k_traj_adc = torch.tensor(k_traj_adc_numpy, dtype=torch.float32)
 
         unique_idxs = {label: torch.unique(getattr(kheader.acq_info.idx, label)) for label in ['k1', 'k2']}
 
-        k1 = len(unique_idxs['k1'])
-        k2 = len(unique_idxs['k2'])
+        n_k1 = len(unique_idxs['k1'])
+        n_k2 = len(unique_idxs['k2'])
 
         n_samples = kheader.acq_info.number_of_samples
         n_samples = torch.unique(n_samples)
         if len(n_samples) > 1:
             raise ValueError('We currently only support constant number of samples')
-        k0 = int(n_samples.item())
+        n_k0 = int(n_samples.item())
 
         sample_size = n_samples.shape[0]
 
         def reshape_pulseq_traj(k_traj: torch.Tensor, encoding_size: int):
             k_traj *= encoding_size / (2 * torch.max(torch.abs(k_traj)))
-            return rearrange(k_traj, '(other k2 k1 k0) -> (other k2 k1) k0', k0=k0, k2=k2, k1=k1, other=sample_size)
+            return rearrange(
+                k_traj,
+                '(other k2 k1 k0) -> (other k2 k1) k0',
+                k0=n_k0,
+                k2=n_k2,
+                k1=n_k1,
+                other=sample_size,
+            )
 
         # rearrange k-space trajectory to match MRpro convention
         kx = reshape_pulseq_traj(k_traj_adc[0], kheader.encoding_matrix.x)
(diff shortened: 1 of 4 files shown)


src/mrpro/data/traj_calculators/_KTrajectoryPulseq.py
```diff
@@ -58,28 +61,35 @@ def __call__(self, kheader: KHeader) -> KTrajectoryRawShape:
...
+        n_k2 = len(unique_idxs['k2'])
 
         n_samples = kheader.acq_info.number_of_samples
         n_samples = torch.unique(n_samples)
         if len(n_samples) > 1:
             raise ValueError('We currently only support constant number of samples')
-        k0 = int(n_samples.item())
+        n_k0 = int(n_samples.item())
 
         sample_size = n_samples.shape[0]
```
fzimmermann89: n_other ?

fzimmermann89: somethink looks wrong here.

Have we tested this with multiple others?

Either n_samples is a single value, then n_k0 = int(n_samples.item()) works
OR
  sample_size = n_samples.shape[0] works.

fzimmermann89: if this indeed works, can you add a comment how and why?

src/mrpro/data/traj_calculators/_KTrajectoryPulseq.py
```diff
@@ -58,28 +61,35 @@ def __call__(self, kheader: KHeader) -> KTrajectoryRawShape:
...
         def reshape_pulseq_traj(k_traj: torch.Tensor, encoding_size: int):
             k_traj *= encoding_size / (2 * torch.max(torch.abs(k_traj)))
-            return rearrange(k_traj, '(other k2 k1 k0) -> (other k2 k1) k0', k0=k0, k2=k2, k1=k1, other=sample_size)
+            return rearrange(
+                k_traj,
+                '(other k2 k1 k0) -> (other k2 k1) k0',
+                k0=n_k0,
+                k2=n_k2,
+                k1=n_k1,
+                other=sample_size,
```
fzimmermann89: If sample_size is only used here, we can get rid of it?
`   return rearrange(k_traj, '(... k0) -> ... k0', k0=n_k0)
`

fzimmermann89 (changes requested):

src/mrpro/data/traj_calculators/_KTrajectoryPulseq.py
schuenke: good point, this should not work for n_other > 1. And obviously it was never tested before, because this is in the code for quite some time. 

How should we do it? 
```
  n_samples_unique = torch.unique(n_samples)
  if len(n_samples_unique) > 1:
      raise ValueError('We currently only support constant number of samples')
  n_k0 = int(n_samples_unique.item())
```

and than just `n_other = n_samples.shape[0]`?
This would at least work for other > 1 with all having the same `n_k0`, no ?!

I am not sure how `n_samples` looks like for multiple `other`.
We should get some test data for this... maybe we can acquire some on Tuesday @JoHa0811

src/mrpro/data/traj_calculators/_KTrajectoryPulseq.py
schuenke: I removed unnecessary code and just made it `return rearrange(k_traj, '(other k0) -> other k0', k0=n_k0)`
Using `...` does not work because `Ellipsis inside parenthesis in the left side is not allowed`

I tested it with the data used in the radial_2D example script from #235 and it works. Currently, I don't see a reason why it should not work in other cases.

src/mrpro/data/traj_calculators/_KTrajectoryPulseq.py
schuenke: Actually not needed at all ... at least at the moment. See other comment by @fzimmermann89 and my answer.

src/mrpro/data/traj_calculators/_KTrajectoryPulseq.py
fzimmermann89: Yes, that was what I meant :)

fzimmermann89 (approved):
thx

PR #233, opened 2024-03-22
Recon examples of Cartesian Siemens scans

### LINKED ISSUE #518: Cartesian Reconstruction example
I think we are missing a cartesian reconstruction example, ideally of the most difficult acquisition we can think of

 (undersampling, partial fourier, oversampling, data outside encoding matrix, etc)

hdillinger:
Fixes #518 

Added file to illustrate the use of matplotlib's imshow for non-isotropic pixel/voxel data.

The output:
grafik


fzimmermann89 (dismissed):
Not sure if this is helpful to have in mrpro in this form:
If we want to have an example that shows how properly plot to somebody how does not know matplotlib in mrpro (not sure if we do), 
 I would instead provide the simplest example without subplots and loops.
And separate creating the random data, recon_matrix (not encoding AFAIK) with the proper type etc from plotting it..
The main point of this example gets somehow lost in the code..

Or is this just an example to showcase something during the hackathon and not meant to end up in main

Edit: OK.. The latter. :D

fzimmermann89:
related: #518

ckolbPTB:
Superseded by #562

PR #234, opened 2024-03-22
Exclude data outside of encoding_matrix

### LINKED ISSUE #217: Mismatch encoding limits - encoding matrix

ckolbPTB:
closes #217


ckolbPTB:
any ideas of what to do now with the exclude info are very welcome

src/mrpro/operators/_CartesianSamplingOp.py
```diff
@@ -78,9 +80,37 @@ def __init__(self, encoding_matrix: SpatialDimension[int], traj: KTrajectory) ->
...
+            & (kz_idx < sorted_grid_shape.z)
+        )
+        if torch.any(~inside_encoding_matrix):
+            warnings.warn(
+                'K-space points lie outside of the encoding_matrix and will be ignored.'
+                ' Increase the encoding_matrix to include these points.',
+                stacklevel=2,
+            )
+
+            inside_encoding_matrix = rearrange(inside_encoding_matrix, '... kz ky kx -> ... 1 (kz ky kx)')
```
ckolbPTB: This part is not very nice. Maybe somebody has a better idea of how to do create the indices.

fzimmermann89: i have a suggestion using non-zero

fzimmermann89:
- rebased to main.
- made test more difficult: two oversampled axes and other dimension (which failed)
- change the index creation to use non-zero
- fix for other dimension, unbreaks more difficult test. uses unsqueeze_left
- tiny performance thing: order of adding broadcasted tensors to figure out the inside_mask
- removed the bool marking telling if inside_matrix indices exist, and instead make it the indices optional/None.
- changed some comments (minor)

@ckolbPTB  did I mess anything up?

I also think there might be a nicer way to do this, but this works for now, lets see if it ever actually becomes a performance bottleneck.

fzimmermann89 (approved):

ckolbPTB:
verified on real data

fzimmermann89:
Thank you for checking on real data! I will merge?

PR #235, opened 2024-03-22
Example script for pulseq radial 2D golden angle acquisition

schuenke:


schuenke:
I did some minor final adjustments and IMO we can merge this now. It's probably not the final example script (or at least not the final example data), but in general everything works fine and I think it's useful to have the example in main so that we can copy+paste the different reco / traj calc options from here. 

I will update the script when we acquired new data with the adjusted / fixed radial 2D sequence that actually samples from -n_x/2 to n_x/2 -1 (and thus samples the k-space center) as we expect it in KTrajectoryRadial2D.

The current final plot looks like this:
grafik

examples/pulseq_2d_radial_golden_angle.py
```diff
@@ -0,0 +1,164 @@
...
+csm_op = SensitivityOp(csm)
+(img_using_pulseq_traj,) = csm_op.adjoint(img)
+
+
+# %% [markdown]
+# ### Plot the different reconstructed images
+# Please note: there is currently a mismatch between the actual trajectory
+# that was used to acquire the data and the trajectory calculated with KTrajectoryRadial2D.
+# This leads to a deviation between the image reconstructed with KTrajectoryRadial2D
+# and the other two methods. In the future, we will upload new measurement data with
```
fzimmermann89, line 128: missing sentence

examples/pulseq_2d_radial_golden_angle.py
```diff
@@ -0,0 +1,164 @@
+# %% [markdown]
+# # Reconstruction of 2D golden angle radial data from pulseq sequence
```
fzimmermann89, line 2: can we remove the copyright from the examples or at least shorten it to 'Copyright 2024 Physikalisch-Technische Bundesanstalt, Apache License 2.0. See License.md"  and move it to the bottom? Its is quite prominent in the notebook and distracts imho.

examples/pulseq_2d_radial_golden_angle.py
```diff
@@ -0,0 +1,164 @@
...
+# This will use the trajectory that is stored in the ISMRMRD file.
+
+# %%
+# Read the raw data and the trajectory from ISMRMRD file
+kdata = KData.from_file(data_file.name, KTrajectoryIsmrmrd())
+
+# Calculate dcf using the trajectory
+dcf = DcfData.from_traj_voronoi(kdata.traj)
+
+# Reconstruct average image for coil map estimation
```
ckolbPTB: I would change this to "Define Fourier operator and reconstruct coil images" (same for the other two)

ckolbPTB (approved):

ckolbPTB:
I would suggest to define a function "reconstruct_images" at the beginning including dcf, fourier_op, csm which returns coil-combined images which is then called three times, rather than repeating the code all the time.

ckolbPTB:
> I will update the script when we acquired new data

Would also be good to have a readout oversampling of 2 for the new acuqisitions to get rid of the "ring"

ckolbPTB:
This cell is actually not needed anymore:

```
If you want to run this notebook in binder you need to still install the MRpro package. This only needs to be done once in a binder session. Open a terminal (File -> New -> Terminal) and run:

pip install -e ".[notebook]"

This will install the MRpro package. Any other required python packages should already be present in this docker image.

```

Could you please also remove it from the other notebook? thanks

fzimmermann89:
Do we want to show lowevel stuff here or should this use the direct reconstruction helper once merged?

fzimmermann89:
If we want to keep it low level, I would keep the repitions, such that each section contains all related code..

Defining a helper in the example would somewhat defeat the purpose of keeping the examples simple..

examples/pulseq_2d_radial_golden_angle.py
schuenke: Would be fine for me. What do you say @ckolbPTB ?

schuenke:
> Do we want to show lowevel stuff here or should this use the direct reconstruction helper once merged?

In general, I would stick with the low-level stuff. For the reconstruction helper functions, I would create (some) separate example scripts.

> If we want to keep it low level, I would keep the repitions, such that each section contains all related code..
> Defining a helper in the example would somewhat defeat the purpose of keeping the examples simple..

I agree. IMO the examples will be mainly used to copy code snippets for personal scripts and thus keeping it low-level without additional helper functions should be most practical.

examples/pulseq_2d_radial_golden_angle.py
ckolbPTB: Fine by me

ckolbPTB:
> Do we want to show lowevel stuff here or should this use the direct reconstruction helper once merged?

I would actually prefer to have one notebook which explains the direct reconstruction helper in detail but then use that for the other notebooks. IMO this would ensure that each notebook focuses on different aspects of the code and does not repeat everything in detail. Also in the long run this would make it easier to maintain all the example notebooks.

For the time being we can also leave it as is and have a look at it again once the reconstruction helper stuff is merged.

examples/pulseq_2d_radial_golden_angle.py
schuenke: So short version at the top (vote with 👍) or long version at the bottom (vote with 👎) ?

PR #237, opened 2024-03-25
Projection operator

fzimmermann89:
A Volume-to-Slice Projection Operator.
The forward samples a 2D slice out of a 3D Volume with a certain offset and rotation, considering a slice profile.
This can, for example, be used for an superresolution reconstruction.

The slice_profile, rotation, offsets can have arbitrary many leading dimensions, as long as they broadcast.
For example you could do:
rotations-shape = 5,1
offsets-shape = 1,3
profile-shape = 1,3

Applying this to a (other, coil,z,y,x) volume would results in 
This would currently result in 2D slices with shape (5,3,other, coil,z=1,y_2d,x_2d), i.e. 5 rotations, each with 3 offsets. each of the 3 offsets has a different slice profile. x_2d and y_2d are the maximum of (z,y,x), such that all rotations result in the same slice shape. Data outside the volume is zero-padded.

The logic is implemented using a sparse matrix multiplication. In the __init__, I construct the projection matrix and reuse it on every .forward/adjoint call.

The operator does multiple slices / rotations etc as one matrix multiplication, and does different volumes as different multiplications. This is somewhat optimized for the superresolution case with more slices than volumes.


fzimmermann89:
Not sure where to put the slice profiles...

fzimmermann89:
@schuenke @ckolbPTB  any suggestions about the slice profiles?

src/mrpro/operators/_SliceProjectionOp.py
```diff
@@ -0,0 +1,449 @@
...
+    def __init__(self, xs: Tensor, weights: Tensor):
+        self._xs = xs
+        self._weights = weights
+
+    def __call__(self, x: torch.Tensor) -> torch.Tensor:
+        return torch.as_tensor(np.interp(x, self._xs.numpy(), self._weights.numpy(), 0, 0))
+
+
+TensorFunction: TypeAlias = Callable[[Tensor], Tensor]
+DefaultGaussianSliceProfile = SliceGaussian(2.0)
```
fzimmermann89: remove

src/mrpro/operators/_SliceProjectionOp.py
```diff
@@ -0,0 +1,449 @@
...
+            [
+                _MatrixMultiplication.apply(x, matrix_adjoint, matrix).reshape(self._domain_shape)
+                for x in x_flatdomainbatch_flatrange
+            ]
+        )
+        y = y_flatdomainbatch.reshape(*x.shape[n_batchdim:-3], *y_flatdomainbatch.shape[1:])
+        return (y,)
+
+    @staticmethod
+    def join_matrices(matrices: Sequence[torch.Tensor]) -> torch.Tensor:
```
fzimmermann89: needs a docstring

src/mrpro/operators/_SliceProjectionOp.py
```diff
@@ -0,0 +1,449 @@
...
+
+    """
+
+    @staticmethod
+    def forward(
+        ctx: torch.autograd.function.FunctionCtx, x: torch.Tensor, matrix: torch.Tensor, matrix_adjoint: torch.Tensor
+    ) -> torch.Tensor:
+        ctx.save_for_backward(matrix_adjoint)
+        ctx.x_is_complex = x.is_complex()  # type: ignore[attr-defined]
+        if x.is_complex() == matrix.is_complex():
```
ckolbPTB, line 49: Why do we need these different cases? Does @ not support float and complex combinations?

src/mrpro/operators/_SliceProjectionOp.py
```diff
@@ -0,0 +1,449 @@
...
+        self.fwhm = torch.as_tensor(fwhm)
+
+    def __call__(self, x: torch.Tensor) -> torch.Tensor:
+        return torch.exp(-(x**2) / (0.36 * self.fwhm**2))
+
+
+class SliceSmoothedRect:
+    """Rectangular Slice Profile with smooth flanks.
+
+    The smaller n, the smoother it is. For n<1, the FWHM might be wrong
```
ckolbPTB: Shall we add a warning in the __init__ for n < 1?

src/mrpro/operators/_SliceProjectionOp.py
```diff
@@ -0,0 +1,449 @@
...
+            else:
+                grad_x = torch.complex(matrix_adjoint @ grad_output[0].real, matrix_adjoint @ grad_output[0].imag)
+        else:  # real grad
+            grad_x = matrix_adjoint.real @ grad_output[0].real
+            if matrix_adjoint.is_complex() and grad_output[0].is_complex():
+                grad_x -= matrix_adjoint.imag @ grad_output[0].imag
+        return grad_x, None, None
+
+
+class SliceGaussian:
```
ckolbPTB: docstrings missing for all slice-classes

src/mrpro/operators/_SliceProjectionOp.py
```diff
@@ -0,0 +1,449 @@
...
+    """Gaussian Slice Profile."""
+
+    def __init__(self, fwhm: float | Tensor):
+        self.fwhm = torch.as_tensor(fwhm)
+
+    def __call__(self, x: torch.Tensor) -> torch.Tensor:
+        return torch.exp(-(x**2) / (0.36 * self.fwhm**2))
```
ckolbPTB: Maybe add comment that 0.36 is the approximation of 1/4ln(2)

src/mrpro/operators/_SliceProjectionOp.py
```diff
@@ -0,0 +1,449 @@
...
+        slice_rotation: Rotation | None = None,
+        slice_shift: float | Tensor = 0.0,
+        slice_profile: TensorFunction | np.ndarray | NestedSequence[TensorFunction] | float = 1.0,
+        optimize_for: Literal['forward', 'adjoint', 'both'] = 'both',
+    ):
+        """Create a module that represents the 'projection' of a volume onto a plane.
+
+        This operation samples from a 3D Volume a slice with a given rotation and shift,
+        according to the slice_profile. It can, for example, be used to describe the slice
+        selection of a 2D MRI sequence from the 3D Volume.-
```
ckolbPTB: Remove -

src/mrpro/operators/_SliceProjectionOp.py
```diff
@@ -0,0 +1,449 @@
...
+        self,
+        input_shape: SpatialDimension[int],
+        slice_rotation: Rotation | None = None,
+        slice_shift: float | Tensor = 0.0,
+        slice_profile: TensorFunction | np.ndarray | NestedSequence[TensorFunction] | float = 1.0,
+        optimize_for: Literal['forward', 'adjoint', 'both'] = 'both',
+    ):
+        """Create a module that represents the 'projection' of a volume onto a plane.
+
+        This operation samples from a 3D Volume a slice with a given rotation and shift,
```
ckolbPTB: relative to the centre of the 3D volume

src/mrpro/operators/_SliceProjectionOp.py
```diff
@@ -0,0 +1,449 @@
...
+        according to the slice_profile. It can, for example, be used to describe the slice
+        selection of a 2D MRI sequence from the 3D Volume.-
+
+        The projection will be done by sparse matrix multiplication.
+
+        Rotation, shift, and profile can have (multiple) batch dimensions. These dimensions will
+        be broadcasted to a common shape and added to the front of the volume.
+        Different settings for different volume batches are NOT supported, consider creating multiple
+        operators if required,
```
ckolbPTB: , -> .

src/mrpro/operators/_SliceProjectionOp.py
```diff
@@ -0,0 +1,449 @@
...
+
+
+        Parameters
+        ----------
+        input_shape:
+            Shape of the 3D volume to sample from
```
ckolbPTB: mention it is only z,y,x ?

src/mrpro/operators/_SliceProjectionOp.py
```diff
@@ -0,0 +1,449 @@
...
+
+        Rotation, shift, and profile can have (multiple) batch dimensions. These dimensions will
+        be broadcasted to a common shape and added to the front of the volume.
+        Different settings for different volume batches are NOT supported, consider creating multiple
+        operators if required,
+
+
+        Parameters
+        ----------
+        input_shape:
```
ckolbPTB: no ":" after parameters, occurs several times

src/mrpro/operators/_SliceProjectionOp.py
```diff
@@ -0,0 +1,449 @@
...
+            Shape of the 3D volume to sample from
+        slice_rotation
+            Rotation that describes the orientation of the plane. If None,
+            an identity rotation is used.
+        slice_shift
+            Offset of the plane in the volume perpendicular plane from the center of the volume.
+        slice_profile:
+            A function that called with a distance x from the slice center should return the
```
ckolbPTB: Maybe: A function that is called with a tensor describing the distance x relative to centre of the volume. It should return the slice profile (i.e. intensity) at the positions in x.

src/mrpro/operators/_SliceProjectionOp.py
```diff
@@ -0,0 +1,449 @@
...
+            intensity along the slice thickness at x. This can also be a nested Sequence or an
+            numpy array of functions.
+            If it is a single float, it will be interpreted as a Gaussian FWHM.
```
ckolbPTB: This seems unintuative to me. If a function is set to a constant value, I would epect it to return a constant value rather than a Gaussian profile.

src/mrpro/operators/_SliceProjectionOp.py
```diff
@@ -0,0 +1,449 @@
...
+        """
+        super().__init__()
+        if isinstance(slice_profile, float):
+            slice_profile = SliceGaussian(slice_profile)
+        slice_profile_array = np.array(slice_profile)
+
+        if slice_rotation is None:
+            slice_rotation = Rotation.identity()
+
+        m = max(input_shape.z, input_shape.y, input_shape.x)
```
ckolbPTB: m, r cdf could be a bit more expressive

src/mrpro/operators/_SliceProjectionOp.py
```diff
@@ -0,0 +1,449 @@
...
+                self.register_buffer('matrix_adjoint', matrix.H.to_sparse_csr())
+                self.register_buffer('matrix', matrix.to_sparse_csr())
+
+            else:
+                raise ValueError("optimize_for must be one of 'forward', 'adjoint', 'both'")
+
+        self._range_shape: tuple[int] = (*batch_shapes, 1, m, m)
+        self._domain_shape = input_shape.zyx
+
+    def forward(self, x: torch.Tensor) -> tuple[torch.Tensor]:
```
ckolbPTB: Parameters missing

src/mrpro/operators/_SliceProjectionOp.py
```diff
@@ -0,0 +1,449 @@
...
+
+        # For the (unusual case) of batched volumes, we will apply for each element in series
+        xflat = torch.atleast_2d(einops.rearrange(x, '... x y z -> (...) (x y z)'))
+        y = torch.stack(
+            [_MatrixMultiplication.apply(x, matrix, matrix_adjoint).reshape(self._range_shape) for x in xflat], -4
+        )
+        y = y.reshape(*y.shape[:-4], *x.shape[:-3], *y.shape[-3:])
+        return (y,)
+
+    def adjoint(self, x: torch.Tensor) -> tuple[torch.Tensor,]:
```
ckolbPTB: Parameters missing

src/mrpro/operators/_SliceProjectionOp.py
```diff
@@ -0,0 +1,449 @@
...
+
+    @staticmethod
+    def join_matrices(matrices: Sequence[torch.Tensor]) -> torch.Tensor:
+        values = []
+        target = []
+        source = []
+        for i, m in enumerate(matrices):
+            if not m.shape == matrices[0].shape:
+                raise ValueError('all matrices should have the same shape')
+            c = m.coalesce()
```
ckolbPTB: Could you add some comments on what coalesce() does here? Maybe superfluous once there is a docstring.

src/mrpro/operators/_SliceProjectionOp.py
```diff
@@ -0,0 +1,449 @@
...
+        slice_function: TensorFunction,
+        rotation_center: torch.Tensor | None = None,
+    ) -> torch.Tensor:
+        """Create a sparse matrix that represents the projection of a volume onto a plane.
+
+        Outside the volume values are approximately zero padded
+
+        Parameters
+        ----------
+        input_shape:
```
ckolbPTB: Clean parameters: no ":", no types

src/mrpro/operators/_SliceProjectionOp.py
```diff
@@ -0,0 +1,449 @@
...
+            Shape of the volume to sample from
+        output_shape:
+            Shape of the resulting plane, 2D.
```
ckolbPTB: mention here that only x and y are used from SpatialDimension?

src/mrpro/operators/_SliceProjectionOp.py
```diff
@@ -0,0 +1,449 @@
...
+            torch.stack(
+                [
+                    torch.arange(-w, w + 1),  # z
+                    torch.zeros(2 * w + 1),  # y
+                    torch.zeros(2 * w + 1),  # x
+                ],
+                dim=-1,
+            )
+        )
+        # In all possible directions for each point aloing the line we consider the eight neighboring points
```
ckolbPTB: along

tests/operators/test_slice_projection_op.py
```diff
@@ -0,0 +1,138 @@
...
+from mrpro.operators._SliceProjectionOp import SliceGaussian
+from mrpro.operators._SliceProjectionOp import SliceInterpolate
+from mrpro.operators._SliceProjectionOp import SliceSmoothedRect
+from mrpro.utils import Rotation
+
+from tests import RandomGenerator
+from tests.helper import dotproduct_adjointness_test
+
+
+def test_slice_projection_op_basic():
```
ckolbPTB: I don't understand this test.

src/mrpro/operators/_SliceProjectionOp.py
```diff
@@ -0,0 +1,449 @@
...
+        Parameters
+        ----------
+        input_shape:
+            Shape of the volume to sample from
+        output_shape:
+            Shape of the resulting plane, 2D.
+        rotation
+            Rotation that describes the orientation of the plane
+        offset: Tensor
+            Offset of the plane in the volume in plane coordinates from the center of the volume
```
ckolbPTB: what do you mean here with "plane coordinates"? I would expect that you first rotate and then shift. Is the shift in voxel coordinates? If so, is a offset =  1 shifted by one voxel of the 3D volume or by the slice thickness, i.e. fwhm?

src/mrpro/operators/_SliceProjectionOp.py
```diff
@@ -0,0 +1,449 @@
...
+            ones = ones.coalesce()
+
+            matrix = torch.sparse_coo_tensor(
+                indices=torch.stack((target_index, source_index)),
+                values=weight.reshape(y * x, -1)[mask],
+                size=(y * x, input_shape.z * input_shape.y * input_shape.x),
+                dtype=torch.float32,
+            ).coalesce()
+
+        # To avoid giving to much weight to points that are duplicated in our logic and summed up by coalesce
```
ckolbPTB: too much

ckolbPTB: maybe "to ensure that the contribution of duplicated points is normalised"?

src/mrpro/operators/_SliceProjectionOp.py
```diff
@@ -0,0 +1,449 @@
...
+        matrix.values()[:] /= ones.values()
+
+        # Normalize
```
ckolbPTB: Normalize by what?

ckolbPTB:
When I try to run the tests they all fail because:

E           RuntimeWarning: overflow encountered in _find_width (vectorized)
../../../Data/CONDA/MRproTorch22/lib/python3.11/site-packages/numpy/lib/function_base.py:2455: RuntimeWarning

ckolbPTB:
I think a bit more information is needed regarding the parameters. Are all the values (fwhm for the slice profile, offset,...) in voxel units of the 3D volume? Could they also be provided in mm? What is the order of rotation and offset?

ckolbPTB:
Would it be possible to add a bit more advanced test by comparing the projection operator with a constant slice profile to a standard radon transform? Then we could at least test different rotation angles. We could also test different slice profiles if we apply them to the object prior to applying the radon transform.

src/mrpro/operators/_SliceProjectionOp.py
fzimmermann89: Not for sparse matrices. Will add a comment.

src/mrpro/operators/_SliceProjectionOp.py
fzimmermann89: I will rename/reparametrize the parameters such that none of them is called fwhm..?

src/mrpro/operators/_SliceProjectionOp.py
fzimmermann89: This is way it is explicitly describe  in the docstring.
Or rename the parameter? 

We could also remove the float option  here, but I found it  useful to be able to only specify a width and have a default...

src/mrpro/operators/_SliceProjectionOp.py
fzimmermann89: Do we want to keep them?
Or how do we handle different profiles?
This was still an open question for me.

tests/operators/test_slice_projection_op.py
fzimmermann89: A slice through a 3d volume of ones should have to correct shape (maximum of z,y,x of the volume in both x and y) and only contain zeros and ones.

The slice is in the center of the volume and in x y orientation.
So there will just be a rectangle of ones in the center (but padded to be square)

Do we need this expansion? It seemed to me quite clear from the description of the operator what the expected result should be.

Or what would help your understanding?

fzimmermann89:
> When I try to run the tests they all fail because:
> 
> E RuntimeWarning: overflow encountered in _find_width (vectorized)
> ../../../Data/CONDA/MRproTorch22/lib/python3.11/site-packages/numpy/lib/function_base.py:2455: RuntimeWarning
> 

All tests? Or only certain slice profiles? 

(.. The slice profiles will need unit tests if we keep them..)


Numpy Version?

Could you replace
 
       widths = np.vectorize(_find_width)(slice_profile_array)

By
 
```
 widths = np.array([_find_width(x) for x in slice_profile_array.ravel()]).reshape(slice_profile_array.shape)
      print(widths) 
```

Does the output look reasonable? Are there errors?

src/mrpro/operators/_SliceProjectionOp.py
ckolbPTB: also a possibility

src/mrpro/operators/_SliceProjectionOp.py
ckolbPTB: I think they are needed/very useful. Maybe put them in utils? Having a slice profile might also be useful for qMRI applications where different flip angles need to be modelled along the slice.

tests/operators/test_slice_projection_op.py
ckolbPTB: The padding was not obvious to me. Maybe add a comment. For this specific case I would have expected a slice with x = 30 and y = 20 with only ones because I am only summing up along z. I do understand that you are doing this padding to ensure it also works for arbitrary angles.

ckolbPTB:
> All tests? Or only certain slice profiles?

No not all tests but adding one more test which verifies that the projection is really working. For an object with 1s everywhere, many things would lead to a slice with only 1s. 

> Numpy version

1.26.3

> Does the output look reasonable? Are there errors?

No errors. Output looks reasonable

fzimmermann89:
> Would it be possible to add a bit more advanced test by comparing the projection operator with a constant slice profile to a standard radon transform? Then we could at least test different rotation angles. We could also test different slice profiles if we apply them to the object prior to applying the radon transform.

I thought about this, but it is difficult, especially to match to a meaningful precision (Or I am missing something)

We could do is check fourier slice theorem:
So create a test volume.
a) set the slice profile to rectangle with width larger than the volume, do the projection operation, 2d fourier transformation
b) do a 3d fourier transform. do the projection operation.

the results should be close-ish, depending on padding the interpolation error (the projection operator does linear interpolation, not sinc interpolation.. ) but this can fail in many different ways.... 

Not sure if this helps as a test... any  more detailed suggestions?

I would maybe add one test where we know what we expect as the output, like a cube cut at 45 degree?

tests/operators/test_slice_projection_op.py
fzimmermann89: yes, this was in the docstring at some point... but somehow it vanished while refactoring :D

src/mrpro/operators/_SliceProjectionOp.py
fzimmermann89: Are do you mean a rectangular slice profile with this width?

ckolbPTB:
> I thought about this, but it is difficult, especially to match to a meaningful precision (Or I am missing something)

I guess it should be similar to here: https://github.com/PTB-MR/mrpro/blob/e8c536cdff3298fe6fea5b983a79c39c8b7b6247/tests/phantoms/test_ellipse_phantom.py#L47
There will be differences but results should be reasonable similar.

> I would maybe add one test where we know what we expect as the output, like a cube cut at 45 degree?

Also an option but again only a very specific case...

src/mrpro/operators/_SliceProjectionOp.py
ckolbPTB: Yes, I meant a rectangular profile with this width

ckolbPTB: Describing something in a docstring is a good option but IMO it is better to provide someting more intuative such that the user doesn't even  have to read the docstring

fzimmermann89: so rectangluar profile. I have no opionion on the default profile shape, so a hard rect is as good as a  gaussian for me.
> This seems unintuative to me. If a function is set to a constant value, I would epect it to return a constant value rather than a Gaussian profile.

--> this is still a issue with the rect profile.
the slice_profilce function for a rectangle would be.. a rectangle. not a constant value. so the issue that it is either 'profile' function, or it is a width float still exists.  a constant function would be an infinite width, uniform profile.
Not sure if there is a better way to do it though if we want to have the option of only setting a width (which I prefer)

src/mrpro/operators/_SliceProjectionOp.py
fzimmermann89: voxel coordinates.
does this make sense (and I only add it to the docstring) or should it be different?

What is meant here is:
"Shift of the plane from the center of the volume, perpendicular the the plane, in units of the 3D volume."

Changing the pixel/voxelsize or shifting the plane in other direction is currently not supported in this version.

fzimmermann89:
Feel free to implement an analytical reference for arbitrary angles with a non trivial shape and slice profile :D

fzimmermann89:
> No not all tests but adding one more test which verifies that the projection is really working. For an object with 1s everywhere, many things would lead to a slice with only 1s.

In the this version, there was actually still an issue where I forgot to switch the axis order, but without rotation it is not visible ;)

src/mrpro/operators/_SliceProjectionOp.py
```diff
@@ -0,0 +1,449 @@
...
+                ),
+            ],
+            dim=-1,
+        )  # coordinates of the 2d output pixels in the coordinate system of the input volume, so shape (y,x,3)
+        if offset is not None:
+            pixel_coord_y_x_zyx = pixel_coord_y_x_zyx + offset
+        if rotation_center is None:
+            # default rotation center is the center of the volume, i.e. for 4 pixels
+            # 0 1 2 3 it is between 0 and 1
+            rotation_center = torch.tensor([input_shape.x / 2 - 0.5, input_shape.y / 2 - 0.5, input_shape.z / 2 - 0.5])
```
fzimmermann89: this needs to be switched to z,y,x

src/mrpro/operators/_SliceProjectionOp.py
ckolbPTB: > Not sure if there is a better way to do it though if we want to have the option of only setting a width (which I prefer)

I can't think of a better way. As far as I understand you are normalising your slice profile anyway so a "a constant function would be an infinite width" would always be the same independet of setting this value to 1 or 100. A rectangular profile seems to be more useful then.

src/mrpro/operators/_SliceProjectionOp.py
ckolbPTB: > "Shift of the plane from the center of the volume, perpendicular the the plane, in units of the 3D volume."

sounds like a good explaination to me

fzimmermann89:
@ckolbPTB I have a logic issue.

A 'rectangular' slice profile  with width 1 currently would return only 1 from -0.5 to 0.5, outside it would be zero.
This is what one would expect from the name.

On the other hand, I use this profile as a weighting of the known data points of the 3D volume.
So, without rotation, to calculate slice shifted by 0.5, I look at the voxels at 0 and 1.
both are 0.5 away from the slice, thus the slice profile says they should not influence the result.

If we want to do an interpolation between those pixels, we would have to use a trapezoidal "slice profile", extended by less than 1 voxel in each direction and call it rectangular. But this would make it somewhat specialized for this use case.

 In practice, this should not be an issue, as usually the slice profiles do not have hard flanks and are not 1 voxel. ...

ckolbPTB:
I agree that this is a very specific corner case and simply adding a warning that width <= 1 for rect probably is not a good idea should be ok. 

Another option would be to define the rect function as 0.5 for x = 0 = 1. At least according to wikipedia this is a valid definition of a rect function.

fzimmermann89:
It is not only this edge case, I think for a true rectangular profile, having some linear interpolation at the edges could maybe always be a better approximation... 

There is no other interpolation of the voxels values in z direction. So for a really thin slice profile, the result would contain lots of zeros. Instead, we could do the interpolation by adjusting the z weights..  
Depends on how we define what the sampling with a rectangular profile should mean wrt interpolation. Infinitly sharp edges in the profile currently mean 'no interpolation', resulting in either hard steps (example: no rotation, profile >1) or even holes in the projection (no rotation, profile <1). With rotation, this could happen up to a width of sqrt(3)  ..

In practices, this does not really realistic slice profiles

I added an exception for width smaller than 1.

(...This was one of the reasons why I defaulted to the Gaussian Profile shape :D having a smoother, more realistic profile helps with these edge cases ...)

fzimmermann89:
Also, not sure what the best solution is with regards to 'center'.
I think our SRR sequence does the rotation around the center of the FOV. So if the FOV has 4 pixel, it will rotate around a center between the pixels. Thus, I consider for an even 3D shape also this as the 'center' of the volume. Thus means without any rotation,
taking a slice of a 4x4x4 volume with shift=0 would be a slice in between the samples.. (or you would need to to the half pixel offset as a slice shift).
The rotation center and slice shift can both be adjusted, so this could easily be changed in the future.

What are your opinions on a good default for this? It is somewhat similar to the k-space center question for radial sampling...

Otherwise, I think I got all remarks, what you could still consider lacking are the more advanced tests...

ckolbPTB (approved):

ckolbPTB:
> What are your opinions on a good default for this

Centre between the voxel seems a good idea. In the end it is the same as with the k-space centre: you need to know what the scanner software does. For the k-space centre this is easy, for a rotation of a slice this is much more challenging. 

> what you could still consider lacking are the more advanced tests...

The new tests with the cube are very useful. I would leave it at that for the moment.

PR #238, opened 2024-03-27
Add from_array_zxy, zyx read-only property and tests to SpatialDimension

fzimmermann89:
- we were missing tests for spatial dimension.
- we had no way to transform a tensor in 'our' dimension order to a SpatialDimension
- zyx property is a replacement for dataclasses.as_tuple with a more obvious order of the axes. this can, for example, be used for iterating over z, y, and x or to transform it into a tensor.


schuenke (approved):
thanks for catching the missing method and tests. lgtm 👍

PR #241, opened 2024-04-01
Motion / Gridsample operator

fzimmermann89:
A 2D / 3D Batched Sampling Operator.

It both forward and adjoint are differentiable with respect to both the inputs and the grid,

Grid is supposed to be of shape
 (*batch_dim, z, y, x, 3) 3D
or (*batch_dim, y, x, 2) 2D

The input should be 
(*batch_dim, *channel_dim, z y x) 3D
(*batch_dim, *channel_dim,  y x) 2D (an existing z will thus be interpreted as a channel dimension)
Both batch and channel can be multiple axes. 

Input can be real or complex.


src/mrpro/operators/_GridSamplingOp.py
```diff
@@ -0,0 +1,307 @@
...
+        self,
+        grid: torch.Tensor,
+        input_shape: SpatialDimension,
+        interpolation_mode: Literal['bilinear', 'nearest', 'bicubic'] = 'bilinear',
+        padding_mode: Literal['zeros', 'border', 'reflection'] = 'zeros',
+        align_corners: bool = False,
+    ):
+        """Initialize Sampling Operator.
+
+        Parameters
```
ckolbPTB, line 176: The order of the parameters in the docu does not match the order above.

tests/operators/test_grid_sampling_op.py
```diff
@@ -0,0 +1,268 @@
+"""Tests for sensitivity operator."""
```
ckolbPTB: grid sampling operator

tests/operators/test_grid_sampling_op.py
```diff
@@ -0,0 +1,268 @@
...
+@pytest.mark.parametrize('padding_mode', ['zeros', 'border', 'reflection'])
+def test_grid_sampling_op_padding_mode(padding_mode):
+    """Test for different padding_modes."""
+    _test_grid_sampling_op_adjoint(padding_mode=padding_mode)
+
+
+@pytest.mark.parametrize('align_corners', ['no_align', 'align'])
+def test_grid_sampling_op_align_mode(align_corners):
+    """Test for different align modes ."""
+    _test_grid_sampling_op_adjoint(align_corners=align_corners == 'align')
```
ckolbPTB: should this not be `align_corners=align_corners` as align_corners is a string and not bool

tests/operators/test_grid_sampling_op.py
```diff
@@ -0,0 +1,268 @@
...
+@pytest.mark.parametrize('channel', ['multi_channel', 'single_channel'])
+def test_grid_sampling_op_dim_batch_channel(dim_str, batched, channel):
+    """Test for different dimensions."""
+    _test_grid_sampling_op_adjoint(dim=int(dim_str[0]), batched=batched, channel=channel)
+
+
+@pytest.mark.parametrize('interpolation_mode', ['bilinear', 'nearest', 'bicubic'])
+def test_grid_sampling_op_interpolation_mode(interpolation_mode):
+    """Test for different interpolation_modes."""
+    _test_grid_sampling_op_adjoint(dim=2, interpolation_mode=interpolation_mode)
```
ckolbPTB, line 45: dim=2 is default and not needed here

tests/operators/test_grid_sampling_op.py
```diff
@@ -0,0 +1,268 @@
...
+
+
+def test_grid_sampling_op_errormsg_gridlastdim():
+    "Test if error message on wrong last dim is raised."
+    grid = torch.ones(1, 2, 3, 4)
+    with pytest.raises(ValueError, match='last dimension'):
+        _ = GridSamplingOp(grid, SpatialDimension(1, 1, 1))
+
+
+def test_grid_sampling_op_errormsg_gridndims():
```
ckolbPTB: better to use parametrize

tests/operators/test_grid_sampling_op.py
```diff
@@ -0,0 +1,268 @@
...
+
+def test_grid_sampling_op_errormsg_complexgrid():
+    "Test if error for complex grid is raised."
+    grid = torch.ones(1, 1, 1, 1, 3) + 0j
+    with pytest.raises(ValueError, match='real'):
+        _ = GridSamplingOp(grid, SpatialDimension(1, 1, 1))
+
+
+def test_grid_sampling_op_errormsg_gridrange():
+    "Test if warnding for grid values outside [-1,1] is raised"
```
ckolbPTB: warning not warnding

tests/operators/test_grid_sampling_op.py
```diff
@@ -0,0 +1,268 @@
...
+
+
+def test_grid_sampling_op_errormsg_complexgrid():
+    "Test if error for complex grid is raised."
+    grid = torch.ones(1, 1, 1, 1, 3) + 0j
+    with pytest.raises(ValueError, match='real'):
+        _ = GridSamplingOp(grid, SpatialDimension(1, 1, 1))
+
+
+def test_grid_sampling_op_errormsg_gridrange():
```
ckolbPTB: op_warningmsg_gridrange

tests/operators/test_grid_sampling_op.py
```diff
@@ -0,0 +1,268 @@
...
+    "Test if warnding for grid values outside [-1,1] is raised"
+    grid = torch.ones(1, 1, 1, 1, 3)
+
+    grid.ravel()[0] = 1.01
```
ckolbPTB: also better to use parametrize

tests/operators/test_grid_sampling_op.py
```diff
@@ -0,0 +1,268 @@
...
+        _ = GridSamplingOp(grid, SpatialDimension(1, 1, 1))
+
+    grid.ravel()[0] = -1.0
+    _ = GridSamplingOp(grid, SpatialDimension(1, 1, 1))
+
+    grid.ravel()[0] = 1.0
+    _ = GridSamplingOp(grid, SpatialDimension(1, 1, 1))
+
+
+def test_grid_sampling_op_errormsg_inputdim():
```
ckolbPTB: can this be rewritten like `test_grid_sampling_op_batchdims`?

src/mrpro/operators/_GridSamplingOp.py
```diff
@@ -0,0 +1,307 @@
...
+    @staticmethod
+    def backward(
+        ctx: torch.autograd.function.FunctionCtx, *grad_output: torch.Tensor
+    ) -> tuple[torch.Tensor | None, torch.Tensor | None, None, None, None, None]:
+        """Backward of the Adjoint Gridsample Operator."""
+        need_y_grad, need_grid_grad, *_ = ctx.needs_input_grad  # type: ignore[attr-defined]
+        grid = ctx.saved_tensors[0]  # type: ignore[attr-defined]
+
+        if need_y_grad:
+            grad_y = torch.grid_sampler(
```
ckolbPTB, line 136: why is `torch.grid_sampler` needed here instead of `torch.nn.functional.grid_sample`?

ckolbPTB (changes requested):

tests/operators/test_grid_sampling_op.py
fzimmermann89: I explicitly added it here as only for dim=2 all Interpolation modes are defined

src/mrpro/operators/_GridSamplingOp.py
fzimmermann89: The functional calls does call the other one internally.

The Functional one takes strings as arguments. 
The gridsample  backward takes ints.
For consistency, I used the variants with int parameters for the interpolation mode etc. throughout the code.

tests/operators/test_grid_sampling_op.py
fzimmermann89: No. I could split it into two tests with each three lines. Using parameterize makes the code more complicated, as 2/3 lines would just be behind an if switch..

tests/operators/test_grid_sampling_op.py
fzimmermann89: Why is it better? It is more ccomplicated for a simple test.

tests/operators/test_grid_sampling_op.py
fzimmermann89: I prefer not to. I had it parameterize, but due to warning/raise/works and really simple tests it was more complicated to read...

fzimmermann89: I could do it as many seperate tests if prefered..

src/mrpro/operators/_GridSamplingOp.py
ckolbPTB: Could you add a comment here to explain this? I guess other people might be wondering about this too...

tests/operators/test_grid_sampling_op.py
ckolbPTB: I understand that parameterization might lead to more verbose or complex code but the problem I see here is that the test can fail for a number of reasons. Depending on when it fails the majority of tests might not actually be run which is not good practice for unit testing.

fzimmermann89: split into different tests

ckolbPTB (approved):

PR #243, opened 2024-04-04
WIP: Rearrange / GetItem / stack

### LINKED ISSUE #382: Easier Indexing / Subsetting of KData
So far, in order to index the data part of kdata, one has to access it via kdata.data[index]
It would be nice if one could simply state kdata[index].
Also, subsetting kdata is relatively tedious.
A perfect case would be if one could subset kdata by usng kdata[subsetting_indices].

Not sure if these are possible right away.

fzimmermann89:
WIP API for shape operations.

- rearrange(pattern, **axes_lengths)
  with an enhanced pattern type
  ... | ... -> (...) | ...
	  -> will reshape to one other dim
	  '|' is the delimter between other-like batch dimensions and k-dimensions. For kdata.data this is the 'coil' dimension.
	  if no | is present, interpret left-to-right, i.e. only apply to other dimensions
	  throw error if reshape does not result in coil k2 k1 k0 dimensions
	  

- `__getitem__`
   Returns a (copy) subset of the data.
   So kdata[0].traj.kz will return the kz values of the first other and is equivalent to kdata.traj.kz[0].
   This supports boolean mask indexing and is used in the next function. .clone() can be used to transform it into a new data 
   object
   
- stack(other, dim)
   stacks another data object in a certain dimension
   
- split(dim, size, step)
   Splits one data object along other dimension into multiple, potentially overlapping, views.
 
   
- select_idx_subset(iabel, idx)
   This is a convenience function, implementing 
   mask= torch.isin( getattr(kdata.header.acq_info.idx, label), idx)
   return self[mask]
   
 - reshape_idx_to_dim(label, dim):
   reshapes a based on a label into a new other dim
    
 - arange_idx_(label, dim)
   Another convenience function, filling in the acqidx with consecutive integers along a dimension.
  

  
Closes #382


fzimmermann89:
Not yet sure if the complexity of introducing the idea of a 'view' is worth it.

Advantages:
   - AcqIdx can be set based on a mask on kdata. 
   - Trajectory could  be changed manually for some subset, i.e.
      kdata.traj[1:] = kdata.traj[0].
      But maybe creating a new trajectory is easier to understand anyways...
   - This avoids the need for a setitem and associated complexity:
      One can justs get a view of kdata and set the properties of the view.

      

Disadvantage:
   - What about broadcasted values?
   - Rearrange will always return a copy. (?)
   - The AcqIdx can also be set by masking on them directly. The only problem are 'convenience' functions

fzimmermann89:
while implementing the rearrange parser for einops, I was started to wonder if
    a) allowing the rearrange k0 is useful
    b) allowing ellipses on the k dims is useful
    c) automatically adjusting such that we always have 3 kdim is useful (or if an error should be thrown)

a-c) I think the answer is NO

  d) if we should support coil-> other reshapes as part of rearrange. I think not... not sure yet.
  e) if we enforce coil, k2, k1 names -> I think not.
  f ) if we allow ... | ..., i.e. an ellipsis on all k dim. -> I think yes if e) is no
  g) If we allow  to leave out "| ..." and default to rearrange only on the batch dims. -> maybe.

fzimmermann89:
superseeded

PR #245, opened 2024-04-10
mono-exponential decay model added

ckolbPTB:
Signal model for T2 or T2* mapping needed for the Challenge.

### DIFF
--- a/src/mrpro/operators/models/_MonoExponentialDecay.py
+++ b/src/mrpro/operators/models/_MonoExponentialDecay.py
@@ -1,0 +1,55 @@
+# Copyright 2023 Physikalisch-Technische Bundesanstalt
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at:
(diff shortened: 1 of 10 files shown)


src/mrpro/operators/models/_MonoExponentialDecay.py
```diff
@@ -0,0 +1,55 @@
...
+    def forward(self, m0: torch.Tensor, decay_constant: torch.Tensor) -> tuple[torch.Tensor,]:
+        """Apply mono-exponential signal model.
+
+        Parameters
+        ----------
+        m0
+            equilibrium signal / proton density
+            with shape (... other, coils, z, y, x)
+        decay_constant
+            exponential decay constant (e.g. T2 or T2*)
```
schuenke: maybe add T1rho as another example

schuenke (approved):

src/mrpro/operators/models/_MonoExponentialDecay.py
ckolbPTB: done

PR #248, opened 2024-04-12
Read in multiple dicom files 

ckolbPTB:
This allows us to read in multiple dicom files from different folders and combine them to a single IData object. 

Needed for the ISMRM qMRI challenge

### DIFF
--- a/src/mrpro/data/_IData.py
+++ b/src/mrpro/data/_IData.py
@@ -104,11 +104,12 @@ class IData(Data):
         filenames
             List of DICOM filenames.
         """
-        if len(list(filenames)) == 0:
-            raise ValueError('No dicom files specified')
-
         # Read in all files
         dataset_list = [dcmread(filename) for filename in filenames]
+
+        # We do the check here to allow for filenames to be a Generator
+        if not dataset_list:
+            raise ValueError('No dicom files specified')
 
         # Ensure they all have the same orientation (same (0019, 1015) SlicePosition_PCS tag)
         def get_unique_slice_positions(slice_pos_tag: TagType = 0x00191015):
--- a/src/mrpro/data/_IData.py
+++ b/src/mrpro/data/_IData.py
@@ -17,6 +17,7 @@
 from __future__ import annotations
 
 import dataclasses
+from collections.abc import Sequence
 from pathlib import Path
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


fzimmermann89 (commented):
Suggestion/question:

does
 from_dicom_files(path(folder).glob(*.dcm))
A) work type wise?
B)  work in practice? 
C) fail nicely if it is empty? 

.. And a sequence of string can be empty I think..?

fzimmermann89 (approved):

ckolbPTB:
A) No, Path.glob returns Generator[Path, None, None]
B) Yes
C) now it does

fzimmermann89:
...can we add Generator[Path, None, None] to the types? :D

ckolbPTB:
done

fzimmermann89:
if len(list(filenames)) == 0:
runs the generator till it is exhausted and does not return any more items

so the next time the generator is accessed within the function, it will not return any elements
you could, for example, do
```
filenames = tuple(fileanemes)
if not filenames:
 raise error
```

(btw: len==0 is redundant, as empty list is falsy..)

or undo and not allow the generator usage and force the caller to convert to tuple.
but it would be quite easy to not pay attention to the types and supply the generator and wonder why the function fails...

ckolbPTB:
whenever I try something like `filenames = tuple(fileanemes)` mypy complains because tuple/list... returns list[object] rather than list[str or Path] needed by dicomread.

fzimmermann89:
yes, sorry, you need to give it a new variable name.

https://mypy-play.net/?mypy=latest&python=3.12&gist=7312d8401e7df7fb70a4bd943ccf20bd

so filenames_tuple=tuple(filenames) or sth similar..

ckolbPTB:
yes, I did that but the problem is then later when calling dicomread because `filenames_tuple` will be of typ list[object]

fzimmermann89:
mhm, no idea why this is requiered, but what about
`    filenames_tuple:tuple[Path|str,...]=tuple(filenames)`
without the explicit type hint it doesnt works for if filenames can also be Sequence[str]...

ckolbPTB:
> filenames_tuple:tuple[Path|str,...]=tuple(filenames)

This also did not work leading to mypy errors: Incompatible types in assignment (expression has type "tuple[object, ...]", variable has type "tuple[str] | tuple[Path]")

My current solution is checking after trying to read the files.

fzimmermann89:
argh. when i tried it, i tried it with:
    filenames_tuple:tuple[str|Path]=(*filenames,)
which works.
but posting here, I decided to use tuple instead.. and did not check it again. No clue why this is a difference, must be something with actually iterating the sequences... ? (It also fails without the generator)

But your solution is nicer, feel free to merge.

Issue #247, opened 2024-04-12
Operators and elementwise multiplication


ckolbPTB:
I am trying to create an acquisition model for Golden Radial Phase Encoding (GRPE) and then using the pseudo-inverse (i.e. adjoint + dcf) to reconstruct and image x0. I tried three different cases of how to combine the operators with the dcf but only one of them works. The other two fail because the operators are always applied first and the dcf is then applied last, which means trying to apply a dcf onto an image.

```
# WORKS
AW = kdcf.data[:,None,...] * cart_sampling_op @ fourier_op @ csm_op
x0 = AW.H(kdata_resp_resolved.data)[0]

# FAILS
A = cart_sampling_op @ fourier_op @ csm_op
AW = kdcf.data[:,None,...] * A
x0 = AW.H(kdata_resp_resolved.data)[0]

# FAILS
A = cart_sampling_op @ fourier_op @ csm_op
AHW = A.H * kdcf.data[:,None,...]
x0 = AHW(kdata_resp_resolved.data)[0]
```
I am not sure if this is an expected outcome or if this is a bug. Any input welcome!

fzimmermann89:
yes, i know, forgot to push the fix for this issue....

fzimmermann89:
I changed the title as this concerns element wise multiplication.

Matrix Multiplication of an Operator is not directly supported, you need to wrap the Matrix as a Linear Operator, for example the einsum op and specify how to perform the multiplication

fzimmermann89:
Mypy will still cause issues due to pytorcs typehints on the multiplication.

In #183 I also added an overload to DCF data to be able to write
dcf@fourierop
no sure if this is nice, as it will only allow for left multiplication...

fzimmermann89:
solved by #262

PR #249, opened 2024-04-13
Operators: elementwise multiplication with non-scalar tensors and typing

fzimmermann89:
We only supported (i.e. tested) for multiplication with a scalar tensor.

This adds tests and logic for elementwise multiplication with nd-tensors.

Related #247 

and adds missing type hints.

### DIFF
--- a/src/mrpro/operators/_LinearOperator.py
+++ b/src/mrpro/operators/_LinearOperator.py
@@ -86,10 +86,10 @@ class LinearOperatorSum(LinearOperator, OperatorSum[torch.Tensor, tuple[torch.Te
 class LinearOperatorElementwiseProductRight(
     LinearOperator, OperatorElementwiseProductRight[torch.Tensor, tuple[torch.Tensor,]]
 ):
-    """Operator elementwise right multiplication with scalar/tensor."""
+    """Operator elementwise right multiplication with a tensor."""
 
     def adjoint(self, x: torch.Tensor) -> tuple[torch.Tensor,]:
-        """Adjoint Operator elementwise multiplication with scalar/tensor."""
+        """Adjoint Operator elementwise multiplication with a tensor."""
         if self._tensor.is_complex():
             return self._operator.adjoint(x * self._tensor.conj())
         return self._operator.adjoint(x * self._tensor)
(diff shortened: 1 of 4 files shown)


fzimmermann89:
(ignore the merge, it will be squashed anyways...)

src/mrpro/operators/_LinearOperator.py
```diff
@@ -82,8 +83,22 @@ def adjoint(self, x: torch.Tensor) -> tuple[torch.Tensor,]:
         return (self._operator1.adjoint(x)[0] + self._operator2.adjoint(x)[0],)
 
 
-class LinearOperatorElementwiseProduct(LinearOperator, OperatorElementwiseProduct[torch.Tensor, tuple[torch.Tensor,]]):
-    """Operator elementwise multiplication with scalar/tensor."""
+class LinearOperatorElementwiseProductRight(
+    LinearOperator, OperatorElementwiseProductRight[torch.Tensor, tuple[torch.Tensor,]]
+):
+    """Operator elementwise right multiplication with scalar/tensor."""
```
ckolbPTB: In the LinearOperator is the only time where you mention tensor and scalar as input. Nevertheless the signature only allows for tensors so I would suggest to remove scalar from the doc string.

tests/operators/test_operators.py
```diff
@@ -128,80 +133,114 @@ def test_sum_operator_linearoperator():
...
+    b = torch.tensor(value)
     c = b * a
-    x = torch.arange(10)
+    x = RandomGenerator(0).complex64_tensor(10)
     (y1,) = c(x)
-    y2 = a(x)[0] * b
+    y2 = b * a(x)[0]
 
     torch.testing.assert_close(y1, y2)
     assert isinstance(c, Operator), 'Operator * scalar should be an Operator'
```
ckolbPTB: To be super accurate this should be scaler * Operator and similarly for the other rproduct tests

ckolbPTB (approved):
Just a few small things, otherwise looks good.

fzimmermann89:
Mhm, this is unfortunatly currently not a viable approach because of mypy and pytorch:

torch has the typehint
`torch.Tensor._mul__(self,other:Any)->Tensor`

even thought, __mul__ does not work with our operators.
so for Tensor*Operator at runtim, our __rmul__ gets  called.
But mypy thinks Tensor.__mul__ will be called and return a Tensor.
So static type of the return does not match the actual type.

I opened an issue at pytorch https://github.com/pytorch/pytorch/issues/124015 . But not sure when and if this will ever get addressed...

Our code is correct, as far as I know. So I would still add it. But we should keep this issue and not use Tensor*operator in mypy checked code ... ...  (Which is annoying, as this is the  case we want to use it for..)

PR #253, opened 2024-04-17
Add .to/.cuda/cpu to KNoise

fzimmermann89:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #261
* #255
* #260
* #259
* #257
* #256
* #254
* __->__ #253
* #252

PR #259, opened 2024-04-17
fix imports and add py.typed marker

fzimmermann89:
* #261
* #255
* #260
* __->__ #259
* #257
* #256
* #254
* #253
* #252


fzimmermann89:
This allows us to use (outside mrpro)
```
import mrpro
mrpro.data.KData(...)
```
and similar with working pyright/vscode annotations.

I also alphabetically ordered some imports where the order now should no longer matter.

PR #267, opened 2024-04-17
fix imports and add py.typed marker

fzimmermann89:
The py.typed file marks our package as typed,
thus if it is imported into another package that uses mypy, mypy will use our type annotations

This should not matter for within mrpro, only if used in conjunction with other code.

https://peps.python.org/pep-0561/


fzimmermann89:
* #270
* #272
* #269
* #268
* __->__ #267

ckolbPTB (approved):
Could you please add a comment in this PR why the py.typed is needed.

fzimmermann89:
> Could you please add a comment in this PR why the py.typed is needed.

If I understand https://peps.python.org/pep-0561/ correctly, this is required for mypy running on other packages to know that our package is typed. It should not matter for within our code base, only if used outside of our environment via an import.

PR #271, opened 2024-04-17
Make TE, TR... to tensors in the header

ckolbPTB:
Not sure about this line here:
https://github.com/PTB-MR/mrpro/blob/9183c2e7573b17106b2f72034bf97429d573ea2f/src/mrpro/data/_IHeader.py#L94

Are we happy with an empty tensor or would it be better to return None?

### DIFF
--- a/src/mrpro/data/IHeader.py
+++ b/src/mrpro/data/IHeader.py
@@ -39,10 +39,10 @@ class IHeader(MoveDataMixin):
 
     # ToDo: decide which attributes to store in the header
     fov: SpatialDimension[float]
-    te: torch.Tensor
-    ti: torch.Tensor
-    fa: torch.Tensor
-    tr: torch.Tensor
+    te: torch.Tensor | None
+    ti: torch.Tensor | None
+    fa: torch.Tensor | None
+    tr: torch.Tensor | None
     misc: dict = dataclasses.field(default_factory=dict)
 
     @classmethod
@@ -89,24 +89,24 @@ class IHeader(MoveDataMixin):
             items = get_items_from_all_dicoms(name)
             return [float(val) if val is not None else None for val in items]
 
-        def make_unique(values: Sequence[float]):
+        def make_unique_tensor(values: Sequence[float]) -> torch.Tensor | None:
             """If all the values are the same only return one."""
             if any(val is None for val in values):
-                return []
+                return None
             elif len(np.unique(values)) == 1:
-                return [values[0]]
+                return torch.as_tensor([values[0]])
             else:
-                return values
+                return torch.as_tensor(values)
 
-        fa = torch.as_tensor(make_unique(get_float_items_from_all_dicoms('FlipAngle')))
-        ti = torch.as_tensor(make_unique(get_float_items_from_all_dicoms('InversionTime')))
-        tr = torch.as_tensor(make_unique(get_float_items_from_all_dicoms('RepetitionTime')))
+        fa = make_unique_tensor(get_float_items_from_all_dicoms('FlipAngle'))
+        ti = make_unique_tensor(get_float_items_from_all_dicoms('InversionTime'))
+        tr = make_unique_tensor(get_float_items_from_all_dicoms('RepetitionTime'))
 
         # get echo time(s). Some scanners use 'EchoTime', some use 'EffectiveEchoTime'
         te_list = get_float_items_from_all_dicoms('EchoTime')
         if all(val is None for val in te_list):  # check if all entries are None
             te_list = get_float_items_from_all_dicoms('EffectiveEchoTime')
-        te = torch.as_tensor(make_unique(te_list))
+        te = make_unique_tensor(te_list)
 
         fov_x_mm = get_float_items_from_all_dicoms('Rows')[0] * float(get_items_from_all_dicoms('PixelSpacing')[0][0])
         fov_y_mm = get_float_items_from_all_dicoms('Columns')[0] * float(
@@ -118,5 +118,5 @@ class IHeader(MoveDataMixin):
         # Get misc parameters
         misc = {}
         for name in MISC_TAGS:
-            misc[name] = make_unique(get_float_items_from_all_dicoms(MISC_TAGS[name]))
+            misc[name] = make_unique_tensor(get_float_items_from_all_dicoms(MISC_TAGS[name]))
         return cls(fov=fov, te=te, ti=ti, fa=fa, tr=tr, misc=misc)
--- a/src/mrpro/data/IHeader.py
+++ b/src/mrpro/data/IHeader.py
@@ -21,6 +21,7 @@ from collections.abc import Sequence
 from dataclasses import dataclass
 
 import numpy as np
+import torch
 from pydicom.dataset import Dataset
 from pydicom.tag import Tag
 from pydicom.tag import TagType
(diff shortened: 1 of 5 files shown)


src/mrpro/data/_KHeader.py
```diff
@@ -108,11 +108,11 @@ def from_ismrmrd(
...
 
         # Conversion functions for units
-        def sequence_ms_to_s(ms: Sequence[float]) -> Sequence[float]:
-            return [el / 1000 for el in ms]
+        def ms_to_s(ms: torch.Tensor) -> torch.Tensor:
+            return ms / 1000
 
-        def sequence_deg_to_rad(deg: Sequence[float]) -> Sequence[float]:
-            return [el * pi / 180 for el in deg]
+        def deg_to_rad(deg: torch.Tensor) -> torch.Tensor:
```
fzimmermann89: we can remove this and use torch.deg2rad

fzimmermann89: (or call deg2rad in this)

fzimmermann89:
> Not sure about this line here:
> 
> https://github.com/PTB-MR/mrpro/blob/9183c2e7573b17106b2f72034bf97429d573ea2f/src/mrpro/data/_IHeader.py#L94
> 
> Are we happy with an empty tensor or would it be better to return None?

Similiar question with KHeader (as the values are not mandatory in ismrmd headers)

Empty list was clear... Empty tensor maybe not?
I would be ok with them being None.

Disadvantage 
  all functions using header.ti etc have to deal with None or mypy will complain.

Advantage
   all functions using header.ti etc have to deal with None or mypy will complain.

ckolbPTB:
> all functions using header.ti etc have to deal with None or mypy will complain.

Wouldn't this mean that we would need something like:

```
class InversionRecovery(SignalModel[torch.Tensor, torch.Tensor]):
    def __init__(self, ti: float | torch.Tensor | None):
         if ti is None:
               raise ValueError
```
This does not look nice so maybe more of a disadvantage rather than advantage

fzimmermann89:
So far we do not test if the signal models work with empty tensors.
T1 would then return an empty tensor as well.
And the fourier operator has to work with an empty tensor etc...

So at some point we might just detect that and throw an exception?
Than mypy could at least warn check if we handle the None Case..

We could also say its neither the job of kdata nor of the signal model to check that ti is not None.
This should be handled by in the code creating the signal operator.
At least if one wants to write correct code that mypy checks.

So for simple reconstruction scripts we don't have to check this at all.
But for a good example or code within mrpro,we have to catch the case that ti is not set before creating the operator.
And mypy would remind us that we need to handle missing ti times somehow. 

So.... thinking about it, my preference would be:
- Set ti to None if missing
- Do not allow None as the ti argument in the signal models

fzimmermann89:
https://mypy-play.net/?mypy=latest&python=3.12&gist=2b31f07eb7db70180370831e13667cc8

but this will not work for "empty tensor"

ckolbPTB:
@fzimmermann89 is this roughly what you had in mind? 

For KHeader I added default values for all fields which are not mandatory in the ismrmrd header. Now this test https://github.com/PTB-MR/mrpro/blob/6338d8127eddb8f7df1ac4aa1aecdbde9d187f78/tests/data/test_kheader.py#L23
passes except for `trajectory`. `trajectory` is defined as `KTrajectoryCalculator` which of course will never be available from ISMRMRD. In our tests we are overwriting trajectory with `xsd.trajectoryType('other')` which seems to be a bug.

Do we really need KTrajectoryCalculator` in the header? We wrote all our code such that we don't need to know the trajectory...

ckolbPTB:
As far as I remember the reason to not use None in KHeader before was, that we said, we want to ensure that certain parameters are defined and we don't have to worry about them later. So we could go ahead with the current approach and allow for None or the other option would be to enforce that all values have to be set in IHeader (using also overwrite and defaults in the same way as for KHeader). Not sure which option I prefer....

fzimmermann89:
> @fzimmermann89 is this roughly what you had in mind?
> 
> For KHeader I added default values for all fields which are not mandatory in the ismrmrd header. Now this test
> 
> https://github.com/PTB-MR/mrpro/blob/6338d8127eddb8f7df1ac4aa1aecdbde9d187f78/tests/data/test_kheader.py#L23
> 
> 
> passes except for `trajectory`. `trajectory` is defined as `KTrajectoryCalculator` which of course will never be available from ISMRMRD. In our tests we are overwriting trajectory with `xsd.trajectoryType('other')` which seems to be a bug.
> Do we really need KTrajectoryCalculator` in the header? We wrote all our code such that we don't need to know the trajectory...

We don't __need__ it, but @schuenke and I thought it might be reallyuseful to know what Trajectory it actually is. 
Yes, the tests is wrong. at some point we should try to enable mypy in the tests as well...


 
 

> As far as I remember the reason to not use None in KHeader before was, that we said, we want to ensure that certain parameters are defined and we don't have to worry about them later. 

so ti would still be None, but we enforce to have a dicom with te, tr, flipangle set (or using an overwrite) ?

None is better than an empty tensor.
But I have no experience how often te,tr,,fa values are missing in dicoms.. Will this be annoying as we always have to guess them and potentially set them to the wrong values?

ckolbPTB:
> None is better than an empty tensor.

Agreed! and I would also say that None is better than a wrong value. 

> But I have no experience how often te,tr,,fa values are missing in dicoms

TE and FA are probably always set, TR maybe not for single-shot stuff like EPI. Main problem is TI. I would prefer to default TI to None if it is not defined rather than forcing the user to set it by hand to a (potentially) wrong value.

fzimmermann89:
So, do we enforce FA and TE.
And default to None for TR and T1?

ckolbPTB:
Maybe better to default to None for TE, TR, FA and TI. I don't see much benefit to treat TE and FA differently to TR and TI but it might lead to unforeseen restrictions (e.g. a spiral does not technically have a TE time, we usually "misuse" TE for the time between RF-pulse and start of the acquisition).

My suggestion would be to default to None for all of them. This would mean that apart from `KTrajectoryCalculator` we can create a KHeader from a minimal ISMRMRD file.

src/mrpro/data/_KHeader.py
```diff
@@ -56,15 +56,15 @@ class KHeader(MoveDataMixin):
...
-    n_coils: int
     acq_info: AcqInfo
-    datetime: datetime.datetime
     h1_freq: float
-    te: list[float]
-    ti: list[float]
-    fa: list[float]
-    tr: list[float]
-    echo_spacing: list[float]
+    n_coils: int | None = None
```
schuenke: we don't test for n_coils = None yet, do we?
Could imagine that this leads to problems...

schuenke (approved):

src/mrpro/data/_KHeader.py
ckolbPTB: Probably, but I don't think we are using this parameter anywhere at the moment.

src/mrpro/data/_KHeader.py
```diff
@@ -79,9 +79,11 @@ class KHeader(MoveDataMixin):
     trajectory_description: TrajectoryDescription = dataclasses.field(default_factory=TrajectoryDescription)
 
     @property
-    def fa_degree(self) -> list[float]:
+    def fa_degree(self) -> torch.Tensor | None:
         """Flip angle in degree."""
-        return [el / pi * 180 for el in self.fa]
+        if not self.fa:
```
fzimmermann89: is it on purpose that this will return None if the flip_angle is set to zero?

fzimmermann89: also, we have now the same if check twice.
can we make it 
```
if self.fa is None:
   warn
   return None
else:
    return torch.rad2deg(self.fa)
```

fzimmermann89 (dismissed):
see comments

fzimmermann89 (commented):
see comments

PR #270, opened 2024-04-17
add direct reconstruction

fzimmermann89:

PR #269, opened 2024-04-17
add fourierop.from_kdata

fzimmermann89:

PR #268, opened 2024-04-17
Improve docstrings and variable names

fzimmermann89:

PR #272, opened 2024-04-17
Allow multiple other dimensions in prewhiten data

fzimmermann89:
and use triangular solve instead of torch.inv as it is supposedly more numerically stable.


fzimmermann89:
* #270
* __->__ #272
* #269
* #268
* #267

tests/algorithms/test_prewhiten_kspace.py
```diff
@@ -32,19 +33,23 @@ def _test_prewhiten_kspace(random_kheader, device):
 
     # Dimensions
     n_coils = 4
-    n_kx = 128
+    ns_k2k1k0 = (4, 5, 32)
```
ckolbPTB: n_k2k1k0 and n_other?

tests/algorithms/test_prewhiten_kspace.py
```diff
@@ -32,19 +33,23 @@ def _test_prewhiten_kspace(random_kheader, device):
...
+    ns_other = (1,)
```
ckolbPTB: would make sense to test for n_other > 1 if this is the feature this PR adds

ckolbPTB (changes requested):

tests/algorithms/test_prewhiten_kspace.py
fzimmermann89: 🙈

ckolbPTB (approved):

PR #273, opened 2024-04-17
Magnitude and Phase operator

ckolbPTB:


fzimmermann89:
Do we want to always return real tensors from abs?

I have a soft preference for yes, and suggest a  x.abs().real  in the forward ( I think 'real' works on both real and complex tensors)

ckolbPTB:
> x.abs().real()

Also fine by me. Then I guess we should call the Operator `RealAbs` to make it clear to the users what it does.

Another option would be to have an Operator which has `func: Callable` as input for init and enables the user to apply any old function on the tensors. Then people could create their own Abs(), Real(), RealAbs()... operators. Not sure what the mathematical sensible name for that would be. `FunctionOperator()`?

fzimmermann89:
Ignore my remark. AFAIK abs*( always returns a real value anyway. Sorry

What about Magnitude instead of Abs?

Re custom operator:'
1. I would call it LambdaOp, similar to https://pytorch.org/vision/main/generated/torchvision.transforms.Lambda.html
2. This would be a different PR

I would still add a magnitude (and maybe also a phase) operator as they are of general use

fzimmermann89:
Small nitpick: the operators were sorted alphabetically in the init..

fzimmermann89 (approved):

PR #279, opened 2024-05-11
Unify .to / .cpu / .cuda for data classes

### LINKED ISSUE #186: Move AcqInfo to device
We should consider if calling KData.cuda etc should also move the AcqInfo, especially the indices, to the GPU.

### LINKED ISSUE #284: Header.to / .cpu / .cuda
We could consider also implementing .to/.cpu/.cuda for KHeader,IHeader, etc 

This should move all tensors to the device.

fzimmermann89:
Implements a new Mixin for moving tensor fields to a device.

Keeps complex data as complex and real as real, etc.
This now also moves the header tensors and all other dataclasses

Also fixes the typing issues

Closes #284 
Closes #186


fzimmermann89:
* #283
* #282
* #280
* __->__ #279

fzimmermann89:
Needs a test for 

- [x]  device
- [x] clone
- [x] overload 3 of .to (calling with a tensor as argument)

fzimmermann89:
I would like to have a consistant rule for the dtype conversion.
So for, the rule is: float is kept as float, complex is kept as  complex and int is kept as int, only the width for float and complex types is adjusted.

The only issue now is what do with an integer trajectory: A cartesian trajectory can currently be of dtype int.
if I call ktraj.to(torch.float64)  on it, it currently is kept as int. This is surprising.
I am wondering what to do here:
- Make the trajectory always a floating dtype? This makes the on_grid check a tiny bit less performant.
- Have a special overwrite for the .to method for the trajectory? This would make it inconsistent with the other dtype changes. Also, should this special rule only apply if the trajectory is explicitly converted or also if we convert a Kdata field to complex128... ?
- Accept it, it is currently documented in the docstring that the type of dtype cannot be changed by calling .to on the dataclass and an int->float conversion must be manually. This would mean changing the test only...

ckolbPTB:
I would prefer 

> Make the trajectory always a floating dtype? This makes the on_grid check a tiny bit less performant.

fzimmermann89:
fixed in #286 by converting an integer trajectory in the KTrajectory construction to float32

src/mrpro/data/_MoveDataMixin.py
```diff
@@ -0,0 +1,208 @@
...
+
+        The conversion will be applied to all Tensor fields of the dataclass,
+        and to all fields that implement the MoveDataMixin.
+
+        The dtype-type, i.e. float/complex/int will always be preserved.
+        Example:
+            If called with dtype=torch.float32 OR dtype=torch.complex64:
+                - A complex128 tensor will be converted to complex64
+                - A float64 tensor will be converted to float32
+                - A bool tensor will remain bool
```
ckolbPTB, line 100: can we add here 
- A int64 tensor will be converted to int32

src/mrpro/data/_MoveDataMixin.py
```diff
@@ -0,0 +1,208 @@
+from __future__ import annotations
```
ckolbPTB, line 17: header missing

src/mrpro/data/_MoveDataMixin.py
```diff
@@ -0,0 +1,208 @@
...
+        memory_format
+            The desired memory format of returned tensor.
+        """
+        return self.to(device='cpu', memory_format=memory_format)
+
+    @property
+    def device(self) -> torch.device | None:
+        """Return the device of the tensors.
+
+        Looks at each field of a dataclass and returns fields implementing a device attribute,
```
ckolbPTB: what do you mean with "fields implementing a device attribute"? Doesn't it return the common device attribute of the dataclass and raise an error of different attributes of the dataclass are on different devices?

src/mrpro/data/_MoveDataMixin.py
fzimmermann89: currently I do not touch int tensors at all.
Should we convert int32->int64? or keep it as is?
(either way, I would add it to the docstring)

fzimmermann89: I would prefer not to convert ints I think... for deep learning, half precision, i.e. float16/bfloat16, is interesting, but int16 could be too small for some indices?

ckolbPTB:
Do all the cuda-tests pass for you?

fzimmermann89:
now they do again.

ckolbPTB:
Could we also add an `is_cuda` and `is_cpu` attribute to the MoveDataMixing class?

fzimmermann89:
added them:
For inconsistant devices both return False.
For no tensor attributes, both return True,

I added tests for kdata only..

tests/data/test_kdata.py
```diff
@@ -184,6 +229,11 @@ def test_KData_cuda(ismrmrd_cart):
     assert kdata_cuda.traj.kz.is_cuda
     assert kdata_cuda.traj.ky.is_cuda
     assert kdata_cuda.traj.kx.is_cuda
+    assert kdata_cuda.header.acq_info.user_int.is_cuda
```
ckolbPTB, line 240: could we add here:
`assert kdata_cuda.header.acq_info.patient_table_position.is_cuda`
to make sure it also work for `SpatialDimension`

tests/data/test_kdata.py
```diff
@@ -195,6 +245,40 @@ def test_KData_cpu(ismrmrd_cart):
     assert kdata_cpu.traj.kz.is_cpu
     assert kdata_cpu.traj.ky.is_cpu
     assert kdata_cpu.traj.kx.is_cpu
+    assert kdata_cpu.header.acq_info.user_int.is_cpu
```
ckolbPTB, line 256: `assert kdata_cuda.header.acq_info.patient_table_position.is_cpu`

ckolbPTB:
Could we add for IData, KNoise and QData in each `test_****_cuda` and `test_****_cpu` the lines `assert idata_cuda.is_cuda`, `assert idata_cuda.is_cpu`, `assert qdata_cuda.is_cpu`, `assert qdata_cuda.is_cuda` and so on?

ckolbPTB:
or maybe even replace e.g. `assert qdata_cuda.data.is_cuda` with `assert qdata_cuda.is_cuda`

src/mrpro/data/_MoveDataMixin.py
```diff
@@ -0,0 +1,394 @@
...
+            copy: bool = False,
+            memory_format: torch.memory_format = torch.preserve_format,
+        ) -> parsedType:
+            return device, dtype, non_blocking, copy, memory_format
+
+        if args and isinstance(args[0], torch.Tensor) or 'other' in kwargs:
+            # overload 3
+            device, dtype, non_blocking, copy, memory_format = parse3(*args, **kwargs)
+        elif args and isinstance(args[0], torch.dtype):
+            # overload 2
```
ckolbPTB: overload 1?

ckolbPTB: or what does 1,2,3 stand for?

src/mrpro/data/_MoveDataMixin.py
```diff
@@ -0,0 +1,394 @@
...
+            for field in dataclasses.fields(self):
+                name = field.name
+                data = getattr(self, name)
+                yield name, data
+        if isinstance(self, torch.nn.Module):
+            yield from self._parameters.items()
+            yield from self._buffers.items()
+            yield from self._modules.items()
+
+    def _to(
```
ckolbPTB, line 164: doc string missing?

src/mrpro/data/_MoveDataMixin.py
```diff
@@ -0,0 +1,394 @@
...
+        memory_format
+            The desired memory format of returned tensor.
+        copy
+            If True, the returned tensor will always be a copy, even if the input was already on the correct device.
+            This will also create new tensors for views
+        """
+        return self._to(dtype=torch.float64, memory_format=memory_format, copy=copy)
+
+    def half(self, *, memory_format: torch.memory_format = torch.preserve_format, copy: bool = False) -> Self:
+        """Convert all float tensors to haf precision.
```
ckolbPTB: half not haf

tests/data/test_movedatamixin.py
```diff
@@ -0,0 +1,115 @@
...
+        assert new_data is not original_data, 'copy requested but not performed'
+    assert torch.equal(new_data, original_data.to(device=expected_device, dtype=expected_dtype))
+    assert new_data.device == expected_device, 'device not set correctly'
+    assert new_data.dtype == expected_dtype, 'dtype not set correctly'
+
+
+@pytest.mark.parametrize('dtype', [torch.float64, torch.complex128])
+@pytest.mark.parametrize('copy', [True, False])
+def test_movedatamixin_float64like(copy: bool, dtype: torch.dtype):
+    original = B()
```
ckolbPTB, line 93: doc string missing

ckolbPTB (approved):

PR #280, opened 2024-05-11
Add EndomorphOperator

### LINKED ISSUE #274: Type hints for operators: n-tensor -> n-tensor
It would be nice to have a specialized Operator class typed to have the same number of inputs as outputs.

This is somewhat difficult to achieve with the current type system.
Either, it is easy to type it as 'same input as output', OR 'all inputs are of this type'. Both is a bit complicated,
https://mypy-play.net/?mypy=latest&python=3.12&gist=d7f6fd4dcbab1e02140c1391e767eeca

The workaround would be to add typing `@overloads` for
1-Tensor -> 1-Tensor
2-Tensors -> 2-Tensors
etc, up to some arbitrary number (10?)

A question is how to name these kind of operators?
EndomorphOperator ? @koflera

fzimmermann89:
Operators with the same number of inputs as outputs

Fixes #274 

As we currently do not have sufficiently sophisticated typing tests that allows us to test for expected type errors, this does not contain any new tests as all functionality is covered by existing tests.


fzimmermann89:
* #283
* #282
* __->__ #280
* #279

fzimmermann89:
An endomorph operator has the same number of inputs as outputs.

As there is currently no good way to have a typevartuple constraint to torch.tensors, this is done by overloading the forward with 1-Tensor -> 1-Tensor, 2-Tensor->2-Tensor etc up to 10-Tensors. More than 10 Tensors as input will thus be considered as resulting in more than 10 Tensors as output, but without a hint for the exact number.

As overloads are not inherited, the overloads would need to be added to each child again.
To avoid this, instead I use a `@endomorph` decorator that adds the overloads to a method. 

Also, matmul has type hints such that EndomorphOp@OtherOp will result in the same output as OtherOp.

This is curently used for the magnitude, phase and constraint ops.

koflera (approved):
Looks ok to me, but the content is faaaar from my area of expertise ;)

PR #282, opened 2024-05-12
Fix types

fzimmermann89:
This fixes all type issues in tests and some other inconsistencies.
- Make the KData mixins return a Self (i.e. KData). So, for example, KData.remove_oversampling will now result in KData (used to be Any)
- Fix issues caused by using numpy in the rotation tests. Widen allowed types in Rotation from only tensor inputs to also allow nested sequences of floats.
- Add type ignores for paths that can never be reached according to the type hints but are tested anyways
- Allow a single int as size for RandomGenerator
- Make return_sensitivity in Rotation.align_vectors a keyword-only argument, and overload the return type conditional on its value


fzimmermann89:
* #283
* __->__ #282
* #280
* #279

src/mrpro/phantoms/_EllipsePhantom.py
```diff
@@ -91,7 +91,7 @@ def image_space(self, image_dimensions: SpatialDimension[int]) -> torch.Tensor:
         """
         # Calculate image representation of phantom
         ny, nx = image_dimensions.y, image_dimensions.x
-        ix, iy = torch.meshgrid(
+        ix, it = torch.meshgrid(
```
ckolbPTB: why `it`? also in line 104 below

ckolbPTB (changes requested):

src/mrpro/phantoms/_EllipsePhantom.py
fzimmermann89: spelling autocorrect fail of the precommit .... iy->it
we should add this as an exception....

fzimmermann89: should be fixed now

ckolbPTB (approved):

PR #283, opened 2024-05-12
Precommit: run mypy on all files, including tests.

fzimmermann89:
Runs mypy on all files. As changes to one file can cause typing issues in unchanged files.


fzimmermann89:
* __->__ #283
* #282
* #280
* #279

fzimmermann89:
As changes to a file can also lead to type errors in other files, only checking committed files does not help.
Also, a lot of non-commited files will be parsed by mypy anyways, if they are imported. Overall, mypy suggested to run on all files
https://github.com/python/mypy/issues/13916

As #282 fixes all remaining issues in the tests, we can now also check the tests.

fzimmermann89:
Mypy cannot infer that the nullcontext or pytest.warns condition always results in a context manager. 

This either needs a type hint or a different logic. 

https://github.com/python/mypy/issues/10109

--> fixed in #282

schuenke (approved):
lgtm

fzimmermann89:
The only issue now is, that if you have some non-staged files with typing issues in the src directory,
the precommit will fail...

not sure if there is an easy way to run on  'staged and already committed files' in precommit..

PR #286, opened 2024-05-14
Make trajectory always float

### LINKED ISSUE #285: Rounding vs truncating of trajectory
In https://github.com/PTB-MR/mrpro/blob/d15579a27c650131d94ca9b058f4d6583293b4d4/src/mrpro/operators/_CartesianSamplingOp.py#L62 we truncate, not round the trajectory. so 0.99 -> 0.
Also in  https://github.com/PTB-MR/mrpro/blob/67a31834ea6d7adebb757de03818c488230c269b/src/mrpro/data/_KTrajectory.py#L159 we check on_grid by the fraction, so 0.999 will not be considered as on_grid.

We should change both.

fzimmermann89:
and fix #285

### DIFF
--- a/src/mrpro/data/_KTrajectory.py
+++ b/src/mrpro/data/_KTrajectory.py
@@ -55,11 +55,15 @@ class KTrajectory:
 
     def __post_init__(self) -> None:
         """Reduce repeated dimensions to singletons."""
+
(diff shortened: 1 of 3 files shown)


fzimmermann89:
* #283
* #282
* #280
* #279
* __->__ #286

ckolbPTB (approved):

PR #287, opened 2024-05-15
Add Iterative SENSE Reconstruction

ckolbPTB:
A first start on an iterative SENSE algorithms. Needs some more design optimisation to avoid code duplication between `DirectReconstruction` and `IterativeSenseReconstruction`. Maybe we should also add a Tikonov-regularisation option already here...


fzimmermann89:
Will we have a separate class for regularized iterative sense? will we add regularization in a second PR to this?

ckolbPTB:
> Will we have a separate class for regularized iterative sense? will we add regularization in a second PR to this?

It will be a separate PR (after solving #288). 
My suggestion would be to only have one IterativeSense-class which also can do regularisation.

koflera:
> > Will we have a separate class for regularized iterative sense? will we add regularization in a second PR to this?
> 
> It will be a separate PR (after solving #288). My suggestion would be to only have one IterativeSense-class which also can do regularisation.

I think I would separate them. If we say that IterativeSENSE is just a special case of a regularized IterativeSENSE for which lambda=0, one could as well say that it is actually the special case of another even more general functional that involves other regularization terms for which the regularization parameter is also set to zero.

For example, one of the sub-probems to be solved when using dictionary learning and sparse coding could be seen as a generalization of the Tikhonov with as many L2-regularization terms as there are patches to be extracted from the image. Still I would refer to that problem formulation as something that has to do with dictionary learning instead of IterativeSENSE.

examples/iterative_sense_reconstruction.ipynb
```diff
@@ -0,0 +1,303 @@
...
+    "Let's assume we have obtained the k-space data $y$ from an image $x$ with an acquisition model (Fourier transforms,\n",
+    "coil sensitivity maps...) $A$ then we can formulate the forward problem as:\n",
+    "\n",
+    "$ y = Ax + n $\n",
+    "\n",
+    "where $n$ describes complex Gaussian noise. Now we want to solve the inverse problem by minimizing\n",
+    "\n",
+    "$ \\min_x \\frac{1}{2}||W^{\\frac{1}{2}}(Ax - y)||_2^2 $\n",
+    "\n",
+    "where $W^\\frac{1}{2}$ is the square root of the density compensation function. We can rewrite this problem as:\n",
```
koflera: I would slightly change the text to: 

"... is the square root of the density compensation function (which corresponds to a diagonal operator)". Setting the derivative of the functional $F$ to zero and rearranging yields

"$ A^H W A x = A^H W y$\n"

which is a linear system $Hx=b$ that needs to be solved for $x$.

examples/iterative_sense_reconstruction.ipynb
```diff
@@ -0,0 +1,303 @@
...
+    "IterativeSenseReconstruction solves the following reconstruction problem:\n",
+    "\n",
+    "Let's assume we have obtained the k-space data $y$ from an image $x$ with an acquisition model (Fourier transforms,\n",
+    "coil sensitivity maps...) $A$ then we can formulate the forward problem as:\n",
+    "\n",
+    "$ y = Ax + n $\n",
+    "\n",
+    "where $n$ describes complex Gaussian noise. Now we want to solve the inverse problem by minimizing\n",
+    "\n",
+    "$ \\min_x \\frac{1}{2}||W^{\\frac{1}{2}}(Ax - y)||_2^2 $\n",
```
koflera: maybe give the functional 1/2 ||W^{\\frac{1}{2}}(Ax - y)||_2^2 a name F(x) such that you can reference it below (as in the next comment).

examples/iterative_sense_reconstruction.ipynb
```diff
@@ -0,0 +1,303 @@
...
+    ")\n",
+    "img = iterative_sense_reconstruction(kdata)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "48f03f8c",
+   "metadata": {},
+   "source": [
+    "### Behind the scenes"
```
koflera, line 113: Maybe mention that the density compensation operator is first estimated based on the trajectories and that needs to be set up.

examples/iterative_sense_reconstruction.ipynb
```diff
@@ -0,0 +1,303 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "c65f82b1",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "cg_operator = acquisition_operator.H @ dcf_operator @ acquisition_operator"
```
koflera: i think I would just call it `operator `instead of `cg_operator `

examples/iterative_sense_reconstruction.ipynb
```diff
@@ -0,0 +1,303 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "41d178f6",
+   "metadata": {},
+   "source": [
+    "##### Conjugate gradient minimisation"
```
koflera: maybe "run cg" instead of "cg minimization"

src/mrpro/algorithms/reconstruction/IterativeSenseReconstruction.py
```diff
@@ -0,0 +1,148 @@
+"""Iterative SENSE Reconstruction by Adjoint Fourier Transform."""
```
koflera: by adjoint fourier transform?

src/mrpro/algorithms/reconstruction/IterativeSenseReconstruction.py
```diff
@@ -0,0 +1,148 @@
...
+        csm: CsmData | None = None,
+        noise: None | KNoise = None,
+        dcf: DcfData | None = None,
+    ) -> None:
+        """Initialize DirectReconstruction.
+
+        Parameters
+        ----------
+        fourier_op
+            Instance of the FourierOperator which adjoint is used for reconstruction.
```
koflera: "whose adjoint" instead of "which adjoint"

src/mrpro/algorithms/reconstruction/IterativeSenseReconstruction.py
```diff
@@ -0,0 +1,148 @@
...
+        (right_hand_side,) = operator.to(device)(kdata.data)
+
+        # Create H = A^H W A
+        operator = operator @ self.fourier_op
+        if self.csm is not None:
+            operator = operator @ self.csm.as_operator()
+        operator = operator.to(device)
+
+        img_tensor = cg(
+            operator, right_hand_side, initial_value=right_hand_side, max_iterations=self.n_iterations, tolerance=0.0
```
koflera: do we prefer the fixed number of iterations here or would it be better touse a tolerance?

koflera (commented):
Looks good to me. I only have few comments regarding the derivation of why we use CG to so solve that particular system.

src/mrpro/algorithms/reconstruction/IterativeSenseReconstruction.py
ckolbPTB: Commonly we want to do early stopping with iterative SENSE. IMO we want to run the algorithm for a set number of iterations rather than up to a certain tolerance.

src/mrpro/algorithms/reconstruction/IterativeSenseReconstruction.py
```diff
@@ -0,0 +1,148 @@
...
+        super().__init__()
+        self.fourier_op = fourier_op
+        self.n_iterations = n_iterations
+        # TODO: Make this buffers once DataBufferMixin is merged
+        self.csm = csm
+        self.noise = noise
+        self.dcf = dcf
+
+    @classmethod
+    def from_kdata(cls, kdata: KData, n_iterations: int, noise: KNoise | None = None) -> Self:
```
fzimmermann89: kdata: KData, noise: KNoise | None = None, *, n_iterations: int = 10

a) would be nice if all reconstructions.from_kdaa could work with just kdata
b) n_iterations should be last parameter, similar to combine_coils in direct reconstruction
c) n_iterations (and combine_coils) should maybe be kwargs only. ( "*" in between)

examples/iterative_sense_reconstruction.py
```diff
@@ -0,0 +1,140 @@
...
+img_manual = mrpro.algorithms.optimizers.cg(
+    operator, right_hand_side, initial_value=right_hand_side, max_iterations=4, tolerance=0.0
+)
+
+
+# %%
+# For comparison we can also carry out a direct reconstruction
+direct_reconstruction = mrpro.algorithms.reconstruction.DirectReconstruction.from_kdata(kdata)
+img_direct = direct_reconstruction(kdata).rss().cpu()
+img_direct = img_direct
```
fzimmermann89: ?

examples/iterative_sense_reconstruction.py
```diff
@@ -0,0 +1,140 @@
...
+
+# %%
+# Display the reconstructed image
+import matplotlib.pyplot as plt
+
+fig, ax = plt.subplots(1, 3, squeeze=False)
+ax[0, 0].imshow(img_direct.data[0, 0, :, :])
```
fzimmermann89: remove .data, use img_direct[0, 0, :, :]

examples/iterative_sense_reconstruction.py
```diff
@@ -0,0 +1,140 @@
...
+ax[0, 0].set_title('Direct Reconstruction', fontsize=10)
+ax[0, 1].imshow(torch.abs(img.data[0, 0, 0, :, :]))
```
fzimmermann89: rss returns a magnitude tensor.

img_direct is a tensor in this example, but img is an IData.
this should be the same in both cases 
consider moving both RSS calls to here, and do  imshow(img.rss[0, 0, 0, :, :])) and imshow(img_direct.rss[0, 0, 0, :, :]))
(remove the RSS above)

examples/iterative_sense_reconstruction.py
```diff
@@ -0,0 +1,140 @@
...
+fname = 'pulseq_radial_2D_402spokes_golden_angle_with_traj.h5'
+# %%
+# Download raw data
+import tempfile
+from pathlib import Path
+
+import requests
+
+data_folder = Path(tempfile.mkdtemp())
+data_file = tempfile.NamedTemporaryFile(dir=data_folder, mode='wb', delete=False, suffix='.h5')
```
fzimmermann89: why do we need data_folder = Path(tempfile.mkdtemp()) ?
tempfile.NamedTemporaryFile wihout the optional dir already uses the temp dir..

src/mrpro/algorithms/reconstruction/IterativeSenseReconstruction.py
```diff
@@ -0,0 +1,148 @@
...
+from mrpro.operators.FourierOp import FourierOp
+from mrpro.operators.LinearOperator import LinearOperator
+
+
+class IterativeSenseReconstruction(Reconstruction):
+    r"""Iterative SENSE reconstruction.
+
+    This algorithm minizes the problem
+
+    .. math:: min_x \\frac{1}{2}||W^\\frac{1}{2} (Ax - y)||_2^2
```
fzimmermann89: remove \\frac{1}{2} ?

examples/iterative_sense_reconstruction.py
```diff
@@ -0,0 +1,140 @@
...
+# IterativeSenseReconstruction solves the following reconstruction problem:
+#
+# Let's assume we have obtained the k-space data $y$ from an image $x$ with an acquisition model (Fourier transforms,
+# coil sensitivity maps...) $A$ then we can formulate the forward problem as:
+#
+# $ y = Ax + n $
+#
+# where $n$ describes complex Gaussian noise. The image $x$ can be obtained by minimising the functionl $F$
+#
+# $ x = \argmin_x F(x) $
```
fzimmermann89: is \argmin is not valid latex

examples/iterative_sense_reconstruction.py
```diff
@@ -0,0 +1,140 @@
...
+# ##### $H = A^H W A$
+
+# %%
+operator = acquisition_operator.H @ dcf_operator @ acquisition_operator
+
+# %% [markdown]
+# ##### Run conjugate gradient
+
+# %%
+import torch
```
fzimmermann89: why?

examples/iterative_sense_reconstruction.py
```diff
@@ -0,0 +1,140 @@
...
+ax[0, 2].imshow(torch.abs(img_manual[0, 0, 0, :, :]))
+ax[0, 2].set_title('"Manual" Iterative SENSE', fontsize=10)
+
+# %% [markdown]
+# ### Check for equal results
+# The two versions result should in the same image data.
+
+# %%
+# If the assert statement did not raise an exception, the results are equal.
+assert torch.allclose(img.data, img_manual)
```
fzimmermann89, line 131: hold with impor ttorch till here

examples/iterative_sense_reconstruction.py
ckolbPTB: In order to delete the folder at the end of the script (this was missing). Alternative would be to use something like shutil.rmtree(os.path.dirname(data_folder)) or simply just delete the file and leave the empty folder. Any preferences?

fzimmermann89:
a named tempory file is automatically deleted when the script quits.

no need for any other cleanup, deletion, creation, etc?

(so it can also be deleted at least from the direct reconstruction example if I am not mistaken?)

examples/iterative_sense_reconstruction.py
fzimmermann89: just don't create the folder!

data_file = tempfile.NamedTemporaryFile(mode='wb', suffix='.h5', delete=False)

or what am I missing here?

examples/iterative_sense_reconstruction.py
ckolbPTB: wouldn't imshow(img.rss[0, 0, 0, :, :])) be confusing because it would suggest that img is a coil-resolved image (same actually applies to img_direct)? Would 

```
img_direct = direct_reconstruction(kdata)

# %%
# Display the reconstructed image

fig, ax = plt.subplots(1, 3, squeeze=False)
ax[0, 0].imshow(torch.abs(img_direct.data[0, 0, :, :]))
ax[0, 0].set_title('Direct Reconstruction', fontsize=10)
ax[0, 1].imshow(torch.abs(img.data[0, 0, 0, :, :]))
ax[0, 1].set_title('Iterative SENSE', fontsize=10)
ax[0, 2].imshow(torch.abs(img_manual[0, 0, 0, :, :]))
ax[0, 2].set_title('"Manual" Iterative SENSE', fontsize=10)
```

not be clearer for users what is a tensor and what is a MRpro object?

examples/iterative_sense_reconstruction.py
fzimmermann89: and to delete it it would be `data_file.delete`

ckolbPTB: I hadn't realised that tempfile folders and files behave differently and that files are deleted automatically whereas folders most of the time are not (except when used in a context)

fzimmermann89:
but stupid questions: 
why does writing  data_file.write(response.content) not require a data.flush() afterwards?
in theory you would have to flush before being sure that it as actually written to disc...

fzimmermann89:
the flush is requiered

` python -c "import tempfile; import pathlib; a=tempfile.NamedTemporaryFile(); a.write(bytes('1'*500, 'utf-8'));import numpy; print(numpy.loadtxt(a.name)>0);" `
fails

also, without the last few bytes might be missing from the h5. this will be difficult to tourbleshoot

examples/iterative_sense_reconstruction.py
fzimmermann89: rss of a single-coil image is the magnitude of the image...

so I would always do img_direct.rss() instead of img_direct.data.abs() (which, in turn, I would prefer over torch.abs(img_direct.data) )

RSS is a function of IData exactly for plotting..

fzimmermann89: and IData always has the coil dimension....

fzimmermann89:
And isn't it named "SENSE" not sense or Sense?
So the Classes and functions should also be capitalized as SENSE ? 

PEP8: When using acronyms in CapWords, capitalize all the letters of the acronym. Thus HTTPServerError is better than HttpServerError.

src/mrpro/algorithms/reconstruction/IterativeSenseReconstruction.py
```diff
@@ -0,0 +1,148 @@
...
+from mrpro.data._kdata.KData import KData
+from mrpro.data.CsmData import CsmData
+from mrpro.data.DcfData import DcfData
+from mrpro.data.IData import IData
+from mrpro.data.KNoise import KNoise
+from mrpro.operators.FourierOp import FourierOp
+from mrpro.operators.LinearOperator import LinearOperator
+
+
+class IterativeSenseReconstruction(Reconstruction):
```
fzimmermann89: IterativeSENSEReconstruction ?

fzimmermann89:
Also, should this perform k-space filtering at the end? (see Prussmann 2001)
Mid-term, it might be nice to have a k-space filtering function that somehow detects the k-space covered by the trajectory and filters outside of it

ckolbPTB:
> Also, should this perform k-space filtering at the end? (see Prussmann 2001)

This seems like a post-processing step to me and I think we should not do this, because the definition of the "k-space covered by the trajectory" is not that easy. For a 2D radial or spiral it might be, but for a random 3D Cartesian trajectory less so. I added a comment in the doc-string.

ckolbPTB:
> the flush is requiered

Fixed for existing examples separately in #374

src/mrpro/algorithms/reconstruction/IterativeSENSEReconstruction.py
```diff
@@ -0,0 +1,130 @@
...
+
+    Note: In [PRU2001]_ a k-space filter is applied as a final step to null all k-space values outside the k-space
+    coverage. This is not done here.
+
+    .. [PRU2001] Pruessmann K, Weiger M, Boernert P, and Boesiger P (2001), Advances in sensitivity encoding with
+       arbitrary k-space trajectories. MRI 46, 638-651. https://doi.org/10.1002/mrm.1241
+
+    """
+
+    n_max_iter: int
```
koflera: this is not used anywhere? also, if I remember correctly, it should be max_iterations

src/mrpro/algorithms/reconstruction/IterativeSENSEReconstruction.py
```diff
@@ -0,0 +1,130 @@
...
+
+        Returns
+        -------
+            the reconstruced image.
+        """
+        device = kdata.data.device
+        if self.noise is not None:
+            kdata = prewhiten_kspace(kdata, self.noise.to(device))
+
+        # Create A^H W
```
koflera: Could lines 110-124 be simplified to something like the following?

```
forward_operator = self.csm.as_operator() @ self.fourier_op if self.csm is not None else self.fourier_op

if self.dcf is not None:
    dcf_operator = self.dcf.as_operator()
    (right_hand_side,) = forward_operator.to(device).H(dcf_operator(kdata.data)[0])
    operator = forward_operator.H @ dcf_operator @ forward_operator
else:
    (right_hand_side,) = forward_operator.to(device).H(kdata.data)
    operator = forward_operator.H @ forward_operator
```

src/mrpro/algorithms/reconstruction/IterativeSENSEReconstruction.py
```diff
@@ -0,0 +1,130 @@
...
+        (right_hand_side,) = operator.to(device)(kdata.data)
+
+        # Create H = A^H W A
+        operator = operator @ self.fourier_op
+        if self.csm is not None:
+            operator = operator @ self.csm.as_operator()
+        operator = operator.to(device)
+
+        img_tensor = cg(
+            operator, right_hand_side, initial_value=right_hand_side, max_iterations=self.n_iterations, tolerance=0.0
```
koflera, line 125: do we want to have `max_iterations `and some tolerance or a fixed number of iterations?

examples/iterative_sense_reconstruction.py
```diff
@@ -0,0 +1,131 @@
...
+    kdata, n_iterations=4
+)
+img = iterative_sense_reconstruction(kdata)
+
+
+# %% [markdown]
+# ### Behind the scenes
+
+# %% [markdown]
+# ##### $W$
```
koflera: Maybe some whole sentence, like "compute the density compensation function to set up the operator $W$" or so?

examples/iterative_sense_reconstruction.py
```diff
@@ -0,0 +1,131 @@
...
+
+# %%
+# The density compensation operator is calculated based on the k-space locations of the acquired data.
+dcf_operator = mrpro.data.DcfData.from_traj_voronoi(kdata.traj).as_operator()
+
+
+# %% [markdown]
+# ##### $A$
```
koflera: Similar as above

examples/iterative_sense_reconstruction.py
```diff
@@ -0,0 +1,131 @@
...
+# Note that operators return a tuple of tensors, so we need to unpack it,
+# even though there is only one tensor returned from adjoint operator.
+img_coilwise = mrpro.data.IData.from_tensor_and_kheader(*fourier_operator.H(*dcf_operator(kdata.data)), kdata.header)
+csm_operator = mrpro.data.CsmData.from_idata_walsh(img_coilwise).as_operator()
+
+# Create the acquisition operator A
+acquisition_operator = fourier_operator @ csm_operator
+
+# %% [markdown]
+# ##### $b = A^H W y$
```
koflera: also here maybe a whole sentence instead of only the formula?

examples/iterative_sense_reconstruction.py
```diff
@@ -0,0 +1,131 @@
...
+
+# %%
+(right_hand_side,) = acquisition_operator.H(dcf_operator(kdata.data)[0])
+
+
+# %% [markdown]
+# ##### $H = A^H W A$
```
koflera: same as above

src/mrpro/algorithms/reconstruction/Reconstruction.py
```diff
@@ -18,3 +38,68 @@ def forward(self, kdata: KData) -> IData:
...
+        self.dcf = DcfData.from_traj_voronoi(kdata.traj)
+        return self
+
+    def recalculate_csm_walsh(self, kdata: KData, noise: KNoise | None | Literal[False] = None) -> Self:
+        """Update (in place) the CSM from KData using Walsh.
+
+        Parameters
+        ----------
+        kdata
+            KData used for pseudo-inverse reconstruction, which is then used for
```
koflera: pseudo inverse has a very specific meaning. For multi-coil acquisitions with enough sampling point, I think "pseudo-inverse" is correct, while for truly undersampled acquisitions, I doubt that we mean the pseudo-inverse but probably just the adjoint?

src/mrpro/algorithms/reconstruction/IterativeSENSEReconstruction.py
ckolbPTB: Good catch, should be `n_iterations`. I left it at `n_iterations` because IMO it is more likely to use iterative SENSE with a fixed number of iterations rather than being able to set some kind of tolerance.

src/mrpro/algorithms/reconstruction/IterativeSENSEReconstruction.py
ckolbPTB: see above

koflera:
Looks good to me!

src/mrpro/algorithms/reconstruction/IterativeSENSEReconstruction.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def __init__(
+        self,
+        fourier_op: LinearOperator,
+        n_iterations: int,
+        csm: CsmData | None = None,
+        noise: None | KNoise = None,
+        dcf: DcfData | None = None,
+    ) -> None:
+        """Initialize DirectReconstruction.
```
koflera: IterativeSENSEReconstruction instead of DirectReconstruction

src/mrpro/algorithms/reconstruction/IterativeSENSEReconstruction.py
```diff
@@ -0,0 +1,128 @@
...
+from mrpro.data.IData import IData
+from mrpro.data.KNoise import KNoise
+from mrpro.operators.FourierOp import FourierOp
+from mrpro.operators.LinearOperator import LinearOperator
+
+
+class IterativeSENSEReconstruction(Reconstruction):
+    r"""Iterative SENSE reconstruction.
+
+    This algorithm minizes the problem :math:`min_x \frac{1}{2}||W^\frac{1}{2} (Ax - y)||_2^2`
```
koflera: typo "minizes"... maybe also better "solves the problem"

src/mrpro/algorithms/reconstruction/IterativeSENSEReconstruction.py
```diff
@@ -0,0 +1,128 @@
...
+        self.dcf = dcf
+
+    @classmethod
+    def from_kdata(cls, kdata: KData, noise: KNoise | None = None, *, n_iterations: int = 10) -> Self:
+        """Create a IterativeSENSEReconstruction from kdata with default settings.
+
+        Parameters
+        ----------
+        kdata
+            KData to use for trajektory and header information
```
koflera: trajeKtory

koflera (approved):
Looks good to me!

PR #293, opened 2024-05-23
Add pgd (fista) algorithm

guastinimara:
I implemented the fista algorithm from https://www.ceremade.dauphine.fr/~carlier/FISTA.

The function fista is with fixed stepsize (given as input) for all iterations. The function fista_with_backtracking updates the stepsize scalar at each iteration following the rule explained in the paper.

I haven't tested it yet because some parts are still missing (functionals, prox map), so I don't know if there are some errors. In the end as comments I put some ideas for some test examples: I copied the wavelet case from the slides just to have it here and then I created a blurring operator to solve a deblurring problem for testing.


fzimmermann89:
suggestions/questions:
- would it make sense to just add a backtracking:bool to Fista? most of the code is the same.
- in the backtracking case, I think there is 2x the amount of work done:
   you can do the step with the stepsize, and only backtrack if the stepsize was to large.
   if it works out, you can continue.
  - I think torch.func.grad has a nicer interface than torch.autograd.grad
  - consider reusing the gradient if y is not updated (in Q)
 
so if would suggest to do inside fista 
f_grad = torch.func.grad(f)
calculate f_grad_y = f_grad(y)
calculate prox_y
if backtracking:
   while not criterion
       update stepsize and  prox_y..

fzimmermann89:
```
import torch
import math
from mrpro.operators._Functional import ProximableFunctional, Functional

def fista(
    f: Functional,
    g: ProximableFunctional,
    initial_value: torch.Tensor, 
    stepsize: float = 1., 
    max_iterations: int = 128,
    backtrack_factor = 1.
) -> torch.Tensor:
    """FISTA algorithm for solving problem min_x f(x) + g(x).
    This relies on the implementation of the proximal map of g.
    
    stepsize is updated in each iteration.
    
    Parameters
    ----------
    f
        convex, differentiable functional
    g
        convex, non-smooth functional with computable proximal map
    initial_value
        initial value for the solution of the algorithm
    stepsize
        initial stepsize for gradient step
    max_iterations
        maximal number of iterations
    backtrack_factor
        must be <=1. if <1., Backtracking rule for stepsize following https://www.ceremade.dauphine.fr/~carlier/FISTA
        is used

    Returns
    -------
        an approximate solution of the minimization problem
    """
    backtracking = not math.isclose(backtrack_factor,1)
    
    x_old = initial_value
    y = initial_value
    t_old = 1.

    for _ in range(max_iterations):
      while stepsize>1e-30:
          f_y, vjp = torch.func.vjp(f,y)
          grad = vjp(torch.ones_like(f_y))
          x = g.prox(y - stepsize * grad, stepsize)
          if not backtracking:
            # no need to check stepsize, continue to next iteration 
            break
          difference = x - y
          Q = f_y+ torch.vdot(difference.flatten(),grad.flatten()).real + 1/(2*stepsize) * difference.abs().square().sum()
          if f(x) <= Q:
            # stepsize is ok, continue to next iteration
            break
          stepsize*=backtrack_factor
      else:
        if backtracking:
          raise RuntimeError("After backtracking, the stepsize became to small.")
        else:
          raise RuntimeError("Stepsize to small.")


        t = (1 + math.sqrt(1 + 4*t_old**2))/2

        # update the solution
        y = x + (t - 1.)/t_update * (x - x_old)

        # update x and stepsize t
        x_old = x
        t_old = t

    return y
```

ckolbPTB:
Add test for fista, maybe denoising of image which can be accurately described by wavelets

ckolbPTB:
This is how Felix fixed the problem with the autograd function:
```

    def grad(x ):
        x=x.clone().detach().requires_grad_(True)
        y=f(x)
        return torch.autograd.grad(y,x)[0]
```

guastinimara:
I checked that fista works on complex variables domain and I implemented it following these lines of code. I took some notes when I was checking fista for the complex domain, I add it here below just fyi. 

I have to work on the tests, I'm still a bit confused about on which image sample I could use it on (for it to be described by the wavelets)

Also, in the pdf I put some calculus on the L1 proximal map for complex domain, and I think that the implementation that we have may not be correct, as it just viewed as real numbers the complex one and computes the proximal on those. What do you think?

[Math_Fista.pdf](https://github.com/user-attachments/files/15605527/Math_Fista.pdf)

fzimmermann89:
We had this discussion briefly at the hackathon. The current L1 functional
should interpret complex numbers as R^2 (correct me if that is wrong)
so L1=abs(real) + abs(imag), not L1=abs(real+i*imag)
Andreas was softly arguing for this.


Maybe we can have either a) both, or, b), for complex numbers always do
L1=abs(real+i*imag) and have a Real and Imag Operator, such that if one
wants to do the R^2 interpretation,t hey can write L1(Real(x)) + L1(Imag(x)
?

The complex L1 proximal should be something like ModRelu instead of the
shoftshrink, not only copying the sign from the input to the output, but
the complex phase?

For the gradients, it should not matter? As the gradient step is the same
if the grad ist done in wirtinger complex, or R^2, correct?

So in the end it mostly matters how you actually define the objective we
are trying to solve..


Btw: skimming through the pdf, it looks like me guestimated interpretation
of Q was correct? :D

Am Do., 6. Juni 2024 um 09:43 Uhr schrieb Mara Guastini <
***@***.***>:

src/mrpro/algorithms/optimizers/_fista.py
```diff
@@ -113,46 +54,70 @@ def fista_with_backtracking(
...
+            # calculate the proximal gradient step
+            gradient = grad(f_y, y)
+            x = g.prox(y - stepsize * gradient, reg_parameter * stepsize)
+
+            if not backtracking:
+                # no need to check stepsize, continue to next iteration
+                break
+            difference = x - y
+            Q = f_y + 1 / (2 * stepsize) * difference.abs().square().sum()
+            if complex_input:
```
fzimmermann89: pretty sure this is not requiered, you can always do .real, even on real numbers.

src/mrpro/algorithms/optimizers/_fista.py
```diff
@@ -0,0 +1,204 @@
...
+#   limitations under the License.
+
+import math
+
+import torch
+
+from mrpro.operators._Functional import Functional
+
+
+def grad(f: Functional, x: torch.Tensor):
```
fzimmermann89: changes to grad_and_value, also returning y.

this saves you a reevaluation of f.

I would like to use as soon as possible [torch.func.grad_and_value](https://pytorch.org/docs/stable/generated/torch.func.grad_and_value.html) here. 
So we should  solve this once and for all for our operators, move to finufft, or at least fix torchkbnufft.

As an intermediate step, we could either add our own, slow, grad_and_value to utils:
```python
def grad_and_value(function, create_graph=False):
     def inner(x):
            if create_graph and x.requires_grad:
                  # shallow clone
                  xg=x.view_as(x))
            xg=x.detach().requires_grad_(True))
            y=function(xg)
            yg=y if isinstance(y,tensor) else y[0]
                 grad=torch.autograd.grad(yg,xc)[0]
            return grad,y
      return inner
```
(or similar, written on github...)

or use torch.autograd.functional.vjp (knowing that the objectives should always return a scalar tensor)

guastinimara:
Wait but if I look at the class L1Norm in its pull request it seems that in the forward it doesn't check whether the input is complex or not, and so if z = x + iy it just takes the abs of it, and not separately, or where should I look?

Yeah I would opt for the b) option but maybe this is something to discuss more, it would make sense to give the option to see them separately but personally I think I would prefer that by default it considers complex values as they are. Why did you opt for the R^2 interpretation in the end at the hackathon?

I just checked ModRelu because I had never used it, and yes it seems exactly it!

The gradient should be the same yes!

So in the end it mostly matters how you actually define the objective we
are trying to solve.. ---> exactly!

Btw: skimming through the pdf, it looks like me guestimated interpretation
of Q was correct? :D --> my memory is too short sorry I think I forgot it ahahah but if you meant that just taking the real part of the complex part would work just as well than yes

src/mrpro/algorithms/optimizers/_fista.py
guastinimara: ah nice!

src/mrpro/algorithms/optimizers/_fista.py
```diff
@@ -0,0 +1,204 @@
...
+    complex_input = torch.is_complex(initial_value)
+    x_old = initial_value
+    y = initial_value
+    t_old = 1.0
+
+    for _ in range(max_iterations):
+        while stepsize > 1e-30:
+            f_y = f(y)
+            # calculate the proximal gradient step
+            gradient = grad(f_y, y)
```
fzimmermann89: here, the values is calculated a second time

src/mrpro/algorithms/optimizers/_fista.py
```diff
@@ -0,0 +1,204 @@
...
+
+    Parameters
+    ----------
+    f
+        differentiable function
+    x
+        point in the domain of f
+    """
+    x = x.clone().detach().requires_grad_(True)
+    y = f(x)
```
fzimmermann89: I believe it should be `y, = f(x)`

fzimmermann89 (commented):
some comments

fzimmermann89:
You are completely right! The forward should have the `view_as_real ` as well!
Or the proximal is wrong. Either way, forward and proximal do not match for complex numbers.

(Hope that at some point somebody will add a test for the functionals that checks if the proximal actually is the proximal...)

src/mrpro/algorithms/optimizers/_fista.py
guastinimara: why?

fzimmermann89: I think the functionals are operators, thus returning a tuple of tensors (even though it is a single tensor)? 
Or does it work as is? If yes, I am surprised and wondering why and how..

src/mrpro/algorithms/optimizers/_fista.py
guastinimara: I just added this to _fista.py for now, I don't really know where I should add it in the utils

src/mrpro/algorithms/optimizers/_fista.py
guastinimara: no it was just my curiosity! I forgot they returned tuple

ckolbPTB:
summary from today's meeting:

- rename to pgd (proximal_gradient_descent)
- test which solves ||x -y||_2^2 + lambda ||x||_1
- test which solves ||FFT(x) -y||_2^2 + lambda ||x||_1
- previous test but with an increased step size to verify backtracking

all tests with random data, test is successful if objective function drecreases

ckolbPTB:
We are not using underscores anymore so it should be src/mrpro/algorithms/optimizers/pgd.py

Please remove fista.py and all the commented code from the files

guastinimara:
Fista is not working but I could not understand why.

I left a commented part at the end of the test consisting of some plots that I used to check. 
Since I didn’t use the operator norm, I looked more into the case with backtracking as it doesn’t depend on it.
It seems to me that the solution of fista  just goes towards the zero tensor somehow and I checked again the code but I cannot see where the implementation of the algorithm is not correct.
I also thought that maybe it depends on the fact that in the example the L1 norm g(x) is huge in comparison to f(x), but I still couldn’t figure it out if there is an error in the implementation.

tests/algorithms/test_pgd.py
```diff
@@ -0,0 +1,167 @@
+"""Tests for the conjugate gradient method."""
```
ckolbPTB: please adapt

tests/algorithms/test_pgd.py
```diff
@@ -0,0 +1,167 @@
...
+from mrpro.data._SpatialDimension import SpatialDimension
+from mrpro.algorithms.optimizers import pgd
+
+
+from mrpro.operators.functionals import MSEDataDiscrepancy
+from mrpro.operators.functionals import L1Norm
+#%%
+
+def test_pgd_convergence_fft_example():
+    dim = SpatialDimension.from_array_xyz((100,100,0))
```
ckolbPTB: why z = 0?

guastinimara: I wanted to have it 2d, how should I do that?

guastinimara:
Currently L1 norm is defined as the sum of the absolute values, whereas in MSEDataDiscrepancy the L2 norm takes the mean and not the sum (and in L2 norm squared I don't understand it), wouldn't it make sense to choose one or the other way (or I don't know give the choice)?

guastinimara:
I needed to use L1Norm(weight=1/(100**2)), now these two tests for the fft example work (and I looked at the image solution and it seemed to work).

For this example, should I add some different cases for ex of reg_parameter and max_iterations?

I will now work on the denoising example

tests/algorithms/test_pgd.py
guastinimara: actually (100,100,0) and (100,100,1) gave me the same result

fzimmermann89: it should be  `SpatialDimension(x=100,y=100,z=1)`

fzimmermann89:
> Currently L1 norm is defined as the sum of the absolute values, whereas in MSEDataDiscrepancy the L2 norm takes the mean and not the sum (and in L2 norm squared I don't understand it), wouldn't it make sense to choose one or the other way (or I don't know give the choice)?

By the name I would assume L1 to be the sum (although pytorch defaults to MAE for l1loss, so mean)
Same for L2.
Although in practice, in most cases you will want to scale it down to the mean. 

So @ckolbPTB @koflera @Stef-Martin @guastinimara : 
should we 
- add a 'divide_by_n:bool=False'  parameter to the functionals (that just scale the weight down when called).
This would allow to use the same functional for differently shaped images without having the recreate it with different weightings.
- or do we say it is fine if we have to scale down the weight

fzimmermann89:
> ’t use the operator norm, I looked more into the case with backtracking as it doesn’t depend on it.

Also, for FFT the operator norm is just `1`

guastinimara:
> > Currently L1 norm is defined as the sum of the absolute values, whereas in MSEDataDiscrepancy the L2 norm takes the mean and not the sum (and in L2 norm squared I don't understand it), wouldn't it make sense to choose one or the other way (or I don't know give the choice)?
> 
> In L2 norm squared it should be the sum, I'd say (not sure what it does) in MSE it should be the mean.

Okay 👍 but then in this case is fine if I have to take the mean of the L1 norm since I'm minimizing f + g right? Otherwise they're very unbalanced

fzimmermann89:
I see. You should maybe also use L2 norm for data consistency in pdg?
And we should reconsider if we want to keep an extra MseDataConsistency 
(or if that should just be an alias for L2 Norm with the mean parameter set to true, or just remove it ?)

And aim for consistency. In practice, the using the mean will most likely be better, right? It will decouple the operator norm from the image size.. ? 

.. Sorry, I forget to click 'update comment' earlier while editing my comment.

ckolbPTB:
> > Currently L1 norm is defined as the sum of the absolute values, whereas in MSEDataDiscrepancy the L2 norm takes the mean and not the sum (and in L2 norm squared I don't understand it), wouldn't it make sense to choose one or the other way (or I don't know give the choice)?
> 
> By the name I would assume L1 to be the sum (although pytorch defaults to MAE for l1loss, so mean) Same for L2. Although in practice, in most cases you will want to scale it down to the mean.
> 
> So @ckolbPTB @koflera @Stef-Martin @guastinimara : should we
> 
>     * add a 'divide_by_n:bool=False'  parameter to the functionals (that just scale the weight down when called).
>       This would allow to use the same functional for differently shaped images without having the recreate it with different weightings.
> 
>     * or do we say it is fine if we have to scale down the weight

I also think that the "mean-version" is probably the most useful. I would suggest to:

- add `divide_by_n:bool=True`
- Remove MseDataConsistency and use L2Squared instead

guastinimara:
Okay I'm editing this comment, I didn't find the target defined on the Functional.

So for now I will rewrite the tests running on L1 and L2 norms and then once there is the divide_by_n parameters I will update them?

fzimmermann89:
> I also thought of using the L2Norm here instead of MSEDataConsistency, but I was confused on how to apply it, in particular:
> 
> I need a functional || F(x) - data ||_2^2, with MSEDataConsistency it was straightforwad to apply @ F to it, so that F was applied only to x because data is fixed. I couldn't see how to write the same if I want to use L2Norm, but still needing to give a functional f to pgd 

L2NormSquared should behave the same as MSE with regard to @. If not, let Stephan know in the functional PR.(A functional should be an operator)
Also, you should be able to set the data as target in the when creating the L2 functional.

guastinimara:
> > I also thought of using the L2Norm here instead of MSEDataConsistency, but I was confused on how to apply it, in particular:
> > I need a functional || F(x) - data ||_2^2, with MSEDataConsistency it was straightforwad to apply @ F to it, so that F was applied only to x because data is fixed. I couldn't see how to write the same if I want to use L2Norm, but still needing to give a functional f to pgd
> 
> L2NormSquared should behave the same as MSE with regard to @. If not, let Stephan know in the functional PR.(A functional should be an operator) Also, you should be able to set the data as target in the when creating the L2 functional.

Yes I realised only after writing my comment! I didn't see the target to use as data, thanks!

koflera:
> > > Currently L1 norm is defined as the sum of the absolute values, whereas in MSEDataDiscrepancy the L2 norm takes the mean and not the sum (and in L2 norm squared I don't understand it), wouldn't it make sense to choose one or the other way (or I don't know give the choice)?
> > 
> > 
> > By the name I would assume L1 to be the sum (although pytorch defaults to MAE for l1loss, so mean) Same for L2. Although in practice, in most cases you will want to scale it down to the mean.
> > So @ckolbPTB @koflera @Stef-Martin @guastinimara : should we
> > ```
> > * add a 'divide_by_n:bool=False'  parameter to the functionals (that just scale the weight down when called).
> >   This would allow to use the same functional for differently shaped images without having the recreate it with different weightings.
> > 
> > * or do we say it is fine if we have to scale down the weight
> > ```
> 
> I also think that the "mean-version" is probably the most useful. I would suggest to:
> 
>     * add `divide_by_n:bool=True`
> 
>     * Remove MseDataConsistency and use L2Squared instead

I didn't have time yet to follow the discussion in detail. But I would say that the implementation of `MSEDataDiscrepancy` is currently not what we would like to use for fista or pdhg or alike. That was back then more thought of a loss function. 

So we somehow have the inherent distinction between loss-functions whose output is really just one single value, while for the functionals to be used for reconstruction algorithms, we typically will want to be able to have the `dim` argument, similar as in the operator-norm implementation.

I will have a look at it later...

src/mrpro/algorithms/optimizers/pgd.py
```diff
@@ -0,0 +1,116 @@
+"""Proximal Gradient Descent algorithm."""
+# Copyright 2024 Physikalisch-Technische Bundesanstalt
```
ckolbPTB: remove copyright

tests/algorithms/test_pgd.py
```diff
@@ -0,0 +1,95 @@
+"""Tests for the proximal gradient descent."""
+
+# Copyright 2024 Physikalisch-Technische Bundesanstalt
```
ckolbPTB: remove copyright

src/mrpro/algorithms/optimizers/pgd.py
```diff
@@ -0,0 +1,116 @@
...
+#   limitations under the License.
+
+import math
+
+import torch
+
+from mrpro.operators.Functional import Functional
+
+
+def grad_and_value(function, x, create_graph=False):
```
ckolbPTB: remove once #307 is merged

src/mrpro/algorithms/optimizers/pgd.py
```diff
@@ -45,8 +33,44 @@ def pgd(
...
+    In this case, ||F^T F|| = 1.
+    
+    
+    Example of usage:
+    fft = FastFourierOp()
+    l2 = L2NormSquared(target=kspace_data)
+    f = l2 @ fft
+    g = L1Norm()
+    fft_norm = 1.
+    suitalbe_stepsize = 0.85 * 1 / fft_norm
```
fzimmermann89: stepsize

fzimmermann89:
```
Example:
    L1 regularized reconstruction::
          fft = ....
```
I think this kind of indentation (and the double colon, ::) makes it show up as a literal block in the docu

src/mrpro/algorithms/optimizers/pgd.py
```diff
@@ -31,46 +32,41 @@ def pgd(
...
     In applications, f is usually of the form f(x) = 1/2 ||Ax - target||^2, where A is a linear operator.
     In this case, t \in (0, 1/||A^T A||) for convergence.
     If no backtracking is used, the fixed stepsize should be given accordingly to the convergence condition.
-    
-    
+
     Case example: image reconstruction.
     Problem formulation: min_x 1/2 ||Fx - target||^2 + ||x||_1,
-    with F being Fast Fourier Tranform, target acquired data \in k-space and x \in image space,
+    with F being the Fast Fourier Transform, target the acquired data \in k-space and x \in image space,
```
fzimmermann89: remove "Fast"  -> Fourier transform

fzimmermann89:
I think the example code should be in an `.. code-block:: python`
or use this format (or similar):
```
Example:
    L1 regularized reconstruction blablabla longer descripton::
          fft = ....
```
with these indents and :: to look good in the docu

guastinimara:
I'm trying to generalize fista for g ProximableFunctionalSeparableSum and f taking more inputs. I understood how to write the g, but I don't get how's the structure of f supposed to be like? 

for image being a tensor for example,  
```
l1 = L1Norm(divide_by_n=True)
g = ProximableFunctionalSeparableSum(l1, l1)
g(image,image)

```
how would it be the same with` f = l2 @ fft` @fzimmermann89 ? I can't find it in the code anywhere as an example

guastinimara:
I pushed a version of fista now generalized to work on separable sums as well. I made the commit with no verify because there was an error from mypy that I didn't know how to solve. It was in the tests, line 106, and said:

tests/algorithms/test_pgd.py:106: error: Unsupported operand types for @ ("L2NormSquared" and "LinearOperatorMatrix")  [operator]

But all the tests worked, and since @ is applied to two operators, I don't understand why it's not fine...

Anyway, tomorrow I'll think about whether to add some other tests

fzimmermann89:
please check how to docstring renders in the documentation (see "View Online" comment above)

src/mrpro/algorithms/optimizers/pgd.py
```diff
@@ -0,0 +1,143 @@
...
+        initial value for the solution of the algorithm
+    stepsize
+        stepsize needed in the gradient step, is constant throughout all
+        iterations
+    reg_parameter
+        regularization parameter that multiplies g
+    max_iterations, optional
+        number of iterations
+    backtrack_factor
+        must be <=1. if <1., Backtracking rule for stepsize following https://www.ceremade.dauphine.fr/~carlier/FISTA
```
fzimmermann89: does tthis render corretyl in docs`?

src/mrpro/algorithms/optimizers/pgd.py
```diff
@@ -0,0 +1,143 @@
...
+    Parameters
+    ----------
+    f
+        convex, differentiable functional
+    g
+        convex, non-smooth functional with computable proximal map
+    initial_value
+        initial value for the solution of the algorithm
+    stepsize
+        stepsize needed in the gradient step, is constant throughout all
```
fzimmermann89: is it constant, doesnt the algrithm backtrack?

tests/algorithms/test_pgd.py
```diff
@@ -0,0 +1,125 @@
...
+        stepsize=2.0,
+        reg_parameter=0.01,
+        max_iterations=200,
+        backtrack_factor=0.75,
+    )
+
+    assert f(*pgd_solution)[0] + g(*pgd_solution)[0] < f(*initial_value)[0] + g(*initial_value)[0]
+
+
+def test_pgd_on_functionals_with_multiple_inputs() -> None:
```
fzimmermann89: should this give the same result as only having one fft and one g?

could this be a useful sanity check?

fzimmermann89 (commented):
please check online documentation

tests/algorithms/test_pgd.py
guastinimara: what do you mean?

src/mrpro/algorithms/optimizers/pgd.py
guastinimara: Yes I'm stupid, I meant when no backtracking, I'll correct it now

src/mrpro/algorithms/optimizers/pgd.py
```diff
@@ -0,0 +1,158 @@
...
+    stepsize: float = 1.0,
+    reg_parameter: float = 0.01,
+    max_iterations: int = 128,
+    backtrack_factor: float = 1.0,
+    callback: Callable[[PGDStatus], None] | None = None,
+) -> tuple[torch.Tensor, ...]:
+    r"""Proximal gradient descent algorithm for solving problem :math:`min_x f(x) + g(x)`.
+
+    f is convex, differentiable, and with L-Lispchitz gradient.
+    g is convex, non-smooth with computable proximal map.
```
koflera: I would write "g is possibly non-smooth", cause it does not necessarily have to be non-smooth.

src/mrpro/algorithms/optimizers/pgd.py
```diff
@@ -0,0 +1,158 @@
...
+
+    For fixed stepsize t, pgd converges globally when :math:`t \in (0, 1/L)`,
+    where L is the Lipschitz constant of the gradient of f.
+    In applications, f is usually of the form :math:`f(x) = 1/2 ||Ax - target||^2`,
+    where :math:`A` is a linear operator.
+    In this case, :math:`t \in (0, 1/||A^T A||)` for convergence.
+    If no backtracking is used, the fixed stepsize should be given accordingly to the convergence condition.
+
+    Example:
+        L1 regularized image reconstruction. Problem formulation: :math:`min_x 1/2 ||Fx - target||^2 + ||x||_1`,
```
koflera: good idea to have an example here. At least, I like it. :)  Should we have it for pdhg as well, @fzimmermann89?

src/mrpro/algorithms/optimizers/pgd.py
```diff
@@ -0,0 +1,158 @@
...
+                # no need to check stepsize, continue to next iteration
+                break
+            difference = tuple(xi - yi for xi, yi in zip(x, y, strict=True))
+            quadratic_approx = (
+                f_y
+                + 1 / (2 * stepsize) * sum(di.abs().square().sum() for di in difference)
+                + sum(torch.vdot(gi.flatten(), di.flatten()).real for gi, di in zip(gradient, difference, strict=True))
+            )
+
+            (f_x,) = f(*x)
```
koflera, line 143: this looks like a possibly computationally expensive operation that seems only to be needed when one wants to use backtracking?
Maybe something like the following?

```
if backtracking:
    ....
```

tests/algorithms/test_pgd.py
```diff
@@ -0,0 +1,159 @@
...
+        f=f,
+        g=g,
+        initial_value=initial_value,
+        stepsize=0.5,
+        reg_parameter=0.01,
+        max_iterations=200,
+        backtrack_factor=1.0,
+    )
+
+    assert f(*pgd_solution)[0] + g(*pgd_solution)[0] < f(*initial_value)[0] + g(*initial_value)[0]
```
koflera: I think this does not imply that the algorithm has converged... this would (most probably) be even the case after a few iterations (if fista monotonically decreases the objective functional - which I am not 100% sure about).
I would suggest to replace the test to check if fista has calculated the true solution to the problem. For certain this problem, the solution would be given by simple soft-thresholding, see also here:

https://github.com/PTB-MR/mrpro/blob/05c259451a62789188492e271f92a8138d780534/tests/algorithms/test_pdhg.py#L101

tests/algorithms/test_pgd.py
```diff
@@ -0,0 +1,159 @@
...
+        f=f,
+        g=g,
+        initial_value=initial_value,
+        stepsize=2.0,
+        reg_parameter=0.01,
+        max_iterations=200,
+        backtrack_factor=0.75,
+    )
+
+    assert f(*pgd_solution)[0] + g(*pgd_solution)[0] < f(*initial_value)[0] + g(*initial_value)[0]
```
koflera: same here for the assert

src/mrpro/algorithms/optimizers/pgd.py
```diff
@@ -0,0 +1,158 @@
...
+    stepsize: float
+    """Current stepsize of the algorithm."""
+
+
+def pgd(
+    f: Operator[torch.Tensor, tuple[torch.Tensor]] | Operator[Unpack[tuple[torch.Tensor, ...]], tuple[torch.Tensor]],
+    g: ProximableFunctional | ProximableFunctionalSeparableSum,
+    initial_value: torch.Tensor | tuple[torch.Tensor, ...],
+    stepsize: float = 1.0,
+    reg_parameter: float = 0.01,
```
koflera: is it necessary to have the reg parameter here? Is it not possible to have, for example, the following?

```
lambda_reg = 0.1
g = lambda_reg * L1Norm()
```

src/mrpro/algorithms/optimizers/pgd.py
```diff
@@ -0,0 +1,158 @@
...
+    max_iterations: int = 128,
+    backtrack_factor: float = 1.0,
+    callback: Callable[[PGDStatus], None] | None = None,
```
koflera, line 32: since it is probably quite straight forward to implement, we could also think of having both ISTA and FISTA with this code? We would need some sort of flag to distinguish the two cases, though. I could imagine that @fzimmermann89 will have an opinion about how nice this is or not? ;)

koflera (commented):
some comments :)

src/mrpro/algorithms/optimizers/pgd.py
fzimmermann89: Good idea. I will also try out the doctest pytest plugin, to check that this example code actually runs (and is not forgotten on changes to the code)

src/mrpro/algorithms/optimizers/pgd.py
fzimmermann89: i think it should work as you would expect of you set `g= 0.01*L1Norm()`

src/mrpro/algorithms/optimizers/pgd.py
```diff
@@ -0,0 +1,158 @@
...
+                break
+            stepsize *= backtrack_factor
+
+        else:
+            if backtracking:
+                raise RuntimeError('After backtracking, the stepsize became to small.')
+            else:
+                raise RuntimeError('Stepsize to small.')
+
+        # update timestep t
```
fzimmermann89: is this a "timestep" ?

src/mrpro/algorithms/optimizers/pgd.py
```diff
@@ -0,0 +1,158 @@
...
+        t = (1 + math.sqrt(1 + 4 * t_old**2)) / 2
+
+        # update the solution
+        y = tuple(xi + (t_old - 1.0) / t * (xi - xi_old) for xi, xi_old in zip(x, x_old, strict=True))
```
fzimmermann89, line 161: remove the commtns that just say what is being done.

src/mrpro/algorithms/optimizers/pgd.py
```diff
@@ -0,0 +1,158 @@
...
+    if isinstance(g, ProximableFunctional):
+        g_sum = ProximableFunctionalSeparableSum(g)
+    else:
+        g_sum = g
+
+    x_old = initial_values
+    y: tuple[torch.Tensor, ...] = initial_values
+    gradient: tuple[torch.Tensor, ...]
+    f_y: torch.Tensor
+    t_old = 1.0
```
fzimmermann89, line 124: of we set this to 0, we get ISTA?

src/mrpro/algorithms/optimizers/pgd.py
```diff
@@ -0,0 +1,158 @@
...
+    Example:
+        L1 regularized image reconstruction. Problem formulation: :math:`min_x 1/2 ||Fx - target||^2 + ||x||_1`,
+        with :math:`F` being the Fourier Transform, target the acquired data \in k-space and x \in image space,
+        :math:`f(x) = 1/2 ||Fx - target||^2, g(x) = ||x||_1`.
+        In this case, :math:`||F^T F|| = 1`. ::
+                fft = FastFourierOp()
+                l2 = L2NormSquared(target=kspace_data)
+                f = l2 @ fft
+                g = L1Norm()
+                fft_norm = 1.
```
fzimmermann89: i would call it operator norm here to avoid any confusion

src/mrpro/algorithms/optimizers/pgd.py
fzimmermann89: are you sure y is supposed to be the solution, not x?
this does not update the solution..., right? 
fista should return x?

src/mrpro/algorithms/optimizers/pgd.py
guastinimara: if `g = lambda_reg * L1Norm()` , then the prox is already done accordingly? If so then I can remove it from the input and then put in the code  
```
x = g_sum.prox(
                *[yi - stepsize * gi for yi, gi in zip(y, gradient, strict=True)], sigma=stepsize
            )
```
as it was the only point where the lambda_reg was used

src/mrpro/algorithms/optimizers/pgd.py
guastinimara: what do you mean? throughout the code I always called it stepsize

src/mrpro/algorithms/optimizers/pgd.py
guastinimara: but isnt it already done only if backtracking ?  I mean the line before is
```
if not backtracking:
                # no need to check stepsize, continue to next iteration
                break
```
doesn't this mean that the lines that you mentioned are already computed only when backtracking?

src/mrpro/algorithms/optimizers/pgd.py
guastinimara: I never noticed that I put y! Yes yes it should ben x not y! thanks

src/mrpro/algorithms/optimizers/pgd.py
guastinimara: the example code should work in theory, but since I wrote it just as a reference example it doesn't actually work because  `kspace_data `and `image_space_shape` are not defined. Should I write that differently?

src/mrpro/algorithms/optimizers/pgd.py
koflera: ah yeah, sorry... didn't read the code correctly

src/mrpro/algorithms/optimizers/pgd.py
guastinimara: you mean if `t_old=0`? If so I wouldn't say so:
if we start with t_1 = 0, y_1=x_0

then at iteration=1:
x_1 = prox(y_1)
t_2 = 1
y_2 = x_1 - (x_1 - x_0) = x_0

and so at iteration=2: 
x_2 = prox(x_0) ?? (would expect x_1 if it was ista)
and t_3 != 1 

or not? maybe it's wrong but I would say it needs to be coded as separate cases if we want both ISTA and FISTA?

guastinimara: or did you mean setting something else to 0?

guastinimara:
I implemented small changes from the comments. 

Still, I need to check the tests from pdhg to add something like this as suggested by @koflera. 

Also, we can for sure add the possibility of running ISTA and not just FISTA (both with either constant stepsize or backtracking). If we agree on the flag then I'll add it, I don't have many ideas on it apart from 'fast=True/False') to be honest

src/mrpro/algorithms/optimizers/pgd.py
ckolbPTB: final dicision: only use FISTA

guastinimara:
I'm facing an error I couldn't understand how to fix, but I think it depends on the wavelet operator. I can't make fista run with it so far, it breaks at line 117 when computing:
`gradient, f_y = grad_and_value_f(y)`
with the error
RuntimeError: Cannot access data pointer of Tensor that doesn't have storage
coming apparently from something in line 369 of the WaveletOp, which is
```
coefficients = torch.split(
            coefficients_stack, [int(np.prod(shape)) for shape in self.coefficients_shape], dim=-1
        )
```

The error came up launching this code here below:

```
import tempfile
import mrpro
import requests

from mrpro.algorithms.optimizers import pgd
from mrpro.algorithms.optimizers.pgd import PGDStatus
from mrpro.operators.functionals import L1NormViewAsReal, L2NormSquared

n_spokes = 24
zenodo_url = 'https://zenodo.org/records/14617082/files/'
fname = f'radial2D_{n_spokes}spokes_golden_angle_with_traj.h5'

data_file = tempfile.NamedTemporaryFile(mode='wb', delete=False, suffix='.h5')
response = requests.get(zenodo_url + fname, timeout=30)
data_file.write(response.content)
data_file.flush()

kdata = mrpro.data.KData.from_file(data_file.name, mrpro.data.traj_calculators.KTrajectoryIsmrmrd())
kdata.header.recon_matrix.x = 256
kdata.header.recon_matrix.y = 256

direct_reconstruction = mrpro.algorithms.reconstruction.DirectReconstruction(kdata)
img_direct = direct_reconstruction(kdata)
iterative_sense_reconstruction = mrpro.algorithms.reconstruction.IterativeSENSEReconstruction(
    kdata,
    n_iterations=8,
    csm=direct_reconstruction.csm,
    dcf=direct_reconstruction.dcf,
)
img_iterative_sense = iterative_sense_reconstruction(kdata)

fourier_operator = direct_reconstruction.fourier_op
assert direct_reconstruction.csm is not None
csm_operator = direct_reconstruction.csm.as_operator()

wavelet_operator = mrpro.operators.WaveletOp(domain_shape=img_direct.data.shape[-2:],
                                             dim=(-2,-1))

acquisition_operator = fourier_operator @ csm_operator @ wavelet_operator.H

regularization_parameter = 0.1
l2 = 0.5 * L2NormSquared(target=kdata.data, divide_by_n=True)
l1 = L1NormViewAsReal(divide_by_n=True)

f = l2 @ acquisition_operator
g = regularization_parameter * l1

initial_values = wavelet_operator(img_iterative_sense.data)

max_iterations = 10
(img_pgd,) =pgd(
        f=f,
        g=g,
        initial_value=initial_values,
        stepsize=0.5,
        reg_parameter=1.,
        max_iterations=max_iterations,
        backtrack_factor=0.75)
```
But I don't get what should be changed, or if there is a stupid error in the code I'm launching that I don't see. 
The same code without wavelet_op.H in the acquisition operator and appropriate initial value doesn't break. 
Do yo have any idea? I can also send the full error message, or write this somewhere else

fzimmermann89:
See #604

fzimmermann89:
@guastinimara I moved the example to the new place and already did some of the adjustments to make it more similar to the other notebooks (download and plotting).
Also, the notebooks now use only package level imports, i.e. only `import mrpro`, not `from mrpo... import ...`

It would be good if you could try to interleave text and code a bit. you can already set up the functionals etc when you introduce the math. This would read a bit nicer than first having text only, followed by a block of code without text in between.

Like in a paper, you would interleave equations and text. I would like to try to handle the code in the examples similiary as one would equations in a paper.

examples/scripts/pgd_wavelet_reconstruction.py
```diff
@@ -0,0 +1,285 @@
+# %% [markdown]
+# # Accelerated Proximal gradient descent (FISTA) reconstruction using wavelets and $\ell_1$-norm minimization
```
guastinimara:
```suggestion
# # Accelerated proximal gradient descent (FISTA) reconstruction using wavelets and $\ell_1$-norm minimization

File: examples/scripts/pgd_wavelet_reconstruction.py [hunk 27]
```diff
@@ -0,0 +1,285 @@
...
+# We consider the following functional $\mathcal{F}$:
+#
+# $ \mathcal{F}(x) = \frac{1}{2}||Ax - y||_2^2 + \lambda \| W x \|_1, \quad \quad \quad (1)$
+#
+# where $W$ is the discretized wavelet operator that maps the image domain to the wavelet domain,
+# and $\lambda >0$ is an appropriate regularization weight.
+#
+# The minimization of the functional $\mathcal{F}$ is non-trivial due to
+# the non-differentiable $\ell_1$-norm and the presence of the wavelet operator.
+# A possible algorithm to approach the solution of (1) would be the primal dual hybrid gradient (PDHG)
```
guastinimara:
```suggestion
# A possible algorithm to approach the solution of $(1)$ would be the primal dual hybrid gradient (PDHG)

File: examples/scripts/pgd_wavelet_reconstruction.py [hunk 28]
```diff
@@ -0,0 +1,285 @@
...
+#
+# The main step of the minimization method is a proximal gradient step, which reads as
+#
+# $x_{k} = \mathrm{prox}_{\sigma g}(x_{k-1} - \sigma \nabla f({x}_{k-1}))$
+#
+# where $\mathrm{prox}$ denotes the proximal operator and $\sigma$ is
+# an appropriate stepsize, ideally $\sigma=\frac{1}{L}$, with $L$ being
+# the Lipschitz constant of $\nabla f$.
+#
+# Additionally, FISTA has an additional step to accelerate the convergence. The following variable $z_{k+1}$
```
guastinimara:
```suggestion
# Moreover, FISTA has an additional step to accelerate the convergence. The following variable $z_{k+1}$

File: examples/scripts/pgd_wavelet_reconstruction.py [hunk 29]
```diff
@@ -0,0 +1,285 @@
...
+    if iteration % 8 == 0:
+        print(
+            f'{iteration}: {optimizer_status["objective"](*solution).item()}, stepsize: {optimizer_status["stepsize"]}'
+        )
+
+
+# %% [markdown]
+# ### Run FISTA for a certain number of iterations
+# Now we can run the FISTA algorithm to solve the minimization problem. As an initial guess,
+# we use the wavelet-coefficients of the # iterative SENSE image to speed up the convergence.
```
guastinimara: is the # a typo or something is missing?

examples/scripts/pgd_wavelet_reconstruction.py
```diff
@@ -0,0 +1,285 @@
...
+# map the solution back to image domain
+(img_pgd_24,) = wavelet_operator.H(img_wave_pgd_24)
+
+# %% [markdown]
+# ```{note}
+# When defining the functional $f$ with the argument `divide_b_n=True`, one needs to be careful when setting
+# the stepsize to be used in FISTA. The reason is that the Lipschitz-constant of the gradient of the functional
+# $f_N(\tilde{x}) = 1/(2N)\,\|\tilde{A}\tilde{x} - y\|_2^2$, where $y\in\mathbb{C}^N$, is no longer given
+# by the squared operator norm of $\tilde{A}$, but rather by the squared operator norm of the scaled
+# operator $1/N \cdot \tilde{A}$. Thus, the Lipschitz constant $L(\nabla f_N))$ must be appropriately scaled,
```
guastinimara:
```suggestion
# operator $1/N \cdot \tilde{A}$. Thus, the Lipschitz constant $L(\nabla f_N)$ must be appropriately scaled,

File: examples/scripts/pgd_wavelet_reconstruction.py [hunk 31]
```diff
@@ -0,0 +1,285 @@
...
+    titles=['402 spokes', '24 spokes (Direct)', '24 spokes (SENSE)', '24 spokes (PGD)'],
+)
+
+
+# %% [markdown]
+# Congratulations! We have successfully reconstructed an image from 24 spokes using wavelets.
+#
+# ### Next steps
+# Not happy with the results? Play around with the regularization weight, the number of iterations, the
+# number of levels in the wavelet decomposition or the backtracing factor to see how they affect the final image.
```
guastinimara: backtracking factor

examples/scripts/pgd_wavelet_reconstruction.py
```diff
@@ -0,0 +1,285 @@
...
+#
+# where $f$ is a convex, differentiable function with $L$-Lipschitz gradient $\nabla f$,
+# and $g$ is convex and possibly non-smooth.
+#
+# The main step of the minimization method is a proximal gradient step, which reads as
+#
+# $x_{k} = \mathrm{prox}_{\sigma g}(x_{k-1} - \sigma \nabla f({x}_{k-1}))$
+#
+# where $\mathrm{prox}$ denotes the proximal operator and $\sigma$ is
+# an appropriate stepsize, ideally $\sigma=\frac{1}{L}$, with $L$ being
```
guastinimara: maybe I would use alread the notation $L(\nabla f)$ here, as used after in the note

```suggestion
# an appropriate stepsize, ideally $\sigma=\frac{1}{L}$, with $L=L(\nabla f)$ being

**Review by guastinimara — COMMENTED** [review]
I added some small corrections/comments on the notebook

**Comment by guastinimara** [comment]
I'll read now the paper that Andreas recommended before (https://inria.hal.science/hal-01060130v3/document) on the convergence so that maybe we can add it also as an option

File: src/mrpro/algorithms/optimizers/pgd.py [hunk 33]
```diff
@@ -28,6 +28,7 @@ def pgd(
     stepsize: float = 1.0,
     max_iterations: int = 128,
     backtrack_factor: float = 1.0,
+    t_for_converging_solution: bool = False,
```
fzimmermann89: convergent_iterates_variant?
convergent_iterates?

i would not mention "t" in the argument name

src/mrpro/algorithms/optimizers/pgd.py
```diff
@@ -80,6 +81,11 @@ def pgd(
     backtrack_factor
         must be :math:`<=1`. if :math:`<1.`, backtracking rule for stepsize following https://www.ceremade.dauphine.fr/~carlier/FISTA
         is used
+    t_for_converging_solution
+        by default, the algorithm updates the variable t as described by Beck and Teboulle in the original FISTA paper.
+        If set to True, the algorithm updates t as suggested by Chambolle and Dossal https://inria.hal.science/hal-01060130v3/document
```
fzimmermann89: make link work in docs.

see how references in docstring are done in other algorithms.

src/mrpro/algorithms/optimizers/pgd.py
```diff
@@ -80,6 +81,11 @@ def pgd(
     backtrack_factor
         must be :math:`<=1`. if :math:`<1.`, backtracking rule for stepsize following https://www.ceremade.dauphine.fr/~carlier/FISTA
```
fzimmermann89: make a proper reference

guastinimara:
In the main function, I added the option to choose a different update for t, as discussed. I also modified the description of the reference articles similarly to what I've seen in other files.

I modified the tests so that now they check that pgd outputs the correct solution for the two example cases considered. The tests run both with and without backtracking. Then I added a test to check that the callback function works (as I've seen in pdhg tests), and a test to check that the two different updates for t lead to the same solution for one example case

tests/algorithms/test_pgd.py
```diff
@@ -0,0 +1,201 @@
...
+from mrpro.phantoms import EllipsePhantom
+from tests import RandomGenerator
+
+
+@pytest.mark.parametrize(('stepsize', 'backtrack_factor'), [(1, 1), (1e5, 0.8)])
+def test_pgd_solution_fourier_l1(stepsize, backtrack_factor) -> None:
+    """ "Set up the problem min_x 1/2*|| Fx - y||_2^2 + lambda * ||x||_1,
+    where F is the full FFT and y is sampled on a Cartesian grid. Thus the
+    problem has a closed-form solution given by soft-thresholding. Test
+    if the espected solution and the one obtained by the pgd are close."""
```
koflera: typo: eSpected

src/mrpro/algorithms/optimizers/pgd.py
```diff
@@ -0,0 +1,176 @@
...
+    stepsize: float = 1.0,
+    max_iterations: int = 128,
+    backtrack_factor: float = 1.0,
+    t_for_converging_solution: bool = False,
+    callback: Callable[[PGDStatus], None] | None = None,
+) -> tuple[torch.Tensor, ...]:
+    r"""Proximal gradient descent algorithm for solving problem :math:`min_x f(x) + g(x)`.
+
+    f is convex, differentiable, and with L-Lispchitz gradient.
+    g is convex, non-smooth with computable proximal map.
```
koflera: possibly non-smooth instead of non-smooth, since L2Squared would also work ;)

src/mrpro/algorithms/optimizers/pgd.py
```diff
@@ -0,0 +1,176 @@
...
+        In this case, :math:`||F^T F|| = 1`. ::
+                kspace_data = torch.randn(3, 10, 10 , dtype=torch.complex64)
+                fft = FastFourierOp()
+                l2 = L2NormSquared(target=kspace_data)
+                f = l2 @ fft
+                reg_parameter = 0.01
+                g = reg_parameter * L1Norm()
+                operator_norm = 1.
+                stepsize = 0.85 * 1 / operator_norm**2
+                initial_value = torch.ones((3,10,10))
```
koflera: `torch.ones(3,10,10)` instead of `torch.ones((3,10,10))`?

examples/scripts/pgd_wavelet_reconstruction.py
```diff
@@ -0,0 +1,285 @@
...
+# of the previous two steps $x_{k}$ and $x_{k-1}$. So, for $t_1=1, z_1=x_0$:
+#
+# $x_{k} = \mathrm{prox}_{\sigma g}(z_{k} - \sigma \nabla f({z}_{k}))$
+#
+# $t_{k+1} = \frac{1 + \sqrt{1 + 4t_k^2}}{2}$
+#
+# $z_{k+1} = x_{k} + \frac{t_k - 1}{t_{k+1}}(x_{k} - x_{k-1}).$
+#
+# As the Lipschitz constant $L$ is in general not known, and
+# the interval of the stepsize $\sigma\in ( 0, \frac{1}{|| \tilde{A}^H \tilde{A}||} )$ is crucial for the convergence,
```
koflera: maybe better:
$\sigma\in ( 0, \frac{1}{|| \tilde{A}||_2^2} )$

koflera (commented):
a few comments

src/mrpro/algorithms/optimizers/pgd.py
guastinimara: yes I can change it, it works with both anyway

tests/algorithms/test_pgd.py
```diff
@@ -0,0 +1,198 @@
...
+from mrpro.algorithms.optimizers import pgd
+from mrpro.data.SpatialDimension import SpatialDimension
+from mrpro.operators import FastFourierOp, LinearOperatorMatrix, ProximableFunctionalSeparableSum, WaveletOp
+from mrpro.operators.functionals import L1NormViewAsReal, L2NormSquared
+from mrpro.phantoms import EllipsePhantom
+from tests import RandomGenerator
+
+
+@pytest.mark.parametrize(('stepsize', 'backtrack_factor'), [(1, 1), (1e5, 0.8)])
+def test_pgd_solution_fourier_l1(stepsize, backtrack_factor) -> None:
```
fzimmermann89: type hints
stepsize:float, backtrack_factor:float

tests/algorithms/test_pgd.py
```diff
@@ -0,0 +1,198 @@
...
+        initial_value=initial_value,
+        stepsize=stepsize,
+        max_iterations=max_iterations,
+        backtrack_factor=backtrack_factor,
+        convergent_iterates_variant=False,
+    )
+    torch.testing.assert_close(pgd_solution, expected, rtol=5e-4, atol=5e-4)
+
+
+@pytest.mark.parametrize(('stepsize', 'backtrack_factor'), [(1, 1), (1e5, 0.8)])
```
fzimmermann89: type hints
stepsize:float, backtrack_factor:float

tests/algorithms/test_pgd.py
fzimmermann89: neee. i meant

```suggestion
def test_pgd_solution_fourier_l1(stepsize:float, backtrack_factor:float) -> None:
```

fzimmermann89: etc

fzimmermann89 (approved):
after adding the type hints, feel free to merge.

PR #294, opened 2024-05-24
Add Wavelet Operator

schuenke:
# Wavelet Operator for wavelet-reco

For wavelet transforms we use [The PyTorch Wavelet Toolbox](https://github.com/v0lta/PyTorch-Wavelet-Toolbox) (ptwt), which extends [PyWavelets](https://pywavelets.readthedocs.io/en/latest/) (pywt). It provides GPU and gradient support via a PyTorch backend.


ckolbPTB:
I get the following error:

worker 'gw3' crashed while running 'tests/data/test_movedatamixin.py::test_movedatamixin_convert[True-dtype1-single]'

As I did not change anything there, I was wondering if @fzimmermann89 has any ideas what this could be about?

src/mrpro/operators/WaveletOp.py
```diff
@@ -0,0 +1,350 @@
...
+            if any(d % 2 for d in domain_shape):
+                raise NotImplementedError('ptwt only supports wavelet transforms for even number of samples.')
+
+            # size of wavelets
+            wavelet_length = torch.as_tensor((Wavelet(wavelet_name).dec_len,) * len(domain_shape))
+
+            # calculate shape of wavelet coefficients at each level
+            current_shape = torch.as_tensor(domain_shape)
+
+            if _check_level(domain_shape, wavelet_length, level) == 0:
```
fzimmermann89: what does this do?

fzimmermann89: check level returns the level?

fzimmermann89: maybe give it a name?
(and reuse it then in l 106)

src/mrpro/operators/WaveletOp.py
```diff
@@ -0,0 +1,350 @@
...
+                self.coefficients_padding = [(0, 0)] * len(current_shape)
+            else:
+                self.coefficients_shape = []
+                self.coefficients_padding = []
+                for _ in range(_check_level(domain_shape, wavelet_length, level)):
+                    # Add padding
+                    for ind in range(len(current_shape)):
+                        padl, padr = _get_pad(current_shape[ind], wavelet_length[ind])
+                        current_shape[ind] += padl + padr
+                    current_shape = torch.floor((current_shape - (wavelet_length - 1) - 1) / 2 + 1).to(
```
fzimmermann89, line 111: suggestion:
(current_shape - wavelet_length) / 2 + 1).to(torch.int64)

src/mrpro/operators/WaveletOp.py
```diff
@@ -0,0 +1,350 @@
...
+            if _check_level(domain_shape, wavelet_length, level) == 0:
+                self.coefficients_shape = [current_shape]
+                self.coefficients_padding = [(0, 0)] * len(current_shape)
+            else:
+                self.coefficients_shape = []
+                self.coefficients_padding = []
+                for _ in range(_check_level(domain_shape, wavelet_length, level)):
+                    # Add padding
+                    for ind in range(len(current_shape)):
+                        padl, padr = _get_pad(current_shape[ind], wavelet_length[ind])
```
fzimmermann89: function returns padr, padl (

fzimmermann89: also ist just  `total_pad = 2*wavelet_length[ind] - 3` according to _get_pad comment

src/mrpro/operators/WaveletOp.py
fzimmermann89: combining (see comment in _get_pad)
# pytorch wavelet toolbox pads such that the output is (data_len + filt_len - 1)//2, matching pywt
((current_shape + wavelet_length- 1)/2).int64()

src/mrpro/operators/WaveletOp.py
```diff
@@ -0,0 +1,350 @@
...
+                        current_shape[ind] += padl + padr
+                    current_shape = torch.floor((current_shape - (wavelet_length - 1) - 1) / 2 + 1).to(
+                        dtype=torch.int64
+                    )
+                    self.coefficients_shape.extend([current_shape.clone()] * self.n_wavelet_directions)
+                    self.coefficients_padding.extend([(padl.clone(), padr.clone())] * self.n_wavelet_directions)
```
fzimmermann89: padl and padr will always be equal for even images, and equal `(2 * filt_len - 3) // 2`

src/mrpro/operators/WaveletOp.py
```diff
@@ -0,0 +1,350 @@
...
+        elif len(self.dim) == 2:
+            coeffs2 = wavedec2(x_real, self.wavelet_name, level=self.level, mode='zero', axes=(-2, -1))
+            coefficients_list = self._format_coeffs_2d(coeffs2)
+        elif len(self.dim) == 3:
+            coeffs3 = wavedec3(x_real, self.wavelet_name, level=self.level, mode='zero', axes=(-3, -2, -1))
+            coefficients_list = self._format_coeffs_3d(coeffs3)
+
+        # stack multi-resolution wavelets along single dimension
+        coefficients_stack = self._coeff_to_stacked_tensor(coefficients_list)
+        coefficients_stack = (
```
fzimmermann89:
```
if x.is_complex:
        coefficients_stac
```k =  torch.view_as_complex(coefficients_stack.moveaxis(0, -1).contiguous())

File: src/mrpro/operators/WaveletOp.py [hunk 6]
```diff
@@ -0,0 +1,350 @@
...
+            torch.view_as_complex(coefficients_stack.moveaxis(0, -1).contiguous())
+            if x.is_complex()
+            else coefficients_stack
+        )
+
+        # move stacked coefficients to first wavelet dimension
+        return (torch.moveaxis(coefficients_stack, -1, min(dim)),)
```
fzimmermann89: i think we could return a contigous array if we move this before the view_as_complex move, right?

src/mrpro/operators/WaveletOp.py
```diff
@@ -0,0 +1,350 @@
...
+        ------
+        ValueError
+            If the domain_shape is not defined.
+        ValueError
+            If the dimensions along which wavelets are to be calculated are not unique.
+        """
+        if self.domain_shape is None:
+            raise ValueError('Adjoint requires to define the domain_shape in init()')
+
+        # normalize axes to allow negative indexing in input
```
fzimmermann89, line 193: change comment?

tests/helper.py
```diff
@@ -82,3 +85,61 @@ def dotproduct_adjointness_test(
...
+    """
+    torch.testing.assert_close(
+        torch.norm(u), torch.norm(operator(u)[0]), rtol=relative_tolerance, atol=absolute_tolerance
+    )
+
+
+def operator_unitary_test(
+    operator: Operator, u: torch.Tensor, relative_tolerance: float = 1e-3, absolute_tolerance=1e-5
+):
+    """Test if an operator is unitary
```
fzimmermann89: period is missing

src/mrpro/operators/WaveletOp.py
```diff
@@ -0,0 +1,350 @@
...
+            1D: [a, d_n, ..., d_1]
+            2D: [aa, ad_n, da_n, dd_n, ..., ad_1, da_1, dd_1]
+            3D: [aaa, aad_n, ada_n, add_n, ..., ..., aad_1, ada_1, add_1, ...]
+        """
+        coefficients = torch.split(
+            coefficients_stack, [int(torch.prod(shape)) for shape in self.coefficients_shape], dim=-1
+        )
+        return [
+            torch.reshape(coeff, coeff.shape[:-1] + tuple(shape))
+            for coeff, shape in zip(coefficients, self.coefficients_shape, strict=False)
```
fzimmermann89: why strict=false?

src/mrpro/operators/WaveletOp.py
```diff
@@ -0,0 +1,350 @@
...
+        elif len(self.dim) == 3:
+            coeffs3 = self._undo_format_coeffs_3d(coefficients_list)
+            data = waverec3(coeffs3, self.wavelet_name, axes=(-3, -2, -1))
+
+        # if we deal with complex coefficients, we also return complex data
+        if coefficients_stack.is_complex():
+            data = torch.view_as_complex(data.moveaxis(0, -1).contiguous())
+
+        # undo swapping of axes
+        data = torch.moveaxis(data, list(range(-len(self.dim), 0)), dim)
```
fzimmermann89: again, con we do this before the view as complex?

fzimmermann89 (changes requested):

src/mrpro/operators/WaveletOp.py
```diff
@@ -0,0 +1,350 @@
...
+            Highest wavelet level. If set to None, the highest possible level is calculated based on the domain shape.
+
+        Raises
+        ------
+        ValueError
+            If wavelets are calculated for more than three dimensions.
+        ValueError
+            If wavelet dimensions and domain shape do not match.
+        NotImplementedError
+            If the domain shape is odd. Adjoint will lead to the wrong domain shape.
```
koflera: do you mean if at least one of the number of pixels in one dimension is odd? what happens for [256,255,255]?
from the code below it's clear, but maybe not from the description here.

src/mrpro/operators/WaveletOp.py
```diff
@@ -0,0 +1,350 @@
...
+
+        if domain_shape is not None:
+            if len(dim) != len(domain_shape):
+                raise ValueError(
+                    'Number of dimensions along which the wavelet transform should be calculated needs to',
+                    'be the same as the domain shape',
+                )
+
+            if any(d % 2 for d in domain_shape):
+                raise NotImplementedError('ptwt only supports wavelet transforms for even number of samples.')
```
koflera: suggestion: maybe something like that

"ptwt only supports wavelet transforms only for tensors with even number of entries for all considered all dimensions"

instead of 

"ptwt only supports wavelet transforms for even number of samples"

src/mrpro/operators/WaveletOp.py
```diff
@@ -0,0 +1,350 @@
...
+        """
+        return coefficients
+
+    def _format_coeffs_2d(
+        self,
+        coefficients: list[torch.Tensor | tuple[torch.Tensor, torch.Tensor, torch.Tensor]],
+    ) -> list[torch.Tensor]:
+        """Format 2D wavelet coefficients to MRpro format.
+
+        Converts from   [aa, (ad_n, da_n, dd_n), ..., (ad_1, da_1, dd_1)]
```
koflera, line 271: maybe add a short comment explaining what aa, ad, da and dd refer to as well as n.

src/mrpro/operators/WaveletOp.py
```diff
@@ -0,0 +1,350 @@
...
+                    self.coefficients_shape.extend([current_shape.clone()] * self.n_wavelet_directions)
+                    self.coefficients_padding.extend([(padl.clone(), padr.clone())] * self.n_wavelet_directions)
+
+                self.coefficients_shape = self.coefficients_shape[::-1]
+                self.coefficients_padding = self.coefficients_padding[::-1]
+                self.coefficients_shape.insert(0, self.coefficients_shape[0])  # shape of a/aa/aaa term
+                self.coefficients_padding.insert(0, self.coefficients_padding[0])  # padding of a/aa/aaa term
+
+    def forward(self, x: torch.Tensor) -> tuple[torch.Tensor,]:
+        """Calculate wavelet coefficients from (image) data.
```
koflera, line 133: I would also include an explanation how complex-valued images are handled, i.e. we compute Wx = [W(Rex), W(Imx)]

src/mrpro/operators/WaveletOp.py
```diff
@@ -0,0 +1,350 @@
...
+        ----------
+        domain_shape
+            Shape of domain where wavelets are calculated. If set to None the shape is taken from the input of the
+            forward operator. The adjoint operator will raise an error.
+        dim
+            Dimensions (axes) where wavelets are calculated
+        wavelet_name
+            Name of wavelets
+        level
+            Highest wavelet level. If set to None, the highest possible level is calculated based on the domain shape.
```
koflera, line 68: could it make sense to still assign a value to self.level, even if it was originally set to None? Without knowing too much about wavelets, if as a user I set level to None, there is no way to find out what the level of the wavelet transform was, unless i transform back the coefficients to the [aa, (ad_n, da_n, ...) , .... ] format.

src/mrpro/operators/WaveletOp.py
ckolbPTB: see comment below

src/mrpro/operators/WaveletOp.py
ckolbPTB: yes, but it does not depend on the image but the current shape of the current level and this can be odd. In this case the padding is (2 * filt_len - 3) // 2 + 1. I could of course include this even/odd check in our code but IMO it makes more sense to call their function

src/mrpro/operators/WaveletOp.py
ckolbPTB: yes, only small disadvantage is that if we do it before view_as_complex it needs to be done differently for complex and non-complex data. but this is fine by me.

ckolbPTB: If this is ok for everyone I will add it to the adjoint

src/mrpro/operators/WaveletOp.py
ckolbPTB: true, but level is only set if domain_shape is defined. If we are happy that self.level is sometimes correctly set and sometimes not then I can add it as a parameter

src/mrpro/operators/WaveletOp.py
fzimmermann89: 1. it is a private function
2. it is a lot easier to understand that the shape will be:
`(current_shape/2).ceil()+wavelet_length//2-1`
and this hole section can just be collapsed.

src/mrpro/operators/WaveletOp.py
fzimmermann89: (still, order is wrong)

src/mrpro/operators/WaveletOp.py
fzimmermann89: why is current_shape a tensor, why is padl and padr tensors?
_getpad has typehints int

src/mrpro/operators/WaveletOp.py
```diff
@@ -0,0 +1,362 @@
...
+                raise NotImplementedError(
+                    'ptwt only supports wavelet transforms for tensors with even number of '
+                    'entries for all considered dimensions.'
+                )
+
+            # size of wavelets
+            wavelet_length = torch.as_tensor((Wavelet(wavelet_name).dec_len,) * len(domain_shape))
+
+            # calculate shape of wavelet coefficients at each level
+            current_shape = torch.as_tensor(domain_shape)
```
fzimmermann89, line 112: why tensor?

src/mrpro/operators/WaveletOp.py
fzimmermann89: (and why the clones..?)

src/mrpro/operators/WaveletOp.py
ckolbPTB: because we are "doing math" with it later

src/mrpro/operators/WaveletOp.py
fzimmermann89: and is      
`self.coefficients_padding.extend([(padl.clone(), padr.clone())] * self.n_wavelet_directions)`
correct? 
padl and padr are updated in the loop above, this should only be the last value? I am really confused by this whole section...

ckolbPTB: the clones are needed because extends does only add a view to the object and hence padl and padr would be overwritten in the loop

ckolbPTB: tensors are there for the math or am I missing something?

fzimmermann89: mazbe either do everything as tensors in this file,
the whole current shape logic couldbe written as (including the padding)
(current_shape/2).ceil()+wavelet_length//2-1

or use the _get_pad function and iterate over all values.

the mix of vectorized tensor logic and the loop seems complicated.

fzimmermann89: - why is coefficients padding not just a list of ints?
-  why is self.coefficients_shape not a list of ints?

I would either do all math in this section on the tensors, then convert to lists of ints to store (so no loop over the elements of the tensor) or just do everything as a loop

fzimmermann89: also... is coefficients_padding ever used?

src/mrpro/operators/WaveletOp.py
```diff
@@ -0,0 +1,362 @@
...
+                self.coefficients_shape = []
+                self.coefficients_padding = []
+                for _ in range(verified_level):
+                    # Add padding
+                    for ind in range(len(current_shape)):
+                        padr, padl = _get_pad(current_shape[ind], wavelet_length[ind])
+                        current_shape[ind] += padl + padr
+                    current_shape = torch.floor((current_shape - wavelet_length) / 2 + 1).to(dtype=torch.int64)
+                    self.coefficients_shape.extend([current_shape.clone()] * self.n_wavelet_directions)
+                    self.coefficients_padding.extend([(padl.clone(), padr.clone())] * self.n_wavelet_directions)
```
fzimmermann89:
```suggestion
                for _ in range(verified_level):
                  current_shape = (current_shape/2).ceil()+wavelet_length//2-1
                  self.coefficients_shape.extend([tuple(current_shape),]*self.n_wavelet_directions)
```

src/mrpro/operators/WaveletOp.py
fzimmermann89: if not, see suggestion below

src/mrpro/operators/WaveletOp.py
```diff
@@ -0,0 +1,362 @@
...
+            wavelet coefficients in the format
+            1D: [a, d_n, ..., d_1]
+            2D: [aa, ad_n, da_n, dd_n, ..., ad_1, da_1, dd_1]
+            3D: [aaa, aad_n, ada_n, add_n, ..., ..., aad_1, ada_1, add_1, ...]
+        """
+        coefficients = torch.split(
+            coefficients_stack, [int(torch.prod(shape)) for shape in self.coefficients_shape], dim=-1
+        )
+        return [
+            torch.reshape(coeff, coeff.shape[:-1] + tuple(shape))
```
fzimmermann89: RUF005 would complain if it understood what we are doing here.
maybe we still want to follow it,  `(*coeff.shape[:-1],*shape)`, for consistency 
(if shape is no longer a tensor, otherwise the tuple(shape) is still required)

fzimmermann89: (it doesnt detect it because shape is a tensor and coeff.shape a torch.size...

fzimmermann89:
Out of curiosity: pytorch wavelet toolbox changed some type hints as we suggested (on their current GitHub version)

I don't remember why we cared - does this help?

Or will this now cause issues with the current waveletop code?

ckolbPTB:
> Out of curiosity: pytorch wavelet toolbox changed some type hints as we suggested (on their current GitHub version)
> I don't remember why we cared - does this help?
> Or will this now cause issues with the current waveletop code?

As far as I can see this fits to your typehinting in the code so it should not be a problem.

src/mrpro/operators/WaveletOp.py
```diff
@@ -0,0 +1,359 @@
...
+
+
+class WaveletOp(LinearOperator):
+    """Wavelet operator class."""
+
+    def __init__(
+        self,
+        domain_shape: Sequence[int] | None = None,
+        dim: tuple[int] | tuple[int, int] | tuple[int, int, int] = (-2, -1),
+        wavelet_name: str = 'db4',
```
fzimmermann89: can we list all possible names as literal?

fzimmermann89: and/or mention them in the docstring

src/mrpro/operators/WaveletOp.py
```diff
@@ -0,0 +1,359 @@
...
+from ptwt.conv_transform_2 import waverec2
+from ptwt.conv_transform_3 import wavedec3
+from ptwt.conv_transform_3 import waverec3
+from pywt import Wavelet
+from pywt._multilevel import _check_level
+
+from mrpro.operators.LinearOperator import LinearOperator
+
+
+class WaveletOp(LinearOperator):
```
fzimmermann89, line 35: maybe we can add a few words aboutthis class?

src/mrpro/operators/WaveletOp.py
```diff
@@ -0,0 +1,359 @@
...
+            wavelet_length = torch.as_tensor((Wavelet(wavelet_name).dec_len,) * len(domain_shape))
+
+            # calculate shape of wavelet coefficients at each level
+            current_shape = torch.as_tensor(domain_shape)
+
+            # if level is None, select the highest possible level.
+            # raise error/warnings if level is not possible or lead to boundary effects
+            verified_level = _check_level(domain_shape, wavelet_length, level)
+
+            if verified_level == 0:
```
fzimmermann89: can you add comment what this means?

src/mrpro/operators/WaveletOp.py
```diff
@@ -0,0 +1,359 @@
...
+        if len(self.dim) == 1:
+            coeffs1 = wavedec(x_real, self.wavelet_name, level=self.level, mode='zero', axis=-1)
+            coefficients_list = self._format_coeffs_1d(coeffs1)
+        elif len(self.dim) == 2:
+            coeffs2 = wavedec2(x_real, self.wavelet_name, level=self.level, mode='zero', axes=(-2, -1))
+            coefficients_list = self._format_coeffs_2d(coeffs2)
+        elif len(self.dim) == 3:
+            coeffs3 = wavedec3(x_real, self.wavelet_name, level=self.level, mode='zero', axes=(-3, -2, -1))
+            coefficients_list = self._format_coeffs_3d(coeffs3)
+
```
fzimmermann89, line 171: else: raise error?

dim is public and could be changed. also, adding this check here does not cost anything...

fzimmermann89: why does this pass mypy?
 coefficients_list could be uninitialized and the self._coeff_to_stacked_tensor would fail

src/mrpro/operators/WaveletOp.py
```diff
@@ -0,0 +1,359 @@
...
+    def _format_coeffs_1d(self, coefficients: list[torch.Tensor]) -> list[torch.Tensor]:
+        """Format 1D wavelet coefficients to MRpro format.
+
+        At the moment, this function just returns the input coefficients as is:
+        [a, d_n, ..., d_1]
+        """
+        return coefficients
+
+    def _undo_format_coeffs_1d(self, coefficients: list[torch.Tensor]) -> list[torch.Tensor]:
```
fzimmermann89: can these all be static if they don't access anything from self? or outside the class (in the same file)

tests/helper.py
```diff
@@ -12,10 +12,13 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
+from __future__ import annotations
```
fzimmermann89: why?

fzimmermann89 (commented):
some comments

schuenke:
For testing the implementation we used some real scanner data already. Should we create an example script / notebook from our test case and directly include it in this PR?

ckolbPTB:
> Should we create an example script / notebook from our test case and directly include it in this PR?

The example we had during the hackathon was already for a FISTA + Wavelet reconstruction. I would therefore suggest to add this example to the FISTA PR.

fzimmermann89:
@CodiumAI-Agent /review
--pr_reviewer.extra_instructions="
In the `possible issues` section, emphasize the following:
- Is the code logic efficient?
- Is it idiopathic pytorch code?
- Are all variable und function names clear and descriptive?
- Are additional comments required to understand the logic?
- Check the docstrings for completeness  
"
--pr_reviewer.inline_code_comments=true
--pr_reviewer.require_score_review=true
--pr_reviewer.num_code_suggestions="8"

src/mrpro/operators/WaveletOp.py
```diff
@@ -0,0 +1,353 @@
...
+            raise ValueError(f'Axis must be unique. Normalized axis are {dim}')
+
+        # swapping the last axes and the axes to calculate wavelets of
+        x = torch.moveaxis(x, dim, list(range(-len(self._dim), 0)))
+
+        # the ptwt functions work only for real data, thus we handle complex inputs as an additional channel
+        x_real = torch.view_as_real(x).moveaxis(-1, 0) if x.is_complex() else x
+
+        if len(self._dim) == 1:
+            coeffs1 = wavedec(x_real, self._wavelet_name, level=self._level, mode='zero', axis=-1)
```
fzimmermann89: coeffs1d?

src/mrpro/operators/WaveletOp.py
fzimmermann89: we could add a WaveType
```
WaveletType = Literal[
    "haar",
    "db1", "db2", "db3", "db4", "db5", "db6", "db7", "db8", "db9",
    "db10", "db11", "db12", "db13", "db14", "db15", "db16", "db17", "db18", "db19",
    "db20", "db21", "db22", "db23", "db24", "db25", "db26", "db27", "db28", "db29",
    "db30", "db31", "db32", "db33", "db34", "db35", "db36", "db37", "db38",
    "sym2", "sym3", "sym4", "sym5", "sym6", "sym7", "sym8", "sym9",
    "sym10", "sym11", "sym12", "sym13", "sym14", "sym15", "sym16", "sym17", "sym18", "sym19", "sym20",
    "coif1", "coif2", "coif3", "coif4", "coif5", "coif6", "coif7", "coif8", "coif9",
    "coif10", "coif11", "coif12", "coif13", "coif14", "coif15", "coif16", "coif17",
    "bior1.1", "bior1.3", "bior1.5", "bior2.2", "bior2.4", "bior2.6", "bior2.8",
    "bior3.1", "bior3.3", "bior3.5", "bior3.7", "bior3.9", "bior4.4", "bior5.5", "bior6.8",
    "rbio1.1", "rbio1.3", "rbio1.5", "rbio2.2", "rbio2.4", "rbio2.6", "rbio2.8",
    "rbio3.1", "rbio3.3", "rbio3.5", "rbio3.7", "rbio3.9", "rbio4.4", "rbio5.5", "rbio6.8",
    "dmey",
    "gaus1", "gaus2", "gaus3", "gaus4", "gaus5", "gaus6", "gaus7", "gaus8",
    "mexh",
    "morl",
    "cgau1", "cgau2", "cgau3", "cgau4", "cgau5", "cgau6", "cgau7", "cgau8",
    "shan",
    "fbsp",
    "cmor"
]
```
and type the wave_name as WaveletType

src/mrpro/operators/WaveletOp.py
```diff
@@ -0,0 +1,353 @@
...
+        ------
+        ValueError
+            If the dimensions along which wavelets are to be calculated are not unique.
+        """
+        # normalize axes to allow negative indexing in input
+        dim = tuple(d % x.ndim for d in self._dim)
+        if len(dim) != len(set(dim)):
+            raise ValueError(f'Axis must be unique. Normalized axis are {dim}')
+
+        # swapping the last axes and the axes to calculate wavelets of
```
fzimmermann89: this does not swap, but move
`# move axes to calculate wavelets of to the end`

fzimmermann89 (approved):

PR #295, opened 2024-05-24
Functionals Base Class, L2 Norm Squared, L1 Norm

JoHa0811:


src/mrpro/operators/_Functional.py
```diff
@@ -0,0 +1,48 @@
...
+
+import torch
+
+from mrpro.operators._Operator import Operator
+
+
+class Functional(Operator[torch.Tensor, tuple[torch.Tensor]]):
+    """Functional Base Class.
+
+    Args:
```
ckolbPTB: complete

src/mrpro/operators/_Functional.py
```diff
@@ -0,0 +1,48 @@
...
+        Operator (_type_): _description_
+    """
+
+    def __init__(
+        self, lam: torch.Tensor | float = 1.0, g: torch.Tensor | None = None, dim: Sequence[int] | None = None
```
ckolbPTB: Please use more descriptive names for lam and g

src/mrpro/operators/_Functional.py
```diff
@@ -0,0 +1,48 @@
...
+    ):
+        """init.
+
+        Args:
```
ckolbPTB: use correct doc strings

src/mrpro/operators/_Functional.py
```diff
@@ -0,0 +1,48 @@
...
+class ProximableFunctional(Functional, ABC):
+    """ProximableFunction Base Class.
+
+    Args:
+        Functional (_type_): _description_
+    """
+
+    @abstractmethod
+    def prox(self, x: torch.Tensor, sigma: torch.Tensor) -> tuple[torch.Tensor]:
+        """Apply forward operator."""
```
ckolbPTB: Apply proximal operator?

src/mrpro/operators/_Functional.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def prox_convex_conj(self, x: torch.Tensor, sigma: torch.Tensor) -> tuple[torch.Tensor]:
+        """Apply forward operator."""
```
ckolbPTB: Apply proximal of convec conjugate of functional

ckolbPTB: add information about this form of the prox_convex_conj

tests/operators/functionals/test_l1_norm.py
```diff
@@ -0,0 +1,74 @@
...
+#   limitations under the License.
+
+import pytest
+import torch
+from mrpro.operators.functionals.l1 import L1Norm
+
+
+@pytest.mark.parametrize(
+    (
+        'x',
```
ckolbPTB: suggestion:
forward_x, prox, prox_complex_conjugate

and then do something like
torch.testing.assert_close(l1_norm.forward(x), forward_x, rtol=1e-3, atol=1e-3)

why are p_forward and pcc_forward needed? the forward of the norm is already tested with forward_x, isn't it?

tests/operators/functionals/test_l2_squared_norm.py
```diff
@@ -0,0 +1,74 @@
+"""Tests for L2-Squared-functional."""
+
+# Copyright 20234 Physikalisch-Technische Bundesanstalt
```
ckolbPTB: 2024

tests/operators/functionals/test_l1_norm.py
```diff
@@ -0,0 +1,74 @@
+"""Tests for L1-functional."""
+
+# Copyright 20234 Physikalisch-Technische Bundesanstalt
```
ckolbPTB: 2024

ckolbPTB:
Please adapt your doc strings to MRpro standards

src/mrpro/operators/functionals/l1.py
```diff
@@ -0,0 +1,72 @@
...
+        -------
+            tuple[torch.Tensor]: prox of data
+        """
+        is_complex = x.is_complex()
+        if is_complex:
+            x = torch.view_as_real(x)
+            threshold = torch.tensor([self.lam * sigma]).unsqueeze(-1)
+        else:
+            threshold = torch.tensor([self.lam * sigma])
+        x = torch.clamp(x, -threshold, threshold)
```
ckolbPTB: x = torch.nn.functional.relu(x - threshold) - torch.nn.functional.relu(- x - threshold)

src/mrpro/operators/functionals/l1.py
```diff
@@ -0,0 +1,72 @@
...
+        -------
+            tuple[torch.Tensor]: prox convex conjugate of data
+        """
+        is_complex = x.is_complex()
+        if is_complex:
+            x = torch.view_as_real(x)
+            complex_lam = torch.tensor([self.lam]).unsqueeze(-1)
+            complex_sigma = torch.tensor([sigma]).unsqueeze(-1)
+            complex_g = torch.tensor([self.g]).unsqueeze(-1)
+            x = torch.clamp((x - complex_g * complex_sigma), -complex_lam, complex_lam)
```
ckolbPTB: probably also needs to be adapted based on relu

src/mrpro/operators/functionals/l1.py
ckolbPTB: there should be a g somewhere in there

src/mrpro/operators/functionals/l1.py
ckolbPTB: double check but should actually be correct

ckolbPTB:
Think of a test which confirms that the prox is actually the argmin

ckolbPTB:
@koflera do we need to adapt anything here with this PR?
https://github.com/PTB-MR/mrpro/blob/554052592324f364f63f79bc441efd8c1fb07518/tests/operators/_OptimizationTestFunctions.py#L19

src/mrpro/operators/_Functional.py
koflera: I would also suggest to remove the parameter "lambda" in the the init, as it is only required when computing the prox, since we are usually interested in computing the prox of lambda*|| ||_1.

src/mrpro/operators/functionals/l1.py
koflera: i think there is no g since here we conly compute the soft-threshold of the input x, i.e. prox_{threshold * || ||_1}, the g, i.e. usually the data, will be mostly present only when we have the l2-norm.

src/mrpro/operators/_Functional.py
koflera: Does it make sense to have the data g as an argument in the init of the class?

fzimmermann89: without the lambda, we have no way to specify lambda*|| ||_1.
it should also be used in the forward.

src/mrpro/operators/functionals/l1.py
fzimmermann89: there should be a g.

fzimmermann89: also, for complex inputs, this should be modrelu, i.e.:

`polar(relu(abs(x)-thres), angle(x))`

fzimmermann89: for complex, this should be modrelu, i.e. something like
```
diff=x-g
thres=lambda*sigma
prox=polar(relu(abs(diff)-thres), angle(d))
```

src/mrpro/operators/functionals/l1.py
```diff
@@ -0,0 +1,72 @@
...
+        """forward.
+
+        Args:
+            x (torch.Tensor): data tensor
+
+        Returns
+        -------
+            tuple[torch.Tensor]: forward of data
+        """
+        return (torch.tensor([(x - self.g).abs().sum(self.dim)]),)
```
fzimmermann89: lambda is missing

src/mrpro/operators/functionals/l2_squared.py
```diff
@@ -0,0 +1,54 @@
...
+
+from mrpro.operators._Functional import ProximableFunctional
+
+
+class L2NormSquared(ProximableFunctional):
+    """Functional for L2 Norm Squared.
+
+    Parameters
+    ----------
+        lambda = 1
```
fzimmermann89: docstring?

src/mrpro/operators/functionals/l2_squared.py
```diff
@@ -0,0 +1,54 @@
...
+        """Forward method.
+
+        Args:
+            x (torch.Tensor): data tensor
+
+        Returns
+        -------
+            tuple[torch.Tensor]: l2 squared of data
+        """
+        return (torch.linalg.vector_norm(x - self.g, ord=2, dim=self.dim, keepdim=True).square(),)
```
fzimmermann89: lambda is missing

tests/operators/functionals/test_l2_squared_norm.py
```diff
@@ -0,0 +1,74 @@
...
+import pytest
+import torch
+from mrpro.operators.functionals.l2_squared import L2NormSquared
+
+
+@pytest.mark.parametrize(
+    (
+        'x',
+        'expected_result_x',
+        'expected_result_p',
```
fzimmermann89: rename

tests/operators/functionals/test_l2_squared_norm.py
```diff
@@ -0,0 +1,74 @@
...
+        'expected_result_pcc',
+        'expected_result_p_forward',
+        'expected_result_pcc_forward',
+    ),
+    [
+        (
+            torch.tensor([1, 1], dtype=torch.complex64),
```
fzimmermann89: where do these come from?
did you check that these are the expected results for proximal and proximal of convex conjugate?
... just in case: using the outputs of the function as expected results does not really help

fzimmermann89 (changes requested):
curenlty wrong results (wrong expressions, lambdas missing, g missing)
complex should better be interpreted as complex functionals (see maras pdf in fista PR), style (dostrings)
tests should tests the expected properties of the functions.

see inline comments

src/mrpro/operators/_Functional.py
koflera: but is this needed in the first place?

fzimmermann89: Yes, if we want to ever allow lambda as maps.

ckolbPTB:
summary from today's meeting:

- tests required
- l1_norm which handles complex data "correctly"
- l1_norm_coponentwise which deals with real and imag part separately

fzimmermann89:
l1_convex_conjugate:
diff = x-sigma*target
`    weight*diff/(torch.max(weight,torch.abs(diff)))
`  oder auch
`    torch.polar(weight.clamp(max=diff.abs()),torch.angle(diff))`

(glaub ich)

src/mrpro/operators/_Functional.py
```diff
@@ -0,0 +1,58 @@
...
+    """Functional Base Class."""
+
+    def __init__(
+        self, weight: torch.Tensor | float = 1.0, target: torch.Tensor | None = None, dim: Sequence[int] | None = None
+    ):
+        """Initialize a Functional.
+
+        Parameters
+        ----------
+            weight (float, optional)
```
ckolbPTB: we do not use optional or type definitions (e.g. float) because usually we forget to update them if we change the types (weight can also be a tensor not just a float).

src/mrpro/operators/_Functional.py
```diff
@@ -0,0 +1,58 @@
...
+    def __init__(
+        self, weight: torch.Tensor | float = 1.0, target: torch.Tensor | None = None, dim: Sequence[int] | None = None
+    ):
+        """Initialize a Functional.
+
+        Parameters
+        ----------
+            weight (float, optional) 
+                weighting of the l1 norm
+            target (optional)
```
ckolbPTB: Maybe it would be good to add an equation in the description of the norm to explain that a bit better

ckolbPTB:
Please rebase to current main and then go through the comments above one by one and click "resolve conversation" after checking that you have implemented the requested change.

ckolbPTB:
we do not use underscores anymore. it should be scr/mrpro/operators/functionals/Functional.py

guastinimara:
shapes

Trying to use L1Norm, L2NormSquared and MSEDataDiscrepancy, I noticed that they act differently in the shapes of (list of) tensor that they return, like in the screenshot here. Would it be better to have one same for all or not?

fzimmermann89:
Yes. 

Either remove the dims, or add a keepdim=False flag to the norms. 

I think I would prefer the latter, as it is often useful.

src/mrpro/operators/Functional.py
```diff
@@ -0,0 +1,64 @@
...
+import torch
+
+from mrpro.operators.Operator import Operator
+
+
+class Functional(Operator[torch.Tensor, tuple[torch.Tensor]]):
+    """Functional Base Class."""
+
+    def __init__(
+        self, weight: torch.Tensor | float = 1.0, target: torch.Tensor | None = None, dim: Sequence[int] | None = None, divide_by_n : bool = False
```
koflera: do we really need weight, target and divide_by_n in the base class? This sounds like something that we would typically want for a norm but seems unintuitive for me for a general functional? At least I would struggle to write a formal description of what these tensors are supposed to be in general without having to refer to norms/data discrepancy/similarity measures as a more specific example of functionals.

src/mrpro/operators/Functional.py
```diff
@@ -0,0 +1,64 @@
...
+from abc import abstractmethod
+from collections.abc import Sequence
+
+import torch
+
+from mrpro.operators.Operator import Operator
+
+
+class Functional(Operator[torch.Tensor, tuple[torch.Tensor]]):
+    """Functional Base Class."""
```
koflera, line 13: Maybe write a short description of what we mean by functionals, i.e. mappings from a vector space (image space, k-space, ...) to the real numbers.

src/mrpro/operators/Functional.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def __init__(
+        self, weight: torch.Tensor | float = 1.0, target: torch.Tensor | None = None, dim: Sequence[int] | None = None, divide_by_n : bool = False
+    ) -> None:
+        """Initialize a Functional.
+
+        Parameters
+        ----------
+            weight
```
koflera: I think the parameter weight comes into play when we think about norms (as also the docstring suggests). Maybe we should have another class Norm that inherits from functional?

Additionally, we maybe should explicitly mention that by providing a weight, we allow for the implementation of a weighted norm only with a weight that corresponds to a diagonal matrix (probably it would also be good to check that the weights are indeed positive). I am not sure if we want to allow norms that are induced by the weighted inner products with more general positive definite matrices, i.e. norms of the form ||x||_M := <x^T, Mx>^(1/2).

src/mrpro/operators/Functional.py
fzimmermann89: These are settings that should exist for all functionals.

You should be able to depend upon that fact.
Also, setting them is the same for all functionals l.

So I would say they are part of the base class.

You will never see the doc string of the base class anyway.

fzimmermann89: What is the issue with 
F:Kn->R
Functional(•):= F(weight*(•-target)) / n ?

koflera: > What is the issue with F:Kn->R Functional(•):= F(weight*(•-target)) / n ?

Just that it sounds a bit arbitrary to provide the base class with those specific choices of hyper-parameters that define the functional.

Would it be a problem to have a base class `functional `that inherits from `operator `and to have a base class `norm` that inherits from `functional ` and for which one can provide `weight`, `target `and `divide_by_n`?

fzimmermann89:
Talking with @koflera :

We are fine with having a joint Functional/Loss class.

We should add a keep_dim = False to the functionals.

We should explicitly mention in the docstring, that only keep_dim=False, dim=None reduces to a single number.

For each functional, the docstring should show what the code does with respect to the weight setting.

src/mrpro/operators/functionals/examples.py
```diff
@@ -0,0 +1,17 @@
+#%%
```
fzimmermann89: needs to be removed befoire merging

tests/operators/functionals/test_l1_norm.py
```diff
@@ -0,0 +1,341 @@
...
+    l1_norm = L1Norm(weight=1, divide_by_n=True)
+    torch.testing.assert_close(
+        l1_norm.forward(x, dim=dim, keepdim=keepdim)[0].shape, shape_forward_x, rtol=1e-3, atol=1e-3
+    )
+
+
+@pytest.mark.parametrize(
+    ('x',),
+    [
+        (torch.rand(10, 10, 10),),
```
fzimmermann89: dontuse torch rand, use the rng helper as done in the other tests.
this is important for reproducible tests.

tests/operators/functionals/test_l1_norm.py
```diff
@@ -0,0 +1,341 @@
...
+            torch.Size([16, 8, 1]),
+        ),
+        (
+            torch.rand(16, 8, 1, 160, 160),
+            None,
+            False,
+            torch.Size([]),
+        ),
+        (
+            torch.rand(16, 8, 1, 160, 160),
```
fzimmermann89: dontuse torch rand, use the rng helper as done in the other tests.
this is important for reproducible tests.

tests/operators/functionals/test_l1_norm.py
```diff
@@ -0,0 +1,341 @@
...
+
+@pytest.mark.parametrize(
+    (
+        'x',
+        'dim',
+        'keepdim',
+        'shape_forward_x',
+    ),
+    [
+        (
```
fzimmermann89: what do these different cases tests? why are the tests there? please add comments and/or rethink what is tested by the tests.

tests/operators/functionals/test_l2_squared_norm.py
```diff
@@ -0,0 +1,176 @@
...
+            False,
+            torch.Size([]),
+        ),
+        (
+            torch.rand(16, 8, 1, 160, 160),
+            None,
+            True,
+            torch.Size([1, 1, 1, 1, 1]),
+        ),
+    ],
```
fzimmermann89: what do these different cases tests? why are the tests there? please add comments and/or rethink what is tested by the tests.

fzimmermann89 (changes requested):
mostly the tests need work.

It would be good to have less tests that test for specific values and more tests that test for properties. 
I.e. at the value x returned by the prox, the forward(x)+1/2||x-v||^2 (or what the porox criterion is, please check...) should at least have autograd-gradient of 0 and small pertubations increase the value (finite different gradient)

src/mrpro/operators/functionals/l1.py
```diff
@@ -0,0 +1,317 @@
...
+    ) -> tuple[torch.Tensor]:
+        """Forward method.
+
+        Compute the l1-norm of the input.
+
+        Parameters
+        ----------
+            x
+                input tensor
+            dim
```
fzimmermann89: what is dim=None?

src/mrpro/operators/functionals/l1.py
```diff
@@ -0,0 +1,317 @@
...
+                dimensions that span the considered vector space
+            divide_by_n
```
fzimmermann89: what is None here? why is None allowed?

src/mrpro/operators/functionals/l1.py
```diff
@@ -0,0 +1,317 @@
...
+    derived assuming the weight vector defines an invertible (block-) diagonal
+    operator with all entries being strictly greater than zero, i.e. only real-valued
+    weight-tensors are allowed.
+    """
+
+    def forward(
+        self,
+        x: torch.Tensor,
+        dim: Sequence[int] | None = None,
+        keepdim: bool | None = None,
```
fzimmermann89: Wby is None allowd here?

JoHa0811:
Remove keep_dim, dim, divide_by_n from forward, prox, prox_convex_conj

src/mrpro/operators/functionals/l1.py
```diff
@@ -0,0 +1,317 @@
...
+        if divide_by_n is None:
+            divide_by_n = self.divide_by_n
+
+        dtype = torch.promote_types(self.target.dtype, x.dtype)
+        x = x.to(dtype)
+        target = self.target.to(dtype)
+        diff = x - target
+        is_complex = diff.is_complex()
+
+        l1norm = L1Norm(weight=self.weight, target=self.target, divide_by_n=divide_by_n)
```
fzimmermann89: move to init

fzimmermann89:
I added a test_functionals.py, which tests all the functionals.

It mainly tests some properties that the functionals should have, and contains a single value check vs. odl (with some example code how to get the reference values from old)

The tests are parameterized to run against ALL proximable functionals.

I changed l2normsquared and l1norm to pass these tests.
This also changed the definition of l2normsquared (I removed the *1/2, l2normquared is now just the l2 norm squared, unscaled)

l1normview as real is unchanged so far..

I renamed the files to match our file name convention and added the imports to the `__init__.py`

As it seemed useful for the tests, I extended the RandomGenerator with two new methods to generate random tensors given either a dtype or a reference tensor. At some point we might move some of the other tests to these new functions as well.

Missing are cuda tests.

fzimmermann89:
I suggest to maybe split this in multiple PRs for review:
- The functional base class and tests cases
- L1Norm
- L2NormSquared
- L1NormViewasReal

PR #296, opened 2024-05-30
No error for missing doc strings

ckolbPTB:


ckolbPTB:
@schuenke @fzimmermann89 Should this not raise an error during pre-commit because there are no doc-strings?

schuenke:
No idea to be honest! 

I tested several settings and observed some weird behavior... For example if I enforce ruff to always run (like with mypy), it detects some additional violations in some of my unstaged files.  But still not the missing docstrings. If I leave the `always_run` activated, but run pre-commit with `--all` , it does NOT detect these violations. 

I also tried to revert the changes from #283 , but it didn't solve the problem...

I'll keep trying, just wanted to give an update...

fzimmermann89:
> No idea to be honest!
> 
> I tested several settings and observed some weird behavior... For example if I enforce ruff to always run (like with mypy), it detects some additional violations in some of my unstaged files. But still not the missing docstrings. If I leave the `always_run` activated, but run pre-commit with `--all` , it does NOT detect these violations.
> 
> I also tried to revert the changes from #283 , but it didn't solve the problem...
> 
> I'll keep trying, just wanted to give an update...

The difference in files is expected. --all would ignore files not tracked by git, afaik,... right?


~~Re docstrings: could it be something related to inheritance? that the rule is satisfied as the parent has a docstring that will be shown? (wild guess....)~~ edit: no... this is not the reason.

fzimmermann89:
Currently, we have all docstring related rules in the tests disabled.

We can discuss if we want to change that...

schuenke:
Explanation: because we use leading underscores in most filenames, they are categorized as private and thus all rules that specifically apply to public methods are not enforced. This is the case for most docstring, but also type annotation rules.

I think we have discuss how we want to "solve" this. IMHO, the leading underscores are actually not necessary or even bad style, because the content of these files is not private. We could just remove them and still provide the import option from the (sub)module inits...

PR #297, opened 2024-05-30
Finite difference operator

ckolbPTB:


src/mrpro/operators/_FiniteDifferenceOp.py
```diff
@@ -0,0 +1,131 @@
...
+        Returns
+        -------
+            Finite difference kernel
+
+        Raises
+        ------
+        ValueError
+            If mode is not central, forward, backward or doublecentral
+        """
+        if mode == 'doublecentral':
```
fzimmermann89: i would drop this one, this was in my old code for backward compatibility

src/mrpro/operators/_FiniteDifferenceOp.py
```diff
@@ -0,0 +1,131 @@
...
+            Dimension along which finite differences are calculated.
+        mode
+            Type of finite difference operator
+        padding_mode
+            Padding to ensure output has the same size as the input
+        """
+        super().__init__()
+        self.dim = dim
+        self.padding_mode = padding_mode
+        self.kernel = self.finite_difference_kernel(mode)
```
fzimmermann89: register buffer

src/mrpro/operators/_FiniteDifferenceOp.py
```diff
@@ -0,0 +1,131 @@
...
+        Raises
+        ------
+        ValueError
+            If the first dimension of y is to the same as the number of dimensions along which the finite differences
+            are calculated
+        """
+        if y.shape[0] != len(self.dim):
+            raise ValueError('Fist dimension of input tensor has to match the number of finite difference directions.')
+        return (
+            torch.sum(
```
fzimmermann89: `sum(_filter_separable(yi, (self.adjoint_kernel,), axis=(dim,)) for dim,yi in zip(self.dim,y))` or similiar?

fzimmermann89: Or what is the reason to first create the stacked intermediate tensor before summing over it?

src/mrpro/operators/FiniteDifferenceOp.py
```diff
@@ -0,0 +1,139 @@
...
+        ------
+        ValueError
+            If the first dimension of y is to the same as the number of dimensions along which the finite differences
+            are calculated
+        """
+        if y.shape[0] != len(self.dim):
+            raise ValueError('Fist dimension of input tensor has to match the number of finite difference directions.')
+        return (
+            torch.sum(
+                torch.stack(
```
fzimmermann89, line 130: Why stack?

src/mrpro/operators/FiniteDifferenceOp.py
```diff
@@ -0,0 +1,139 @@
...
+            kernel = torch.tensor((-1, 1, 0))
+        else:
+            raise ValueError(f'mode should be one of (central, forward, backward), not {mode}')
+        return kernel
+
+    def __init__(
+        self,
+        dim: Sequence[int],
+        mode: Literal['central', 'forward', 'backward'] = 'central',
+        pad_mode: Literal['zeros', 'circular'] = 'zeros',
```
fzimmermann89, line 69: Pad mode or padding_mode (pytorch conv calls it padding_mode) ?

src/mrpro/operators/FiniteDifferenceOp.py
```diff
@@ -0,0 +1,139 @@
...
+        mode
+            Type of finite difference operator
+        pad_mode
+            Padding to ensure output has the same size as the input
+        """
+        super().__init__()
+        self.dim = dim
+        self.pad_mode: Literal['constant', 'circular'] = 'constant' if pad_mode == 'zeros' else pad_mode
+        self.register_buffer('kernel', self.finite_difference_kernel(mode))
+        self.register_buffer('adjoint_kernel', torch.flip(self.kernel, dims=(-1,)))
```
fzimmermann89: Is it really worth to save this one as well?

src/mrpro/operators/FiniteDifferenceOp.py
fzimmermann89: This is way more expensive than the flip :)

tests/operators/test_finite_difference_op.py
```diff
@@ -0,0 +1,53 @@
...
+from tests import RandomGenerator
+from tests.helper import dotproduct_adjointness_test
+
+
+@pytest.mark.parametrize('mode', ['central', 'forward', 'backward'])
+def test_finite_difference_op_forward(mode):
+    """Test correct finite difference of simple object."""
+    # Test object with linear gradient in real and imaginary part
+    linear_gradient_object = torch.arange(1, 21)[None, :].to(dtype=torch.float32)
+    linear_gradient_object = linear_gradient_object + 1j * linear_gradient_object
```
fzimmermann89: Small suggestion: do it with different direction in real and image part. So we would notice some mixup there.

fzimmermann89: And maybe with more than onenaxis in the input.

src/mrpro/operators/FiniteDifferenceOp.py
ckolbPTB: I changed it to pad_mode because this is what we call it now in separable_filter. But I am also happy to change it back to padding_mode. Then we should also change it in separable_filter...

src/mrpro/operators/FiniteDifferenceOp.py
ckolbPTB: I tried 
```
sum([filter_separable(yi, (self.adjoint_kernel,), axis=(dim,), pad_mode=self.pad_mode, pad_value=0.0)
            for dim, yi in zip(self.dim, y, strict=False)
            ])
```
but the output type of sum is Tensor | Literal[0] which then leads to a mypy problem. Is there a clever way to overcome this?

fzimmermann89: ah, I already forgot about that issue. 
the alternatives, functools.reduce, type annotating, or giving an explicit start point for the sum all a bit ugly as well, so I just shut up and agree that you most likely found the best solution

src/mrpro/operators/FiniteDifferenceOp.py
fzimmermann89: up to you

fzimmermann89 (approved):

koflera:
The only thing I would maybe do is to rename the operator to `_GradientOperator` instead of finite-differences operator.

PR #302, opened 2024-06-04
Rename all files to categorize them as 'public' for linters

schuenke:
Files with a leading underscore in the filename are in general categorized as private. Thus, several linting rules that target public functions/methods/classes were not enforced until now. This was escpecially the case for several pydocstyle (c.f. #296) and flake8-annotation rules that we actually want to enforce with ruff.


schuenke:
In general, everything is working now. However, there are still a few docstrings missing that @fzimmermann89 and I didn't add yet:

```
src/mrpro/operators/EndomorphOperator.py:214:9: D102 Missing docstring in public method
src/mrpro/operators/EndomorphOperator.py:218:9: D102 Missing docstring in public method
src/mrpro/operators/GridSamplingOp.py:163:7: D101 Missing docstring in public class
src/mrpro/operators/LinearOperator.py:132:9: D107 Missing docstring in `__init__`
src/mrpro/operators/Operator.py:40:9: D102 Missing docstring in public method
src/mrpro/operators/Operator.py:63:9: D107 Missing docstring in `__init__`
src/mrpro/operators/Operator.py:76:9: D107 Missing docstring in `__init__`
src/mrpro/operators/Operator.py:89:9: D107 Missing docstring in `__init__`
src/mrpro/operators/Operator.py:103:9: D107 Missing docstring in `__init__`
```

ckolbPTB (approved):

ckolbPTB:
Just a small thing: Why did we remove ""D415", # first line should end with punctuation" ? 

Not that I particularly care about the full stop at the end of the first line in the doc-string but we have done it so far, so maybe we should just continue for consistency sake.

fzimmermann89:
We already enforce a dot at the end. This rule was redundant.

PR #307, opened 2024-06-09
Adjoint as gradient for linear operator

### LINKED ISSUE #68: Make operators and pytorch.autograd work together
For most of our operators an adjoint/derivative method will be defined. pytorch.autograd should use that. If the method is not defined it should default to standard pytorch.autgrad. More details can be found in the discussion of #8.

### LINKED ISSUE #303: additional check for adjoint/derivate of linear operators
Besides the existing dot product test for adjointness, it would be nice to have a tests that checks the derivate identities.
So testing if 
a) forward and reverse mode autograd using torch.func work on both the .forward and the .adjoint 
and b) the adjoint does the  same for a random vector as the  vjp of the forward (and vice-versa)

ckolbPTB:
closes #303 
closes #68

also addresses #8 

First test implementation to make autograd use adjoint of linear operators. 

Mypy does not like it, but I am not sure this can be fixed...
src/mrpro/operators/LinearOperator.py:41: error: Signature of "forward" incompatible with supertype "_SingleLevelFunction"  [override]
src/mrpro/operators/LinearOperator.py:41: note:      Superclass:
src/mrpro/operators/LinearOperator.py:41: note:          def forward(ctx: Any, *args: Any, **kwargs: Any) -> Any
src/mrpro/operators/LinearOperator.py:41: note:      Subclass:
src/mrpro/operators/LinearOperator.py:41: note:          @staticmethod
src/mrpro/operators/LinearOperator.py:41: note:          def forward(fw: Callable[[Tensor], Tensor], bw: Callable[[Tensor], Tensor], x: Tensor) -> Any

@fzimmermann89 could you explain a bit more what you meant with "a) forward and reverse mode autograd using torch.func work on both the .forward and the .adjoint" in #303?

### DIFF
--- a/src/mrpro/operators/LinearOperator.py
+++ b/src/mrpro/operators/LinearOperator.py
@@ -59,6 +59,10 @@ class _AutogradWrapper(torch.autograd.Function):
     def backward(ctx: Any, *grad_output: torch.Tensor) -> tuple[None, None, Any | None]:  # noqa: ANN401
         return None, None, _AutogradWrapper.apply(ctx.bw, ctx.fw, grad_output[0])
 
+    @staticmethod
+    def jvp(ctx: Any, *grad_inputs: Any) -> Any:  # noqa: ANN401
+        return _AutogradWrapper.apply(ctx.fw, ctx.bw, grad_inputs[-1])
+
 
 class LinearOperator(Operator[torch.Tensor, tuple[torch.Tensor]]):
     """General Linear Operator.
--- a/src/mrpro/operators/LinearOperator.py
+++ b/src/mrpro/operators/LinearOperator.py
@@ -16,9 +16,9 @@
 
 from __future__ import annotations
 
-from abc import abstractmethod
 from collections.abc import Callable
 from collections.abc import Sequence
+from typing import Any
 from typing import overload
 
 import torch
@@ -31,6 +31,33 @@ from mrpro.operators.Operator import OperatorSum
 from mrpro.operators.Operator import Tin2
 
 
+class _AutogradWrapper(torch.autograd.Function):
+    """Wrap forward and adjoint functions for autograd."""
+
+    # If the forward and adjoint implementation are vmap-compatible,
+    # the function can be marked as such to enable vmap support.
+    generate_vmap_rule = True
+
+    @staticmethod
+    def forward(
+        fw: Callable[[torch.Tensor], torch.Tensor], bw: Callable[[torch.Tensor], torch.Tensor], x: torch.Tensor
+    ) -> Any:
+        return fw(x)
+
+    @staticmethod
+    def setup_context(
+        ctx: Any,
+        inputs: tuple[Callable[[torch.Tensor], torch.Tensor], Callable[[torch.Tensor], torch.Tensor], torch.Tensor],
+        output: torch.Tensor,
+    ) -> torch.Tensor:
+        ctx.fw, ctx.bw, x = inputs
+        return output
+
+    @staticmethod
+    def backward(ctx, *grad_output: torch.Tensor):
+        return None, None, _AutogradWrapper.apply(ctx.bw, ctx.fw, grad_output[0])
+
+
 class LinearOperator(Operator[torch.Tensor, tuple[torch.Tensor]]):
     """General Linear Operator.
 
@@ -39,10 +66,13 @@ class LinearOperator(Operator[torch.Tensor, tuple[torch.Tensor]]):
     with a,b scalars and x,y tensors.
     """
 
-    @abstractmethod
-    def adjoint(self, x: torch.Tensor) -> tuple[torch.Tensor,]:
-        """Adjoint of the operator."""
-        ...
+    def forward(self, x: torch.Tensor) -> tuple[torch.Tensor]:
+        """Apply the operator to x."""
+        return (_AutogradWrapper.apply(self._forward_implementation, self._adjoint_implementation, x),)
+
+    def adjoint(self, x: torch.Tensor) -> tuple[torch.Tensor]:
+        """Apply the adjoint of the Operator to x."""
+        return (_AutogradWrapper.apply(self._adjoint_implementation, self._forward_implementation, x),)
 
     @property
     def H(self) -> LinearOperator:  # noqa: N802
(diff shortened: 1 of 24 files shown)


fzimmermann89:
i am sorry. I went for a giro and my computer turned off again. I will not start writing a response a forth time.

fzimmermann89:
we also need a 
```
@staticmethod
def jvp(ctx, *grad_inputs):
   """Forward Mode AD"""
   return _AutogradWrapper.apply(ctx.fw, ctx.bw, grad_inputs[-1])
```

tests/helper.py
```diff
@@ -82,3 +93,130 @@ def dotproduct_adjointness_test(
...
+    AssertionError
+        if autodiff fails
+
+
+    """
+    # Forward-mode autodiff using jvp
+    assert torch.func.jvp(operator.forward, u, u)
+
+    # Backward-mode autodiff using vjp
+    assert torch.func.vjp(operator.forward, *u)
```
fzimmermann89: consider 
https://pytorch.org/docs/stable/autograd.html#torch.autograd.detect_anomaly

fzimmermann89: torch.func.vjp(operator.forward, *u) does afaik not actually start the gradient calculation.
it returns function, that takes an element of the range and calculates the vjp.
you can use the frist output of  the torch.func.jvp(operator.forward, u, u) above , which will be an element of the range,

ckolbPTB:
Once we are happy with the overall structure I will extend it to all Operators.

src/mrpro/operators/LinearOperator.py
```diff
@@ -29,6 +30,39 @@
...
+    # If the forward and adjoint implementation are vmap-compatible,
+    # the function can be marked as such to enable vmap support.
+    generate_vmap_rule = True
+
+    @staticmethod
+    def forward(  # type: ignore [override]
+        fw: Callable[[torch.Tensor], torch.Tensor],
+        bw: Callable[[torch.Tensor], torch.Tensor],  # noqa: ARG004
+        x: torch.Tensor,
+    ) -> Any:  # noqa: ANN401
```
fzimmermann89: can this be typed to return torch.Tensor?

src/mrpro/operators/LinearOperator.py
```diff
@@ -29,6 +30,39 @@
...
+    def setup_context(
+        ctx: Any,  # noqa: ANN401
+        inputs: tuple[Callable[[torch.Tensor], torch.Tensor], Callable[[torch.Tensor], torch.Tensor], torch.Tensor],
+        output: torch.Tensor,
+    ) -> torch.Tensor:
+        ctx.fw, ctx.bw, x = inputs
+        return output
+
+    @staticmethod
+    def backward(ctx: Any, *grad_output: torch.Tensor) -> tuple[None, None, Any | None]:  # noqa: ANN401
```
fzimmermann89: Can we add type hints for the output? Or would this conflict with torch?

src/mrpro/operators/LinearOperator.py
```diff
@@ -37,10 +71,13 @@ class LinearOperator(Operator[torch.Tensor, tuple[torch.Tensor]]):
...
-    @abstractmethod
-    def adjoint(self, x: torch.Tensor) -> tuple[torch.Tensor,]:
-        """Adjoint of the operator."""
-        ...
+    def forward(self, x: torch.Tensor) -> tuple[torch.Tensor]:
+        """Apply the operator to x."""
+        return (_AutogradWrapper.apply(self._forward_implementation, self._adjoint_implementation, x),)
+
+    def adjoint(self, x: torch.Tensor) -> tuple[torch.Tensor]:
+        """Apply the adjoint of the Operator to x."""
```
fzimmermann89: capitalisation should be same as for foward

tests/operators/models/test_mono_exponential_decay.py
```diff
@@ -56,3 +57,11 @@ def test_mono_exponential_decay_shape(parameter_shape, contrast_dim_shape, signa
...
     (signal,) = model_op.forward(m0, decay_constant)
     assert signal.shape == signal_shape
+
+
+@pytest.mark.filterwarnings('ignore:Anomaly Detection has been enabled')
+def test_autodiff_exponential_decay():
+    """Test autodiff works for mono-exponential decay model."""
+    model = MonoExponentialDecay(decay_time=20)
+    m0, decay_constant = create_parameter_tensor_tuples(parameter_shape=(2, 5, 10, 10, 10), number_of_tensors=2)
+    autodiff_of_operator_test(model, m0, decay_constant)
```
fzimmermann89, line 67: what is this error??

test_autodiff_exponential_decay

DeprecationWarning: ast.Str is deprecated and will be removed in Python 3.14; use ast.Constant instead

fzimmermann89:
one  possible issue (wip,will do more detailed review end of week):
  _forward_implementation will not be added to docu and forward's docstring will always be 'apply linear operator'
  _I think I am fine with that._ 
  
  otherwise, lgtm! will review again once it is applied to all operators. good job :)

fzimmermann89:
maybe we can split the linear op changes from the signal model tests?

I.e. have one PR that includes the general tests (which should pass for all operators)
and a second PR which then does the adjoint-as-gradient changes?

I am also wondering what MOLLI should actually return for the signal and the value for some of the edge cases. And if it is a useful signal model with` a,b, and t1`, as it is difficult to use constraints on these parameters.

ckolbPTB:
> I.e. have one PR that includes the general tests (which should pass for all operators)
> and a second PR which then does the adjoint-as-gradient changes?

Do you mean to have 
- one PR which includes mainly the `autodiff_of_operator_test` for all operators including linear operators and 
- a second PR then with all the things specific for the LinearOperator

tests/operators/models/test_mono_exponential_decay.py
ckolbPTB: > DeprecationWarning: ast.Str is deprecated and will be removed in Python 3.14; use ast.Constant instead

where did you get this error?

fzimmermann89: It was auto annotated by the pytest action at the time I did the review (if there is no force push, you should be able to see it). Maybe a pytest update resolved the issue?

ckolbPTB:
Open discussion in #68

Issue #88, opened 2023-10-12
Using Coil Sensitivity Reference Scan (Siemens)


fzimmermann89:
We could at some point consider to have an option to use the AdjCoilSense data created by the scanner as a CSM method.

This would mean reading in the data and reformatting it to match the image data.

ckolbPTB:
On our Cima.X it seems as if AdjCoilSense is saved in dataset_idx=0 and contains interleaved data of body coil and local receiver coils. This currently leads to an error because acquisitions are e.g. of shape (2,100) and (32,100).

fzimmermann89:
Do they have different flags?

ckolbPTB:
> Do they have different flags?

not that I can tell

fzimmermann89:
Should we instead of trying to read in the adjustment data as KData, just add a csmdata.from_siemens_adjustement_scan(filename, dataset_id) method (or similar) that loads in the body coil and the coil elements, does the calculation etc..

fzimmermann89:
What is the status here?

fzimmermann89:
or at least add an example how to read it it... @Stef-Martin  und @ckolbPTB did this recently for some data iirc

fzimmermann89:
it still needs some resampling, correct? the adjustment scan is always with te same orientation or somehting? Or am I misremembering the issue?

PR #309, opened 2024-06-12
Select k-space data based on n_coils

ckolbPTB:
Currently we get the number of coils (n_coils) as a header parameter but we do not check that the raw data actually has these many coils.

This leads to problems for coil reference scans, because these scans contain acquisitions obtained with the surface coils (e.g. 16-element cardiac coil) and body coil (2-element coil). This data cannot be stacked and hence yields an error.

This PR only tries to stack acquisitions into a tensor where n_coils fits to the number of coil elements. This now also allows to set the desired number of coils and get the data from the surface coils or the body coil.


src/mrpro/data/_kdata/KData.py
```diff
@@ -121,6 +121,39 @@ def from_file(
...
+            ismrmrd_header.acquisitionSystemInformation is not None
+            and ismrmrd_header.acquisitionSystemInformation.receiverChannels is not None
+        ):
+            n_coils = ismrmrd_header.acquisitionSystemInformation.receiverChannels
+        if n_coils is not None:
+            if n_coils not in n_coils_available:
+                raise ValueError(f'No acquisitions for {n_coils} coil elements found.')
+            acquisitions = [acq for acq in acquisitions if acq.data.shape[0] == n_coils]
+            if len(n_coils_available) > 1:
+                warnings.warn(
```
ckolbPTB: Not sure if we need this. Also happy to remove it.

fzimmermann89:
An idea (not fully thought out):

We already have predicates for selecting the acquisition based on flags.
We could extend this mechanism, such that the full acquisition is given to the predicates.
Then we could also filter on the number of coils.
And maybe predefine either a bodycoil filter of a n_coils filter.

tests/data/_IsmrmrdRawTestData.py
```diff
@@ -236,9 +237,19 @@ def __init__(
...
 
+        # Add acquisitions obtained with a 2-element body coil (e.g. used for adjustment scans)
+        if add_bodycoil_acquisitions:
+            acq.resize(n_freq_encoding, 2, trajectory_dimensions=2)
+            for _ in range(8):
+                acq.scan_counter = counter
+                acq.clearAllFlags()
+                acq.data[:] = torch.randn(2, n_freq_encoding, dtype=torch.complex64)
+                dataset.append_acquisition(acq)
+                counter += 1  # increment the scan counter
```
fzimmermann89: call it acquisition_counter and remove the comments?

src/mrpro/data/_kdata/KData.py
```diff
@@ -109,6 +109,40 @@ def from_file(
...
+        n_coils_available = {acq.data.shape[0] for acq in acquisitions}
+
+        # We need to handle the n_coils parameter of header_overwrites here separately rather than in
+        # KHeader.from_ismrmrd, because the acquisitions have to be selected before we can calculate the header
+        if len(n_coils_available) > 1:
+            if header_overwrites is not None and 'n_coils' in header_overwrites:
+                if isinstance(header_overwrites['n_coils'], int | float):
+                    n_coils = int(header_overwrites['n_coils'])
+                else:
+                    raise ValueError(f'n_coils has to be numeric but is {type(header_overwrites["n_coils"])}')
```
fzimmermann89: why do we allow float here?

src/mrpro/data/_kdata/KData.py
```diff
@@ -109,6 +109,40 @@ def from_file(
             modification_time = datetime.datetime.fromtimestamp(mtime)
 
         acquisitions = [acq for acq in acquisitions if acquisition_filter_criterion(acq)]
+
+        # Select only acquisitions with the desired number of receiver channels/coils.
```
fzimmermann89, line 113: is this comment still correct?

src/mrpro/data/_kdata/KData.py
```diff
@@ -109,6 +109,40 @@ def from_file(
...
+        # If the number of coils is not defined use all acquisitions
```
fzimmermann89: i think think comment is wrong
`# Ensure consistent number of receiver coils by filtering, default to the number matching the header
`
and move the` # We need to handle the n_coils parameter of header_overwrites here separately rather than` ...
into the if statement where you use use it.

and add a `best-effort guess` or something like that to the max(n_coils_available) line?

src/mrpro/data/_kdata/KData.py
```diff
@@ -109,6 +109,40 @@ def from_file(
...
+            warnings.warn(
+                f'Acquisitions with different number {n_coils_available} of receiver coil elements detected.'
+                'Data with {n_coils} receiver coil elements will be used.',
+                stacklevel=1,
+            )
+
+            # Select acquisition with n_coils receiver coils
+            acquisitions = [acq for acq in acquisitions if has_n_coils(n_coils, acq)]
+
+        # Make sure there are any acquisitions left
```
fzimmermann89: nit-picking: does this comment help to understand what is going on here, considering the clearness of the following two lines ?

fzimmermann89 (commented):
minor comments

tests/data/_IsmrmrdRawTestData.py
fzimmermann89: oder scan_counter?

ckolbPTB:
pre-commit.ci autofix

fzimmermann89:
#88

fzimmermann89:
we want to drop n_coils from the header.
maybe we can do this first, already in this PR, before restructuring the rest of the headers?
This would simplify this PR can we can merge it.

fzimmermann89:
if your are fine with the changes and maybe after  another look if I messed something up, feel free to merge.

or, undo my
 commits and we wait.

fzimmermann89 (approved):

fzimmermann89:
This should be merged into main after unit_conversion_utils has been merged

src/mrpro/data/_kdata/KData.py
```diff
@@ -110,6 +110,29 @@ def from_file(
...
+        # we need the same number of receiver coils for all acquisitions
+        n_coils_available = {acq.data.shape[0] for acq in acquisitions}
+        if len(n_coils_available) > 1:
+            if (
+                ismrmrd_header.acquisitionSystemInformation is not None
+                and ismrmrd_header.acquisitionSystemInformation.receiverChannels is not None
+            ):
+                n_coils = int(ismrmrd_header.acquisitionSystemInformation.receiverChannels)
+            else:
+                # most likely, these the coils used for imaging
```
fzimmermann89:
```suggestion
                # most likely, highest number of elements are the coils used for imaging
```

PR #312, opened 2024-06-16
Tests for signal models separated

ckolbPTB:
The tests for each signal model should be separated. 

Having the tests to verify the shapes of all the signal models in one file is different to all of our other tests. In addition, more complex signal models cannot easily be integrated into this one-for-all shape test.


schuenke (approved):
adapted the style of the copyright section to match the one in the src files. Should be done for all files in `tests` at some point.

PR #315, opened 2024-06-19
Select ISMRMRD acquisitions with a function instead of ignore flags

### LINKED ISSUE #305: Flags for reading in coil map data
In order to be able to calculate coil maps for a Cartesian SENSE scan, we need to read in only the data with flag == ACQ_IS_PARALLEL_CALIBRATION and flag == ACQ_IS_PARALLEL_CALIBRATION_AND_IMAGING. 

Currently we can only set an ignore_flag but it is not possible to ignore the data without any flag because NO_FLAG has a value of 0 which is tricky to encode in the bit-logic. 

Suggestion: Give NO_FLAG the highest possible value in MRpro and then have a separate check in KData (similar to ACQ_IS_PARALLEL_CALIBRATION_AND_IMAGING/ACQ_IS_PARALLEL_CALIBRATION logic) for this case.

ckolbPTB:
closes #305


fzimmermann89:
Suggestion: consider putting the predicates either in an acqfilter namespace (i.e. .py) within data or in kdata.py.
These serve a single purpose within the data logic, utils does not feel like a good place for them.

Also, consider renaming them to is_* as they are returning a bool if the condition is matched

And the parameters to acquisition_predicate or acq_filter_criterion - it is not a selection function.

..and there should be no fun. ;)

schuenke (approved):
lgtm

PR #318, opened 2024-06-21
Docker build for main

### LINKED ISSUE #244: Docker Cache
Not sure if it applies, just came across
 https://blacksmith.sh/blog/cache-is-king-a-guide-for-docker-layer-caching-in-github-actions

Tldr

   - name: Build and push Docker image
        uses: docker/build-push-action@v5
        with:
          context: .
          push: true
          tags: user/app:latest
          cache-from: type=gha
          cache-to: type=gha,mode=max

For caching layers

ckolbPTB:
evaluate #244

Wrong pathname meant that the docker build was not triggered if docker.yml was modified -> fixed

Currently the docker build pipeline is only run when certain files (files in the docker folder, docker.yml or pyproject.toml) are changed. For docker.yml or any other file in the docker folder this is all fine. 

Changes to pyproject.toml (e.g. adding a new package) on the other hand do not have any effect on the container build because the packages inside the container a created by a checkout of MRpro main branch. Afterwards we do run pytest inside the container which uses the new pyproject.toml. Afterwards we push the container image based on the old pyproject.toml to the registry and this container image is then used for all subsequent tests. If we never update any of the above mentioned files which trigger this docker pipeline, the container will never be updated to the latest pyproject.toml.

Therefore I removed the trigger which starts a new container build when pyproject.toml is changed (because it doesn't do anything, and it is tested anyway in the normal pytest pipeline) and I added a new trigger for all pushes to main. This will mean that we will build the container image more often than we need to but at least it will be up-to-date.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -68,6 +68,9 @@ jobs:
           dockerfile=${{ matrix.dockerfile }}
           echo "image_name=${dockerfile/Dockerfile_/ghcr.io/ptb-mr/mrpro_}" >> $GITHUB_OUTPUT
 
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
       - name: Login to GitHub Packages
         uses: docker/login-action@v3
         with:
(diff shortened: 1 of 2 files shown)


schuenke (approved):
lgtm, but we only know if it works as expected when we try it, right? So lets go...

PR #319, opened 2024-06-24
Transient steady state model

ckolbPTB:
This signal model describes a continuous data acquisition after a preparation pulse assuming perfect spoiling (i.e. ignoring any T2 effects). 

This signal model describes the data aquisition used by Simone and for the 2s-T1-mapping proposed by Kirsten.


src/mrpro/operators/models/TransientSteadyStateWithPreparation.py
```diff
@@ -0,0 +1,123 @@
...
+
+        Parameters
+        ----------
+        m0
+            equilibrium signal / proton density
+            with shape (... other, coils, z, y, x)
+        t1
+            longitudinal relaxation time T1
+            with shape (... other, coils, z, y, x)
+        flip_angle
```
schuenke, line 102: Maybe add the info (again), that flip angles are supposed to be in rad

src/mrpro/operators/models/TransientSteadyStateWithPreparation.py
```diff
@@ -0,0 +1,123 @@
...
+        sampling_time: float | torch.Tensor,
+        repetition_time: float,
+        m0_scaling_preparation: float = 1.0,
+        delay_after_preparation: float = 0.0,
+    ):
+        """Initialize transient steady state signal model.
+
+        Parameters
+        ----------
+        sampling_time
```
schuenke, line 70: sampling_time = 0 is directly AFTER the spoiler from the inversion block, no?! Maybe add a more detailed description here.

src/mrpro/operators/models/TransientSteadyStateWithPreparation.py
```diff
@@ -0,0 +1,123 @@
...
+            with shape (time ... other, coils, z, y, x)
+        """
+        delta_ndim = m0.ndim - (self.sampling_time.ndim - 1)  # -1 for time
+        sampling_time = self.sampling_time[..., *[None] * (delta_ndim)] if delta_ndim > 0 else self.sampling_time
+
+        # effect of preparation pulse
+        m_start = m0 * self.m0_scaling_preparation
+
+        # relaxation towards M0
+        t1 = torch.where(t1 == 0, 1e-10, t1)
```
schuenke: I remember @Stef-Martin reporting that using torch.where() might be problematic in some cases? What's the status here? Does @fzimmermann89 know something about this?

schuenke (approved):

guastinimara:
I tried it on Kirsten data and it works perfectly fine

src/mrpro/operators/models/TransientSteadyStateWithPreparation.py
fzimmermann89: Why is this required?

I would only ever add these hacks if in the end they result in the same value and are only required for numerical reasons.
It is arbitrary that the signal model returns something if t1 is exactly 0.

tests/operators/models/test_transient_steady_state_with_preparation.py
```diff
@@ -0,0 +1,86 @@
...
+        torch.testing.assert_close(signal[0, ...], m0)
+    # Assert closeness to 0
+    elif result == '0':
+        torch.testing.assert_close(signal[0, ...], torch.zeros_like(m0))
+    # Assert closeness to -m0
+    elif result == '-m0':
+        torch.testing.assert_close(signal[0, ...], -m0)
+    # Assert closensess to m0*
+    elif result == 'm0*':
+        t1 = torch.where(t1 == 0, 1e-10, t1)
```
fzimmermann89, line 53: why?

src/mrpro/operators/models/TransientSteadyStateWithPreparation.py
fzimmermann89: if that is the case, we might add a comment

also, what about t1=1e-30?

fzimmermann89:
```
lncostr=torch.log(torch.cos(flip_angle)) / repetition_time
r1star=1/t1 - lncostr
m0star=m0/(1-t1*lncostr)
signal = m0_star + (m_start - m0_star) * torch.exp(-sampling_time*r1_star)
```

PR #322, opened 2024-06-25
Select number of cores for pytest in VSCode automatically

### LINKED ISSUE #87: Suggestion: Run tests in parallel

ckolbPTB:
closes #87 

@schuenke @koflera @fzimmermann89 @Stef-Martin @guastinimara and who ever has time please checkout main and run all tests in VSCode. Then checkout this branch and run it again and check that 

- all your tests are still run
- there are no strange error/warnings in the _Test Results_ - terminal even though you get a green tick in the VSCode test overview
- the tests are finished actually faster than before

On MacOs with VSCode Version: 1.89.1 it seems to be working.


schuenke:
For me and my high-end PTB notebook it works, but it's only about 10% faster (which might actually not be reproducible 😁 )

schuenke:
Let's merge!? All tests were positive until now and if it does break something somehow, we can still revert it?!

schuenke (approved):

PR #324, opened 2024-06-27
Remove pytest filter warning for optimizers

ckolbPTB:
Check if filterwarning is still needed.


schuenke (approved):
not needed, so it can be merged, no?!

PR #332, opened 2024-07-02
Update documentation

schuenke:
Our documentation only included parts of our code so far.

I added the remaining submodules "Algorithms" and "Operators" and renamed/reordered the menu entries. 

supersedes #313


ckolbPTB:
We are currently using MRpro, mrpro and MRPro in the documentation. Would be great to unify this. I would suggest MRpro

ckolbPTB:
Maybe we should review and merge this PR first #313

schuenke:
> Maybe we should review and merge this PR first #313

Sorry, I missed that one. Approved it, so that we can merge it here afterwards. 

BTW: Was there any specific reason why we chose the `pydata_sphinx_theme`?
From the style (especially the menu) I would prefer the very common Read The Docs Theme (`sphinx_rtd_theme`), with an adjusted maximum width of the main content window (e.g. 75%). This would look like this:

grafik

docs/source/api.rst
```diff
@@ -3,8 +3,18 @@ API
 
 This is the mrpro api documentation.
```
ckolbPTB: MRpro

ckolbPTB:
> BTW: Was there any specific reason why we chose the `pydata_sphinx_theme`?

I don't think so or at least I cannot remember it. I am fine with the `sphinx_rtd_theme` style.

ckolbPTB (approved):

PR #334, opened 2024-07-02
Remove torch.nn.module... from api doc

### LINKED ISSUE #333: Clean-up inherited functions in docu
We should remove all the inherited torch.nn.module functions from our docu and only link to pytorch.
Maybe we can have an exception list somewhere which ones to keep in the docu.
A lot of these functions might be confusing, do not work for us or do nothing useful for operators.
Even in pytorch, the list of functions of torch.nn.module is confusing because of lots of legacy functions..

ckolbPTB:


ckolbPTB:
I thought that would be easy to fix but it is not...

fzimmermann89:
Do you know (or speculate) why this does not work?

ckolbPTB:
Any changes I made in `class_template.rst` seem to only affect the overview of the class but not the listed members. Need to better understand how this recursive autosummary approach of sphinx works. 

So I can go from
grafik

to get rid of everything in the overview on the right but still present in the list below:
grafik

ckolbPTB:
Doing something like 
```
   .. autosummary::
   {% for item in methods %}
   {% if not item == 'add_module' %}
      ~{{ name }}.{{ item }}
   {%- endif %}
   {%- endfor %}
   {% endif %}
   {% endblock %}
```
in class_template.rst removes `add_module` from the list but this seems a very tedious solution.

fzimmermann89:
We could also jsut remove the autosummary.
The are listed below by autodoc anyways (which can be controlled by the change in this PR)
We might want to keep autosummary for the attributes only (then we only have to remove 2-3 pytorch attributes), or add attributes also to the autodoc output.
Either way, I would prefer the attributes to be listed in a table. 

So this is a larger thing to clean up our doc template...

fzimmermann89:
But this PR still lgtm to partially address the issue

PR #337, opened 2024-07-02
Remove optional and default in doc string

ckolbPTB:


schuenke (approved):
There are several more cases where the default value is mentioned in the docstring, but as these are not obvious and thus also include additional explanations, we keep them for the moment.

src/mrpro/data/KTrajectory.py
```diff
@@ -52,7 +52,7 @@ class KTrajectory(MoveDataMixin):
     """tolerance of how close trajectory positions have to be to integer grid points."""
 
     repeat_detection_tolerance: float | None = 1e-3
-    """tolerance for repeat detection, by default 1e-3. Set None to disable."""
+    """tolerance for repeat detection. Set None to disable."""
```
fzimmermann89: Set to None?

src/mrpro/utils/Rotation.py
```diff
@@ -441,7 +441,6 @@ def from_euler(cls, seq: str, angles: torch.Tensor | _NestedSequence[float] | fl
             (`degrees` is True).
         degrees
             If True, then the given angles are assumed to be in degrees.
-            Default is False.
```
fzimmermann89, line 444: Otherwise they are assumed to be in radians

fzimmermann89 (approved):

Issue #311, opened 2024-06-16
homogenize delayed annotation evaluation in all files


fzimmermann89:
Might be useful for consistency to always  `from __future__ import annotations`

This can be done using ruff:

```
[tool.ruff.lint.isort]
required-imports = ["from __future__ import annotations"]
```
@schuenke : what do you think?

..or only do it when it is really required. currently, we do it in a lot of random files.

schuenke:
do we still need it when we use the `from typing import Self` style?

fzimmermann89:
..or remove them from all files and see where we actually need them.

schuenke:
If we switch to the `from typing import Self` annotation style, we might need to specify that the mypy pre-commit hook uses a python version >=3.11. This was an Issue for me in another repo... this is how the pre-commit config might need to look like:

```yaml
  - repo: https://github.com/pre-commit/mirrors-mypy
    rev: v1.10.0
    hooks:
      - id: mypy
        pass_filenames: false
        always_run: true
        args: [src, tests, examples, --python-version=3.11]
```

fzimmermann89:
> If we switch to the `from typing import Self` annotation style, we might need to specify that the mypy pre-commit hook uses a python version >=3.11. This was an Issue for me in another repo... this is how the pre-commit config might need to look like:
> 
> ```yaml
>   - repo: https://github.com/pre-commit/mirrors-mypy
>     rev: v1.10.0
>     hooks:
>       - id: mypy
>         pass_filenames: false
>         always_run: true
>         args: [src, tests, examples, --python-version=3.11]
> ```

No, this only tells mypy what python version to consider, not which version is used to run mypy.
not sure what the actual problem was that you experienced, but this should not be required for our code. also, we already use many python 3.11 only typing features.

PR #343, opened 2024-07-04
Remove dependencies of pre-commit

### LINKED ISSUE #339: Potential missmatch between dependency versions in pre-commit and PyTest
In our [pyproject.toml](https://github.com/PTB-MR/mrpro/blob/main/pyproject.toml) we define our dependencies including version limitations, e.g.
`numpy>=1.23,<2.0` or `torch>=2.3,<3.0`

In our [.pre-commit-config.yaml](https://github.com/PTB-MR/mrpro/blob/main/.pre-commit-config.yaml), we explicitly list the additional dependencies for mypy, e.g. `numpy` or `torch>=2.3.0`

This mismatch in the specified versions might actually lead to different versions being used in our tests (e.g. numpy <2.0) and in the pre-commit checks like mypy (e.g. numpy > 2.0). I realized this in my [BMCTool Repo](https://github.com/schuenke/BMCTool/pull/28), where I now use the same dependency limits in the pre-commit configuration.

Maybe we can find a better way to adjust these dependencies automatically...

ckolbPTB:
closes #339


fzimmermann89:
No.

precommit runs in a clean environment by design.

all dependencies must either be installed or mypy set to ignore missing type information (which we don't want as it would also ignore all related typing errors)

this is not a viable approach.

also, the dependencies are different than the runtime types (we don't need request-types at runtime and we don't need pypulseq for mypy)

schuenke:
> No.
> 
> precommit runs in a clean environment by design.
> 
> all dependencies must either be installed or mypy set to ignore missing type information (which we don't want as it would also ignore all related typing errors)
> 
> this is not a viable approach.
> 
> also, the dependencies are different than the runtime types (we don't need request-types at runtime and we don't need pypulseq for mypy)

Exactly. And this can actually mean that different package or even python versions are used in the mypy pre-commit environment as mentioned in #339 and #311

fzimmermann89:
please also consider local usage. the pre-commit is mainly meant as a local pre-commit check.

your local pre-commit will only update if the pre-commit config is changed. 
we require pytorch>=2.3 for our type hints. this should be increased if we require a newer version for the type hints to work.

fzimmermann89:
> Exactly. And this can actually mean that different package or even python versions are used in the mypy pre-commit environment as mentioned in #339 and #311

No, the  regular python depending on your path settings should be used be pre-commit...

ckolbPTB:
Why would I not want to run mypy in my local MRpro-environment?

ckolbPTB:
My understanding is that mypy creates a new environment using the _regular python depending on your path settings_. If we run mypy in the container it will default to the versions installed there and if you run it locally it will default to the versions in your local environment. 

If this is correct I see two options:
- not specifying any versions in mypy. this will work in our CI pipeline and also locally assuming you run mypy in your local MRpro-environment.

- defining all the versions in mypy too. The advantage would be that you would get the correct versions in any environment you run mypy. For this to work I don't think that `pytorch>=2.3` is enough because you could have pytorch 3.1 in your local environment and then this would pass but MRpro would still require <3.0. So the disadvantage would be that we would have to copy over the exact same requirements from your pyproject.toml (and always remember to adapt them because in our CI pipeline it would always work independent if we are doing it right or not). 

Any opinions?

PS I think mypy currently fails because there was a matplotlib update.

fzimmermann89:
If we want to keep using pre-commit, we must specify dependencies and versions for mypy. Otherwise, the local dev experience will suffer.

Or we remove the pre-commit config, everybody uses their setup to call mypy, ruff etc and we add separate github actions for these steps.

my suggestion would be to fix all dependencies in the mypy precommit to concrete versions and manually bump them. 
everything else would just result in more yak shaving for minimal benefit.

ckolbPTB:
> Otherwise, the local dev experience will suffer.

Why?

My main concern with the manual adaption of the version is, that I don't see a way of detecting any mistakes we make and if something is manual then we will make mistakes eventually.

schuenke:
I did some more reading about additional dependencies in pre-commit hooks again and I might have found an elegant solution.

Currently, we don't define any `language` in our mypy pre-commit config, which then defaults to `python`, which means that mypy is run in the mentioned isolated environment.

However, you can specify `language: system` to use the environment from which the pre-commit command was called. This works fine locally for me and should also work in our CI pipeline, but it might lead to problems when using a (GUI) Git client... dunno...

If this sounds reasonable to @ckolbPTB and @fzimmermann89 I can push the changes here. 

# EDIT:
**It seems that it's NOT working with the Source Control Management in VSCode...** 😞

fzimmermann89:
I would still just leave this as is...

schuenke:
> I would still just leave this as is...

I agree. Should we just close this PR?

ckolbPTB:
Fine by me - shall we then also close #339?

PR #345, opened 2024-07-07
Csm Inati 2D and 3D

### LINKED ISSUE #64: Csm inati method

ckolbPTB:
supersedes #173 
closes #64 

Too many things changes since we last worked on this so I started a fresh PR.

### DIFF
--- a/src/mrpro/algorithms/csm/inati.py
+++ b/src/mrpro/algorithms/csm/inati.py
@@ -52,11 +52,11 @@ def inati(
     # Covariance with shape (z y x coils coils)
     coil_images_covariance = torch.einsum('i...j,k...j->...ik', coil_images_roi.conj(), coil_images_roi)
     singular_vector = torch.sum(coil_images_roi, dim=-1)  # coils z y x
-    singular_vector /= singular_vector.abs().square().sum(0, keepdim=True).sqrt()
+    singular_vector /= singular_vector.norm(dim=0, keepdim=True)
     for _ in range(max_power_iterations):
         singular_vector_old = singular_vector
         singular_vector = torch.einsum('...ij,j...->i...', coil_images_covariance, singular_vector)  # coils z y x
-        singular_vector /= singular_vector.abs().square().sum(0, keepdim=True).sqrt()
+        singular_vector /= singular_vector.norm(dim=0, keepdim=True)
         if torch.isclose(singular_vector, singular_vector_old, rtol=1e-1).all():
             break
 
--- a/src/mrpro/algorithms/csm/inati.py
+++ b/src/mrpro/algorithms/csm/inati.py
@@ -9,6 +9,7 @@ from mrpro.utils.sliding_window import sliding_window
 def inati(
     coil_images: torch.Tensor,
     smoothing_width: SpatialDimension[int] | int,
+    max_power_iterations: int = 1,
 ) -> torch.Tensor:
     """Calculate a coil sensitivity map (csm) using an the Inati method [INA2013]_ [INA2014]_.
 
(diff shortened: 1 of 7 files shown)


ckolbPTB:
I verified the result with the output of the old PtbPyRecon inati code for several examples.

ckolbPTB:
One last problem is using `torch.is_close`:
UserWarning: There is a performance drop because we have not yet implemented the batching rule for aten::isclose. 

Maybe somebody with more pytorch-knowledge could comment if this is a problem and we should change this or accept the warning in the tests.

src/mrpro/data/CsmData.py
```diff
@@ -53,6 +54,31 @@ def from_idata_walsh(
...
+        ----------
+        idata
+            IData object containing the images for each coil element.
+        smoothing_width
+            Size of the smoothing kernel.
+        """
+        from mrpro.algorithms.csm.inati import inati
+
+        # convert smoothing_width to SpatialDimension if int
+        if isinstance(smoothing_width, int):
```
fzimmermann89: why?

inati also accepts int.

fzimmermann89 (approved):
I am wondering if in the end it might be smarter to just set the number of iterations to a fixed value of say 10 and use vmap? 

Would this be faster if run for say a 32 image batch on GPU?
In case we ever want to use this during a NN training..

Could you test the performance difference @ckolbPTB or should I do it?
It might be either not worth it are a large difference, I really do not know.

Why is the relative change currently set to 0.1 (this seems high?) ? how many iterations would this corrpsopond to on some real data example?

src/mrpro/algorithms/csm/inati.py
```diff
@@ -0,0 +1,67 @@
...
+    if any(ks % 2 != 1 for ks in [smoothing_width.z, smoothing_width.y, smoothing_width.x]):
+        raise ValueError('kernel_size must be odd')
+
+    ks_halved = [ks // 2 for ks in smoothing_width.zyx]
+    padded_coil_images = torch.nn.functional.pad(
+        coil_images,
+        (ks_halved[-1], ks_halved[-1], ks_halved[-2], ks_halved[-2], ks_halved[-3], ks_halved[-3]),
+        mode=padding_mode,
+    )
+    # Get the voxels in an ROI defined by the smoothing_width around each voxel leading to shape (coils z y x ks*ks)
```
fzimmermann89: this comment might be wrong? is it ks*ks`?  (or only in 2d)?

fzimmermann89:
2d images are now also replicate padded. this does not change the result, only makes it a bit less efficient? Or does it change the result?

ckolbPTB:
> Why is the relative change currently set to 0.1 (this seems high?) ? how many iterations would this corrpsopond to on some real data example?

For my test data (2 phantoms + 1 in-vivo) this leads to around 30 iterations.

> I am wondering if in the end it might be smarter to just set the number of iterations to a fixed value of say 10 and use vmap?

I will check that

> Would this be faster if run for say a 32 image batch on GPU? In case we ever want to use this during a NN training..

For 2D cases, vmap leads to a significant speedup compared to smap (2-6 for n_batch=10). For 3D cases (even if it is just a few slices) there is no difference anymore between smap and vmap.

ckolbPTB:
> I am wondering if in the end it might be smarter to just set the number of iterations to a fixed value of say 10 and use vmap?

10 seems to be a good value. For my test data there are only minimal changes in the csm in the background between 10 and 20 iterations.

src/mrpro/algorithms/csm/inati.py
ckolbPTB: It is prod(smoothing_width)

ckolbPTB:
Currently if `smoothing_width` is `int` then we pad + smooth also along the z direction even if the data is 2D. I guess we do not want that...

PR #346, opened 2024-07-08
Minimize listed inherited functions/attributes in doc

### LINKED ISSUE #333: Clean-up inherited functions in docu

ckolbPTB:
Supersedes #334
closes #333


ckolbPTB:
I managed to clean up most of the problems in your documentation. I now also added a flag such that any warning in the sphinx-build results in an error. 

These things are still to be solved:

from ... import ... leads to problems in the sphinx-build because it seems to create the docu again for the imported module/function in the file where it is imported to:
- [ ] acq_filters: `from mrpro.data.enums import AcqFlags` 
- [ ] EndomorphOperator: `from mrpro.operators.Operator import Operator`

At other places I managed to solve this by simply saying e.g. `import mrpro.data.enums` and then using `mrpro.data.enums.AcqFlags` in the code. For these too cases it do not work with the sphinx-build error message 'mrpro module not found'.

For some modules the path to the typehinting modules is wrong and not properly linked. I could not find a pattern what makes these modules different to the rest. The once I found where it doesn't work:

- [ ] KData
- [ ] KHeader
- [ ] IHeader (IData works fine)
- [ ] EncodingLimits
- [ ] all reconstruction algorithms. Strangely in reconstruction, the correct linking works for `direct_reconstruction` and `forward` but e.g. not for `recalculate_csm`

Update: I did find a pattern, it seems to affect functions where the default value is more complex, e.g. `dataclasses.field(default_factory=dict)` or `CsmData.from_idata_walsh`

fzimmermann89:
> I managed to clean up most of the problems in your documentation. I now also added a flag such that any warning in the sphinx-build results in an error. 
> 
> These things are still to be solved:
> 
> from ... import ... leads to problems in the sphinx-build because it seems to create the docu again for the imported module/function in the file where it is imported to:
> - [ ] acq_filters: `from mrpro.data.enums import AcqFlags` 
> - [ ] EndomorphOperator: `from mrpro.operators.Operator import Operator`
> 
> At other places I managed to solve this by simply saying e.g. `import mrpro.data.enums` and then using `mrpro.data.enums.AcqFlags` in the code. For these too cases it do not work with the sphinx-build error message 'mrpro module not found'.
> 
> For some modules the path to the typehinting modules is wrong and not properly linked. I could not find a pattern what makes these modules different to the rest. The once I found where it doesn't work:
> 
> - [ ] KData
> - [ ] KHeader
> - [ ] IHeader (IData works fine)
> - [ ] EncodingLimits
> - [ ] all reconstruction algorithms. Strangely in reconstruction, the correct linking works for `direct_reconstruction` and `forward` but e.g. not for `recalculate_csm`
> 
> Update: I did find a pattern, it seems to affect functions where the default value is more complex, e.g. `dataclasses.field(default_factory=dict)` or `CsmData.from_idata_walsh`

I would only use mrpro.enums.AcqFlags.. when it is absolutely necessary to avoid circular imports.
Changing valid code to make our somewhat cursed sphinx config work feels like having the tail wag the dog.

Is there no better way to fix it in the docu?
What kinds of problems?

ckolbPTB:
still a bug here
https://github.com/PTB-MR/mrpro/blob/f349c41ddfd07a3c92ba6e1ea36b3ebd21c5199a/.github/workflows/docs.yml#L181

fzimmermann89:
We could try disabling autosummary_imported_members ?

ckolbPTB:
Related issue #333 has been closed.

PR #349, opened 2024-07-09
Moved 1D,2D,3D dcf with  voronoi calculations to algorithms

### LINKED ISSUE #96: Refactor CSM + DCF calculation functions into module?
I am wondering if we want to keep the CSM/DCF calculation as partof the CSMData / DCFData classes or if we should move them into an algorithm/... namespace and have only the read/write/store functionality in the mrpro.data-classes.

Currently, they are part of the *Data-Classes. But for the reconstruction, we plan to have a separate namespace (not an IData.from_KData_cg....) 
So KData/IData/QData are different in scope than the CSM/DCF classes.
I think I prefer the KData/IData/QData way and have most of the more complex algorithmic/math stuff not in the data classes.

(We could also go a hybrid way and have the logic as part of a different module, but have small convenience classmethods in DCF/CSM that just call the logic if think it might be too cumbersome to have `CSM=mrpro.csm_estimation.walsh(images)` in an reconstruction script.

Stef-Martin:
closes #96


koflera (commented):
looks good to me

ckolbPTB:
Please also separate the tests, such that the algorithms are tested by tests in tests/algorithms/dcf/test_dcf_voronoi.py

JoHa0811 (approved):

ckolbPTB (commented):
example_traj_rpe and example_traj_spiral_2d not used in test_dcf_voronoi.py

example_traj_rad_2d not used in test_dcf_data.py

PR #353, opened 2024-07-09
Add callback to lbfgs and adam

### LINKED ISSUE #194: Progress bars / verbose flag / callback functions / ... etc

koflera:
includes a callable `callback `in `lbfgs `and `adam` to fix #194.

### DIFF
--- a/src/mrpro/algorithms/optimizers/adam.py
+++ b/src/mrpro/algorithms/optimizers/adam.py
@@ -61,7 +61,7 @@ def adam(
     decoupled_weight_decay
         whether to use Adam (default) or AdamW (if set to true) [1]_
     callback
-        user-provided function to be called after each iteration
+        function to be called after each iteration
 
     Returns
     -------
--- a/src/mrpro/algorithms/optimizers/adam.py
+++ b/src/mrpro/algorithms/optimizers/adam.py
@@ -20,6 +20,7 @@ from collections.abc import Sequence
 import torch
 from torch.optim import Adam
 
+from mrpro.algorithms.optimizers.OptimizerStatus import OptimizerStatus
 from mrpro.operators.Operator import Operator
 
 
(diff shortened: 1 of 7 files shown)


koflera:
not sure if it makes sense to emply a callback in `DirectReconstruction `as well.

src/mrpro/algorithms/optimizers/lbfgs.py
```diff
@@ -75,13 +80,23 @@ def lbfgs(
...
+        nonlocal iteration
         optim.zero_grad()
         (objective,) = f(*parameters)
         objective.backward()
+        if callback is not None:
+            state = optim.state[optim.param_groups[0]['params'][0]]
+            if state['n_iter'] > iteration:
+                callback({'solution': parameters, 'iteration_number': iteration})
+
+            iteration = state['n_iter']
```
ckolbPTB: can this be inside the if?

src/mrpro/algorithms/optimizers/adam.py
```diff
@@ -52,12 +54,14 @@ def adam(
...
     weight_decay
-        weight decay (L2 penalty)
+        weight decay (L2 penalty if decoupled_weight_decay is False)
     amsgrad
         whether to use the AMSGrad variant of this algorithm from the paper
         `On the Convergence of Adam and Beyond`
     decoupled_weight_decay
         whether to use Adam (default) or AdamW (if set to true) [1]_
+    callback
+        user-provided function to be called after each iteration
```
ckolbPTB: i would remove "user-provided" - all parameters are user-provided...

ckolbPTB (approved):

PR #354, opened 2024-07-09
Consider user indices when sorting kdata

### LINKED ISSUE #32: user idx 5 and 6 in siemens ismrmrd data
the siemens to ismrmrd converter seems to abuse these for storing something that is not an acqIdx:
https://github.com/ismrmrd/siemens_to_ismrmrd/issues/114
https://github.com/ismrmrd/siemens_to_ismrmrd/blob/545ee4dfb39db491f4d19e5b843225d08f216ae4/main.cpp#L1176

Do we remove these from our code?
Do we detect this case and unset them?

Our AcqIdx should have to same meaning for Philips and siemens data...

Pierrickkk:
We ignore user index 5 and 6.
These are abused by the ismrmrd converter to store other information.
We now print a warning if these are used and use all other user idx for sorting

closes #32

### DIFF
--- a/src/mrpro/data/_kdata/KData.py
+++ b/src/mrpro/data/_kdata/KData.py
@@ -124,10 +124,10 @@ class KData(KDataSplitMixin, KDataRearrangeMixin, KDataSelectMixin, KDataRemoveO
 
         acqinfo = AcqInfo.from_ismrmrd_acquisitions(acquisitions)
 
-        user_indices = (acqinfo.idx.user5, acqinfo.idx.user6)
+        user_index = (acqinfo.idx.user5, acqinfo.idx.user6)
 
-        for ind in user_indices:
-            if len(torch.unique(ind)) > 1:
+        for indx in user_index:
+            if len(torch.unique(indx)) > 1:
                 warnings.warn(
                     'The Siemens to ismrmrd converter currently (ab)uses '
                     'the user 5 and 6 indices for storing the kspace center line and partition number.\n'
--- a/src/mrpro/data/_kdata/KData.py
+++ b/src/mrpro/data/_kdata/KData.py
@@ -124,16 +124,16 @@ class KData(KDataSplitMixin, KDataRearrangeMixin, KDataSelectMixin, KDataRemoveO
 
         acqinfo = AcqInfo.from_ismrmrd_acquisitions(acquisitions)
 
-        for i in (5, 6):
-            user_indices = getattr(acqinfo.idx, f'user{i}', [])
-            for j in user_indices:
-                if j != 0:
-                    warnings.warn(
-                        'The Siemens to ismrmrd converter currently (ab)uses '
-                        f'the user {i} indices for storing the kspace center line and partition number\n'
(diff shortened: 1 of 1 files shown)


JoHa0811 (approved):

src/mrpro/data/_kdata/KData.py
```diff
@@ -96,6 +124,17 @@ def from_file(
 
         acqinfo = AcqInfo.from_ismrmrd_acquisitions(acquisitions)
 
+        for i in (5, 6):
+            user_indices = getattr(acqinfo.idx, f'user{i}', [])
+            for j in user_indices:
```
ckolbPTB: don't check every label (time consuming) but do something like if len(torch.unique(user_indices)) > 1

ckolbPTB (changes requested):

src/mrpro/data/_kdata/KData.py
fzimmermann89: Also, just write idx.user5 and idx.user6 explicitly instead of constructing a string and using getattr.

src/mrpro/data/_kdata/KData.py
```diff
@@ -96,6 +124,17 @@ def from_file(
 
         acqinfo = AcqInfo.from_ismrmrd_acquisitions(acquisitions)
 
+        user_indices = (acqinfo.idx.user5, acqinfo.idx.user6)
+
+        for j in user_indices:
```
ckolbPTB: for user_index in [acqinfo.idx.user5, acqinfo.idx.user6]:

src/mrpro/data/_kdata/KData.py
```diff
@@ -96,6 +124,17 @@ def from_file(
 
         acqinfo = AcqInfo.from_ismrmrd_acquisitions(acquisitions)
 
+        user_index = (acqinfo.idx.user5, acqinfo.idx.user6)
```
fzimmermann89: This is bad because it is only a certain subset of the user indices..

src/mrpro/data/_kdata/KData.py
```diff
@@ -96,6 +124,17 @@ def from_file(
...
+
+        for indx in user_index:
+            if len(torch.unique(indx)) > 1:
```
fzimmermann89: I would just write both checks explicitly:
If len(torch.unique(acqinfo.idx.user6) >1 or len(torch.unique(acqinfo.idx.user6) >1:


This is imho easier to see what is actually checked and it is only two cases. I don't see an advantage of a 'smarter' loop for only two special cases.

fzimmermann89 (changes requested):

fzimmermann89:
For future reference (this is what ismrmrd converter does)
    ismrmrd_acq.idx().user[5] = scanhead.ushKSpaceCentreLineNo;
    ismrmrd_acq.idx().user[6] = scanhead.ushKSpaceCentrePartitionNo;

fzimmermann89 (approved):
nice

ckolbPTB (approved):

Issue #304, opened 2024-06-05
discussion: SNR in source files


fzimmermann89:
Hi everybody,

I am still a bit dissatisfied with the amount of screen space the intro to all our source code files takes up.

1. I would prefer to NOT enforce one-import-per-line, ie. use from `typing import Any, overload, TypeVar`
    If you want to know where an import comes from, our IDE should help you anyways. You will not 
    remember that it was listed in a page long import section.
Similiar for our examples. The imports take up way to much space. This would also still be consistent with https://peps.python.org/pep-0008/#imports and https://google.github.io/styleguide/pyguide.html#typing-imports
Furthermore, we could reevaluate all of the imports if we really want to use the `from package import symbol` style. a) There are some inconsistencies in our code and b) it might be easier to understand `dataclasses.fields` than `fields` somewhere in the code (similar for other uncommon, non-mrpro and non-typing symbols)

2. Reconsider the copyright statement. There is seems to be no legal requirement to add it to all files. See also torch, numpy, scipy, apple swift,...
 At we could least consider shortening it to "This file is part of MRPro, copyright Physikalisch-Technische-Bundesanstalt  2024, (all other copyrights), and licensed under Apache License 2.0, see LICENSE)
(and fix the copyright in the license file...)

PR #355, opened 2024-07-09
Cleanup imports

### LINKED ISSUE #311: homogenize delayed annotation evaluation in all files

JoHa0811:
closes part 1 of #304 
closes #311


Stef-Martin (commented):
Looks good to me

schuenke (approved):
would be nice if we could enforce multiple imports per line, but ruff does not provide this option atm. This means we should check it manually in new PRs.

PR #359, opened 2024-07-10
Allow more than one 'other' dimension

### LINKED ISSUE #138: Make "other" dimension more flexible

guastinimara:
some initial changes to deal with some of the other dimension for issue #138.

Still working on this at the moment, just wanted to be sure that I understood it and that this is the way that we would like to have it


src/mrpro/data/IData.py
```diff
@@ -146,13 +146,15 @@ def get_unique_slice_positions(slice_pos_tag: TagType = 0x00191015):
...
         # stack required due to mypy: einops rearrange list[tensor]->tensor not recognized
         idata = torch.stack([_dcm_pixelarray_to_tensor(ds) for ds in dataset_list])
-        idata = rearrange(
-            idata,
-            '(other coils z) y x -> other coils z y x',
-            other=len(idata),
-            coils=1,
-            z=1,
-        )
+        # idata = rearrange(
```
fzimmermann89: this can just be kept. no need to change anything here.

src/mrpro/data/KNoise.py
```diff
@@ -67,6 +67,6 @@ def from_file(
         noise_data = torch.stack([torch.as_tensor(acq.data, dtype=torch.complex64) for acq in acquisitions])
 
         # Reshape to standard dimensions
-        noise_data = rearrange(noise_data, 'other coils (k2 k1 k0)->other coils k2 k1 k0', k1=1, k2=1)
+        noise_data = rearrange(noise_data, '... coils (k2 k1 k0)-> ... coils k2 k1 k0', k1=1, k2=1)
```
fzimmermann89: ...  (k2 k1 k0)-> ...  k2 k1 k0

src/mrpro/operators/CartesianSamplingOp.py
```diff
@@ -130,7 +130,7 @@ def adjoint(self, y: torch.Tensor) -> tuple[torch.Tensor,]:
         if not self._needs_indexing:
             return (y,)
 
-        y_kflat = rearrange(y, '... coil k2 k1 k0 -> ... coil (k2 k1 k0)')
+        y_kflat = rearrange(y, '... coils k2 k1 k0 -> ... coils (k2 k1 k0)')
```
fzimmermann89: '...  k2 k1 k0 -> ... (k2 k1 k0)'

fzimmermann89:
I addresse some of my previous remarks in the merge commit and found a couple of other instances

ckolbPTB:
I tested and fixed as many things as I could think of - please list any other potential issues for multiple other dimensions.

One important part is still our traj-tests:
https://github.com/PTB-MR/mrpro/blob/9dc116731154cade0029c8c58ee97013918d1a98/tests/conftest.py#L266

Here we also should add a case for multiple other dimensions. Currently this fails for the non-uniform part and will be added with #463

fzimmermann89:
We decided to close this for now.

If anybody wants to take it up again or notices an issue with limited other dimensions, consider reopening (or copy-and-pasting changes into a new PR)

PR #360, opened 2024-07-10
Separate conftest into subdirectories

### LINKED ISSUE #211: PyTest fixture definitions/locations
Currently, we have a single `conftest.py` file in our main `tests` directory including the majority of our fixtures.

However, some fixtures are defined directly in our test files (e.g. https://github.com/PTB-MR/mrpro/blob/9abccff816a5a55d657cfb592af69e08f96dac56/tests/data/test_kdata.py#L71), although they are used in other test files (in this case for example in https://github.com/PTB-MR/mrpro/blob/9abccff816a5a55d657cfb592af69e08f96dac56/tests/utils/test_modify_acq_info.py#L22)

If a fixture is not in `conftest.py`, we have to import it.
However, Ruff will detect and remove these imports as unused.

My suggestions would be to move ALL fixtures into `conftest.py`, but to make use of the possibility to have [multiple `conftest.py` files in multiple nested directories](https://docs.pytest.org/en/7.1.x/reference/fixtures.html#conftest-py-sharing-fixtures-across-multiple-files) to prevent the main `conftest.py` file to get to overcrowded and chaotic:

 
>  You can have multiple nested directories/packages containing your tests, and each directory can have its own conftest.py with its own fixtures, adding on to the ones provided by the conftest.py files in parent directories.

ckolbPTB:
closes #211

### DIFF
--- a/tests/algorithms/test_optimizers.py
+++ b/tests/algorithms/test_optimizers.py
@@ -16,7 +16,7 @@ import pytest
 import torch
 from mrpro.algorithms.optimizers import adam, lbfgs
 from mrpro.operators import ConstraintsOp
-from tests.operators._OptimizationTestFunctions import Rosenbrock
+from tests.operators import Rosenbrock
 
 
 @pytest.mark.parametrize('enforce_bounds_on_x1', [True, False])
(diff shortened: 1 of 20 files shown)


schuenke (approved):
👍

PR #361, opened 2024-07-10
Add test for minimum dependencies

schuenke:
Things to do:
- [ ] change to scheduled run (once every day/week?) when finished testing

Required changes so far:
- increase mininum numpy version from 1.23 to 1.26 (required to install numpy in py311/py312)
- increase minium torch version from 2.3 to 2.3.1 to pass https://github.com/PTB-MR/mrpro/blob/bdf849c7e876d6ec96810a43479ae64a24bd0956/tests/data/test_movedatamixin.py#L90

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -41,6 +41,25 @@ jobs:
         run: |
           echo "final list of images with tag latest: ${{ steps.set-matrix.outputs.imagenames }}"
 
+  test_min_dependency:
+    name: Run tests with minimum dependencies
+    needs: get_dockerfiles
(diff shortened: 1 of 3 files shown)


schuenke:
bookmark: https://stackoverflow.com/questions/64795367/method-to-determine-lowest-required-versions-of-python-packages-for-a-project-pa

fzimmermann89:
any news?

fzimmermann89:
suggestion:
Add a github action where we manually install the minimum version as currently specified. I don't think an automatic approach will robust.

fzimmermann89:
made obsolete by #709

PR #362, opened 2024-07-10
Update readme

### LINKED ISSUE #329: Update README: add some content
At the moment our README / landing page is pretty empty and we should start thinking about a style / structure that we want to use here. Different big packages like [numpy](https://github.com/numpy/numpy/blob/main/README.md) or [pytorch](https://github.com/pytorch/pytorch/blob/main/README.md) do this completely different. 

Personally, I prefer a more extensive version with more content and examples in the README.

ckolbPTB:
closes #329

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,17 +1,68 @@
 # MRpro
 
-MR image reconstruction and processing package specifically developed for PyTorch.
-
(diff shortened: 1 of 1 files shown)


schuenke (approved):
I don't really like the style of the **Main Features** section yet, but imo it's already a nice improvement and a good starting point for a even nicer README 😄

PR #363, opened 2024-07-10
Remove unused parameters in adam

### LINKED ISSUE #203: differentiable in adam optimizer // optimizer arguments
We have a `differentiable `flag in the adam optimizer. This will not work as expected as we always detach the parameters.

Also, we should reconsider which of the optimizer flags we actually support. I removed the `capturable`, in #193 as this is not useful for us.

JoHa0811:
closes #203

### DIFF
--- a/src/mrpro/algorithms/optimizers/adam.py
+++ b/src/mrpro/algorithms/optimizers/adam.py
@@ -57,7 +57,7 @@ def adam(
         whether to use the AMSGrad variant of this algorithm from the paper
         `On the Convergence of Adam and Beyond`
     decoupled_weight_decay
-        whether to use Adam or AdamW [1]_
+        whether to use Adam (default) or AdamW (if set to true) [1]_
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


koflera (approved):
Looks good to me.

src/mrpro/algorithms/optimizers/adam.py
```diff
@@ -59,39 +56,26 @@ def adam(
...
-        whether `foreach` implementation of optimizer is used
-    maximize
-        maximize the objective with respect to the params, instead of minimizing
-    differentiable
-        whether autograd should occur through the optimizer step. This is currently not implemented.
-    fused
-        whether the fused implementation (CUDA only) is used. Currently, torch.float64, torch.float32,
-        torch.float16, and torch.bfloat16 are supported.
+    decoupled_weight_decay
+        whether to use Adam or AdamW [1]_
```
fzimmermann89: when will itbe used ? :D

JoHa0811: good point

PR #364, opened 2024-07-10
Add __repr__ method to dataclasses and header

### LINKED ISSUE #250: __repr__ for our dataclasses
Currently when we do someting like
```
idata = IData.from....
print(idata)
```
then we get a lot of mainly useless information. It would be nice if this print statement would lead to some helpful information, e.g. 
_IData with shape [1 1 20 30 40]_

This would require to implement `__repr__` method for our dataclasses. See e.g. https://www.pythontutorial.net/python-oop/python-__repr__/

Stef-Martin:
closes #250


src/mrpro/data/IData.py
```diff
@@ -176,3 +176,15 @@ def from_dicom_folder(cls, foldername: str | Path, suffix: str | None = 'dcm') -
             raise ValueError(f'No dicom files with suffix {suffix} found in {foldername}')
 
         return IData.from_dicom_files(filenames=file_paths)
+
+    def __repr__(self):
+        """Representation method for IData class."""
+        fov = self.header.fov if self.header.fov is not None else 'none'
```
ckolbPTB: This part should also be part of IHeader

src/mrpro/data/KHeader.py
```diff
@@ -235,3 +236,29 @@ def mm_to_m(m: float) -> float:
...
         return instance
+
+    def __repr__(self):
+        """Representation method for KHeader class."""
+        fov = self.encoding_fov if self.encoding_fov is not None else 'none'
+        te = str(np.round(self.te.item(), 4)) if self.te is not None else 'none'
+        ti = str(np.round(self.ti.item(), 4)) if self.ti is not None else 'none'
+        fa = str(np.round(self.fa.item(), 4)) if self.fa is not None else 'none'
+        out = (
+            f'KHeader with\n'
```
ckolbPTB: I would remove "Kheader with" and also remove the "," if you are using \n anyway

src/mrpro/data/KHeader.py
```diff
@@ -235,3 +236,29 @@ def mm_to_m(m: float) -> float:
                 'Consider setting them via the defaults dictionary',
             ) from None
         return instance
+
+    def __repr__(self):
+        """Representation method for KHeader class."""
+        fov = self.encoding_fov if self.encoding_fov is not None else 'none'
+        te = str(np.round(self.te.item(), 4)) if self.te is not None else 'none'
```
ckolbPTB: te, ti, fa... can be tensors, so this should be something like [str(te.item()) for te in torch.round(self.te, decimals=4)]

using torch.round would also avoid the numpy import

ckolbPTB (dismissed):

JoHa0811 (approved):

ckolbPTB:
Follow up from the hackathon (as far as I understood it, so please chip in if I missed something):

- Rename PR to something useful to make sure we do not forget that before the squash-merge
- Reuse exisiting __repr__ functionality, e.g. we could do `str(kdata.header.te).replace('tensor', '')` and str(kdata.header.encoding_matrix).replace('SpatialDimension', '') or something similar
- Add units

ckolbPTB:
Could we also add something for KDcf (just the shape) and KTrajectory (also just the shape of z,y,x) and then also print KTrajectory for KData?

fzimmermann89:
maybe consider adding dtype and device. you can call .device on self to obtain it - self.device  will check for all child tensors. if an inconsitantdeviceerror is raised, print "mixed" for device.

src/mrpro/data/_kdata/KData.py
```diff
@@ -234,3 +234,17 @@ def sort_and_reshape_tensor_fields(input_tensor: torch.Tensor):
             ) from None
 
         return cls(kheader, kdata, ktrajectory_final)
+
+    def __repr__(self):
+        """Representation method for KData class."""
+        traj = str(self.traj.as_tensor().shape).replace('torch.Size', '')
+        device_traj = 'cuda' if self.traj.device == torch.device('cuda') else 'cpu'
```
ckolbPTB: Maybe we could replace this by something like
```
try:
    device = self.device
except InconsistentDeviceError:
    device= 'mixed'
```
The main advantage is that it also works if there are other tensors apart from traj and data. We could then use the same for QData+IData (or adding it to Data), KNoise and DcfData.

fzimmermann89: I think this answers your question @Stef-Martin, correct?

fzimmermann89: you can also just do str(device) to the device string:
for CPU it will be CPU
for GPU it will be cuda, followed by the GPU number (which is also useful)

Stef-Martin: yeah, i was just confused by InconsistentDeviceError - but turns out I overthought this a bit and this just gives a RuntimeError, which is what i included as the check now :D

src/mrpro/data/_kdata/KData.py
```diff
@@ -234,3 +234,18 @@ def sort_and_reshape_tensor_fields(input_tensor: torch.Tensor):
             ) from None
 
         return cls(kheader, kdata, ktrajectory_final)
+
+    def __repr__(self):
+        """Representation method for KData class."""
+        traj = str(self.traj.as_tensor().shape).replace('torch.Size', '')
```
fzimmermann89: Either use  broadcasted_shape of traj here to get one shape
Or print the shape for kx, ky, kz.

Also, consider adding a __repr__ to Ktraectory that prints kz, ky, kx shape -- and consider reusing this here.

src/mrpro/data/IHeader.py
```diff
@@ -122,3 +122,12 @@ def deg_to_rad(deg: torch.Tensor | None) -> torch.Tensor | None:
         for name in MISC_TAGS:
             misc[name] = make_unique_tensor(get_float_items_from_all_dicoms(MISC_TAGS[name]))
         return cls(fov=fov, te=te, ti=ti, fa=fa, tr=tr, misc=misc)
+
+    def __repr__(self):
+        """Representation method for IHeader class."""
+        fov = str(self.fov).replace('SpatialDimension', '')
+        te = str(self.te).replace('tensor', '')
```
fzimmermann89: what does this print if these are on GPU or have a different dtype?

fzimmermann89:
I am not to fond of the replace(str(object),...) idiom.

For tensors, consider looking at torch._tensor_str.get_summarized_data
to summarize the data.

Consider adding a __str__ to spatal dimension that returns the values as a string in the format you want to use. 
(__repr__ will still include the SpatialDimension(...)

src/mrpro/data/_kdata/KData.py
Stef-Martin: Yes, i was thinking about that too. I already included a repr method in KTrajectory, but then i would need to use 
`traj = KTrajectory(self.traj.kz, self.traj.ky, self.traj.kx)`
in the repr method of kdata to access it, no? Was not sure about it but i certainly can do it like that if you think it is better :)

src/mrpro/data/IData.py
```diff
@@ -153,3 +153,15 @@ def from_dicom_folder(cls, foldername: str | Path, suffix: str | None = 'dcm') -
...
         return cls.from_dicom_files(filenames=file_paths)
+
+    def __repr__(self):
+        """Representation method for IData class."""
+        try:
+            device = str(self.device)
+        except RuntimeError:
+            device = 'mixed'
+        out = (
+            f'IData with shape: {list(self.data.shape)!s} and dtype {self.data.dtype}\n'
```
ckolbPTB: Can we use something like `type(self)` here instead of `IData` such that it also works for derived classes? Applies to the others too.

fzimmermann89: `type(self)).__name__`

src/mrpro/utils/summarize_tensorvalues.py
```diff
@@ -0,0 +1,21 @@
+"""Summarize the values of a tensor to a string."""
+
+import torch
+
+
+def summarize_tensorvalues(tensor: torch.Tensor | None, summarization_threshold: int = 0) -> str:
```
fzimmermann89: minor nitpick:
maybe set the default to 7 or also decrease the number of  edgeitems?https://pytorch.org/docs/stable/generated/torch.set_printoptions.html

with threshold smaller than 7 and default edgeitems of 3,
torch.arange(7) will be 0,1,2,...4,5,6 which is not shorter than the non abbreviated version

fzimmermann89 (approved):
only change from using Idata etc to using the class name (see Christophs comment

src/mrpro/utils/summarize_tensorvalues.py
Stef-Martin: yeah, you are right. I would include sth like edgeitems is usually ` (summarization_threshold - 1) // 2` but min 1 and max 3, you think that would make sense?

Stef-Martin: `edgeitems = 1 if summarization_threshold < 4 else (2 if summarization_threshold < 7 else 3)`
or just like that maybe

fzimmermann89: Good idea, but I would not write it as a three-way inline if statement but in multiple lines..

PR #366, opened 2024-07-11
Run pytest also for changes to .md

ckolbPTB:
Currently we do not run pytest for the readme but we require that pytest has run for all PRs before they can be merged. Therefore the current PR #362 cannot be merged. 

If we require pytest to run before merge then we need to run it always.


schuenke (approved):
😁

PR #367, opened 2024-07-12
SI unit conversion in dicom and unit info in docstring

ckolbPTB:
As a follow up from the hackathon I verified that dicom requires ms for MR timing info and deg for flip angles. This is now converted to s and rad in MRpro. I added the unit info to the doc strings. Please check that I didn't miss anything.

Open points:

- `acquisition_time_stamp`: in ISMRMRD this is defined in Siemens time steps which is 2.5ms and is of type uint. One option would be to check if the file is from a Siemens scanner and then convert it to [s]. The same applies to `physiology_time_stamp`.

- `sample_time_us`: as the name suggests this is defined in [us]. Do we want to convert this to [s] and change the name of the variable to `sample_time`?

### DIFF
--- a/src/mrpro/data/AcqInfo.py
+++ b/src/mrpro/data/AcqInfo.py
@@ -133,7 +133,7 @@ class AcqInfo(MoveDataMixin):
     """Unique ID corresponding to the readout."""
 
     number_of_samples: torch.Tensor
-    """Number of readout sample points per readout (readouts may have different number of sample points)."""
+    """Number of sample points per readout (readouts may have different number of sample points)."""
 
     patient_table_position: SpatialDimension[torch.Tensor]
     """Offset position of the patient table, in LPS coordinates [m]."""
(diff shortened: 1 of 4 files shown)


src/mrpro/data/AcqInfo.py
```diff
@@ -120,26 +133,25 @@ class AcqInfo(MoveDataMixin):
     """Unique ID corresponding to the readout."""
 
     number_of_samples: torch.Tensor
-    """Number of readout sample points per readout (readouts may have different
-    number of sample points)."""
+    """Number of readout sample points per readout (readouts may have different number of sample points)."""
```
schuenke: Maybe change to "Number of ~~readout~~ sample points per readout (...)"

schuenke (approved):
lgtm 👍

ckolbPTB:
Regarding:
`acquisition_time_stamp`: I added a comment explaining that one Siemens time step is 2.5ms
`physiology_time_stamp`: Similar to above but did not mention 2.5ms because this varies depending on the sensor used (e.g. ECG or respiratory belt)
`sample_time_us`: I kept this in us

PR #375, opened 2024-07-30
Remove copyright statement from files

### LINKED ISSUE #304: discussion: SNR in source files

ckolbPTB:
closes  #304

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2023 Physikalisch-Technische Bundesanstalt
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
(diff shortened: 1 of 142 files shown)


schuenke (approved):
Looks like you found them all 😁

PR #378, opened 2024-08-01
Allow tensor input for transient steady state model

ckolbPTB:
So far most of the parameters to the transient steady state model were only allowed to be scalars but there is not reason why we should not also allow tensors.


src/mrpro/operators/SignalModel.py
```diff
@@ -12,3 +12,26 @@
...
+        The inversion times are commonly the same for each voxel and hence `ti` could be of shape (6,) and the T1 and M0
+        map could be of shape (100,100,100). To make sure `ti` can be broadcasted to the maps it needs to be extended to
+        the shape (6,1,1,1) which then yields a signal of shape (6,100,100,100).
+
+        Parameters
+        ----------
+        parameter
+            Parameter (e.g with shape (m,n))
+        n_dim_to_expand
+            Number of dimensions to expand. If < 0 then parameter is not changed.
```
schuenke: should be `If <= 0, `

src/mrpro/operators/models/TransientSteadyStateWithPreparation.py
```diff
@@ -45,34 +45,44 @@ class TransientSteadyStateWithPreparation(SignalModel[torch.Tensor, torch.Tensor
...
             with shape (time, ...)
         repetition_time
             repetition time
         m0_scaling_preparation
-            scaling of the equilibrium magnetisation due to the preparation pulse before the data acquisition
+            Scaling of the equilibrium magnetisation due to the preparation pulse before the data acquisition.
         delay_after_preparation
-            Time between preparation pulse and start of data acquisition.
-            During this time standard longitudinal relaxation occurs.
+            Time between preparation pulse and start of data acquisition. During this time standard longitudinal
```
schuenke: You might wanna add a comma after "During this time"

src/mrpro/operators/models/TransientSteadyStateWithPreparation.py
```diff
@@ -45,34 +45,44 @@ class TransientSteadyStateWithPreparation(SignalModel[torch.Tensor, torch.Tensor
...
-            time points when model is evaluated. A sampling_time of 0 describes the first acquired data point after the
+            Time points when model is evaluated. A sampling_time of 0 describes the first acquired data point after the
             inversion pulse and spoiler gradients. To take the T1 relaxation during the delay between inversion pulse
             and start of data acquisition into account, set the delay_after_preparation > 0.
             with shape (time, ...)
         repetition_time
             repetition time
         m0_scaling_preparation
-            scaling of the equilibrium magnetisation due to the preparation pulse before the data acquisition
+            Scaling of the equilibrium magnetisation due to the preparation pulse before the data acquisition.
```
schuenke: We use magnetization (AE) in all the example files. Maybe we want to adjust this everywhere in `src/mrpro/operators/models/TransientSteadyStateWithPreparation.py`

schuenke (approved):

PR #379, opened 2024-08-02
Adapt it sense functions

ckolbPTB:
Here I changed the `from_kdata` such that also a `CsmData` can be provided and it is more flexible how the `CsmData` is calculated (assuming we will have espirit, inati... in the future). Especially for IterativeSENSE we probably will calculate the CSM before calling `from_kdata`.

I also separted the `forward` from creating `H` and `b`. Having one single function which returns `(H, b)` together would save a bit of duplicate code. Not sure which version I prefer...


fzimmermann89:
My idea was to keep these somewhat simple and opinionated, as there is always the possibility to just the the reconstruction manually.

Also, the reconstruction can be initiated using the init with any csm.

I would oppose to having a if switch in from_kdata.
Either do dependency inversion and allow to supply the CSM function somewhere. I would like to not have string literals to switch between different functionality.

Or just use the init, and have from_kdata choose some sensible defaults.

This PR moves in the wrong direction imho, as it blurs the lines between the different "levels" of how to do the reconstruction with mrpro.



Gesendet von Outlook für Android<https://aka.ms/AAb9ysg>
________________________________
From: Christoph Kolbitsch ***@***.***>
Sent: Friday, August 2, 2024 9:03:57 PM
To: PTB-MR/mrpro ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [PTB-MR/mrpro] Adapt it sense functions (PR #379)


Here I changed the from_kdata such that also a CsmData can be provided and it is more flexible how the CsmData is calculated (assuming we will have espirit, inati... in the future). Especially for IterativeSENSE we probably will calculate the CSM before calling from_kdata.

I also separted the forward from creating H and b. Having one single function which returns (H, b) together would save a bit of duplicate code. Not sure which version I prefer...

________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/PTB-MR/mrpro/pull/379

Commit Summary

  *   b466da2<https://github.com/PTB-MR/mrpro/pull/379/commits/b466da279cb61bb92ab4df741427041cea5853f0> create H and b in sep functions
  *   1d2a243<https://github.com/PTB-MR/mrpro/pull/379/commits/1d2a243a65e1e748d11d22fea2d7e697a1ec62d6> from_kdata adapted

File Changes

(2 files<https://github.com/PTB-MR/mrpro/pull/379/files>)

  *   M src/mrpro/algorithms/reconstruction/DirectReconstruction.py<https://github.com/PTB-MR/mrpro/pull/379/files#diff-06f95671dcfe9f42a25915c548054cd4597f51a485af25da926dec0c7383fa80> (16)
  *   M src/mrpro/algorithms/reconstruction/IterativeSENSEReconstruction.py<https://github.com/PTB-MR/mrpro/pull/379/files#diff-04a3c2dee6c279749519bf74d9546cd016d1c7e2dec0edae0c94838f5d16519e> (89)

Patch Links:

  *   https://github.com/PTB-MR/mrpro/pull/379.patch
  *   https://github.com/PTB-MR/mrpro/pull/379.diff

ckolbPTB:
> Also, the reconstruction can be initiated using the init with any csm.

True, but I find `from_kdata` a very useful function and I would like to utilise it also as best as possible for the iterative reconstructions. Especially as we will also have to add `CartesianSamplingOp` the `__init__` of each reconstruction if we want to enable Cartesian reconstructions. 

One other option would be to leave `DirectReconstruction.from_kdata()` as it was before and change `IterativeSENSEReconstruction.from_kdata(kdata: KData, csm: CsmData | None, noise: KNoise | None, n_iteartions: int)`, i.e. CSM needs to be provided. IMO this is the most useful case for this iterative recon.

fzimmermann89:
Just to summarize the discussion:
The init should now have an additional argument kdata.
If kdata is specified and none of the other arguments are specified, the former from_kdata logic should apply and everything should be estimated from kdata.
If, for example, a fourieroperator is given, it will be used instead.
If neither kdata nor a fouriroperator is given, an error is raised.

(I think @ckolbPTB already implementedthis in the last commits..)

src/mrpro/algorithms/reconstruction/DirectReconstruction.py
```diff
@@ -18,55 +17,56 @@ class DirectReconstruction(Reconstruction):
...
-        dcf = DcfData.from_traj_voronoi(kdata.traj)
-        fourier_op = FourierOp.from_kdata(kdata)
-        self = cls(fourier_op, None, noise, dcf)
-        if coil_combine:
-            # kdata is prewhitened
-            self.recalculate_csm_walsh(kdata, noise=False)
-        return self
+        self.noise = noise
+
+        if csm is None or type(csm) is CsmData:
```
fzimmermann89: this should be` isinstance(csm, CsmData)`
https://switowski.com/blog/type-vs-isinstance/

fzimmermann89: this might already fix the mypy error

src/mrpro/algorithms/reconstruction/DirectReconstruction.py
```diff
@@ -18,55 +17,56 @@ class DirectReconstruction(Reconstruction):
...
+            self.csm = csm
+        else:
+            if kdata is None:
+                raise ValueError('kdata needs to be defined to calculate the sensitivity maps.')
+            else:
```
fzimmermann89: you could remove the else here

src/mrpro/algorithms/reconstruction/DirectReconstruction.py
```diff
@@ -18,55 +17,54 @@ class DirectReconstruction(Reconstruction):
...
         ----------
+        kdata
+            KData. If kdata is provided and fourier_op or dcf are None, then fourier_op and dcf are estimated based on
+            kdata. Otherwise fourier_op and dcf are used as provided.
         fourier_op
-            Instance of the FourierOperator which adjoint is used for reconstruction.
+            Instance of the FourierOperator used for reconstruction. If None, set up based on kdata.
         csm
-            Sensitivity maps for coil combination. If None, no coil combination will be performed.
+            Sensitivity maps for coil combination. If a callable is provided, then kdata needs also to be provided.
```
fzimmermann89: Explain waht the callable would need to da and what None means

fzimmermann89 (approved):

PR #377, opened 2024-08-01
Golden Radial T1 mapping example

ckolbPTB:
This was the first example I added to test the python-script to notebook pipeline. It never included the T1 mapping part. This is fixed now. 

Things still to do:

- [x] Upload correct data to zenodo
- [x] Enable zenodo download
- [x] Fix float/Tensor problem in TransientSteadyStateWithPreparation #378 
- [x] Fix float/Tensor problem in ConstraintsOp
- [x] Requires #379 and #287
- [x] Verify notebook rendering

PR #380, opened 2024-08-04
Improve readability of dimension operations

### LINKED ISSUE #71: Add expected dimension comments
_Originally posted by @schuenke in https://github.com/ckolbPTB/mrpro/pull/70#pullrequestreview-1650087452_ :
Maybe we can add some more infos about the (expected) dimensions. Not just for the new KTrajectory dataclass, but also for all the other data classes and methods/functions.

 
 So the task might be to add useful dimension comments:
 - to our docstrings
 - to the variables in dataclasses as a comment.
 - In our code whenever we do some reshaping/rearranging that is not completely self-documenting (I.e. an einops call with named axes is fine, a moveaxis or `x[..,None,:,None:]` could get a small comment in the spirit of 
 `'... kx ky ->... coils=1,kx, kz=1,ky'`

ckolbPTB:
closes #71

If anybody finds some more cases where the dimension operations could be improved please adapt or post the code line here in the PR. 

Note: This PR is supposed to improve on the readability of dimension operations. Making them more general will be the aim of #359

### DIFF
--- a/examples/pulseq_2d_radial_golden_angle.ipynb
+++ b/examples/pulseq_2d_radial_golden_angle.ipynb
@@ -26,10 +26,9 @@
     "\n",
     "import matplotlib.pyplot as plt\n",
     "import requests\n",
-    "import torch\n",
-    "from mrpro.data import CsmData, DcfData, IData, KData\n",
-    "from mrpro.data.traj_calculators import KTrajectoryIsmrmrd, KTrajectoryPulseq, KTrajectoryRadial2D\n",
-    "from mrpro.operators import FourierOp, SensitivityOp"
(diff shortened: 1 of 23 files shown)


ckolbPTB:
This https://github.com/PTB-MR/mrpro/blob/296fd305e6ce2a16b2e95b38df18f8e8269b6a23/examples/t1_mapping_with_grad_acq.py#L41

will be fixed in #377

schuenke (approved):
The changes make sense and improve the readability in many cases.
However, I have no overview if this PR catches all cases. I trust you there 😄 

There are a few lines where splitting the logic into two seperate steps would be an option to make the code easier to read, but not sure if it's needed. For me it's fine as it is.

PR #389, opened 2024-09-04
Add addition of Operator with Tensor

### LINKED ISSUE #387: Add OperatorTensorSum
Allow for the addition of a tensor with an operator, e.g. useful for Tikonov-regularised SENSE where we need `A^HA + lambda*UnityMatrix`

lambda will be broadcasted to the output of the operator if possible.

ckolbPTB:
closes #387

### DIFF
--- a/src/mrpro/operators/LinearOperator.py
+++ b/src/mrpro/operators/LinearOperator.py
@@ -152,23 +152,14 @@ class LinearOperator(Operator[torch.Tensor, tuple[torch.Tensor]]):
         else:
             return OperatorComposition(self, other)
 
-    @overload
-    def __radd__(self, other: LinearOperator | torch.Tensor) -> LinearOperator: ...
-
-    @overload
-    def __radd__(
-        self, other: Operator[torch.Tensor, tuple[torch.Tensor,]]
-    ) -> Operator[torch.Tensor, tuple[torch.Tensor,]]: ...
-
-    def __radd__(self, other: torch.Tensor|LinearOperator | Operator[torch.Tensor, tuple[torch.Tensor,]]):
+    def __radd__(self, other: torch.Tensor) -> LinearOperator:  # type: ignore[misc]
         """Operator addition.
 
-        Returns lambda x: self(x) + other(x) if other is a operator,
-        lambda x: self(x) + other if other is a tensor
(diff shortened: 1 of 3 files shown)


fzimmermann89:
Do we want 
Operator + lambda to mean
a) (Operator+ lambda)(x)=Operator(x) + lambda(x) or
b) Operator(x) + lambda

For regularized reconstruction we will need a)

fzimmermann89:
B) would not be a linear operator 

Something I realized after writing the code for it.... ...

fzimmermann89:
And a)ight indeed br better served by a identity operator, see #390, last test.

src/mrpro/operators/LinearOperator.py
```diff
@@ -151,21 +152,32 @@ def __matmul__(
         else:
             return OperatorComposition(self, other)
 
-    @overload
-    def __add__(self, other: LinearOperator) -> LinearOperator: ...
+    def __radd__(self, other: torch.Tensor) -> LinearOperator:  # type: ignore[misc]
+        """Operator addition.
+
+        Returns lambda x: lambda x: self(x) + other
```
ckolbPTB: should this be self(x) + other*x (similar in line 176)?

src/mrpro/operators/LinearOperator.py
```diff
@@ -151,21 +152,32 @@ def __matmul__(
         else:
             return OperatorComposition(self, other)
 
-    @overload
-    def __add__(self, other: LinearOperator) -> LinearOperator: ...
+    def __radd__(self, other: torch.Tensor) -> LinearOperator:  # type: ignore[misc]
+        """Operator addition.
+
+        Returns lambda x: lambda x: self(x) + other*x
```
schuenke: one `lambda x:` to much?

schuenke (approved):
Just did a very very quick review, but lgtm and tests seem to work as expected.

fzimmermann89 (changes requested):
Please hold with merge

fzimmermann89:
This is one way to do it, via a separate (Linear)OperatorTensorSum class.
Alternativly, we could also reuse the IdentityOp:
LinOp + Tensor = LinOpSum(Op, Identity*Tensor), see #430

PR #390, opened 2024-09-05
Add Identity LinearOperator

fzimmermann89:
Add a LinearOperator Identity, i.e. Identity with a single input tensor.

(This might be extended for multiple inputs)

### DIFF
--- a/src/mrpro/operators/IdentityOp.py
+++ b/src/mrpro/operators/IdentityOp.py
@@ -16,7 +16,12 @@ class IdentityOp(LinearOperator):
         super().__init__()
 
     def forward(self, x: torch.Tensor) -> tuple[torch.Tensor]:
-        """Identity of input `x`.
+        """Identity of input.
+
+        Parameters
+        ----------
+        x
+            input tensor
 
         Returns
         -------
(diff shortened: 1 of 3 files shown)


fzimmermann89:
This allows us to write for example
H=A.H@A+lam*Identity()
In CG sense

fzimmermann89:
We could also add a identity instance to the init/make it static.

This would allow us to write just "Identity" or "I" instead of IdentityOp().
This is possible if all identity operator instances do exactly the same thing..

src/mrpro/operators/IdentityOp.py
```diff
@@ -0,0 +1,34 @@
...
+
+    A Linear Operator that returns a single input unchanged.
+    """
+
+    def __init__(self) -> None:
+        """Initialize Identity Operator."""
+        super().__init__()
+
+    def forward(self, x: torch.Tensor) -> tuple[torch.Tensor]:
+        """Identity of input `x`.
```
ckolbPTB: The ` before and after x will lead to x being printed slightly italic in the docu. Is this what you want here or did you want it to be a literal? 

see e.g. https://bashtage.github.io/sphinx-material/rst-cheatsheet/rst-cheatsheet.html

src/mrpro/operators/IdentityOp.py
```diff
@@ -0,0 +1,34 @@
...
+
```
ckolbPTB, line 25: Maybe add 
```
Parameters
----------
```
for sake of completeness

ckolbPTB:
I don't think we should change the EinsumOp in this PR even if it is just doc string formatting.

fzimmermann89:
I am impartial about it and can remove that change.

ckolbPTB (approved):

PR #391, opened 2024-09-09
Fix errors due to pydicom 3.0

ckolbPTB:
Pydicom 3.0 was released which led to several errors due to deprecation, see e.g. https://github.com/pydicom/pydicom/pull/1567/files#diff-9b5a2e48457143cde31b1bbe1c7546a6733e577c306cfe582d56099851e12468

### DIFF
--- a/src/mrpro/data/IHeader.py
+++ b/src/mrpro/data/IHeader.py
@@ -64,10 +64,8 @@ class IHeader(MoveDataMixin):
 
         def get_item(dataset: Dataset, name: TagType):
             """Get item with a given name or Tag from a pydicom dataset."""
-            tag = Tag(name) if isinstance(name, str) else name  # find item via value name
-
             # iterall is recursive, so it will find all items with the given name
-            found_item = [item.value for item in dataset.iterall() if item.tag == tag]
+            found_item = [item.value for item in dataset.iterall() if item.tag == Tag(name)]
 
             if len(found_item) == 0:
                 return None
--- a/src/mrpro/data/IHeader.py
+++ b/src/mrpro/data/IHeader.py
@@ -62,7 +62,7 @@ class IHeader(MoveDataMixin):
             list of dataset objects containing the DICOM file.
         """
 
-        def get_item(dataset: Dataset, name: str | TagType):
+        def get_item(dataset: Dataset, name: TagType):
             """Get item with a given name or Tag from a pydicom dataset."""
             tag = Tag(name) if isinstance(name, str) else name  # find item via value name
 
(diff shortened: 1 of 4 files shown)


ckolbPTB:
Currently all warnings are raised as errors in pytest (see https://github.com/PTB-MR/mrpro/blob/ca2f07b7030fe8ce3674f23207f974efcbc09fd3/pyproject.toml#L66)

This means that e.g DeprecationWarnings in other packages lead to errors. @fzimmermann89 @schuenke : Is this the behaviour we want, or should it only lead to errors for warning raised directly by our mrpro code?

fzimmermann89:
DeprecationWarnings as errors makes sense IMHO.

It would be better if we could somehow set a depth for this, i.e. only if we call deprecated code it should error, not if one of our dependencies calls a deprecated function. But I think that might not be possible to detect..

fzimmermann89:
(If  third party code throws a DeprecationWarning that we cannot fix, we could add an ignore to the toml filter warning setting for this deprecating warning)

fzimmermann89:
@CodiumAI-Agent /review

src/mrpro/data/IHeader.py
```diff
@@ -62,7 +62,7 @@ def from_dicom_list(cls, dicom_datasets: Sequence[Dataset]) -> Self:
             list of dataset objects containing the DICOM file.
         """
 
-        def get_item(dataset: Dataset, name: str | TagType):
+        def get_item(dataset: Dataset, name: TagType):
```
fzimmermann89, line 65: why did you remove the str here?

src/mrpro/data/IHeader.py
```diff
@@ -62,7 +62,7 @@ def from_dicom_list(cls, dicom_datasets: Sequence[Dataset]) -> Self:
...
             """Get item with a given name or Tag from a pydicom dataset."""
             tag = Tag(name) if isinstance(name, str) else name  # find item via value name
```
fzimmermann89: but kept it here

fzimmermann89 (commented):
why did you remove the str type hints?

src/mrpro/data/IHeader.py
ckolbPTB: TagType is a union of different types already including str so it is not needed here

ckolbPTB: Well to be more accurate it used to be a union so str | TagType was redundant but ok and now it is a TagAlias: https://github.com/pydicom/pydicom/blob/b2073cccbd92d537dfb307c4d351a267e602069e/src/pydicom/tag.py#L37

src/mrpro/data/IHeader.py
ckolbPTB: Probably this can be removed, yes

fzimmermann89:
Got it, TagType includes str

...and you already answered the question :D sorry

fzimmermann89 (approved):

fzimmermann89 (approved):

PR #396, opened 2024-09-16
Add Functional base class

fzimmermann89:
Add the base class for (proximable) functionals and the tests for
functional without adding any functionals yet.

Continued from #295


tests/operators/functionals/test_functionals.py
```diff
@@ -0,0 +1,404 @@
...
+    f.cuda()
+    (fx,) = f(x)
+    assert fx.is_cuda
+
+
+@pytest.mark.cuda()
+@pytest.mark.parametrize('parameters', ['scalar', 'none', 'tensor'])
+@pytest.mark.parametrize('functional', PROXIMABLE_FUNCTIONALS)
+def test_functional_cuda_prox(functional: type[ProximableFunctional], parameters: Literal['scalar', 'none', 'tensor']):
+    """Test if the forward pass works on the GPU."""
```
fzimmermann89: should be prox and prox_convex_conj

fzimmermann89:
I tried to clean up the tests and join them for all functionals.

Open to suggestions which tests are missing or which of the previous tests in the history of #295 should be added back

fzimmermann89:
I added both those you worked on #295 as reviewers to ensure I did not miss any test case, and some ppl who have not seen the code..

src/mrpro/operators/Functional.py
```diff
@@ -0,0 +1,82 @@
...
+        for some functional :math:`phi`.
+
+        Parameters
+        ----------
+        weight
+            weighting of the norm (see above)
+        target
+            element to which distance is taken - often data tensor
+        dim
+            dimension over which norm is calculated
```
ckolbPTB: dimension(s)

src/mrpro/operators/Functional.py
```diff
@@ -0,0 +1,82 @@
...
+        target: torch.Tensor | None | complex = None,
+        dim: int | Sequence[int] | None = None,
+        divide_by_n: bool = False,
+        keepdim: bool = False,
+    ) -> None:
+        r"""Initialize a Functional.
+
+        We assume that functionals are given in the form
+            :math:`f(x) = \phi( weight ( x - target))`
+        for some functional :math:`phi`.
```
ckolbPTB: \phi instead of phi

src/mrpro/operators/Functional.py
```diff
@@ -0,0 +1,82 @@
...
+        Parameters
+        ----------
+        weight
+            weighting of the norm (see above)
+        target
+            element to which distance is taken - often data tensor
+        dim
+            dimension over which norm is calculated
+        divide_by_n
+            True: norm calculated with mean
```
ckolbPTB: ...mean. (fullstop after mean) because in the docu a line break is ignored

src/mrpro/operators/Functional.py
```diff
@@ -0,0 +1,82 @@
...
+        if isinstance(dim, int):
+            dim = (dim,)
+        elif isinstance(dim, Sequence):
+            dim = tuple(dim)
+        self.dim = dim
+        self.divide_by_n = divide_by_n
+        self.keepdim = keepdim
+
+    def _divide_by_n(self, x: torch.Tensor, shape: None | Sequence[int]) -> torch.Tensor:
+        """Compute factor for normalization."""
```
ckolbPTB: Parameters doc string missing

src/mrpro/operators/Functional.py
```diff
@@ -0,0 +1,82 @@
...
+
+class ProximableFunctional(Functional, ABC):
+    """ProximableFunction Base Class."""
+
+    @abstractmethod
+    def prox(self, x: torch.Tensor, sigma: torch.Tensor | float = 1.0) -> tuple[torch.Tensor]:
+        """Apply proximal operator."""
+
+    def prox_convex_conj(self, x: torch.Tensor, sigma: torch.Tensor | float = 1.0) -> tuple[torch.Tensor]:
+        """Apply proximal of convex conjugate of functional."""
```
ckolbPTB: Parameters doc string missing

src/mrpro/operators/Functional.py
```diff
@@ -0,0 +1,82 @@
...
+        if not isinstance(sigma, torch.Tensor):
+            sigma = torch.as_tensor(1.0 * sigma, device=self.target.device)
```
ckolbPTB, line 160: Why 1.0 * sigma? Would it make sense to use x.device here?

src/mrpro/operators/Functional.py
```diff
@@ -0,0 +1,82 @@
...
+        sigma[sigma.abs() < 1e-8] += 1e-6
+        return (x - sigma * self.prox(x * 1 / sigma, 1 / sigma)[0],)
```
ckolbPTB: maybe just self.porx(x / sigma, 1 / sigma)? would we get a speed up if we calculated 1 / sigma only once?

tests/_RandomGenerator.py
```diff
@@ -199,3 +199,19 @@ def int64_tuple(self, size: int, low: int = -1 << 63, high: int = 1 << 63) -> tu
 
     def ascii(self, size: int) -> str:
         return ''.join([chr(self.uint8(32, 127)) for _ in range(size)])
+
+    def rand_like(self, x: torch.Tensor) -> torch.Tensor:
+        """Generates a tensor with the same shape as x filled with uniform random numbers."""
```
ckolbPTB: uniform random would be symmetric around 0. Maybe mention "in the intervall [0, 1)" or something like this

tests/operators/functionals/test_functionals.py
```diff
@@ -0,0 +1,404 @@
...
+
+import pytest
+import torch
+from mrpro.operators import Functional, ProximableFunctional, functionals
+
+from tests import RandomGenerator
+
+
+@dataclass
+class FunctionalTestCase:
```
ckolbPTB: Please add doc-strings here and also to functional_test_cases, test_functional_shape, test_functional_devided_by_n

src/mrpro/operators/Functional.py
fzimmermann89: re device: self.target.devices mathes the devices of the pytorch module, as target is a buffer. 
re 1.0 *: common way to convert int to float.

if sigma is already a tensor, we want to keep if it is 32bit or 64bit.
if sigma is a float, we should use pytorch default bit width
if sigma is an int, we should use pytorch default bit width (and not int)

all other ways are longer and more complicated and need multiple ifs...
```suggestion
        if shape is None:

File: src/mrpro/operators/Functional.py [hunk 11]
```diff
@@ -0,0 +1,177 @@
...
+            isinstance(x, torch.Tensor) and not x.dtype.is_complex and (x >= 0).all()
+        ):
+            return
+        raise ValueError(message)
+
+
+class ElementaryFunctional(Functional):
+    r"""Elementary functional base class.
+
+    An elementary functional is a functional that can be written as
```
koflera: I would specifically mention, that WE assume an elementary functional to be of this form. otherwise it sounds like this is well-known concept and might be a bit confusing.

src/mrpro/operators/Functional.py
```diff
@@ -0,0 +1,177 @@
...
+        self,
+        weight: torch.Tensor | complex = 1.0,
+        target: torch.Tensor | None | complex = None,
+        dim: int | Sequence[int] | None = None,
+        divide_by_n: bool = False,
+        keepdim: bool = False,
+    ) -> None:
+        r"""Initialize a Functional.
+
+        We assume that functionals are given in the form
```
koflera, line 52: ok, since it's already mentioned here below, maybe it might be good to instead of saying above that it has this form, just to mention that elementary functionals do not require another functionals for initialization.

src/mrpro/operators/Functional.py
```diff
@@ -0,0 +1,177 @@
...
+        else:
+            size = list(shape)
+        return x / math.prod(size)
+
+
+class ProximableFunctional(Functional, ABC):
+    r"""ProximableFunctional Base Class.
+
+    A proximable functional is a functional :math:`f(x)` that has a prox implementation,
+    i.e. a function that solves the problem :math:`\min_x f(x) + 1/(2\sigma) ||x - y||^2`.
```
koflera: actually, we use the definition with 
:math:`\min_x sigma * f(x) + 1/(2) ||x - y||^2`
right?
And also, I am not sure about the factor 1/2 since, as far as I remember, you preferred to remove it? If yes, are we at least consistent now, i.e. for the L2SquaredNorm do we consider
:math:`\min_x  1/2 ||x-y||^2 + sigma */2 ||x||^2`   (1)
:math:`\min_x  ||x-y||^2 + sigma */2 ||x||^2`         (2)
:math:`\min_x  1/2 ||x-y||^2 + sigma * ||x||^2`      (3)
:math:`\min_x  ||x-y||^2 + sigma * ||x||^2`            (4)
I think we should have either (1) (my preferred definition) or (4), but for sure not (2) or (3).
Please check.

src/mrpro/operators/Functional.py
```diff
@@ -0,0 +1,177 @@
...
+    """
+
+    @abstractmethod
+    def prox(self, x: torch.Tensor, sigma: torch.Tensor | float = 1.0) -> tuple[torch.Tensor]:
+        r"""Apply proximal operator.
+
+        Applies :math:`prox_{\sigma f}(x) = argmin_{p} (f(p) + 1/(2*sigma) \|x-p\|^{2}`.
```
koflera: same as above

src/mrpro/operators/Functional.py
```diff
@@ -0,0 +1,177 @@
...
+
+        Returns
+        -------
+            Proximal operator applied to the input tensor.
+        """
+
+    def prox_convex_conj(self, x: torch.Tensor, sigma: torch.Tensor | float = 1.0) -> tuple[torch.Tensor]:
+        r"""Apply proximal of convex conjugate of functional.
+
+        Applies :math:`prox_{\sigma f*}(x) = argmin_{p} (f(p) + 1/(2*sigma) \|x-p\|^{2}`,
```
koflera: same as above

tests/operators/functionals/conftest.py
```diff
@@ -0,0 +1,89 @@
+import inspect
```
koflera, line 1: what does the name "conftest.py" stand for? maybe consider renaming it to something more obvious?

src/mrpro/operators/Functional.py
koflera: maybe, also 
"i.e. a function that yields/implements :math:`\argmin_x ...`.
instead of
"i.e. a function that solves the problem :math:`\min_x ...`.

koflera (changes requested):
some minor comments

src/mrpro/operators/Functional.py
fzimmermann89: checked.
acutually, the definition is correct and the same as odl uses.
the 1/2 here is different from the 1/2 in L2NormSquared.
For L2Norm, it is also `min_x ||x-y||^2 + 1/(2*sigma) * ||x-y||^2`
The L2NormSquared used to be defined as 1/2*||x||_^2, it is now ||x||_^2. Matching forward and prox is also tested..

tests/operators/functionals/conftest.py
fzimmermann89: conftest is a pytest name, it should be named like this.

src/mrpro/operators/Functional.py
fzimmermann89: (not that the definition really matters, as `argmin_x sigma * f(x) + 1/(2) ||x - y||^2 ==  argmin_x f(x) + 1/(2s) ||x - y||^2`

koflera: yes, sure... I just find the one where you scale the function of interest a bit more intuitive and, as far as I remember, we had started to work with that one

src/mrpro/operators/Functional.py
```diff
@@ -0,0 +1,177 @@
...
+
+        Input is scaled by the number of elements of either the input
+        or optional shape.
+
+        Parameters
+        ----------
+        x
+            input to be scaled.
+        shape
+            input will be divided by the product these numbers.
```
koflera: by the product OF these numbers

src/mrpro/operators/Functional.py
```diff
@@ -0,0 +1,177 @@
...
+            If None, it divides by the number of elements of the input.
```
koflera: "If None, it is divided .. " or "If None, the input is divided ... "

koflera (commented):
just a few small comments... otherwise looks good to me

src/mrpro/operators/Functional.py
```diff
@@ -0,0 +1,177 @@
...
+        r"""Initialize a Functional.
+
+        We assume that functionals are given in the form
+            :math:`f(x) = \phi( weight ( x - target))`
+        for some functional :math:`\phi`.
+
+        Parameters
+        ----------
+        weight
+            weighting of the norm (see above)
```
koflera: "weighting of the norm" seems to be related to specific choices of functionals... maybe better: just "weight parameter (see above"?

src/mrpro/operators/Functional.py
```diff
@@ -0,0 +1,177 @@
...
+        target
+            element to which distance is taken - often data tensor (see above)
```
koflera: same as above... the term "distance" suggests a particular type of functional

src/mrpro/operators/Functional.py
```diff
@@ -0,0 +1,177 @@
...
+        dim
+            dimension(s) over which norm is calculated.
+            All other dimensions of  `weight ( x - target))` will be treated as batch dimensions.
+        divide_by_n
+            if True, the result is scaled by the number of elements of the dimensions index by `dim` in
+            the tensor `weight ( x - target))`. If true, the norm is thus calculated as the mean,
+            else the sum.
```
koflera, line 68: also here the word "norm"

src/mrpro/operators/Functional.py
```diff
@@ -0,0 +1,177 @@
...
+        keepdim
+            if true, the dimension(s) of the input indexed by dim are mainted and collapsed to singeltons,
```
koflera: typo: mainted --> maintained

src/mrpro/operators/Functional.py
```diff
@@ -0,0 +1,177 @@
...
+    @abstractmethod
+    def prox(self, x: torch.Tensor, sigma: torch.Tensor | float = 1.0) -> tuple[torch.Tensor]:
+        r"""Apply proximal operator.
+
+        Yields :math:`prox_{\sigma f}(x) = argmin_{p} (sigma f(p) + 1/2 \|x-p\|^{2}` given `x` and `sigma`
+
+        Parameters
+        ----------
+        x
+            input tensor
```
koflera, line 133: not sure whether this has some impact on the documentation or so... sometimes there is a "." after the description of the input/return parameters, sometimes not.

koflera (approved):
only a few small comments

src/mrpro/operators/Functional.py
```diff
@@ -0,0 +1,177 @@
...
+
+import torch
+
+from mrpro.operators.Operator import Operator
+
+
+class Functional(Operator[torch.Tensor, tuple[torch.Tensor]]):
+    """Functional Base Class."""
+
+    def _throw_if_negative_or_complex(
```
ckolbPTB, line 15: I would call it _exception_if_...

src/mrpro/operators/Functional.py
```diff
@@ -0,0 +1,177 @@
...
+        self, x: torch.Tensor | float, message: str = 'sigma must be real and contain only positive values'
+    ):
+        """Throw an exception if any element of x is negative or complex.
+
+        Raises a ValueError if x contains negative or complex values.
```
ckolbPTB: I would combine this line with the first line as it does not add any additional information except for the ValueError

src/mrpro/operators/Functional.py
```diff
@@ -0,0 +1,177 @@
...
+            dim = tuple(dim)
+        self.dim = dim
+        self.divide_by_n = divide_by_n
+        self.keepdim = keepdim
+
+    def _divide_by_n(self, x: torch.Tensor, shape: None | Sequence[int]) -> torch.Tensor:
+        """Apply factor for normalization.
+
+        Input is scaled by the number of elements of either the input
+        or optional shape.
```
ckolbPTB: or product of the optional shape entries

ckolbPTB (approved):

src/mrpro/operators/Functional.py
```diff
@@ -0,0 +1,177 @@
...
+    A proximable functional is a functional :math:`f(x)` that has a prox implementation,
+    i.e. a function that yields :math:`argmin_x sigma f(x) + 1/2 ||x - y||^2`
+    and a prox_convex_conjugate, yielding the prox of the convex conjugate.
+    """
+
+    @abstractmethod
+    def prox(self, x: torch.Tensor, sigma: torch.Tensor | float = 1.0) -> tuple[torch.Tensor]:
+        r"""Apply proximal operator.
+
+        Yields :math:`prox_{\sigma f}(x) = argmin_{p} (sigma f(p) + 1/2 \|x-p\|^{2}` given `x` and `sigma`
```
ckolbPTB: This does not render properly in the docu

src/mrpro/operators/Functional.py
```diff
@@ -0,0 +1,177 @@
...
+
+class ElementaryProximableFunctional(ElementaryFunctional, ProximableFunctional):
+    r"""Elementary proximable functional base class.
+
+    Here, an *elementary* functional is a functional that can be written as
+    :math:`f(x) = \phi( weight ( x - target))`, returning a real value.
+    It does not require another functional for initialization.
+
+    A proximable functional is a functional :math:`f(x)` that has a prox implementation,
+    i.e. a function that yields :math:`argmin_x sigma f(x) + 1/2 ||x - y||^2`.
```
ckolbPTB: does not render properly probably \sigma

fzimmermann89:
pre-commit.ci autofix

PR #398, opened 2024-09-16
Add L2NormSquared Functional

fzimmermann89:
The L2NormSquared Functional from #295


koflera (approved):
Looks good to me.

Maybe one thing that came to my mind now and holds also for the L1Norm and the L1NormViewAsReal as well.

Strictly speaking, if any of the elements of the weight tensor is exactly zero, the function does not implement a norm because the weight tensors are not positive definite. Maybe we should either include a short explanation in the docstring or even raise a warning?

PR #399, opened 2024-09-16
Add L1NormViewAsReal Functional

fzimmermann89:
The L1NormViewAsReal Functional from #295


fzimmermann89:
The meaning of a complex weight is slightly different here (mentioned in the docstring)
A complex weight will also be viewed-as-R2 and element-wise multiplication is performed.
This felt more natural in the context of view-as-R2 than complex multiplication and greatly simplified the code (To be honest, I am not 100% sure what the prox would be for a complex weight otherwise)

src/mrpro/operators/functionals/L1NormViewAsReal.py
```diff
@@ -0,0 +1,84 @@
...
+import torch
+
+from mrpro.operators.Functional import ProximableFunctional
+
+
+class L1NormViewAsReal(ProximableFunctional):
+    r"""Functional class for the L1 Norm, where C is identified with R^2.
+
+    This implements the functional given by
+        f: C^N --> [0, \infty), x ->  \|Wr.*Re(x-b) )\|_1 + \|( Wi.*Im(x-b) )\|_1
```
koflera: I would write
`f: C^N --> [0, \infty), x ->  \|W.*(x-b)\|_1  := \|Wr.*Re(x-b) )\|_1 + \|( Wi.*Im(x-b) )\|_1`

to be able to refer to the weight W as done further below in the docstrings.

src/mrpro/operators/functionals/L1NormViewAsReal.py
```diff
@@ -0,0 +1,84 @@
...
+
+    If the weight parameter W is real-valued, Wr and Wi are both set to W.
+    If it is complex-valued, Wr and Wi are set to the real and imaginary part of W, respectively.
+
+    The norm of the vector is computed along the dimensions given by "dim".
+    Further, it is possible to scale the functional by N, i.e. by the number voxels of
+    the elements of the vector space that is spanned by the dimensions indexed by "dim".
+    If "dim" is set to None and "keepdim" to False, the result is a single number, which
+    is typically of interest for computing loss functions.
+
```
koflera, line 17: Also here, I would explicitly mention that, similar to L1, the weights (i.e. either W already, or, if W is complex-valued, Re(W) and Im(W), have to be strictly positive, i.e. 

"Note that here, the proximal mapping and the convex conjugate of the proximal mapping are
derived assuming the weight vector defines an invertible (block-) diagonal
operator with all entries of the real and imaginary part being strictly positive."

src/mrpro/operators/functionals/L1NormViewAsReal.py
```diff
@@ -0,0 +1,84 @@
+"""L1 Norm."""
+
+import torch
+
+from mrpro.operators.Functional import ProximableFunctional
+
+
+class L1NormViewAsReal(ProximableFunctional):
```
koflera: maybe I missed something, but why is there no `prox_conv_conj ` anymore? :D

src/mrpro/operators/functionals/L1NormViewAsReal.py
```diff
@@ -0,0 +1,84 @@
...
+
+        if self.divide_by_n:
+            return (torch.mean(value, dim=self.dim, keepdim=self.keepdim),)
+        else:
+            return (torch.sum(value, dim=self.dim, keepdim=self.keepdim),)
+
+    def prox(self, x: torch.Tensor, sigma: torch.Tensor | float = 1.0) -> tuple[torch.Tensor]:
+        """Proximal Mapping of the L1 Norm.
+
+        Compute the proximal mapping of the L1-norm with C identified as R^2.
```
koflera: I think I would write
"Apply the proximal mapping of the L1-norm with C identified as R^2"
instead of 
"Compute the proximal mapping of the L1-norm with C identified as R^2."
since "computing" the proximal mapping sounds more like the process of deriving it rather than applying it to an input tensor. Similar for the prox conv conj.

src/mrpro/operators/functionals/L1NormViewAsReal.py
koflera: maybe we could also raise a warning/error if the provided weights (or the real and the imaginary parts of the weights, for the complex case) do not fulfill the critertion of being strictly positive?

src/mrpro/operators/functionals/L1NormViewAsReal.py
```diff
@@ -0,0 +1,84 @@
...
+
+        Parameters
+        ----------
+        x
+            input tensor
+        sigma
+            real valued scaling factor
```
koflera: sigma should probably also be >= 0, otherwise the considered functional ||x-z||_2^2 + sigma * f(z) is not convex if f is convex (which should be by assumption), thus the proximal operator not well-defined.

maybe also worth raising an error?

tests/operators/functionals/test_functionals.py
```diff
@@ -276,6 +277,28 @@ class NumericCase(TypedDict):
```
koflera, line 240: Further tests that could be included are maybe the properties of the 
- precomposition
- postcomposition
- affine addition

(see p.130 in https://web.stanford.edu/~boyd/papers/pdf/prox_algs.pdf)

as well as the contraction property, i.e.
that 
||prox(x) - prox(y)||_2 \leq ||x-y||_2 
for all x,y.

koflera (changes requested):
some comments

src/mrpro/operators/functionals/L1NormViewAsReal.py
fzimmermann89: Re(W) and Im(W) do not have to be positive.

src/mrpro/operators/functionals/L1NormViewAsReal.py
fzimmermann89: moreau identity. uses the prox. a explicit implementation was not much faster and replicated the is-complex logic from the prox, so we just use the base classes implemntation

tests/operators/functionals/test_functionals.py
fzimmermann89: the precomposition is tested, via the scaling test with random  targets tests.
postcomposition is also tested: the scaling property via the sigma check, the shift is not something we can simply tests: adding a shift to the forward is not something we support directly. and all tests I can think of would then mainly tests this new shift impelemation, not a property of the underlying functional.

I will add a test for random ||prox(x) - prox(y)||_2 \leq ||x-y||_2, that's a great idea!

src/mrpro/operators/functionals/L1NormViewAsReal.py
fzimmermann89: I would rather write in the next following paragraph somewhere W:=Wr+iWi

fzimmermann89: I will update it and you can check again

src/mrpro/operators/functionals/L1NormViewAsReal.py
koflera: ah yeah, I had missed that indeed.

one general thought, though: isn't this something that in general we should avoid to do? What I mean is that the Moreau identity is some sort of result, similar to 1=sin^2(x) + cos^2(x).

So if for a functional one derives a prox operator that is wrong but for some reason still passes all the tests (could this happen?), one would define the prox_conv_conj by using the wrong prox. However, the Moreau-identity would be fulfilled by construction. Like defining the cosine by cos^2(x) = 1- sin^2(x) instead of cos(x) = adjacent/hypotenuse and having 1=sin^2(x) + cos^2(x) as a result.

fzimmermann89: but most of the codes to actually calculate the sin/cos use these properties to calculate it :)

it is equally likely to have a wrong prox_convex_conj that passes moreau identity as having a wrong prox. imho it is not a bug but a feature that prox_convex_conj will automatically be correct if prox is correct.

it only makes sense to manually implement it if there are some simplifications that can be done in the code. 
but here, the dtype checks would be most of the code of the prox_convex_conj.

src/mrpro/operators/functionals/L1NormViewAsReal.py
koflera: maybe I am mistaken, but if I remember correctly, at some point, @guastinimara and I were convinced that for the derivation of the prox for the L1-norm, it made sense to require the weights to be positive. Thus, I thought it would make sense here too. @guastinimara do you happen to remember the reason why? (maybe you are closer than I am to the notes we had written down :D )

koflera: btw, if negative values are ok as well, I think it also should be changed in the docstrings of the L1-norm, where we currently have

"Note that here, the proximal mapping and the convex conjugate of the proximal mapping are
derived assuming the weight vector defines an invertible (block-) diagonal
operator with all entries being strictly greater than zero, i.e. only real-valued
weight-tensors are allowed."

src/mrpro/operators/functionals/L1NormViewAsReal.py
koflera: ok, but sin/cos is much more trivial than the case here. I would have thought that it is very unlikely to be able to implement a (probably wrong) prox_conv_conj that compensates for the wrongly implemented prox such that the Moreau identity passes. Thus, I thought that, whenever possible, it could be better to have both implementations and maybe for the case where one of prox or prox_conv_conv is much more computationally intensive than the other, to rely on the Moreau-identity to compute the desired operator.

And sure, I also agree the Moreau identity is a "feature", I was just wondering if coding-wise it might potentially lead to wrong implementations.

src/mrpro/operators/functionals/L1NormViewAsReal.py
fzimmermann89: I convinced mara that the prox is valid with negative and zero weights as well.
see the precomposition property of the prox.
we also test for it to work.

fzimmermann89: or, in the case of the l1 norm derivation: |a*b| =  |a||b| for a,b, € C is all you need

koflera (approved):
Looks good to me

PR #404, opened 2024-09-17
Allow adjoint to be used as autograd backward

fzimmermann89:
Alternative to #307 using init_subclass as I suggested in #68

### DIFF
--- a/tests/operators/test_autograd_linop.py
+++ b/tests/operators/test_autograd_linop.py
@@ -1,0 +1,91 @@
+"""Test for the adjoint as backward option in the LinearOperator class."""
+
+import pytest
+import torch
+from mrpro.operators import LinearOperator
+from torch.autograd.gradcheck import GradcheckError
+
+from tests import RandomGenerator
+from tests.helper import dotproduct_adjointness_test
+
+
+class NonDiffentiableOperator(LinearOperator, adjoint_as_backward=False):
+    """Dummy operator for testing purposes."""
+
+    def forward(self, x):
+        python_float = x.item()
+        tensor = torch.tensor([2 * python_float], dtype=x.dtype)  # breaks the autograd
+        return (tensor,)
+
+    def adjoint(self, x):
+        python_float = x.item()
+        tensor = torch.tensor([2 * python_float], dtype=x.dtype)  # breaks the autograd
+        return (tensor,)
+
+
+class DiffentiableOperator(LinearOperator, adjoint_as_backward=True):
+    """Dummy operator for testing purposes. Should be differentiable due to adjoint_as_backward=True."""
+
+    def forward(self, x):
+        python_float = x.item()
+        tensor = torch.tensor([2 * python_float], dtype=x.dtype)  # breaks the autograd
+        return (tensor,)
+
+    def adjoint(self, x):
+        python_float = x.item()
+        tensor = torch.tensor([2 * python_float], dtype=x.dtype)  # breaks the autograd
+        return (tensor,)
+
+
+def test_linop_autograd_dummy_operator():
+    """Ensure correct dummy operator."""
+    rng = RandomGenerator(seed=0)
+    nondiff = NonDiffentiableOperator()
+    diff = DiffentiableOperator()
+
+    u = rng.float32_tensor((1,))
+    v = rng.float32_tensor((1,))
+
+    # Differentiable operator and non differentiable operator should be the same
+    torch.testing.assert_close(nondiff.adjoint(u), nondiff.adjoint(u))
+    torch.testing.assert_close(diff.adjoint(u), diff.adjoint(u))
+
+    # and the adjoin should be correct
+    dotproduct_adjointness_test(nondiff, u, v)
+
+
+def test_linop_autograd_nondifferentiable():
+    """Test the forward of the non differentiable operator."""
+    nondiff = NonDiffentiableOperator()
+    # The non differentiable operator should raise an error,
+    # because the forward is non differentiable (python float)
+    x = torch.tensor(1.0, requires_grad=True, dtype=torch.float64)
+    with pytest.raises(GradcheckError):
+        torch.autograd.gradcheck(nondiff, x)
+
+
+def test_linop_autograd_differentiable_forward():
+    """Test the backward of the forward of the differentiable operator."""
+    diff = DiffentiableOperator()
+    # The differentiable operator should work, as
+    # the adjoint can be used as backward.
+    x = torch.tensor(1.0, requires_grad=True, dtype=torch.float64)
+    torch.autograd.gradcheck(diff, x)
+
+
+def test_linop_autograd_differentiable_adjoint():
+    """Test the backward of the adjoint of the differentiable operator."""
+    diff = DiffentiableOperator()
+    # Should work, as the forward will be used as the backward
+    # of the adjoint.
+    x = torch.tensor(1.0, requires_grad=True, dtype=torch.float64)
+    torch.autograd.gradcheck(diff.H, x)
+
+
+def test_linop_autograd_differentiable_gradgrad():
+    """Test the gradgrad of the differentiable operator."""
+    diff = DiffentiableOperator()
+    # The backward of the backward should be the working
+    # as it should be using the forward
+    x = torch.tensor(1.0, requires_grad=True, dtype=torch.float64)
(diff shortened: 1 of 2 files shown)


fzimmermann89:
@ckolbPTB

fzimmermann89:
I removed all typing information from init_subclass. Myp will most likely never be able to check inside the autograd wrapper anyways and if I am not mistaken,currently mypy will works as-if there was no adjoint wrapper. 
The forward, the adjoint and all calls of these will still be type checked.

fzimmermann89:
There is a fundamental issue with using the adjoint as the backward:

During the forward inside the wrapper, autograd is disabled. Thus gradients from any parameters of self do not flow to the output.
This means, looking at for example the gridsampleop, the gradient wrt x of the operator will work, the gradient wrt the grid will not work.

this is difficult to solve without using a custom solution for each linearop.

We might be able to use it for operators that do not depend on any other tensor. But most of our operators will then not be covered by the solution: 
 - Fourier: Should be differentiable wrt trajectory
 - CSM wrt csms
 - gridsample wrt grid
- we need to disable it for all linear operator-magic operators, for the composition of linops to be differentiable to some attributes of one of the operators...

no idea how to pursue.

maybe we just do a custom autograd for the fourier operator?

@schuenke @ckolbPTB

ckolbPTB:
> There is a fundamental issue with using the adjoint as the backward:

That is a shame!

> maybe we just do a custom autograd for the fourier operator?

Can we provide an custom gradient wrt to x but use pytorch autograd wrt to e.g. traj or would we have to provide our own autograd functionality for each parameter?

fzimmermann89:
Torchkbnuftt also does not work for traj requiring gradients afaik.

We could
-  fix torchkbnuftt (a started doing that, but there are many issues and it would require a substantial change - which would also speed up the nufft by a factor of 2..)
- use finnuft (the torch bindings are not in pypi)
- implementing a custom autograd function that works for traj and x,  maybe supports different nufft backends and uses mri conventions. 

As I started doing that work in the torchkbnuftt repo already and we can look up the equations in https://arxiv.org/abs/2111.02912 and the finnuft pytorch wrapper.

I will try to mock something up tonight or tomorrow morning

fzimmermann89:
Now this can be enabled for operators, using the adjoint_as_backward setting, i.e.
`class Op(LinearOperator, adjoint_as_backward=True)`. The default is false, it has to be enabled for each Operator that should use this.

ckolbPTB (commented):
Would be good to have a test with a simple operator here.

tests/operators/test_autograd_linop.py
```diff
@@ -0,0 +1,91 @@
...
+import pytest
+import torch
+from mrpro.operators import LinearOperator
+from torch.autograd.gradcheck import GradcheckError
+
+from tests import RandomGenerator
+from tests.helper import dotproduct_adjointness_test
+
+
+class NonDiffentiableOperator(LinearOperator, adjoint_as_backward=False):
```
ckolbPTB: Differentiable

tests/operators/test_autograd_linop.py
```diff
@@ -0,0 +1,91 @@
...
+    """Ensure correct dummy operator."""
+    rng = RandomGenerator(seed=0)
+    nondiff = NonDiffentiableOperator()
+    diff = DiffentiableOperator()
+
+    u = rng.float32_tensor((1,))
+    v = rng.float32_tensor((1,))
+
+    # Differentiable operator and non differentiable operator should be the same
+    torch.testing.assert_close(nondiff.adjoint(u), nondiff.adjoint(u))
```
ckolbPTB: compare nondiff.adjoint to diff.adoint?

tests/operators/test_autograd_linop.py
```diff
@@ -0,0 +1,91 @@
...
+    torch.testing.assert_close(diff.adjoint(u), diff.adjoint(u))
```
ckolbPTB: not sure what is tested here. did you want to compare the forward of diff and nondiff?

tests/operators/test_autograd_linop.py
```diff
@@ -0,0 +1,91 @@
...
+
+    # and the adjoin should be correct
```
ckolbPTB: adjoint

tests/operators/test_autograd_linop.py
```diff
@@ -0,0 +1,91 @@
...
+    dotproduct_adjointness_test(nondiff, u, v)
+
+
+def test_linop_autograd_nondifferentiable():
+    """Test the forward of the non differentiable operator."""
```
ckolbPTB: Test the backward of the forward of the non-differentialbe operator?

tests/operators/test_autograd_linop.py
```diff
@@ -0,0 +1,91 @@
...
+    # because the forward is non differentiable (python float)
+    x = torch.tensor(1.0, requires_grad=True, dtype=torch.float64)
+    with pytest.raises(GradcheckError):
+        torch.autograd.gradcheck(nondiff, x)
+
+
+def test_linop_autograd_differentiable_forward():
+    """Test the backward of the forward of the differentiable operator."""
+    diff = DiffentiableOperator()
+    # The differentiable operator should work, as
```
ckolbPTB, line 70: The gradient of the differentiable?

tests/operators/test_autograd_linop.py
```diff
@@ -0,0 +1,91 @@
...
+    # Should work, as the forward will be used as the backward
+    # of the adjoint.
+    x = torch.tensor(1.0, requires_grad=True, dtype=torch.float64)
+    torch.autograd.gradcheck(diff.H, x)
+
+
+def test_linop_autograd_differentiable_gradgrad():
+    """Test the gradgrad of the differentiable operator."""
+    diff = DiffentiableOperator()
+    # The backward of the backward should be the working
```
ckolbPTB: no "the"

tests/operators/test_autograd_linop.py
```diff
@@ -0,0 +1,91 @@
...
+from tests import RandomGenerator
+from tests.helper import dotproduct_adjointness_test
+
+
+class NonDiffentiableOperator(LinearOperator, adjoint_as_backward=False):
+    """Dummy operator for testing purposes."""
+
+    def forward(self, x):
+        python_float = x.item()
+        tensor = torch.tensor([2 * python_float], dtype=x.dtype)  # breaks the autograd
```
ckolbPTB: breaks the autograd backward functionality?

tests/operators/test_autograd_linop.py
```diff
@@ -0,0 +1,91 @@
...
+        tensor = torch.tensor([2 * python_float], dtype=x.dtype)
+        return (tensor,)
+
+
+class DifferentiableOperator(LinearOperator, adjoint_as_backward=True):
+    """Dummy operator for testing purposes. Should be differentiable due to adjoint_as_backward=True."""
+
+    def forward(self, x):
+        python_float = x.item()  # would break autograd as floats do not track gradients
+        tensor = torch.tensor([2 * python_float], dtype=x.dtype)  # breaks the autograd
```
ckolbPTB: I guess x.item() from above breaks the autograd and not this line

fzimmermann89: ups, forgot to remove second line. fixed

ckolbPTB (approved):

tests/operators/test_autograd_linop.py
ckolbPTB: This still needs fixing

fzimmermann89: Ahhh.. sorry, I was fixating an the first two 'the'.
This took me some time ..

tests/operators/test_autograd_linop.py
```diff
@@ -0,0 +1,91 @@
...
+    # Should work, as the forward will be used as the backward
+    # of the adjoint.
+    x = torch.tensor(1.0, requires_grad=True, dtype=torch.float64)
+    torch.autograd.gradcheck(diff.H, x)
+
+
+def test_linop_autograd_differentiable_gradgrad():
+    """Test the gradgrad of the differentiable operator."""
+    diff = DifferentiableOperator()
+    # backward of backward should be the working
```
fzimmermann89:
```suggestion
    # backward of backward should be working
```

PR #405, opened 2024-09-18
Add tests for autodiff of operators

ckolbPTB:
Adds the gradient checks for operators from #307


ckolbPTB:
The MOLLI test fails. Currently we have:
`signal(a,b,t1) = a * (1 - b/a * torch.exp(ti / t1 * (1 - b/a)))`

The b/a term is what is causing problems. My suggestion would be to change this to 
`signal(a,c,t1) = a * (1 - c * torch.exp(ti / t1 * (1 - c)))`

a and b don't have any physiological meaning and this should also make it easier to find good constraints for c. 

See #406

ckolbPTB:
by the way I think it should be `ti / t1 * (b/a - 1)`

Messroghli et al:

grafik

tests/operators/models/test_inversion_recovery.py
```diff
@@ -39,3 +40,11 @@ def test_inversion_recovery_shape(parameter_shape, contrast_dim_shape, signal_sh
     m0, t1 = create_parameter_tensor_tuples(parameter_shape, number_of_tensors=2)
     (signal,) = model_op.forward(m0, t1)
     assert signal.shape == signal_shape
+
+
+@pytest.mark.filterwarnings('ignore:Anomaly Detection has been enabled')
```
fzimmermann89: can we isntead filter the warnings inside the test helper?

[i.e. use  filterwarnings](https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings)

or ignore this warning globally in the tests (set it in the pyproject.toml)

tests/helper.py
```diff
@@ -73,19 +79,19 @@ def dotproduct_adjointness_test(
     torch.testing.assert_close(dotproduct_range, dotproduct_domain, rtol=relative_tolerance, atol=absolute_tolerance)
 
 
-def operator_isometry_test(
-    operator: Operator, u: torch.Tensor, relative_tolerance: float = 1e-3, absolute_tolerance=1e-5
+def linear_operator_isometry_test(
+    linear_operator: LinearOperator, u: torch.Tensor, relative_tolerance: float = 1e-3, absolute_tolerance=1e-5
```
fzimmermann89: why is this limited to linear_operator?
this tests should work for  any operators first output

tests/helper.py
```diff
@@ -26,9 +28,13 @@ def relative_image_difference(img1: torch.Tensor, img2: torch.Tensor) -> torch.T
 
 
 def dotproduct_adjointness_test(
-    operator: Operator, u: torch.Tensor, v: torch.Tensor, relative_tolerance: float = 1e-3, absolute_tolerance=1e-5
+    linear_operator: LinearOperator,
```
fzimmermann89: not to fond of this change.
that it is a linear_operator is already given by the type info.
does it ake the code easier to understand to read linear_operator than operator?
if talking about it, I would always say, for example: "Given a linear operator, you can always get the adjoint of the _operator_", i.e. drop the adjective.


but if you have strong preference for linear_operator here, I wont care.

tests/helper.py
```diff
@@ -128,4 +134,37 @@ def operator_unitary_test(
     AssertionError
         if the adjointness property does not hold
     """
-    torch.testing.assert_close(u, operator.adjoint(operator(u)[0])[0], rtol=relative_tolerance, atol=absolute_tolerance)
+    torch.testing.assert_close(
+        u, linear_operator.adjoint(linear_operator(u)[0])[0], rtol=relative_tolerance, atol=absolute_tolerance
+    )
+
+
+def autodiff_of_operator_test(
```
fzimmermann89:
```suggestion
def autodiff_test(
```

tests/helper.py
```diff
@@ -128,4 +134,37 @@ def operator_unitary_test(
...
+    operator: Operator[*tuple[torch.Tensor, ...], tuple[torch.Tensor, ...]],
+    *u: Any,
+):
+    """Test if autodiff of an operator is working.
+    This test does not check that the gradient is correct but simply that it can be calculated using autodiff.
```
fzimmermann89: ..using both torch.func.jvp and torch.func.vjp

fzimmermann89:
remind me again, we we have no grad_check in the helpers, testing the numeric values of the gradients?

ckolbPTB:
> remind me again, we we have no grad_check in the helpers, testing the numeric values of the gradients?

I don't think we have discussed this yet. If we use pytorch autograd functionality, would this test not merely test pytorch functionality?

pyproject.toml
```diff
@@ -66,7 +66,8 @@ notebook = [
 testpaths = ["tests"]
 filterwarnings = [
     "error",
-    "ignore:'write_like_original':DeprecationWarning:pydicom:"
+    "ignore:'write_like_original':DeprecationWarning:pydicom:",
+    "ignore:Anomaly Detection has been enabled:UserWarning", #torch.autograd
```
ckolbPTB: I am a bit puzzled here because I thought that something like 
`ignore:Anomaly Detection has been enabled:UserWarning:torch`
or
`ignore:Anomaly Detection has been enabled:UserWarning:torch.*`
should work, but I could not get it to work. If anybody knows how to get it to work with specifying the module correctly I would be very happy to know.

ckolbPTB:
@fzimmermann89 could you have a look at this again please

tests/operators/models/test_inversion_recovery.py
```diff
@@ -3,6 +3,7 @@
 import pytest
 import torch
 from mrpro.operators.models import InversionRecovery
+from tests.helper import autodiff_test
```
fzimmermann89: we could consider importing all these in `tests/__init__.py`
Such that we can just do from tests import ..., similar as we do with the randomgenerator. 
I think we realistically only ever import from 'helper' of tests anyways...

fzimmermann89 (commented):
remind me, what was the plan here?
should we add a test for each operator and its adjoint?

fzimmermann89:
So far, what is there look fine to me

ckolbPTB:
> remind me, what was the plan here?
> should we add a test for each operator and its adjoint?

With this PR we wanted to add autodiff tests for all non-linear operators. autodiff (and some more) tests for linear operators would then we added in PR #407

tests/helper.py
```diff
@@ -129,3 +138,32 @@ def operator_unitary_test(
         if the adjointness property does not hold
     """
     torch.testing.assert_close(u, operator.adjoint(operator(u)[0])[0], rtol=relative_tolerance, atol=absolute_tolerance)
+
+
+def autodiff_test(
+    operator: Operator[*tuple[torch.Tensor, ...], tuple[torch.Tensor, ...]],
+    *u: Any,
```
fzimmermann89: why is this any?

fzimmermann89: shoulnt this match *tuple[torch.Tensor, ...] from operator?

fzimmermann89: i will try if that works

fzimmermann89: also, this is not py310 compatible, we would have to use Unpack instead of splitting with *

fzimmermann89:
Will need #515 to pass mypy

fzimmermann89:
> Will need #515 to pass mypy

I stand corrected -- #515 would make the type hints more useful, but it passes without the changes

fzimmermann89 (approved):

PR #408, opened 2024-09-20
Move GridSampleOp to new autograd.Function syntax

fzimmermann89:
Allows for torch.func


fzimmermann89:
should be rebased to main before merge

ckolbPTB (approved):

PR #409, opened 2024-09-20
Move SliceProjectionOp to new autograd.Function syntax

fzimmermann89:
Also clean up type hints.

This change is requiered for torch.func to work -- but
torch.func still fails due to pytorch/pytorch#136357 🙈


ckolbPTB (approved):

fzimmermann89:
@ckolbPTB

PR #411, opened 2024-09-23
Add ScaledFunctional implementing scalar*Functional

fzimmermann89:


fzimmermann89:
currently  torch.tensor(1.) * L2SquaredFunctional() will result in mypy type Tensor|Functional.
This is caused by wrong type hints in pytorch. 
I should fix https://github.com/pytorch/pytorch/pull/1353 and get it merged..

src/mrpro/operators/Functional.py
```diff
@@ -162,6 +170,12 @@ def prox_convex_conj(self, x: torch.Tensor, sigma: torch.Tensor | float = 1.0) -
         sigma[sigma < 1e-8] += 1e-6
         return (x - sigma * self.prox(x / sigma, 1 / sigma)[0],)
 
+    def __rmul__(self, scalar: torch.Tensor | complex) -> ProximableFunctional:
```
guastinimara, line 174: is there a reason why scalar is either a tensor or complex? Because in the __rmul__ for Functional is either tensor or float and it made more sense to me

src/mrpro/operators/Functional.py
```diff
@@ -173,3 +187,80 @@ class ElementaryProximableFunctional(ElementaryFunctional, ProximableFunctional)
     A proximable functional is a functional :math:`f(x)` that has a prox implementation,
     i.e. a function that yields :math:`argmin_x \sigma f(x) + 1/2 \|x - y\|^2`.
     """
+
+
+class ScaledFunctional(Functional):
+    """Functional scaled by a scalar."""
+
+    def __init__(self, functional: Functional, scale: torch.Tensor | complex) -> None:
```
guastinimara: similarly here, I don't understand why we give the option of complex if in the parameters we write that it must be real and positive?

guastinimara: or is this a constraint only for the prox to be well-defined? In that case I would write it separately in the description of ScaledFunctional and ScaledProximableFunctional

fzimmermann89: Good catch.
The reason is that all ProximalFunctionals are Operators.
We can multiply all Operators with complex numbers.
Thus, from a type hinting perspective, we must also allow multiplication of ProximableFunctional with complex.
I will change it to make it consistent complex everywhere.
In the end, the code will error if you call .prox on something that after applying all scalings etc is still complex or negative.

So (1j*(-1j)*Functional).prox() would work, but (1j*Functional).prox() would fail.

Does that make sense?

src/mrpro/operators/Functional.py
fzimmermann89: in the rmul, I had to add a type ignore because of a pytorch bug. That type ignore also silenced that it should be complex (not float). Not that multiplication with a complex number makes much sense, but because all Operators can be multiplied by complex numbers, and Functionals are Operators.. so you can still multiply them -- just the prox does not give a correct result, but throws an exception.

src/mrpro/operators/Functional.py
```diff
@@ -12,6 +14,12 @@
 class Functional(Operator[torch.Tensor, tuple[torch.Tensor]]):
     """Functional Base Class."""
 
+    def __rmul__(self, scalar: torch.Tensor | float) -> Functional:
```
fzimmermann89:
```suggestion
    def __rmul__(self, scalar: torch.Tensor | complex) -> Functional:
```

src/mrpro/operators/Functional.py
```diff
@@ -173,3 +187,80 @@ class ElementaryProximableFunctional(ElementaryFunctional, ProximableFunctional)
...
+        Returns
+        -------
+            Proximal mapping applied to the input tensor
+        """
+        self._throw_if_negative_or_complex(
+            self.scale, 'For prox to be defined, the scaling factor must be real and non-negative'
+        )
+        return (self.functional.prox(x, sigma * self.scale)[0],)
+
+    def prox_convex_conj(self, x: torch.Tensor, sigma: torch.Tensor | float = 1.0) -> tuple[torch.Tensor]:
```
fzimmermann89, line 266:
```suggestion
    def prox_convex_conj(self, x: torch.Tensor, sigma: torch.Tensor | complex = 1.0) -> tuple[torch.Tensor]:
```

fzimmermann89:
Scratch my previous responses.
Scale must be float.
The type hints in rmul must be complex,
but rmul should return not implemented for complex.

fzimmermann89:
pre-commit.ci autofix

guastinimara:
nice! I like it like this :)

fzimmermann89:
Please remember to do review

guastinimara (approved):

PR #416, opened 2024-09-27
Add arithmetic for SpatialDimension

ckolbPTB:
When saving our data to Dicom we need to recalculate positions and orientations. This would provide the arithemtic to do this more easily. 

Current disadvantage: `z,y,x` were `int | float | Tensor` but are now always `Tensor`. This makes lots of things easier but would require quite some changes to our code (especially the SliceProjectionOp). We could of course keep it as `int | float | Tensor` but then we would need quite a few checks in the arithmetic part.

Another option would be to have `z,y,x` as `int | float` and then things like `read_dir `would be a list of SpatialDimension objects.

### DIFF
--- a/src/mrpro/data/SpatialDimension.py
+++ b/src/mrpro/data/SpatialDimension.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 from collections.abc import Callable
 from dataclasses import dataclass
-from typing import Generic, Protocol, TypeVar
+from typing import Any, Generic, Protocol, TypeVar
 
 import numpy as np
 import torch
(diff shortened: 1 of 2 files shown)


fzimmermann89:
I would prefer to keep it as int|float|Tensor.
Especially the int part is important for mypy.
We use spatal dimensions for image sizes etc where only ints are valid. This should be type-checked.

fzimmermann89:
can you give an example where where this is required?

ckolbPTB:
The image centre for dicom is the top left corner of the image so we have to do something like `iheader.position + iheader.fov.x/2 * iheader.read_dir +... ` for every other dimension

would be nice if we could simply say `iheader.position[idx,...] + iheader.fov.x/2 * iheader.read_dir[idx,...] + `

currently that is not so straight forward.

Maybe we could add something like `.as_xyz_tensor` and `.as_zyx_tensor`?

fzimmermann89:
the original idea was to be explicit with x,y, and z.
So` SpatialDimension(x=iheader.position.x[idx]+iheader.fov.x/2 * iheader.read_dir,... )` on purpose. 

Nevertheless, I will do a draft with the correct type hints of this, so we can do it in your way.
give me a  couple minutes.

Furthermore, this might be a good opportunity to add a orientation as a Rotation property to the header.
I put this to the side in the past, as we did not have any code accessing read_dir, etc anyways. 
This should do the the rotation of the fov then as iheader.rotation(iheader.fov)

ckolbPTB:
superseded by #417

PR #417, opened 2024-09-27
Add arithmetic to SpatialDimension

fzimmermann89:
Follow up to #416 with correct type hints


fzimmermann89:
I also moved Rotation out of utils (as it depended on SpatialDimension) and TYPE_CHECKING guarded modify_acq_info to avoid circular imports. Functions in utils should not depend on data classes (otherwise, we can't use them in data classes)
As modify_acq_info will be deprecated anyways once we move that functionality into acqinfo, I only added the future annotations and guard.

These changes might be moved to a separate (stacked) PR .


I removed xyz to force us to always be zyx or be explicit about it and write (spatialdim.x,spatialdim.y,spatialdim.z). Otherwise, the temptation to wrongly use xyz is high..

fzimmermann89:
This passes all tests from #416  and passes mypy

fzimmermann89:
(To be merged after #418  into main)

src/mrpro/data/SpatialDimension.py
```diff
@@ -101,3 +102,136 @@ def zyx(self) -> tuple[T, T, T]:
...
+    @overload
+    def __mul__(self: SpatialDimension[int], other: int | SpatialDimension[int]) -> SpatialDimension[int]: ...
+    @overload
+    def __mul__(self: SpatialDimension[int], other: float | SpatialDimension[float]) -> SpatialDimension[float]: ...
+    @overload
+    def __mul__(self: SpatialDimension[float], other: float | SpatialDimension[float]) -> SpatialDimension[float]: ...
+
+    @overload
+    def __mul__(
+        self: SpatialDimension[torch.Tensor], other: float | T | SpatialDimension
```
schuenke: why do we need the an explicit `float` here? Shouldn't `float` be covered by `T` already?

src/mrpro/data/SpatialDimension.py
```diff
@@ -101,3 +102,136 @@ def zyx(self) -> tuple[T, T, T]:
...
+    ) -> SpatialDimension[torch.Tensor]: ...
+
+    def __mul__(self: SpatialDimension[T], other: T | float | SpatialDimension) -> SpatialDimension:
```
schuenke: see `float` / `T` comment above. If we do need it, using the same order everywhere would be nice. 

Please also check the rest of the code... appeards several times

schuenke (approved):
lgtm. IMO all cases are covered. Didn't go through all the tests, but they seem to cover everhing as well.

Please check my specific comments about the `float | T` type hints.

src/mrpro/data/SpatialDimension.py
fzimmermann89: no, it is required, you can write it out as, for example:
SpatialDimension[int], other: int | float | SpatialDimension
All spatial dimensions can be multiplied by a scalar (float, which then includes int).
But only SpatialDimension[np.ndarray] can be multiplied by np.ndarray , only SpatialDimensin[tensor] by tensor...
 
 or am I missing something?

fzimmermann89: i can go through and sort the order

schuenke: but T is `int | float | torch.Tensor` already, no?
so `float | T` is `float | int | float | torch.Tensor` , no?

Or am I actually missing something here?

schuenke: ah, got it... it's needed. Sorry for the confusion! Hmm... but maybe not in the last overload... I will also check it again 😁

fzimmermann89: but maybe your are right an the cases can be cleaned up, let me check

fzimmermann89:
i cleaned the type hints a bit.

also, I removed the conversion option from the class methods.
Instead, there is now an apply_ that is used after creation to do the conversion.
This will anyways be required for rearranging etc..

fzimmermann89:
Also, I added typing asserts as tests.
This is in the same test_spatialdimension.py file, even though the tests is not actually run by pytest -- just parsed by mypy.

maybe @schuenke can have another look ...

src/mrpro/data/SpatialDimension.py
```diff
@@ -70,34 +69,342 @@ def from_array_xyz(
...
+        self: SpatialDimension[T_co_scalar] | SpatialDimension[T_co_vector],
+        other: SpatialDimension[T_co_scalar] | SpatialDimension[T_co_vector],
+    ) -> bool | T_co_vector:
+        """Check if self is greater than or equal to other."""
+        if not isinstance(other, SpatialDimension):
+            return NotImplemented
+        return (self.x >= other.x) & (self.y >= other.y) & (self.z >= other.z)
+
+    def __post_init__(self):
+        """Ensure that the data is of the same type."""
```
ckolbPTB: this actually checks the shape and not the type

ckolbPTB:
Why do we want to support `np.ndarray` for SpatialDimension? It would be nice if that would be supported in the init and maybe from_-functions but it could then be converted to tensors.

fzimmermann89:
> Why do we want to support `np.ndarray` for SpatialDimension? It would be nice if that would be supported in the init and maybe from_-functions but it could then be converted to tensors.

from_array* will convert to tensors.
I mostly added nd.array type hints to see if the hints are correct or if errors are just hidden by the fact that there is only one vector type supported.  The biggest uses I see is for compatibility and object arrays (example: array of callables).
I would leave it in as long as it doesn't cause issues.

ckolbPTB:
> I would leave it in as long as it doesn't cause issues.

It seems strange to me why we would support ndarrays here when everywhere else we ensure torch.tensors. I would go the other way and remove it unless we really need it. Otherwise we should also support it for getitem/setitem

fzimmermann89:
> Otherwise we should also support it for getitem/setitem

I added the overloads for getitem and setitem and a test for these.

fzimmermann89:
> > Otherwise we should also support it for getitem/setitem
> 
> I added the overloads for getitem and setitem and a test for these.

Scratch that (_Was kümmert mich mein Geschwätz von gestern_). 
I removed np.ndarray support, as I wont use a SpatialDimension of callables in the use case I had in mind anyway.

src/mrpro/data/SpatialDimension.py
```diff
@@ -3,16 +3,38 @@
...
 
 import numpy as np
 import torch
 from numpy.typing import ArrayLike
 
+import mrpro.utils.typing as type_utils
 from mrpro.data.MoveDataMixin import MoveDataMixin
 
-T = TypeVar('T', int, float, torch.Tensor)
+# Change here to add more types
```
ckolbPTB, line 18: could we add a sentence why we need the types to be covariant?

src/mrpro/data/SpatialDimension.py
```diff
@@ -3,16 +3,38 @@
...
+T = TypeVar('T', torch.Tensor, int, float)
+T_co = TypeVar('T_co', torch.Tensor, int, float, covariant=True)
+T_co_float = TypeVar('T_co_float', float, torch.Tensor, covariant=True)
+T_co_vector = torch.Tensor
+T_co_scalar = TypeVar('T_co_scalar', int, float, covariant=True)
+
+
+def _as_vectortype(x: ArrayLike) -> VectorTypes:
+    """Convert ArrayLike to VectorType."""
+    if isinstance(x, VectorTypes) and type(x) in get_args(VectorTypes):
```
ckolbPTB, line 32: is `type(x) in get_args(VectorTypes)` not enough here?

src/mrpro/data/SpatialDimension.py
```diff
@@ -24,80 +46,401 @@ class XYZ(Protocol[T]):
...
         data
             shape (..., 3) in the order (x,y,z)
-        conversion
-            will be called for each value to convert it
         """
-        if not isinstance(data, np.ndarray | torch.Tensor):
-            data = np.asarray(data)
-
+        data_ = _as_vectortype(data)
         if np.size(data, -1) != 3:
```
ckolbPTB: I would have this check before _as_vectortype

src/mrpro/data/SpatialDimension.py
```diff
@@ -24,80 +46,401 @@ class XYZ(Protocol[T]):
...
         ----------
         data
             shape (..., 3) in the order (z,y,x)
-        conversion
-            will be called for each value to convert it
         """
-        data = torch.flip(torch.as_tensor(data), (-1,))
-        return SpatialDimension.from_array_xyz(data, conversion)
+        data = _as_vectortype(data)
+        if np.size(data, -1) != 3:
```
ckolbPTB: same as above

src/mrpro/data/SpatialDimension.py
```diff
@@ -24,80 +46,401 @@ class XYZ(Protocol[T]):
...
+        Returns
+        -------
+            Empty tuple if x, y, and z are scalar types, otherwise shape
+
+        Raises
+        ------
+            ValueError if the shapes are not equal
+        """
+        if isinstance(self.x, ScalarTypes) and isinstance(self.y, ScalarTypes) and isinstance(self.z, ScalarTypes):
+            return ()
```
ckolbPTB, line 440: would I not expect an error if I call shape on a scalar?

ckolbPTB (approved):
some minor last comments

src/mrpro/data/SpatialDimension.py
fzimmermann89: At runtime yes. But mypy curently only understands the former.

src/mrpro/data/SpatialDimension.py
fzimmermann89: It should be this order.

ArrayLike is anything that numpy can convert to an Array, for example a list of lists.
Np.shape on objects that do not have .shape is plemented by converting to an Array, then checking it's size.
So switching the commands here would result in double conversion and has no advantages?

src/mrpro/data/SpatialDimension.py
fzimmermann89: Calling shape on scalar numpy arrays or tensors returns also an empty tuple.

src/mrpro/data/SpatialDimension.py
ckolbPTB: should it then be np.size(data_, -1)?

fzimmermann89: YES! 
thank you

fzimmermann89:
pre-commit.ci autofix

fzimmermann89:
pre-commit.ci autofix

PR #421, opened 2024-09-28
Add export of KData to ismrmrd file

ckolbPTB:
Save the KData object as mrd raw data file including all the header parameters supported by MRpro and the trajectory.


src/mrpro/data/AcqInfo.py
```diff
@@ -263,3 +273,53 @@ def spatialdimension_2d(
             version=tensor_2d(headers['version']),
         )
         return acq_info
+
+    def add_ismrmrd_acquisition_info(
+        self, acquisition: ismrmrd.Acquisition, other: int, k2: int, k1: int
```
fzimmermann89: Multiple other?

src/mrpro/data/AcqInfo.py
```diff
@@ -263,3 +273,53 @@ def spatialdimension_2d(
...
+            [m_to_mm(getattr(self.patient_table_position, label)[other, k2, k1, 0]) for label in ['x', 'y', 'z']]
+        )
+        acquisition.phase_dir = tuple([getattr(self.phase_dir, label)[other, k2, k1, 0] for label in ['x', 'y', 'z']])
+        acquisition.physiology_time_stamp = tuple(self.physiology_time_stamp[other, k2, k1, :])
+        acquisition.position = tuple(
+            [m_to_mm(getattr(self.position, label)[other, k2, k1, 0]) for label in ['x', 'y', 'z']]
+        )
+        acquisition.read_dir = tuple([getattr(self.read_dir, label)[other, k2, k1, 0] for label in ['x', 'y', 'z']])
+        acquisition.sample_time_us = self.sample_time_us[other, k2, k1, 0]
+        acquisition.scan_counter = self.scan_counter[other, k2, k1, 0]
```
fzimmermann89: Self.slicedir.x ..
Seems simpler than  getattr and loop

src/mrpro/data/KHeader.py
```diff
@@ -12,7 +12,7 @@
 import torch
 
 from mrpro.data import enums
-from mrpro.data.AcqInfo import AcqInfo, mm_to_m, ms_to_s
+from mrpro.data.AcqInfo import AcqInfo, m_to_mm, mm_to_m, ms_to_s, s_to_ms
```
fzimmermann89: We could move the conversions to utils.conversions and add some for hz/Tesla/ppm ?
(And amy other common scaling factors.)

src/mrpro/data/KHeader.py
```diff
@@ -171,13 +171,13 @@ def from_ismrmrd(
             parameters['n_coils'] = header.acquisitionSystemInformation.receiverChannels
 
         if header.sequenceParameters is not None:
-            if header.sequenceParameters.TR:
+            if any(header.sequenceParameters.TR):
```
fzimmermann89: Why any?
I think this checks for empty list. We would save  values that are set to 0. Any if it is not set, we ignore it.

src/mrpro/data/KHeader.py
```diff
@@ -210,6 +210,8 @@ def from_ismrmrd(
 
         if enc.trajectory is not None:
             parameters['traj_type'] = enums.TrajectoryType(enc.trajectory.value)
+        if enc.trajectoryDescription is not None:
+            parameters['trajectory_description'] = TrajectoryDescription.from_ismrmrd(enc.trajectoryDescription)
```
fzimmermann89: Why do we add this?

src/mrpro/data/KHeader.py
```diff
@@ -251,6 +253,8 @@ def from_ismrmrd(
 
         # Dump everything into misc
         parameters['misc'] = dataclasses.asdict(header)
+        # Remember encoding number
```
fzimmermann89: Maybe the comment could explain why, instead of what...? (I don't know why we save it)

src/mrpro/data/KHeader.py
```diff
@@ -270,6 +274,180 @@ def from_ismrmrd(
...
+        we update the values based on the attributes of the KHeader.
+
+        Returns
+        -------
+            ISMRMRD header
+        """
+        header = ismrmrdschema.ismrmrdHeader()
+        header.version = self.misc['version']
+
+        def create_from_misc_dictionary(xsd_type, misc_dictionary_entry: dict):  # noqa: ANN001
```
fzimmermann89: I would prefer not to do this .

Either we move these into our header, switch to using the ismrmrd header completely or be ok with loosing some of these on save.

The misc dictionary is a dumping ground for stuff that weigjt want to move into our header without any guarantees.
If some of these are so important or required for a valid ismrmd, we should move them into our header

Otherwise, This now makes.the dict it an untyped part of our API.
So it is the worst of all possible ways to deal with the header information...

src/mrpro/data/AcqInfo.py
ckolbPTB: add_to_ismrmrd_acquisition

make other as tuple and unpack tuple inside

src/mrpro/data/KHeader.py
ckolbPTB: done in #425

src/mrpro/data/AcqInfo.py
```diff
@@ -249,17 +268,91 @@ def spatialdimension_2d(
...
-            slice_dir=spatialdimension_2d(headers['slice_dir']),
             trajectory_dimensions=tensor_2d(headers['trajectory_dimensions']).fill_(3),  # see above
             user_float=tensor_2d(headers['user_float']),
             user_int=tensor_2d(headers['user_int']),
             version=tensor_2d(headers['version']),
         )
         return acq_info
+
+    def add_to_ismrmrd_acquisition(
+        self, acquisition: ismrmrd.Acquisition, other: Any, k2: int, k1: int
```
ckolbPTB: TypeHint for `other` should be `typeing_utils.IndexerType `once #417 is merged

ckolbPTB: Need to merge #425 before this can be merged

src/mrpro/data/AcqInfo.py
```diff
@@ -258,3 +259,51 @@ def spatialdimension_2d(data: np.ndarray) -> SpatialDimension[torch.Tensor]:
             version=tensor_2d(headers['version']),
         )
         return acq_info
+
+    def add_to_ismrmrd_acquisition(
+        self, acquisition: ismrmrd.Acquisition, other: type_utils.TorchIndexerType, k2: int, k1: int
```
ckolbPTB: I guess this should be 
`other: tuple[_SingleTorchIndexerType, ...]`

would it make sense to make `_SingleTorchIndexerType` from utils.typing also globally available?

ckolbPTB:
pre-commit.ci autofix

src/mrpro/data/KHeader.py
```diff
@@ -255,6 +255,97 @@ def from_ismrmrd(
...
+        subj = ismrmrdschema.subjectInformationType()
+        subj.patientName = self.patient_name
+        header.subjectInformation = subj
+
+        # Measurement information
+        meas = ismrmrdschema.measurementInformationType()
+        meas.protocolName = self.protocol_name
+        meas.measurementID = self.measurement_id
+        if self.datetime is not None:
+            meas.seriesTime = str(self.datetime.time())
```
fzimmermann89, line 272: can this be None?

fzimmermann89: i thought we set it to file modify data if not set in the header or somehting like that..

fzimmermann89: (need to check)

src/mrpro/data/KHeader.py
```diff
@@ -255,6 +255,97 @@ def from_ismrmrd(
...
+
+        # Acquisition system information
+        sys = ismrmrdschema.acquisitionSystemInformationType()
+        sys.systemModel = self.model
+        sys.systemVendor = self.vendor
+        header.acquisitionSystemInformation = sys
+
+        # Sequence information
+        seq = ismrmrdschema.sequenceParametersType()
+        if isinstance(self.tr, torch.Tensor):
```
fzimmermann89: why not "is not None" ?

fzimmermann89:
I think there was a bug:
`self.patient_table_position[*idx][0].apply_(m_to_mm)`

getitem returns a view.
apply_ changes that view in place.

#547

fzimmermann89:
Fixed the idx and simplified a tiny bit.

this now should pass mypy and work in py310

ckolbPTB:
> I think there was a bug:

Good catch! thanks!

fzimmermann89 (dismissed):

fzimmermann89:
blocked by #560

fzimmermann89:
needs adjustment to current header structure

fzimmermann89 (dismissed):

ckolbPTB:
@fzimmermann89 should we also rename "to/from file" to `to_ismrmrd` and `from_ismrmrd` or `to_mrd` and `from_mrd` to somewhat distinguish between a ismrmrd object and a mrd-file?

src/mrpro/data/AcqInfo.py
```diff
@@ -327,3 +341,67 @@ def spatialdimension_5d(data: np.ndarray) -> SpatialDimension[torch.Tensor]:
...
             additional_values = tuple(tensor_5d(headers[field]) for field in additional_fields)
             return acq_info, additional_values
+
+    def write_single_acquisition_to_ismrmrd_acquisition(
+        self,
+        acquisition: ismrmrd.Acquisition,
+        convert_time_stamp: Callable[[float], int] = convert_time_stamp_to_siemens,
+    ) -> ismrmrd.Acquisition:
+        """Overwrite ISMRMRD acquisition information for single acquisition."""
+        if np.prod(self.shape) != 1:
```
fzimmermann89: self.shape.numel()

src/mrpro/data/KHeader.py
```diff
@@ -232,3 +240,87 @@ def from_ismrmrd(
...
+            m_to_mm(self.encoding_fov.x), m_to_mm(self.encoding_fov.y), m_to_mm(self.encoding_fov.z)
+        )
+        encoding_matrix = ismrmrdschema.matrixSizeType(
+            self.encoding_matrix.x, self.encoding_matrix.y, self.encoding_matrix.z
+        )
+        encoding_space.matrixSize = encoding_matrix
+        encoding_space.fieldOfView_mm = encoding_fov
+
+        # Recon space
+        recon_space = ismrmrdschema.encodingSpaceType()
```
fzimmermann89, line 313: encoding.reconSpace?

src/mrpro/data/KHeader.py
```diff
@@ -232,3 +240,87 @@ def from_ismrmrd(
...
+        recon_fov = ismrmrdschema.fieldOfViewMm(
+            m_to_mm(self.recon_fov.x), m_to_mm(self.recon_fov.y), m_to_mm(self.recon_fov.z)
+        )
+        recon_matrix = ismrmrdschema.matrixSizeType(self.recon_matrix.x, self.recon_matrix.y, self.recon_matrix.z)
+        recon_space.matrixSize = recon_matrix
+        recon_space.fieldOfView_mm = recon_fov
+
+        # Set encoded and recon spaces
+        encoding.encodedSpace = encoding_space
```
fzimmermann89, line 322: either directly set the encodedSpace or move in the respecitve blocks?

src/mrpro/data/KData.py
```diff
@@ -324,6 +332,84 @@ def sort(x: T) -> T:
...
+
+        acq.available_channels = acq_shape[1]
+        acq.version = int(version('ismrmrd').split('.')[0])
+        acq.scan_counter = 0
+
+        for other in np.ndindex(self.data.shape[:-4]):
+            for k2 in range(self.data.shape[-3]):
+                for k1 in range(self.data.shape[-2]):
+                    # Select current readout for all coils
+                    single_acquisition = self[(*other, slice(None), k2, k1, slice(None))]  # type: ignore[index]
```
fzimmermann89: why does this give a typing error?

src/mrpro/data/KData.py
```diff
@@ -324,6 +332,84 @@ def sort(x: T) -> T:
...
+                    single_acquisition.header.acq_info.write_single_acquisition_to_ismrmrd_acquisition(
+                        acq, convert_time_stamp
+                    )
+
+                    acq.traj[:] = torch.squeeze(single_acquisition.traj.as_tensor(-1)).numpy()[:, ::-1]  # zyx -> xyz
```
fzimmermann89: should this maybe be in the write single acquisition

fzimmermann89:
There is some issue with the cetner sample or flipped readouts in general. @ckolbPTB will investigate and either address it here, or  if it gets more involved, remove the center sample adjustment here, merge this PR first an then add an issue for the flipped readouts.

fzimmermann89:
i just removed the correction here and we keep this as a todo.

fzimmermann89 (approved):

fzimmermann89:
Any reason not to merge this? @ckolbPTB

PR #422, opened 2024-09-29
Extend parameters of IHeader

ckolbPTB:


src/mrpro/data/IHeader.py
```diff
@@ -38,6 +48,48 @@ class IHeader(MoveDataMixin):
...
+    phase_dir: SpatialDimension[torch.Tensor] | None
+    """Directional cosine of phase encoding (2D)."""
+
+    position: SpatialDimension[torch.Tensor] | None
+    """Center of the excited volume, in LPS coordinates relative to isocenter [m]."""
+
+    read_dir: SpatialDimension[torch.Tensor] | None
+    """Directional cosine of readout/frequency encoding."""
+
+    slice_dir: SpatialDimension[torch.Tensor] | None
```
fzimmermann89: Consider using a rotation here.

If not, "readout_direction" etc ?

src/mrpro/data/IHeader.py
```diff
@@ -14,18 +17,25 @@
...
     """MR image data header."""
 
-    # ToDo: decide which attributes to store in the header
+    b0: float | None
+    """Magnetic field strength [T]."""
+
     fov: SpatialDimension[float]
     """Field of view [m]."""
 
+    h1_freq: float | None
```
fzimmermann89: Can we give a better name?
At least write out frequency?
Proton_lamor_frequency?

Also, is this redundant  to b0?

src/mrpro/data/IHeader.py
```diff
@@ -38,6 +48,48 @@ class IHeader(MoveDataMixin):
...
+    read_dir: SpatialDimension[torch.Tensor] | None
+    """Directional cosine of readout/frequency encoding."""
+
+    slice_dir: SpatialDimension[torch.Tensor] | None
+    """Directional cosine of slice normal, i.e. cross-product of read_dir and phase_dir."""
+
+    patient_table_position: SpatialDimension[torch.Tensor] | None = None
+    """Offset position of the patient table, in LPS coordinates [m]."""
+
+    n_coils: int | None = None
```
fzimmermann89: Why? Shouldn't this be detected by the data?
This might give rise to inconsistencies or needs updating on subsetting.
I would say this is not worth the effort.

src/mrpro/data/IHeader.py
```diff
@@ -38,6 +48,48 @@ class IHeader(MoveDataMixin):
...
+    """Number of receiver coils."""
+
+    datetime: datetime.datetime | None = None
+    """Date and time of acquisition."""
+
+    echo_train_length: int | None = 1
+    """Number of echoes in a multi-echo acquisition."""
+
+    seq_type: str = UNKNOWN
```
fzimmermann89: Is this helpful?
Also, please write out the names.

src/mrpro/data/IHeader.py
```diff
@@ -92,35 +162,93 @@ def make_unique_tensor(values: Sequence[float]) -> torch.Tensor | None:
...
+            else:
+                raise ValueError(f'Inconsistent entries for {dicom_tag} found.')
+
+        @overload
+        def ensure_unique_number(values: Sequence[int], dicom_tag: TagType) -> int: ...
+
+        @overload
+        def ensure_unique_number(values: Sequence[float], dicom_tag: TagType) -> float: ...
+
+        def ensure_unique_number(values: Sequence[float | int], dicom_tag: TagType) -> float | int:
```
fzimmermann89: You could use a type var here instead of the overloads

fzimmermann89: from typing import TypeVar

T = TypeVar('T', int, float)

def ensure_unique_number(values: Sequence[T], dicom_tag: TagType) -> T:
    """Ensure there is only one unique number."""
    if len(set(values)) == 1:
        return values[0]
    else:
        raise ValueError(f'Inconsistent entries for {dicom_tag} found.'

Or even just ensure_unique without bounds on T:

T = TypeVar('T')
def ensure_unique(values: Sequence[T], dicom_tag: TagType) -> T:
    """Ensure there is only one unique element"""
    if len(set(values)) == 1:
        return values[0]
    else:
        raise ValueError(f'Inconsistent entries for {dicom_tag} found.'


The typevar line could be removed if we limit ourselves to python 3.12, but is required for 3.11 compatibility.

fzimmermann89: This way,, for a  sequence of Str, the function will return a string, for floats a float, etc.

src/mrpro/data/IHeader.py
ckolbPTB: remove also from KHeader

src/mrpro/data/IHeader.py
ckolbPTB: h1_frequency and remove b0

src/mrpro/data/IHeader.py
ckolbPTB: combine string, floats and int and use `if string return UKNOWN else return None`

ckolbPTB:
Add ImageIdx from AcqIdx

fzimmermann89:
Related #591

fzimmermann89:
I think this might be superseeded by #591 , if not,  feel free to reopen.

PR #424, opened 2024-10-01
Gitignore common data file types

ckolbPTB:


fzimmermann89:
Dat, Pth, mp4, PNG, pdf maybe?

fzimmermann89 (approved):

fzimmermann89:
Sorry, I forgot: Nifti and dicom maybe?

ckolbPTB:
one other option would be to ignore everything and only add exceptions for file endings which we need...

PR #425, opened 2024-10-01
Adapt KHeader parameters

ckolbPTB:


src/mrpro/data/AcqInfo.py
```diff
@@ -236,6 +219,23 @@ def spatialdimension_2d(
             user7=tensor(idx['user'][:, 7]),
         )
 
+        # Calculate orientation as rotation matrix from directional cosines
+        def orientation_from_directional_cosine(
```
ckolbPTB: Not sure if this is correct, please double check

src/mrpro/data/TrajectoryDescription.py
```diff
@@ -1,29 +0,0 @@
-"""TrajectoryDescription dataclass."""
```
ckolbPTB: I removed this because we are not using it anywhere and I do not see much use in supporting that

ckolbPTB:
One problem: `acq_info.orientation: Rotation` is not modified by `modify_acq_info`. Adapting `modify_acq_info` to adapt `acq_info.orientation: Rotation` would lead to a circular import...

maybe easier to keep the orientation as read_dir, phase_dir and slice_dir?

fzimmermann89:
Suggestion:
Rename cmodify_acq_info to ._apply
and move it to the acq info class.
It does pretty much the same as the _apply on a torch.module - apply a function to the tensor fields.
We could,to be more consistent,also name it _apply_ as it does in place modification.

Soon(tm) we will most likely need this kind of apply also on the other objects for the fancy indexing..

fzimmermann89:
Alternatively (or additionally) the modify acq info should be changed (dependence injections):
Move the is dataclass condition up.
Do the other path on else (the previous tensor path)

Now, in the modify functions you can do a match case on what kind of object you get.
If it is something you do not want to change, do nothing.
For most modifications, rotation should behave the same as a tensor -  and you can just index into it in the modify function.

fzimmermann89:
(Ps: sorry for the spelling, hope you can guess what I try to say. I can only communicate using  my phone and one hand at most, as my main job still seems to be a heating pad.. )

src/mrpro/data/_kdata/KDataSelectMixin.py
```diff
@@ -49,10 +52,13 @@ def select_other_subset(
         other_idx = torch.cat([torch.where(idx == label_idx[:, 0, 0])[0] for idx in subset_idx], dim=0)
 
         # Adapt header
-        def select_acq_info(info: torch.Tensor):
-            return info[other_idx, ...]
+        def select_acq_info(field: T) -> T:
```
ckolbPTB: adapt after #417 merge

src/mrpro/data/_kdata/KDataSplitMixin.py
```diff
@@ -94,10 +99,17 @@ def split_acq_info(acq_info: torch.Tensor) -> torch.Tensor:
         kheader = copy.deepcopy(self.header)
 
         # Update shape of acquisition info index
-        def reshape_acq_info(info: torch.Tensor):
-            return rearrange(split_acq_info(info), rearrange_pattern_acq_info)
-
-        kheader.acq_info = modify_acq_info(reshape_acq_info, kheader.acq_info)
+        def split_acq_info(field: T) -> T:
```
ckolbPTB: adapt after #417 merge

ckolbPTB:
The _apply_ function works nicely. I moved it to DRAFT because we should first merge  #417 and then we can remove the special case for SpatialDimension in KDataSplitMixin and KDataSelectMixin

ckolbPTB:
Something is not right with the orientation. 

Siemens saves the orientation as a quaternion. This is then translated to read_dir, slice_dir and phase_dir in siemens_to_ismrmrd here:
https://github.com/ismrmrd/siemens_to_ismrmrd/blob/cda30e2f4fbace05dce666a781e20a11eb28c7e6/main.cpp#L1139C14-L1139C46
using
https://github.com/ismrmrd/ismrmrd/blob/4037c41fcdcb8870234baaeb46ce4848aac1579a/libsrc/ismrmrd.c#L667

To me it seems that the signs are not necessarily correct or I am misunderstanding what these directions are supposed to represent.

src/mrpro/data/TrajectoryDescription.py
fzimmermann89: Isn't this one of the things 8.14 would use?
I thought they store something in these user values?

fzimmermann89: I will ask max next week

src/mrpro/data/AcqInfo.py
fzimmermann89: Isn't this transposed?

Applying the orientation to a 1,0,0 vector should give the slice_dir.
So slice dir hascto be the first column.

fzimmermann89:
> Something is not right with the orientation. 
> 
> Siemens saves the orientation as a quaternion. This is then translated to read_dir, slice_dir and phase_dir in siemens_to_ismrmrd here:
> https://github.com/ismrmrd/siemens_to_ismrmrd/blob/cda30e2f4fbace05dce666a781e20a11eb28c7e6/main.cpp#L1139C14-L1139C46
> using
> https://github.com/ismrmrd/ismrmrd/blob/4037c41fcdcb8870234baaeb46ce4848aac1579a/libsrc/ismrmrd.c#L667
> 
> To me it seems that the signs are not necessarily correct or I am misunderstanding what these directions are supposed to represent. 

How and why did you notice? Where is the issue in your opinion? 

I faintly remember that there was something stupid going on either in ismrmd or in the Siemens convection regarding a sign of a single direction..

ckolbPTB:
> How and why did you notice? Where is the issue in your opinion?

I noticed that when I do ismrmrd-slice-phase-readout-dir -> mrpro quaternion -> ismrmrd-slice-phase-readout-dir then I don't get the same result as I put in. 

Siemens is a quaternion which corresponds to a rotation matrix with columns phase - readout - slice encoding. If I take that into consideration (latest commit) I get the same out as I put in from the ismrmrd file. But I guess this still does not fit to our slice - phase - readout convention...

src/mrpro/data/TrajectoryDescription.py
ckolbPTB: any update from Max?

src/mrpro/data/AcqInfo.py
ckolbPTB: should be fixed now

fzimmermann89: I think I added a function to rotation to create it from spatial dimensions...

src/mrpro/data/AcqInfo.py
```diff
@@ -251,17 +270,37 @@ def spatialdimension_2d(
             flags=tensor_2d(headers['flags']),
             measurement_uid=tensor_2d(headers['measurement_uid']),
             number_of_samples=tensor_2d(headers['number_of_samples']),
+            orientation=orientation_from_directional_cosine(
+                spatialdimension_2d(headers['slice_dir']),
```
fzimmermann89: Suggestion: move the apply of conversion to here 
Such that SpatialDimension2d only creates a 2d SpatialDimension,  and the conversion is done explicitly in each line. Makes it easier to see what is happening

fzimmermann89: I can't comment on the SpatialDimension2d code where it should be changed..

src/mrpro/data/AcqInfo.py
ckolbPTB: very handy

fzimmermann89:
Maybe we can discuss UserParameters again I we make `misc` more private and unsupported

src/mrpro/data/AcqInfo.py
```diff
@@ -251,17 +249,37 @@ def spatialdimension_2d(
...
             scan_counter=tensor_2d(headers['scan_counter']),
-            slice_dir=spatialdimension_2d(headers['slice_dir']),
             trajectory_dimensions=tensor_2d(headers['trajectory_dimensions']).fill_(3),  # see above
             user_float=tensor_2d(headers['user_float']),
             user_int=tensor_2d(headers['user_int']),
             version=tensor_2d(headers['version']),
         )
         return acq_info
+
+    def _apply_(self, modify_acq_info_field: Callable) -> None:
```
fzimmermann89: I think it would make sense it instead use the apply_ from the movedatamixin, as it should do the same, right?

ckolbPTB: yes, but I think we need #489 first, don't we?

fzimmermann89: Yes.. I wrote that in slack.. give me a bit of time to order things tonight and then merge them in a sane order

fzimmermann89: I am currently figuring out for me what a good order could be

fzimmermann89:
I f'ed this up. sorry.

@ckolbPTB can you create a new PR from the unit_conversion_utils branch to main?

fzimmermann89:
test

fzimmermann89:
New PR: #506

PR #414, opened 2024-09-26
Add  separable sum of proximable functions

fzimmermann89:
Closes #308

PR #426, opened 2024-10-02
Add Primal Dual Hybrid Gradient algorithm

fzimmermann89:
PDHG algorithm,
needs #432 #433 #431 #414


fzimmermann89:
After merging the other PRs, this should be rebased.

src/mrpro/algorithms/optimizers/pdhg.py
```diff
@@ -0,0 +1,178 @@
...
+        elif primal_stepsize is None:
+            primal_stepsize_ = 1 / (operator_norm * dual_stepsize)
+        elif dual_stepsize is None:
+            dual_stepsize_ = 1 / (operator_norm * primal_stepsize)
+    else:
+        primal_stepsize_ = primal_stepsize
+        dual_stepsize_ = dual_stepsize
+
+    primals_relaxed = initial_values if initial_relaxed is None else initial_relaxed
+    duals = (0 * operator_matrix)(initial_values) if initial_duals is None else initial_duals
```
fzimmermann89:
```suggestion
    duals = (0 * operator_matrix)(*initial_values) if initial_duals is None else initial_duals
```

fzimmermann89:
Nice tests!


Still (maybe) required:
- Test of callback -- maybe combine it with a check if n_iterations is used? (similar to what @koflera did for cg)
- Test of only supplying one of the stepsizes
- Test of the errors

- mentioned that operator norm is expensive to calculate in docstring
- consider #545 : can the "stepsize" be a tensor?
- should we add a stopping criterion besides the number of iterations?

fzimmermann89:
@koflera did you check if the algorithm implemented is pdhg? like if there are any logical / algorithm mistakes...

koflera:
> Nice tests!
> 
> Still (maybe) required:
> 
>     * Test of callback -- maybe combine it with a check if n_iterations is used? (similar to what @koflera did for cg)
> 
>     * Test of only supplying one of the stepsizes
> 
>     * Test of the errors
> 
>     * mentioned that operator norm is expensive to calculate in docstring
> 
>     * consider [k-space preconditioning for pdhg #545](https://github.com/PTB-MR/mrpro/issues/545) : can the "stepsize" be a tensor?
> 
>     * should we add a stopping criterion besides the number of iterations?

yes, I was hoping to be able to do a few more today... still work in progress ;)

one thing that I noticed is that the operator norm is calculated using the `*initial_values` as starting point. However, for the calculation of the operator norm, one would typically take a random vector. For PDHG, as far as I know, one would most probably on the other hand take an initial guess of the solution or a zero-vector (which on the other hand, is not allowed for the power iteration, cause it causes division by zero). So the question is how to handle this case... as far as I remember, you had a strong opinion against using a random vector as initialization for the power method if one provides a zero-vector. how should we handle it here, i.e. when one wants to use zero as initialization for the pdhg?

koflera:
> @koflera did you check if the algorithm implemented is pdhg? like if there are any logical / algorithm mistakes...

yes, as far as I can tell, the algorithm seems to be correct. Also, I guess the fact that the first 4 tests pass makes it unlikely that it is wrong ;)

fzimmermann89:
> how should we handle it here, i.e. when one wants to use zero as initialization for the pdhg?

Good catch!
I still think it would be surprising behaviour to use a random vector in operator_norm.
But for pdhg I would be completely fine with using a random vector (always using a random vector?) 
I just used the pdhg initial_value because it has the right shape and the same name ...

koflera:
> > how should we handle it here, i.e. when one wants to use zero as initialization for the pdhg?
> 
> Good catch! I still think it would be surprising behaviour to use a random vector in operator_norm. But for pdhg I would be completely fine with using a random vector (always using a random vector?) I just used the pdhg initial_value because it has the right shape and the same name ...

I agree with the random behavior. But nevertheless, a random vector is what one is probably encouraged to use anyways as starting point for the power iteration. For PDHG on the other hand, zero seems the much better choice than a random vector. I think in general, there are starting points that one should avoid to have a slow convergence for the power method. But those have to be explicitly constructed. As far as I know, using random initial values is the most straight forward way to avoid those. Thus, I would be more in favour of having random values for the power method and zeros as default initial value for PDHG.

fzimmermann89:
no, you missunderstood me:

Suggestion: keep operator_norm as is. it is user error to supply zeros.
But in phdg, don't use initial values for the operator norm, but random values like operator norm! 
of course, still use the initial_values for the pdhg iterations.

I think we agree, but misunderstand

koflera:
> no, you missunderstood me:
> 
> Suggestion: keep operator_norm as is. it is user error to supply zeros. But in phdg, don't use initial values for the operator norm, but random values like operator norm! of course, still use the initial_values for the pdhg iterations.
> 
> I think we agree, but misunderstand

ok, yes... we agree :)

koflera:
> * should we add a stopping criterion besides the number of iterations?

Regarding that, I thought it could be useful to have the primal dual gap. However, for that we would also need the convex conjugate of f and g, respectively, while we currently only have the prox of the convex conjugate. Until then, I think the best is to be able to track the value of the primal functional and to stopp the iterations do not change that value too much anymore. What do you think?

fzimmermann89:
Relative Change of the of the solution sounds reasonable?

We don't actually calculate the full objective within pdhg, so it would almost double the calculations to track it..


That's the reason why the objective is a callable in the callback -- when one wants to know it in the callback it must be calculated.

examples/tv_minimization_reconstruction_pdhg.py
```diff
@@ -0,0 +1,190 @@
...
+# where $\mathrm{prox}$ denotes the proximal operator and $f^{\ast}$ denotes the convex conjugate of the
+# functional $f$, $\theta\in [0,1]$ and step sizes $\sigma, \tau$ such that $\sigma \tau < 1/L^2$, where
+# $L=\|K\|_2$ is the operator norm of the operator $K$.
+#
+# The first step is to recast problem (1) into the general form of (2) and then to apply the steps above
+# in an iterative fashion.
+#
+# A possible and intuitive (but unfortunately not efficient) identification is the following
+#
+# $f(x) = \lambda \| x\|_1,$
```
fzimmermann89: I would remove this.

It goes beyond the scope of an example ways how not to do it.

examples/tv_minimization_reconstruction_pdhg.py
```diff
@@ -0,0 +1,190 @@
...
+
+# %%
+import mrpro
+
+# %% [markdown]
+# ##### Read-in the raw data
+
+# %%
+# Use the trajectory that is stored in the ISMRMRD file
+trajectory = mrpro.data.traj_calculators.KTrajectoryIsmrmrd()
```
fzimmermann89: I would not save this in a variable named trajectory. consider not saving it as a variable and only giving it as an input to from_file.

What you have there is not the trajectory, it only tells from_file how to get the trajectory.

examples/tv_minimization_reconstruction_pdhg.py
```diff
@@ -0,0 +1,190 @@
...
+
+# %% [markdown]
+# ### Set up the operator $A$
+# Estimate coil sensitivity maps and density compensation function. Also run a direct (adjoint)
+# reconstruction and iterative SENSE as methods of comparison.
+
+# %%
+# Set up direct reconstruction class to estimate coil sensitivity maps, density compensation etc
+direct_reconstruction = mrpro.algorithms.reconstruction.DirectReconstruction(kdata)
+img_direct = direct_reconstruction(kdata).data
```
fzimmermann89: i would keeo it as idata, not taking .data

examples/tv_minimization_reconstruction_pdhg.py
```diff
@@ -0,0 +1,190 @@
...
+
+# run tierative SENSE to get an initiual guess of the solution
+iterative_sense_reconstruction = mrpro.algorithms.reconstruction.IterativeSENSEReconstruction(
+    kdata, n_iterations=8, csm=direct_reconstruction.csm
+)
+img_iterative_sense = iterative_sense_reconstruction(kdata).data
```
fzimmermann89: not .data

fzimmermann89: only when you want to plot it your input it into pdgh, do .data

examples/tv_minimization_reconstruction_pdhg.py
```diff
@@ -0,0 +1,190 @@
...
+
+# Define Fourier operator using the trajectory and header information in kdata
+fourier_operator = mrpro.operators.FourierOp.from_kdata(kdata)
+
+# estimate density compensation
+dcf_operator = mrpro.data.DcfData.from_traj_voronoi(kdata.traj).as_operator()
```
fzimmermann89: you could also obtain them from iterative_sense_reconstruction

examples/tv_minimization_reconstruction_pdhg.py
```diff
@@ -0,0 +1,190 @@
...
+
+# Calculate coil maps
+img_coilwise = mrpro.data.IData.from_tensor_and_kheader(*fourier_operator.H(*dcf_operator(kdata.data)), kdata.header)
+csm_operator = mrpro.data.CsmData.from_idata_walsh(img_coilwise).as_operator()
```
fzimmermann89: you could also obtain them from iterative_sense_reconstruction

fzimmermann89: this is not really relevant to this examole

fzimmermann89: this, you could ave on the dcf and csm calculation by getting the csms from iterative sense.

and say "Here, we reuse the csms obtained using Walsh's method inside the iterative SENSE reconstruction"

examples/tv_minimization_reconstruction_pdhg.py
```diff
@@ -0,0 +1,190 @@
...
+# Set up the problem by using the previously described identification
+from mrpro.algorithms.optimizers import pdhg
+from mrpro.operators import LinearOperatorMatrix, ProximableFunctionalSeparableSum
+from mrpro.operators.functionals import L1NormViewAsReal, L2NormSquared, ZeroFunctional
+
+# Regularization parameter for the $\ell_1$-norm
+regularization_parameter = 0.5
+
+# Define the functionals
+kdata_tensor = kdata.data
```
fzimmermann89: i would also remove this varaible and just use target=kdata.data

examples/tv_minimization_reconstruction_pdhg.py
```diff
@@ -0,0 +1,190 @@
...
+l2 = 0.5 * L2NormSquared(target=kdata_tensor, divide_by_n=True)
```
fzimmermann89: you could use MSE here

examples/tv_minimization_reconstruction_pdhg.py
```diff
@@ -0,0 +1,190 @@
...
+l1 = regularization_parameter * L1NormViewAsReal(divide_by_n=True)
+
+# Define the functionals f and g and the operator K
```
fzimmermann89: Either make it headins or consider more helpful comments, describing why you do something. Not what.
it is quite obvious that you define f,g, and operator.

examples/tv_minimization_reconstruction_pdhg.py
```diff
@@ -0,0 +1,190 @@
...
+f = ProximableFunctionalSeparableSum(l2, l1)
+g = ZeroFunctional()
+operator = LinearOperatorMatrix(((acquisition_operator,), (nabla_operator,)))
```
fzimmermann89: consider K (or change thename in the description)

examples/tv_minimization_reconstruction_pdhg.py
```diff
@@ -0,0 +1,190 @@
...
+
+# %%
+# ### Compare the results
+# Display the reconstructed images
+import matplotlib.pyplot as plt
+
+fig, ax = plt.subplots(1, 3, squeeze=False)
+clim = [0, 1e-3]
+ax[0, 0].set_title('Adjoint Recon', fontsize=10)
+ax[0, 0].imshow(img_direct.abs()[0, 0, 0, :, :], clim=clim)
```
fzimmermann89: if you dont do .data before, you can also just use img_direct.rss() to obtain the abs directly from IData

tests/algorithms/test_pdhg.py
```diff
@@ -0,0 +1,327 @@
...
+
+    data_shape = (32, 32)
+    data = random_generator.float32_tensor(size=data_shape)
+
+    regularization_parameter = 0.1
+
+    l2 = 0.5 * L2NormSquared(target=data, divide_by_n=False)
+    l1 = regularization_parameter * L1Norm(divide_by_n=False)
+
+    f = l2
```
fzimmermann89, line 30: why?

fzimmermann89: just use f=l2, g=l1 etc in line 38

tests/algorithms/test_pdhg.py
```diff
@@ -0,0 +1,327 @@
...
+
+
+def test_fourier_l2_l1_():
+    """Set up the problem min_x 1/2*|| Fx - y||_2^2 + lambda * ||x||_1,
+    where F is the full FFT sampled on a Cartesian grid. Thus, again, the
+    problem has a closed-form solution given by soft-thresholding.
+    """
+    random_generator = RandomGenerator(seed=0)
+
+    image_shape = (32, 48, 48)
```
fzimmermann89, line 82: does this have to be 3d?

I haven't run it, is it fast? or can we reduce it to 2d?

fzimmermann89: (same questions applies to all 3d examples)

fzimmermann89 (commented):
some comments

examples/tv_minimization_reconstruction_pdhg.py
koflera: I would like to keep it... it's just a few lines and compared to the space it takes, it's highly instructional :)

The first time I wanted to use pdhg for TV, I remember being quite puzzled by why one would not use the first identification.

examples/tv_minimization_reconstruction_pdhg.py
koflera: good points... actually I had first done something like this. But then, I was unable to make mypy stop complaining. I'll give it another try

examples/tv_minimization_reconstruction_pdhg.py
fzimmermann89: Can we maybe move it down into its own section?

I would like to reduce the wall of text before the first code.

Also, for the identification we use, could you conaidee interleaving the explanation with the code a bit more?

I think this could make it more educational

examples/tv_minimization_reconstruction_pdhg.py
koflera: true, but I think I would like to have it like this such that it is consistent with L1, instead of having a mixture.

examples/tv_minimization_reconstruction_pdhg.py
koflera: but then they differ, no? as far as I understand it, direct reconstruction is x_direct = A^H y (i.e. it depends on the csms), while img_direct.rss() should be (at least from the name) the RSS-recon?

tests/algorithms/test_pdhg.py
koflera: yes, everything is fast... the idea was just to have a few different dimensionalities

examples/tv_minimization_reconstruction_pdhg.py
fzimmermann89: .rss on IData is the root sum of squares of the images in IData.

If you perform a CSM-aware reconstruction, you only have one complex valued image. thus RSS is the root of square, the abs.
.rss always returns a magnitude tensor ready to plot.

examples/tv_minimization_reconstruction_pdhg.py
```diff
@@ -0,0 +1,184 @@
...
+#
+# $g(x) = \frac{1}{2}\|Ax  - y\|_2^2,$
+#
+# $K(x) = \nabla x.$
+#
+# However, although $\mathrm{prox}_{\sigma f^\ast}$ has a simple form, calculations show that
+# to be able to compute $\mathrm{prox}_{\tau g}$, one would need to solve a linear system at each
+# iteration. We will therefore use a way more efficient identification.
+#
+# In the following, we load some 2D radial MR data and PDHG to reconstruct it.
```
ckolbPTB:
```suggestion
# In the following, we load 2D radial MR data and set-up PDHG to reconstruct it.
```

examples/tv_minimization_reconstruction_pdhg.py
```diff
@@ -0,0 +1,184 @@
...
+# ### Set up the operator $A$
+# Estimate coil sensitivity maps and density compensation function. Also run a direct (adjoint)
+# reconstruction and iterative SENSE as methods of comparison.
+
+# %%
+# Set up direct reconstruction class to estimate coil sensitivity maps, density compensation etc
+direct_reconstruction = mrpro.algorithms.reconstruction.DirectReconstruction(kdata)
+img_direct = direct_reconstruction(kdata)
+
+# run tierative SENSE to get an initiual guess of the solution
```
ckolbPTB:
```suggestion
# run iterative SENSE to get an initiual guess of the solution
```

examples/tv_minimization_reconstruction_pdhg.py
```diff
@@ -0,0 +1,184 @@
...
+kdata.header.recon_matrix.x = 256
+kdata.header.recon_matrix.y = 256
+
+# %% [markdown]
+# ### Set up the operator $A$
+# Estimate coil sensitivity maps and density compensation function. Also run a direct (adjoint)
+# reconstruction and iterative SENSE as methods of comparison.
+
+# %%
+# Set up direct reconstruction class to estimate coil sensitivity maps, density compensation etc
```
ckolbPTB: We don't need the direct reconstruction to get the dcf and csm but I guess it makes sense to reuse it for later to save time. Maybe reformulate to
```suggestion
# Set up direct reconstruction class. The estimate coil sensitivity maps and density compensation values can be reused later to save time.
```

examples/tv_minimization_reconstruction_pdhg.py
```diff
@@ -0,0 +1,184 @@
...
+# run tierative SENSE to get an initiual guess of the solution
+iterative_sense_reconstruction = mrpro.algorithms.reconstruction.IterativeSENSEReconstruction(
+    kdata,
+    n_iterations=8,
+    csm=direct_reconstruction.csm,
+    dcf=direct_reconstruction.dcf,
+)
+img_iterative_sense = iterative_sense_reconstruction(kdata)
+
+# Define Fourier operator and csm-operator re-using the ones already constructed in the
```
ckolbPTB:
```suggestion
# Define Fourier operator and CSM operator re-using the ones already constructed in the
```

examples/tv_minimization_reconstruction_pdhg.py
```diff
@@ -0,0 +1,184 @@
...
+# direct reconstruction class
+fourier_operator = direct_reconstruction.fourier_op
+assert direct_reconstruction.csm is not None
+csm_operator = direct_reconstruction.csm.as_operator()
+
+# Create the entire acquisition operator A to be used in the operator K=[A, \nabla]^T for PDHG
+acquisition_operator = fourier_operator @ csm_operator
+
+# adjoint reconstruction as additional method of comparison
+(img_adjoint,) = acquisition_operator.H(kdata.data)
```
ckolbPTB: Do we need this? Should be the same as img_direct

examples/tv_minimization_reconstruction_pdhg.py
```diff
@@ -0,0 +1,184 @@
...
+)
+img_iterative_sense = iterative_sense_reconstruction(kdata)
+
+# Define Fourier operator and csm-operator re-using the ones already constructed in the
+# direct reconstruction class
+fourier_operator = direct_reconstruction.fourier_op
+assert direct_reconstruction.csm is not None
+csm_operator = direct_reconstruction.csm.as_operator()
+
+# Create the entire acquisition operator A to be used in the operator K=[A, \nabla]^T for PDHG
```
ckolbPTB: I would add the "efficient identification" already at the top, otherwise this operator K appears very suddenly and not clear why we need it.

examples/tv_minimization_reconstruction_pdhg.py
```diff
@@ -0,0 +1,184 @@
...
+acquisition_operator = fourier_operator @ csm_operator
+
+# adjoint reconstruction as additional method of comparison
+(img_adjoint,) = acquisition_operator.H(kdata.data)
+
+
+# %% [markdown]
+# ### Recast the problem to be able to apply PDHG
```
ckolbPTB: See comment above - I would move this to the intro

examples/tv_minimization_reconstruction_pdhg.py
```diff
@@ -0,0 +1,184 @@
...
+# for which one can show that both $\mathrm{prox}_{\sigma f^{\ast}}$ and $\mathrm{prox}_{\tau g}$ are
+# given by simple and easy-to-compute operations, see for example [Sidky et al, PMB 2012].
+#
+# In the following, we first identify the functionals $f$, $g$ and the operator $K$ and then run PDHG.
+
+# %%
+# Define the gradient operator \nabla to be used in the operator K=[A, \nabla]^T for PDHG
+from mrpro.operators import FiniteDifferenceOp
+
+# The operator computes the directional derivatives along the penultimate and last dimensions (x,y)
```
ckolbPTB:
```suggestion
# The operator computes the directional derivatives along the the last two dimensions (x,y)
```

examples/tv_minimization_reconstruction_pdhg.py
```diff
@@ -0,0 +1,184 @@
...
+
+# initialize PDHG with iterative SENSE solution for warm start
+initial_values = (img_iterative_sense.data,)
+
+# %% [markdown]
+# ### Run PDHG for a certain number of iterations
+
+# %%
+max_iterations = 64
+(img_pdhg,) = pdhg(f=f, g=g, operator=K, initial_values=initial_values, max_iterations=max_iterations)
```
ckolbPTB: Can we also use the OptimizerStatus here to showcase how one would use it?

src/mrpro/algorithms/optimizers/pdhg.py
```diff
@@ -0,0 +1,204 @@
...
+    r"""Primal-Dual Hybrid Gradient Algorithm (PDHG).
+
+    Solves the minimization problem
+        :math:`\min_x g(x) + f(A x)`
+    with linear operator A and proximable functionals f and g.
+
+    The operator is supplied as a matrix (tuple of tuples) of linear operators,
+    f and g are supplied as tuples of proximable functionals interpreted as separable sums.
+
+    Thus, problem solved is
```
ckolbPTB:
```suggestion
    Thus, the problem to be solved is
```

src/mrpro/algorithms/optimizers/pdhg.py
```diff
@@ -0,0 +1,204 @@
...
+            :math:`\min_x \sum_i,j g_j(x_j) + f_i(A_ij x_j)`.
+
+    If neither primal nor dual step size are not supplied, they are chose as :math:`1/||A||_{op}`.
```
ckolbPTB:
```suggestion
    If neither primal nor dual step size are supplied, they are chosen as :math:`1/||A||_{op}`.
```

src/mrpro/algorithms/optimizers/pdhg.py
```diff
@@ -0,0 +1,204 @@
...
+    If either is supplied, the other is chosen such that
```
ckolbPTB:
```suggestion
    If one of them is supplied, the other is chosen such that
```

src/mrpro/algorithms/optimizers/pdhg.py
```diff
@@ -0,0 +1,204 @@
...
+    initial_duals: Sequence[torch.Tensor] | None = None,
+    callback: Callable[[PDHGStatus], None] | None = None,
+) -> tuple[torch.Tensor, ...]:
+    r"""Primal-Dual Hybrid Gradient Algorithm (PDHG).
+
+    Solves the minimization problem
+        :math:`\min_x g(x) + f(A x)`
+    with linear operator A and proximable functionals f and g.
+
+    The operator is supplied as a matrix (tuple of tuples) of linear operators,
```
ckolbPTB: The operator can also be supplied as a LinearOperator

src/mrpro/algorithms/optimizers/pdhg.py
```diff
@@ -0,0 +1,204 @@
...
+    primal_stepsize: float | None = None,
+    dual_stepsize: float | None = None,
+    relaxation: float = 1.0,
+    initial_relaxed: Sequence[torch.Tensor] | None = None,
+    initial_duals: Sequence[torch.Tensor] | None = None,
+    callback: Callable[[PDHGStatus], None] | None = None,
+) -> tuple[torch.Tensor, ...]:
+    r"""Primal-Dual Hybrid Gradient Algorithm (PDHG).
+
+    Solves the minimization problem
```
ckolbPTB, line 55: Can we add some more math here and then also add the math description of the steps of the iteration further below in the code? Similar to the example script?

tests/algorithms/test_pdhg.py
```diff
@@ -0,0 +1,327 @@
...
+    expected = torch.nn.functional.softshrink(data, regularization_parameter)
+
+    max_iterations = 128
+    (pdhg_solution,) = pdhg(f=f, g=g, operator=operator, initial_values=initial_values, max_iterations=max_iterations)
+    torch.testing.assert_close(pdhg_solution, expected, rtol=5e-4, atol=5e-4)
+
+
+def test_fourier_l2_l1_():
+    """Set up the problem min_x 1/2*|| Fx - y||_2^2 + lambda * ||x||_1,
+    where F is the full FFT sampled on a Cartesian grid. Thus, again, the
```
ckolbPTB:
```suggestion
    where F is the full FFT and y is sampled on a Cartesian grid. Thus, again, the
```

tests/algorithms/test_pdhg.py
```diff
@@ -0,0 +1,327 @@
...
+
+def test_callback():
+    """Check that the callback function is called."""
+    random_generator = RandomGenerator(seed=0)
+    f = ZeroFunctional()
+    g = None
+    operator = None
+    initial_values = (random_generator.complex64_tensor(size=(8,)),)
+
+    # callback function; should not be called since PDHG should exit
```
ckolbPTB: should or should not be called?

tests/algorithms/test_pdhg.py
```diff
@@ -0,0 +1,327 @@
...
+    l2 = 0.5 * L2NormSquared(target=data, divide_by_n=False)
+    l1 = regularization_parameter * L1Norm(divide_by_n=False)
+
+    f = ProximableFunctionalSeparableSum(l2, l1)
+    g = None  # corresponds to ZeroFunctional()
+    operator = LinearOperatorMatrix(((IdentityOp(),), (IdentityOp(),)))
+
+    expected = torch.nn.functional.softshrink(data, regularization_parameter)
+
+    # callback function that should not be called since pdhg should exit the for-loop
```
ckolbPTB, line 316:
```suggestion
    # callback function that should not be called since pdhg should exit the for-loop beforehand
```

tests/algorithms/test_pdhg.py
```diff
@@ -0,0 +1,327 @@
...
+    """Check that the callback function is called."""
+    random_generator = RandomGenerator(seed=0)
+    f = ZeroFunctional()
+    g = None
+    operator = None
+    initial_values = (random_generator.complex64_tensor(size=(8,)),)
+
+    # callback function; should not be called since PDHG should exit
+    def callback(solution):
+        assert True
```
fzimmermann89: This does nothing

Consider setting as
Callnack_was_called variable to true in the callback and assert this variable then after running pdhg.

examples/tv_minimization_reconstruction_pdhg.py
koflera: I used to have it like this, but Felix' suggestion was to interleave the text a bit with some code in order to not to have a first very long explanation and all the code afterwards. @fzimmermann89, @ckolbPTB: please come to an agreement :)

examples/tv_minimization_reconstruction_pdhg.py
koflera: same as above... I am fine with either with maybe a slight preference of having first the text and then all the code

examples/tv_minimization_reconstruction_pdhg.py
fzimmermann89: In your first draft, you had all explanation at the top, including what we don't do, followed by some code.

I still stand by the point that it is from an education point better to interleave code with explanations.
But of course, anything that is used in the code must be explained before hand -- but it doesn't have to be at the beginning of the file, just before it is used.

koflera: ok, I think I understood now the point Christoph wanted to make. I removed the operator K=[A, \nabla]^T in the comment. Then, everything should be consistent.

ckolbPTB (approved):

src/mrpro/algorithms/optimizers/pdhg.py
```diff
@@ -0,0 +1,217 @@
...
+    :math:`z_{k+1} = \mathrm{prox}_{\sigma f^{\ast}}(z_k + \sigma K \bar{x}_k),`
+    :math:`x_{k+1} = \mathrm{prox}_{\tau g}(x_k - \tau K^H z_{k+1}),`
+    :math:`\bar{x}_{k+1} = x_{k+1} + \theta(x_{k+1} - x_k),`
+
+    where :math:`\mathrm{prox}` denotes the proximal operator and :math:`f^{\ast}`is the
+    convex conjugate of the functional :math:`f`. Thereby, :math:`\tau` and :math:`\sigma`
+    are the primal and dual step sizes, respectively (see further below) and :math:`\theta\in [0,1]`.
+
+    The operator is supplied as a LinearOperator or as a matrix (tuple of tuples) of linear operators,
+    f and g are supplied as tuples of proximable functionals interpreted as separable sums.
```
fzimmermann89: This is outdated

src/mrpro/algorithms/optimizers/pdhg.py
```diff
@@ -73,16 +73,16 @@ def pdhg(
     as separable sums.
 
     Thus, the problem to be solved is
-            :math:`\min_{x=(x_1,\ldots,x_n)} \sum_{i=1}^m \sum_{j=1}^n g_j(x_j) + f_i(K_{i,j} x_{j})`.
+            :math:`\min_{x=(x_1,\ldots,x_n)} \sum_{i=1}^m \sum_{j=1}^n g_j(x_j) + f_i\big( (Kx)_j\big)`.
```
fzimmermann89: `\min_{x=(x_1,\ldots,x_n)}  \sum_{j=1}^n g_j(x_j) + \sum_{i=1}^m  f_i( \sum_{j=1}^n K_{i,j} x_{j})`
?

ckolbPTB:
Would it make sense to use undersampled data in the example such that there is actually a difference between iterative SENSE and TV-reg recon?

ckolbPTB:
and I would not fix clim to the same value for the adjoint reco

fzimmermann89:
> and I would not fix clim to the same value for the adjoint reco

Maybe not for now, but I want to fix the scaling such  that these two actually have the same scale..

fzimmermann89:
> Would it make sense to use undersampled data in the example such that there is actually a difference between iterative SENSE and TV-reg recon?

Maybe we can also put an under sampled dataset in zenoodo and use that for the examples?

@schuenke or @ckolbPT,  could you do that?

The retrospective undersampling is only noise for understanding the example, imo.
We could just use undersampled data here and in iterative sense.
And, _maybe_, we can continue to use also the fully sampled data for reference - but I don't even think this is really necessary for a good example if it is obvious that TV and iterative sense are more realistic than the direct reco.

koflera:
> > Would it make sense to use undersampled data in the example such that there is actually a difference between iterative SENSE and TV-reg recon?
> 
> Maybe we can also put an under sampled dataset in zenoodo and use that for the examples?
> 
> @schuenke or @ckolbPT, could you do that?
> 
> The retrospective undersampling is only noise for understanding the example, imo. We could just use undersampled data here and in iterative sense. And, _maybe_, we can continue to use also the fully sampled data for reference - but I don't even think this is really necessary for a good example if it is obvious that TV and iterative sense are more realistic than the direct reco.

I also think this would be nice indeed. Otherwise, we could also just add a decent amount of noise to the k-space data... iterative SENSE should then amplify that noise while TV should supress it.

schuenke:
> > Would it make sense to use undersampled data in the example such that there is actually a difference between iterative SENSE and TV-reg recon?
> 
> Maybe we can also put an under sampled dataset in zenoodo and use that for the examples?
> 
> @schuenke or @ckolbPT, could you do that?
> 
> The retrospective undersampling is only noise for understanding the example, imo. We could just use undersampled data here and in iterative sense. And, _maybe_, we can continue to use also the fully sampled data for reference - but I don't even think this is really necessary for a good example if it is obvious that TV and iterative sense are more realistic than the direct reco.

What kind of data would you prefer? Radial 2D? "Fully sampled" to fulfil nyquist would require `pi x N/2` spokes (e.g. 402 for a 256x256 matrix). Which undersampling factor (or hamy many spokes) would you want for the undersampled data set? I should be able to acquire the data next week.

fzimmermann89:
> > > Would it make sense to use undersampled data in the example such that there is actually a difference between iterative SENSE and TV-reg recon?
> > 
> > 
> > Maybe we can also put an under sampled dataset in zenoodo and use that for the examples?
> > @schuenke or @ckolbPT, could you do that?
> > The retrospective undersampling is only noise for understanding the example, imo. We could just use undersampled data here and in iterative sense. And, _maybe_, we can continue to use also the fully sampled data for reference - but I don't even think this is really necessary for a good example if it is obvious that TV and iterative sense are more realistic than the direct reco.
> 
> What kind of data would you prefer? Radial 2D? "Fully sampled" to fulfil nyquist would require `pi x N/2` spokes (e.g. 402 for a 256x256 matrix). Which undersampling factor (or hamy many spokes) would you want for the undersampled data set? I should be able to acquire the data next week.

I was thinking about just using the existing data. creating a new isrmrd file with only some of the acquisitions.
actual prospective undersampling would of course also be nice if you find the time to do it at the scanner.
choose a number of spokes that results in some noticeable artifacts ;)

koflera:
> > > > Would it make sense to use undersampled data in the example such that there is actually a difference between iterative SENSE and TV-reg recon?
> > > 
> > > 
> > > Maybe we can also put an under sampled dataset in zenoodo and use that for the examples?
> > > @schuenke or @ckolbPT, could you do that?
> > > The retrospective undersampling is only noise for understanding the example, imo. We could just use undersampled data here and in iterative sense. And, _maybe_, we can continue to use also the fully sampled data for reference - but I don't even think this is really necessary for a good example if it is obvious that TV and iterative sense are more realistic than the direct reco.
> > 
> > 
> > What kind of data would you prefer? Radial 2D? "Fully sampled" to fulfil nyquist would require `pi x N/2` spokes (e.g. 402 for a 256x256 matrix). Which undersampling factor (or hamy many spokes) would you want for the undersampled data set? I should be able to acquire the data next week.
> 
> I was thinking about just using the existing data. creating a new isrmrd file with only some of the acquisitions. actual prospective undersampling would of course also be nice if you find the time to do it at the scanner. choose a number of spokes that results in some noticeable artifacts ;)

I would also say no necessity to acquire new data (although obviously nice ;) ). From my experience, with 36 spokes (for a 320x320 image) is already quite drastic. So, maybe 48 or so? i.e. ~ undersampling of 8?

schuenke:
> > > > > Would it make sense to use undersampled data in the example such that there is actually a difference between iterative SENSE and TV-reg recon?
> > > > 
> > > > 
> > > > Maybe we can also put an under sampled dataset in zenoodo and use that for the examples?
> > > > @schuenke or @ckolbPT, could you do that?
> > > > The retrospective undersampling is only noise for understanding the example, imo. We could just use undersampled data here and in iterative sense. And, _maybe_, we can continue to use also the fully sampled data for reference - but I don't even think this is really necessary for a good example if it is obvious that TV and iterative sense are more realistic than the direct reco.
> > > 
> > > 
> > > What kind of data would you prefer? Radial 2D? "Fully sampled" to fulfil nyquist would require `pi x N/2` spokes (e.g. 402 for a 256x256 matrix). Which undersampling factor (or hamy many spokes) would you want for the undersampled data set? I should be able to acquire the data next week.
> > 
> > 
> > I was thinking about just using the existing data. creating a new isrmrd file with only some of the acquisitions. actual prospective undersampling would of course also be nice if you find the time to do it at the scanner. choose a number of spokes that results in some noticeable artifacts ;)
> 
> I would also say no necessity to acquire new data (although obviously nice ;) ). From my experience, with 36 spokes (for a 320x320 image) is already quite drastic. So, maybe 48 or so? i.e. ~ undersampling of 8?

The old data was acquired from -127.5 to +127.5 instead of -128 to 127 and is thus not (perfectly) compatible with our radial 2D traj calculator. I will acquire some new data with 402, 96, 48, 36, 24 spokes next Monday before the X-mas party. Let me know if you want other / additional scans.

koflera:
> > > > > > Would it make sense to use undersampled data in the example such that there is actually a difference between iterative SENSE and TV-reg recon?
> > > > > 
> > > > > 
> > > > > Maybe we can also put an under sampled dataset in zenoodo and use that for the examples?
> > > > > @schuenke or @ckolbPT, could you do that?
> > > > > The retrospective undersampling is only noise for understanding the example, imo. We could just use undersampled data here and in iterative sense. And, _maybe_, we can continue to use also the fully sampled data for reference - but I don't even think this is really necessary for a good example if it is obvious that TV and iterative sense are more realistic than the direct reco.
> > > > 
> > > > 
> > > > What kind of data would you prefer? Radial 2D? "Fully sampled" to fulfil nyquist would require `pi x N/2` spokes (e.g. 402 for a 256x256 matrix). Which undersampling factor (or hamy many spokes) would you want for the undersampled data set? I should be able to acquire the data next week.
> > > 
> > > 
> > > I was thinking about just using the existing data. creating a new isrmrd file with only some of the acquisitions. actual prospective undersampling would of course also be nice if you find the time to do it at the scanner. choose a number of spokes that results in some noticeable artifacts ;)
> > 
> > 
> > I would also say no necessity to acquire new data (although obviously nice ;) ). From my experience, with 36 spokes (for a 320x320 image) is already quite drastic. So, maybe 48 or so? i.e. ~ undersampling of 8?
> 
> The old data was acquired from -127.5 to +127.5 instead of -128 to 127 and is thus not (perfectly) compatible with our radial 2D traj calculator. I will acquire some new data with 402, 96, 48, 36, 24 spokes next Monday before the X-mas party. Let me know if you want other / additional scans.

ok, cool! thanks a lot :)

examples/scripts/tv_minimization_reconstruction_pdhg.py
```diff
@@ -0,0 +1,231 @@
+# %% [markdown]
+# # Total-variation (TV)-minimization reconstruction
+
+# %% [markdown]
+# ### Image reconstruction
+# Here, we use the Primal Dual Hybrid Gradient (PDHG) algorithm to reconstruct an image from 2D radial k-space data
+# data.
```
ckolbPTB:
```suggestion
```

examples/scripts/tv_minimization_reconstruction_pdhg.py
```diff
@@ -0,0 +1,231 @@
+# %% [markdown]
+# # Total-variation (TV)-minimization reconstruction
+
+# %% [markdown]
+# ### Image reconstruction
+# Here, we use the Primal Dual Hybrid Gradient (PDHG) algorithm to reconstruct an image from 2D radial k-space data
```
ckolbPTB:
```suggestion
# Here, we use the Primal Dual Hybrid Gradient (PDHG) algorithm to reconstruct an image from 2D radial k-space data.
```

examples/scripts/tv_minimization_reconstruction_pdhg.py
```diff
@@ -0,0 +1,231 @@
...
+# of the 24 spokes data. we use it as a comparison and as an initial guess for the TV-minimization reconstruction.
+# %%
+sense_reconstruction = mrpro.algorithms.reconstruction.IterativeSENSEReconstruction(
+    kdata_24spokes,
+    n_iterations=8,
+    csm=direct_reconstruction_24.csm,
+    dcf=direct_reconstruction_24.dcf,
+)
+img_sense_24 = sense_reconstruction(kdata_24spokes)
+
```
ckolbPTB, line 102:
```suggestion

# %% [markdown]
```

examples/scripts/tv_minimization_reconstruction_pdhg.py
```diff
@@ -0,0 +1,231 @@
...
+# ### Set up the operator $A$
+# Now, to set up the Problem, we need to define the acquisition operator $A$, consisting of a
```
ckolbPTB:
```suggestion
# Now, to set up the problem, we need to define the acquisition operator $A$, consisting of a
```

examples/scripts/tv_minimization_reconstruction_pdhg.py
```diff
@@ -0,0 +1,231 @@
...
+(img_pdhg_24,) = mrpro.algorithms.optimizers.pdhg(
+    f=f,
+    g=g,
+    operator=K,
+    initial_values=(img_sense_24.data,),
+    max_iterations=257,
+    callback=callback,
+)
+
+# %%
```
ckolbPTB:
```suggestion
# %% [markdown]
```

examples/scripts/tv_minimization_reconstruction_pdhg.py
```diff
@@ -0,0 +1,231 @@
...
+    img_sense_24.rss().squeeze(),
+    img_pdhg_24.abs().squeeze(),
+    titles=['402 spokes', '24 spokes (direct)', '24 spokes (SENSE)', '24 spokes (PDHG)'],
+)
+
+# %% [markdown]
+# Hurrah! We have successfully reconstructed an image from 24 spokes using TV-minimization.
+#
+# ### Next steps
+# Play around with the regularization weight and the number of iterations to see how they affect the final image.
```
ckolbPTB, line 232:
```suggestion
# In the above example we used quite a high regularization weight $\lambda$. Play around with the regularization weight and the number of iterations to see how they affect the final image.
```

ckolbPTB (approved):
Just a few small things

fzimmermann89:
Sorry, I started to update the notebook but got stuck at the output of the callback being rendered in different cells..

I will incorporate your suggestions and fix some of the rendering issues.

ckolbPTB:
@fzimmermann89 was merging #612 on purpose? I wanted to try to get the text on top of a hidden code cell (Show code cell content) to show something more meaningful but I did not succeed so I don't think this code contains anything useful...

fzimmermann89:
> @fzimmermann89 was merging #612 on purpose? I wanted to try to get the text on top of a hidden code cell (Show code cell content) to show something more meaningful but I did not succeed so I don't think this code contains anything useful...

My bad. I wanted to merge your changes., but instead merged the PR..

PR #431, opened 2024-10-06
Add ZeroOp as neutral elements of addition, make IdentityOp neutral element of composition

fzimmermann89:
Add ZeroOp (returns zeros)
Makes ZeroOp neutral element of OperatorAddion, IdentityOp of composition, i.e.

Op + zeroop = op
op@identityop = op


fzimmermann89:
Curently in this pr, ZeroOp by default returns zeros of the same shape and type as the input. By setting the flag to false  it can be changed to instead return a scalar 0 tensors in forward and adjoint

I am considering changing the default of the ZeroOp to always return a scalar 0 tensor (which can broadcast/type promote to domain and range)

This is the most useful form of the operator ( as a neutral element of addition), for example in the matrix operator. On its own, it does not pass the adjointess test, but ZeroOp+any linearoperator does. 
I cannot think of a use case where retuning a scalar tensor does not work.

fzimmermann89:
I am considering changing the default of the ZeroOp to always return a scalar integer 0 in forward and adjoint.
This is the most useful form of the operator ( as a neutral element of addition), for example in the matrix operator. On its own, it does not pass the adjointess test, but ZeroOp+any linearoperator does. 
I cannot think of a use case where retuning a scalar tensor does not work. 
And whenever somebody needs a zeroop, they actually need this broadcasting form..

fzimmermann89:
@ckolbPTB @schuenke

ckolbPTB:
> I am considering changing the default of the ZeroOp to always return a scalar integer 0 in forward and adjoint.

sounds reasonable to me

src/mrpro/operators/LinearOperator.py
```diff
@@ -145,46 +148,93 @@ def __matmul__(
 
         Returns lambda x: self(other(x))
         """
-        if isinstance(other, LinearOperator):
+        if isinstance(other, mrpro.operators.IdentityOp):
```
ckolbPTB, line 206: Any special reasons for why if is used for the first and if-elif for the rest?

src/mrpro/operators/LinearOperator.py
```diff
@@ -145,46 +148,93 @@ def __matmul__(
...
+            return other
+        elif isinstance(other, LinearOperator):
             # LinearOperator@LinearOperator is linear
             return LinearOperatorComposition(self, other)
-        else:
-            return OperatorComposition(self, other)
+        elif isinstance(other, Operator):
+            # cast due to https://github.com/python/mypy/issues/16335
+            return OperatorComposition(self, cast(Operator[*Tin2, tuple[torch.Tensor,]], other))
+        return NotImplemented  # type: ignore[unreachable]
```
ckolbPTB, line 217: Is this needed for something specific?

src/mrpro/operators/LinearOperator.py
ckolbPTB: similarly in matmul of the EndomorphOperator

src/mrpro/operators/LinearOperator.py
```diff
@@ -218,7 +268,12 @@ class LinearOperatorElementwiseProductRight(
 
     def adjoint(self, x: torch.Tensor) -> tuple[torch.Tensor,]:
         """Adjoint Operator elementwise multiplication with a tensor."""
-        return self._operator.adjoint(x * self._tensor.conj())
+        conj: complex | torch.Tensor
```
ckolbPTB: `conj = self._scalar.conj() if isinstance(self._scalar, torch.Tensor) else self._scalar.conjugate()` ?

src/mrpro/operators/Operator.py
```diff
@@ -31,21 +34,46 @@ def __matmul__(self: Operator[*Tin, Tout], other: Operator[*Tin2, tuple[*Tin]])
...
+        elif isinstance(other, Operator):
+            return OperatorSum(
+                cast(Operator[*Tin, Tout], other), self
+            )  # cast due to https://github.com/python/mypy/issues/16335
+        return NotImplemented  # type: ignore[unreachable]
+
+    def __mul__(self, other: torch.Tensor | complex) -> Operator[*Tin, Tout]:
         """Operator multiplication with tensor.
 
         Returns lambda x: self(other*x)
```
ckolbPTB: should this actually be `x*other`?

src/mrpro/operators/ZeroOp.py
```diff
@@ -0,0 +1,64 @@
...
+        """Apply the operator to the input.
+
+        Parameters
+        ----------
+        x
+            Input tensor.
+
+        Returns
+        -------
+        zeros_like(x)
```
ckolbPTB: or 0 scalar

tests/operators/test_zero_op.py
```diff
@@ -0,0 +1,81 @@
...
+    generator = RandomGenerator(seed=0)
+    x = generator.complex64_tensor(2, 3, 4)
+    operator = ZeroOp(keep_shape=True)
+    (actual,) = operator(x)
+    expected = torch.zeros_like(x)
+    torch.testing.assert_close(actual, expected)
+
+
+def test_zero_op_scalar():
+    """Test that the zero operator returns zeros."""
```
ckolbPTB: single zero

ckolbPTB (commented):
just some minor comments

Some of the doc strings need updating because we do not just support tensors now but also scalar values.

src/mrpro/operators/LinearOperator.py
fzimmermann89: returning notimplemented is the suggested way to signify, that the magic method does not implement this specific combination and python should try rmatmul on the right hand operator.

src/mrpro/operators/LinearOperator.py
fzimmermann89: in endomorph operator it helped hiding a mypy buig resulting in a crash, but casting to any afew lines furhter down should also now solve it :D

src/mrpro/operators/LinearOperator.py
```diff
@@ -145,46 +148,93 @@ def __matmul__(
...
+            return LinearOperatorSum(self, other)
+        elif isinstance(other, Operator):
+            # for general operators
+            return OperatorSum(self, other)
+        else:
+            return NotImplemented  # type: ignore[unreachable]
 
-    def __mul__(self, other: torch.Tensor) -> LinearOperator:
+    def __mul__(self, other: torch.Tensor | complex) -> LinearOperator:
         """Operator elementwise left multiplication with tensor.
```
ckolbPTB: multiplication with tensor or scalar.

src/mrpro/operators/LinearOperator.py
```diff
@@ -145,46 +148,93 @@ def __matmul__(
...
+            else:
+                return LinearOperatorElementwiseProductLeft(self, other)
+        elif isinstance(other, torch.Tensor):
+            return LinearOperatorElementwiseProductLeft(self, other)
+        else:
+            return NotImplemented  # type: ignore[unreachable]
 
-    def __rmul__(self, other: torch.Tensor) -> LinearOperator:
+    def __rmul__(self, other: torch.Tensor | complex) -> LinearOperator:
         """Operator elementwise right multiplication with tensor.
```
ckolbPTB: multiplication with tensor or scalar.

src/mrpro/operators/LinearOperator.py
```diff
@@ -218,7 +268,8 @@ class LinearOperatorElementwiseProductRight(
 
     def adjoint(self, x: torch.Tensor) -> tuple[torch.Tensor,]:
         """Adjoint Operator elementwise multiplication with a tensor."""
```
ckolbPTB: tensor or scalar

src/mrpro/operators/LinearOperator.py
```diff
@@ -231,7 +282,8 @@ class LinearOperatorElementwiseProductLeft(
...
```
ckolbPTB: tensor or scalar

tests/operators/test_zero_op.py
```diff
@@ -0,0 +1,81 @@
...
+    generator = RandomGenerator(seed=0)
+    x = generator.complex64_tensor(2, 3, 4)
+    operator = ZeroOp(keep_shape=True)
+    (actual,) = operator(x)
+    expected = torch.zeros_like(x)
+    torch.testing.assert_close(actual, expected)
+
+
+def test_zero_op_scalar():
+    """Test that the zero operator return single zero."""
```
ckolbPTB: return**s**

ckolbPTB (approved):
some doc string stuff

src/mrpro/operators/LinearOperator.py
fzimmermann89:
```suggestion
        """Operator elementwise left multiplication with tensor/scalar.
```

src/mrpro/operators/LinearOperator.py
fzimmermann89:
```suggestion
        """Operator elementwise right multiplication with tensor/scalar.
```

src/mrpro/operators/LinearOperator.py
fzimmermann89:
```suggestion
        """Adjoint Operator elementwise multiplication with a tensor/scalar."""
```

src/mrpro/operators/LinearOperator.py
fzimmermann89:
```suggestion
        """Adjoint Operator elementwise multiplication with a tensor/scalar."""
```

tests/operators/test_zero_op.py
fzimmermann89:
```suggestion
    """Test that the zero operator returns single zero."""
```

PR #432, opened 2024-10-07
Add matrix of LinearOperators

fzimmermann89:
A matrix opf LinearOperators, as discussed with @koflera to be able to implement TV, TGV etc with both horizontally and vertically stacked operators.

So horizontally stacking n operators results in n-tensor -> 1 tensor matrix (separable sum of linear operators),
vertically stacking results in 1 tensor-> n-tensor (results of applying different operators are stacked on top of each other).

Might also be useful for dynamic MRI to stack Fourier Operators diagonally (see .from_diagonal), to apply different operators to different inputs.

I only implemented the linear case, not a general operator case. If we want to stack non-linear operators, for example to have a diagonal operator of different t1 mapping objectives, we would need an extension t general operators. 

For now, I want to avoid any lengthy discussion about whether a matrix of non-linear operators, a diagonal non linear operator etc makes sense to everybody involved.

Also, we do not have type hinting for the number of rows and cols. This got quite lengthy to be able to tell mypy what kind of operators takes which number of arguments and returns which number. We can revisit this once higher kinded types are type arithmetic is supported in python.


fzimmermann89:
should be merged into main. Please review only dfff to zero_op.

tests/operators/test_linearoperatormatrix.py
```diff
@@ -0,0 +1,322 @@
...
+import torch
+from mrpro.operators import EinsumOp, LinearOperator, MagnitudeOp
+from mrpro.operators.LinearOperatorMatrix import LinearOperatorMatrix
+
+from tests import RandomGenerator
+from tests.helper import dotproduct_adjointness_test
+
+
+def random_linearop(size, rng):
+    """Create a random LinearOperator."""
```
koflera, line 13: maybe "Create a random LinearOperator corresponding to the common batched matrix-vector multiplication"?

src/mrpro/operators/LinearOperatorMatrix.py
```diff
@@ -0,0 +1,354 @@
...
+        ----------
+        x
+            Input tensors. Requires the same number of tensors as the operator has columns.
+
+        Returns
+        -------
+            Output tensors. The same number of tensors as the operator has rows.
+        """
+        if len(x) != self.shape[1]:
+            raise ValueError('Input should have the same length as the operator has columns.')
```
koflera: Maybe a clearer and more precise comment like "The input should have the same number of tensors as the matrix of LinearOperatorMatrix has linear operators in the column-dimension." ?

src/mrpro/operators/LinearOperatorMatrix.py
```diff
@@ -0,0 +1,354 @@
...
+
+        Returns
+        -------
+            Output tensors. The same number of tensors as the operator has columns.
+        """
+        return self.H(*x)
+
+    @classmethod
+    def from_diagonal(cls, *operators: LinearOperator):
+        """Create a diagonal LinearOperatorMatrix.
```
koflera, line 238: maybe better: "Create a block-diagonal LinearOperatorMatrix"

src/mrpro/operators/LinearOperatorMatrix.py
```diff
@@ -0,0 +1,354 @@
...
+
+        Create a square LinearOperatorMatrix with the given Linear Operators on the diagonal.
```
koflera: Maybe it could be helpful to allow to also be able to construct sort of block-diagonal operators with an additional argument "diagonal=0" that lets you select on which diagonal you want to put the operators?

src/mrpro/operators/LinearOperatorMatrix.py
```diff
@@ -0,0 +1,354 @@
...
+    def operator_norm(
+        self,
+        *initial_value: torch.Tensor,
+        dim: Sequence[int] | None = None,
+        max_iterations: int = 20,
+        relative_tolerance: float = 1e-4,
+        absolute_tolerance: float = 1e-5,
+        callback: Callable[[torch.Tensor], None] | None = None,
+    ) -> torch.Tensor:
+        """Upper bound of operator norm of the Matrix.
```
koflera, line 262: why only upper bound?

src/mrpro/operators/LinearOperatorMatrix.py
```diff
@@ -0,0 +1,354 @@
...
+
+        Uses the bounds :math:`||[A, B}^T|||<=sqrt(||A||^2 + ||B||^2)` and :math:`||[A, B]|||<=||A||+||B||`
```
koflera: typo: [A, B]^T

src/mrpro/operators/LinearOperatorMatrix.py
koflera: i don't see why we should only use the upper-bound for computing the operator-norm... is it computationally so much cheaper than computing the actual one?

src/mrpro/operators/LinearOperatorMatrix.py
koflera: I think I do not really in which context these inequalities are used to compute the upper bound of the operator-norm.
Could we maybe give an example for an operator of the form, say, 
```
Op = [A,B,C]
     [D,E,F]
```
?

src/mrpro/operators/LinearOperatorMatrix.py
```diff
@@ -0,0 +1,354 @@
...
+                max_iterations=max_iterations,
+                relative_tolerance=relative_tolerance,
+                absolute_tolerance=absolute_tolerance,
+                callback=callback,
+            )
+
+        if len(initial_value) != self.shape[1]:
+            raise ValueError('Initial value should have the same length as the operator has columns.')
+        norms = torch.tensor(
+            [[_singlenorm(op, iv) for op, iv in zip(row, initial_value, strict=True)] for row in self._operators]
```
koflera, line 306: maybe I misunderstand the code, but for example, for this operator,
```
Op = [A,0,0]
     [D,B,0]
     [0,0,C]
```
having the estimate given by ||Op|| < sqrt(||A||**2 + .... ||D||**2) seems to be quite bad, see the following example, where you can change `lower_diagonal_flag  ` to either include a lower-diagonal operator or not. Regardless of that, the estimate seems rather bad, shower. For example, running

```
import torch 
from math import sqrt

torch.manual_seed(0)
n=32

lower_diagonal_flag = False     
A = torch.randn(n,n)
B = torch.randn(n,n)
C = torch.randn(n,n)

#block-diagonal matrix
operator_list = [A,B,C]
Mat = torch.block_diag(*operator_list)

operator_names = ["A","B","C"]

if lower_diagonal_flag:
    D = torch.randn(n,n)
    operator_list.append(D)
    operator_names.append("D")
    
    #put some matrix in the lower diagonal
    Mat[n:n+n,:n] = D 


Mat_norm = torch.linalg.matrix_norm(Mat,ord=2)
for op_name, operator in zip(operator_names, operator_list):
    print("op norm of ",op_name, torch.linalg.matrix_norm(operator,ord=2).item())

def upper_bound_op_norm(operators):
    return sqrt(sum([torch.linalg.matrix_norm(op,ord=2).item()**2 for op in operators]))

print("op norm of Mat", Mat_norm.item())
print("est upper bound: ",upper_bound_op_norm(operator_list))
```
gives me

```
op norm of  A 11.429673194885254
op norm of  B 10.949110984802246
op norm of  C 10.883465766906738
op norm of Mat 11.42967414855957
est upper bound:  19.20859931902445
```
which also shows that, if we have purely block-diagonal operator, taking the max of the operator-norms of the operators gives the exact operator norm (we had also a test-example for that).

Thus, maybe it is worth distinguishing for the cases where we have a block-diagonal operator or not?

koflera (changes requested):
somme comments

src/mrpro/operators/LinearOperatorMatrix.py
koflera: additionally, if we still opt for calculating only an upper bound for the operator norm instead of the actual operator norm, we could/should at least raise a warning?

tests/operators/test_linearoperatormatrix.py
fzimmermann89: it really does not matter here, it is just a tiny test helper.

fzimmermann89: any linear operator would suffice

src/mrpro/operators/LinearOperatorMatrix.py
fzimmermann89: why?

src/mrpro/operators/LinearOperatorMatrix.py
fzimmermann89: depends, i would like to add a analytical operator norms to  some operators where we know them.
also, the norm is a bit annoying if the inputs have different shapes.

src/mrpro/operators/LinearOperatorMatrix.py
fzimmermann89: you are right, for horzontally stacked operators, it could be the max.
I just thought the result has to be bigger than either, so sum .. but using the max would of course be better to achieve that.
so root-sum-square of rows and max of cols.
this would make the bound a bit tighter, especially for diagonl operators.

```python
import einops
norms=torch.linalg.matrix_norm(einops.rearrange(Mat,"(r n) (c m)-> r c n m", n=n, m=n),ord=2)
print('norms',norms)
norm=norms.square().sum(-2).sqrt().amax(-1)
print('tighter', norm.item())
```
```
op norm of Mat 14.396413803100586
norms       [[11.4297,  0.0000,  0.0000],
            [10.7493, 10.9491,  0.0000],
            [ 0.0000,  0.0000, 10.8835]]
tighter bound 15.690311431884766
```

fzimmermann89: good catch!

src/mrpro/operators/LinearOperatorMatrix.py
koflera: > depends, i would like to add a analytical operator norms to some operators where we know them.< 
definitely agree

>also, the norm is a bit annoying if the inputs have different shapes.<
Also agree, although I think that typically, one will know anyway the shape of the tensors to be used (e.g. in iterative algorithms), thus only needing to create `torch.randn_like` tensors as starting values. Also, something like this has to happen anyway given that each sub-operator in the matrix requires a starting value, no?

src/mrpro/operators/LinearOperatorMatrix.py
fzimmermann89: i put it in the second line

src/mrpro/operators/LinearOperatorMatrix.py
fzimmermann89: made clearer

src/mrpro/operators/LinearOperatorMatrix.py
fzimmermann89: This would just add a permutation to the output, and I am not sure we ever want to have this -- it would be quite confusing

src/mrpro/operators/LinearOperatorMatrix.py
fzimmermann89: added more exlantion and fixed typo

fzimmermann89:
@koflera any thoughts?

src/mrpro/operators/LinearOperatorMatrix.py
```diff
@@ -0,0 +1,357 @@
...
+class LinearOperatorMatrix(Operator[Unpack[tuple[torch.Tensor, ...]], tuple[torch.Tensor, ...]]):
+    r"""Matrix of Linear Operators.
+
+    A matrix of Linear Operators, where each element is a Linear Operator.
+
+    This matrix can be applied to a sequence of tensors, where the number of tensors should match
+    the number of columns of the matrix. The output will be a sequence of tensors, where the number
+    of tensors will match the number of rows of the matrix.
+    The j-th output tensor is calculated as
+    :math:`\sum_i \text{operators}[i][j](x[i])` where :math:`\text{operators}[i][j]` is the linear operator
```
koflera: should be 
:math:`\sum_i \text{operators}[j][i](x[i])` where :math:`\text{operators}[j][i]`

src/mrpro/operators/LinearOperatorMatrix.py
```diff
@@ -0,0 +1,357 @@
...
+    in the i-th row and j-th column and :math:`x[i]` is the i-th input tensor.
```
koflera: should be "in the j-th row and i-th column and :math:`x[i]` is the i-th input tensor."

src/mrpro/operators/LinearOperatorMatrix.py
```diff
@@ -0,0 +1,357 @@
...
+            raise ValueError('Other should have the same length as the operator has columns.')
+        else:
+            other_ = other
+        operators = []
+        for row in self._operators:
+            operators.append([op * o for op, o in zip(row, other_, strict=True)])
+        return self.__class__(operators)
+
+    def __rmul__(self, other: torch.Tensor | Sequence[torch.Tensor] | complex) -> Self:
+        """Tensor*LinearOperatorMatrix multiplication."""
```
koflera: what is this operation supposed to do? say for example x*A for A = [ A1, A2]?

koflera (approved):
looks good to me

src/mrpro/operators/LinearOperatorMatrix.py
fzimmermann89: A = [ A1, A2]
 x*A = [x*A1, x*A2]

koflera: I think I still do not understand the purpose... what is x*A in general if A is a linear operator? For A being a matrix ok, but in general?

fzimmermann89: It is not `x*A` but `Lambda*A`, `A.H*DCF`, etc.
It's is just pre and post multiplication.

PR #433, opened 2024-10-07
Zero functional

fzimmermann89:


fzimmermann89:
should be merged after #396 into main

fzimmermann89:
should be merged into main, not functional_base.
I only selected functional_base here for a clearer code review.

src/mrpro/operators/functionals/ZeroFunctional.py
```diff
@@ -0,0 +1,84 @@
...
+            new_shape = [1 if i in normal_dim else s for i, s in enumerate(x.shape)]
+        else:
+            new_shape = [s for i, s in enumerate(x.shape) if i not in normal_dim]
+
+        return (torch.zeros(new_shape, dtype=dtype, device=self.target.device),)
+
+    def prox(self, x: torch.Tensor, sigma: float | torch.Tensor = 1.0) -> tuple[torch.Tensor,]:
+        """Apply the proximal operator to a tensor.
+
+        Always returns x.
```
koflera: given that the formula is simple, maybe we could explicitly write that the proximal of the zero-functional is the identity.

src/mrpro/operators/functionals/ZeroFunctional.py
```diff
@@ -0,0 +1,84 @@
...
+            Result of the proximal operator applied to x
+        """
+        self._throw_if_negative_or_complex(sigma)
+        dtype = torch.promote_types(torch.promote_types(x.dtype, self.weight.dtype), self.target.dtype)
+        return (x.to(dtype=dtype),)
+
+    def prox_convex_conj(self, x: torch.Tensor, sigma: float | torch.Tensor = 1.0) -> tuple[torch.Tensor,]:
+        """Apply the proximal operator of the convex conjugate of the functional to a tensor.
+
+        Returns x if sigma is exactly 0, otherwise returns 0.
```
koflera: similarly here, maybe explicitly mention that, since the convex conjugate of the zero-functional is the indicator function over the set C^N\{0} or R^N\{0}, the prox of the convex conjugate is 0 if sigma>0 and the identity otherwise.

src/mrpro/operators/functionals/ZeroFunctional.py
koflera, line 47: maybe explicitly mention that the prox of the zero-functional is the identity, thus always returns x.

src/mrpro/operators/functionals/ZeroFunctional.py
koflera, line 67: maybe explicitly mention that the prox of the convex conj of the zero-functional is the indicator function over the set C\{0}, and thus it returns x if sigma is exactly 0, otherwise returns 0.

koflera (approved):
Looks good to me!

src/mrpro/operators/functionals/ZeroFunctional.py
fzimmermann89: " Always returns x."
```suggestion
        Always returns x, as the proximal operator of a constant functional is the identity.
```
?

src/mrpro/operators/functionals/ZeroFunctional.py
fzimmermann89: `The convex conjugate of the zero functional is the indicator function over the set {0}. If sigma>0, the proximal operator of the scaled convex conjugate is constant zero, otherwise it is the identity`

?

fzimmermann89:
```suggestion
        The convex conjugate of the zero functional is the indicator function over the set {0}. If sigma>0,
        the proximal operator of the scaled convex conjugate is constant zero, otherwise it is the identity.
```

PR #438, opened 2024-10-08
Outline for new header structure 

ckolbPTB:
So far IData and QData has very limited header parameters. I am trying to extend this now and I am struggling a bit of how we can efficiently reuse things. 

@fzimmermann89 @schuenke do you think it makes sense to intruduce a base class for the AcqIdx and AcqInfo and reuse that also for the IData and QData? We would then have something like

KHeader including AcqIdx and AcqInfo with
DataIdx <- AcqIdx
DataInfo <- AcqInfo

I don't know what to call the overarching QData and IData space, so let's just call it X

XHeader including XIdx and XInfo with
DataIdx <- XIdx
DataInfo <- XInfo
this will include functionality such as from_kheader and from_dicom

and then QHeader and IHeader which will be child of XHeader with a few additional parameters (e.g. TE or TR for IHeader)

I would then rename Data to XData which would then also have the common functionality such as from_dicom

I would also do this as a separate PR to main before extending the IHeader and QHeader parameters.

### DIFF
--- a/src/mrpro/data/AcqInfo.py
+++ b/src/mrpro/data/AcqInfo.py
@@ -1,40 +1,21 @@
 """Acquisition information dataclass."""
 
-import dataclasses
 from collections.abc import Callable, Sequence
 from dataclasses import dataclass
-from typing import Self, TypeVar
+from typing import Self
 
 import ismrmrd
 import numpy as np
 import torch
-from einops import rearrange
 
-from mrpro.data.MoveDataMixin import MoveDataMixin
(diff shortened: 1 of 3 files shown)


fzimmermann89:
LEts talk about it in the next meeting, might be a good idea.

Some initial thoughts:
 data_info: DataInfo is really uninformative
 the whole header is info about the data..
 
 
 the tensors will have different shapes for AcqIdx and DataIdx

fzimmermann89:
I think this can be closed, reopen if you disagree

PR #439, opened 2024-10-09
Add continuous deployment (CD) pipeline

### LINKED ISSUE #125: Release strategy
In order to be able to publish the package we need a release strategy and also include proper versioning of the code and documentation.

schuenke:
# Summary

- the current version of MRpro is now defined in src/mrpro/VERSION
- the current version format is **0.YYMMDD** (no semantic versioning yet)
- `mrpro.__version__` is now defined
- the CD pipeline will be triggered if the VERSION file is changed
- if VERSION is changed in a dev branch, only a TestPyPi deployment is triggered
- for TestPyPi releases, the format is <current version>rc<timestamp> (e.g. `0.241009rc1728492371`)
- if merged to main, both, a TestPyPi and PyPi release is triggered
- if you build the package locally with modified files compared to the current branch, the version will be <current version>+dirty<sha> (e.g. `0.241009+dirty8c038f38`)
- you can check which version will be used for the build process using `python -m setuptools_git_versioning`

# To-do

- [x] add tag and release job
- [x] include `_version.py` module
- [x] change version suffix from `pre` to `rc` in workflow
- [x] ensure TestPyPi version from workflow is installed

Closes #125


.github/workflows/cd_testpypi.yml
```diff
@@ -0,0 +1,100 @@
...
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.x"
+
+      - name: Read the current version
+        id: get_version
+        run: echo "version=$(cat VERSION)" >> $GITHUB_ENV
+
+      - name: Append current time to the version
+        run: |
```
fzimmermann89: das brauchst du glaub ich nicht wenn du statt dessen im template das ENV drin hast, {env:MRPROVERSIONSUFFIX:''}
und MRPROVERSIONSUFFIX dann auf "-time" setzt.

fzimmermann89: das hat den vorteil, dass die dirty detection funktionieren kann.
d.h. versionen die nicht commitete änderungen haben, werden als dirty makiert von setup tools git

fzimmermann89: wenn du die versions datei lokal änderst, wird sonst jede version dirty

pyproject.toml
```diff
@@ -1,14 +1,24 @@
 [build-system]
-requires = ["setuptools>=42", "wheel"]
+requires = ["setuptools>=42", "wheel", "setuptools-git-versioning>=2.0,<3"]
 build-backend = "setuptools.build_meta"
 
+[tool.setuptools-git-versioning]
+enabled = true
+version_file = "VERSION"
+template = "{tag}"
```
fzimmermann89: {tag}{env:MRPROVERSIONSUFFIX:''}

fzimmermann89 (commented):
Ich würde noch
```
from importlib.metadata import version, PackageNotFoundError
try:
    __version__ = version("package-name")
except PackageNotFoundError:
    # package is not installed
    __version__ = Path(__file__).parent.joinpath("VERSION").read_text()

```
zur maiin `__init__.py` hinzufügen damit mrpro.__version__ korrekt ist.
vllt muss man in der docu conf.py dann auch ein overwrite für die version machen und da auch mrpro.__version__ als overwrite nehmen

fzimmermann89:
```
from mrpro  import __version__ as project_version
config = SphinxConfig("../../pyproject.toml", globalns=globals(), config_overrides = {"version": project_version})
```

schuenke:
Everything should work as expected now. The version for the PyPi release seems to be correct (check https://github.com/PTB-MR/mrpro/actions/runs/11270262817/job/31340730315?pr=439). If we merge to main, the deployment step (which is currently the only one that is skipped to ensure everything else works as expected) should be executed as well.

Last thing I will change now is to trigger the whole pipeline only when the VERSION file is changed. This was disabled for test purposes atm. 

**EDIT**: maybe we want to move the `if: github.ref == 'refs/heads/main'` condition from `step` to `job` level? Else GitHub Actions will always report that we successfully deployed to PyPi although we didn't 😁

fzimmermann89:
any idea why it says "@[schuenke](https://github.com/schuenke) schuenke [deployed](https://github.com/PTB-MR/mrpro/actions/runs/11270751643/job/31342300557) to [pypi](https://pypi.org/project/mrpro) [1 hour ago](https://github.com/PTB-MR/mrpro/pull/439#event-14587349639) — with  [GitHub Actions](https://github.com/apps/github-actions)"

src/mrpro/__init__.py
```diff
@@ -1,2 +1,10 @@
+from pathlib import Path
```
fzimmermann89: suggestion:
create a version.py file with the content and import its __version__ in the init
as it currently is (sorry, was my suggestion), Path etc will end up in mrpro namespace

.github/workflows/deployment.yml
```diff
@@ -0,0 +1,126 @@
...
+    steps:
+      - name: Download PyPi distribution
+        uses: actions/download-artifact@v4
+        with:
+          name: pypi-package-distribution
+          path: dist/
+
+      - name: Publish to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        if: github.ref == 'refs/heads/main'
```
fzimmermann89: can we move this up such that the whole job only runs if the condition is true?

fzimmermann89: maybe this would also resolve the wrong deployed to pypi message..

schuenke: Yeah, this is what I suggested above already:

> EDIT: maybe we want to move the if: github.ref == 'refs/heads/main' condition from step to job level? Else GitHub Actions will always report that we successfully deployed to PyPi although we didn't 😁

schuenke: I did put it at the step level during development to ensure all other steps work

fzimmermann89: I did not see the edit :D

src/mrpro/__init__.py
schuenke: Do you want to rename `VERSION` to `version.py` or `_version.py` and define the version only there? 

Or do you want to add an additional `_version.py` file, where we put the code that is currently in the main `__init__.py`, so that we can import `__version__` from there?

Both should work, but for the first option we need to adapt the workflow as well.

fzimmermann89: The latter.
I would leave VERSION as a text file, as is.
I would create an additional _version.py and move the version parsing logic there and import it in init.
such that the only change in this pr to init is:
`from _version import __version__`

pyproject.toml
```diff
@@ -1,14 +1,27 @@
...
+version_file = "src/mrpro/VERSION"
+template = "{tag}{env:MRPROVERSIONSUFFIX:''}"
+dev_template = "{tag}{env:MRPROVERSIONSUFFIX:''}"
+dirty_template = "{tag}+dirty{sha}"
+
+[tool.setuptools]
+include-package-data = true
+
+[tool.setuptools.package-data]
+"mrpro" = ["VERSION"]
```
fzimmermann89, line 17: LICENSE ?

schuenke: According to https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license listing the license in the classifiers is enough, or did I get that wrong?

fzimmermann89: both of us are wrong: LICENSE should be automatically included by setuptools when building the distribution.

fzimmermann89:
👍

ckolbPTB:
have we already added "MRpro Team" and more generic email address?

schuenke:
> have we already added "MRpro Team" and more generic email address?

Nope, didn't do it yet. I can add it tomorrow.

.github/workflows/deployment.yml
```diff
@@ -0,0 +1,171 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.x"
+
+      - name: Set environment variable for version suffix
+        run: echo "MRPROVERSIONSUFFIX=.pre$(date +%s)" >> $GITHUB_ENV
```
fzimmermann89: maybe already call it rc here?

.github/workflows/deployment.yml
```diff
@@ -0,0 +1,171 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.x"
+
+      - name: Install MRpro from TestPyPI
+        run: |
+          python -m pip install --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple/ mrpro
```
fzimmermann89: i am not sure this does what we think it does.
wont this default to installing the last released version?
at the minimum, we might need a --pre here to allow rc versions.

Better would be to actually try to install the version we just uploaded.

fzimmermann89:
in the the testpypi install
```bash
VERSION=$(cat VERSION)
SUFFIX=${MRPROVERSIONSUFFIX#.

pip install mrpro=={VERSION}${SUFFIX} --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple/
```
(assuming we change the MRPROVERSONSUFFIX from pre to rc)

.github/workflows/deployment.yml
schuenke: done ✔️

.github/workflows/deployment.yml
schuenke: should be the case now... I store the version and suffix variables in the `build-testpypi-package` job now:

```
VERSION=$(cat src/mrpro/VERSION)
SUFFIX=rc$(date +%s)
echo "suffix=$SUFFIX" >> $GITHUB_OUTPUT
echo "version=$VERSION" >> $GITHUB_OUTPUT
```
and then use it later in the `test-install-from-testpypi` job:

```
VERSION=${{ needs.build-testpypi-package.outputs.version }}
SUFFIX=${{ needs.build-testpypi-package.outputs.suffix }}
python -m pip install mrpro==$VERSION$SUFFIX --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple/
```

schuenke:
This PR should now be ready for final review @fzimmermann89 and @ckolbPTB

ckolbPTB:
On testpypi we have got some versions without rc in the name and they are always shown as the latest version. Is it possible to delete those versions from testpypi?

schuenke:
> On testpypi we have got some versions without rc in the name and they are always shown as the latest version. Is it possible to delete those versions from testpypi?

Done ✔️ 

Deleted the files with the wrong version style uploaded during debugging and marked 0.0.1 as a [Yank Release](https://pypi.org/help/#yanked)

grafik

.github/workflows/deployment.yml
```diff
@@ -0,0 +1,183 @@
...
+          python-version: "3.x"
+
+      - name: Install setuptools_git_versioning
+        run: |
+          python -m pip install setuptools-git-versioning
+
+      - name: Get current version
+        id: get_version
+        run: |
+          VERSION=$(python -m setuptools_git_versioning)
```
fzimmermann89, line 111: 👍

fzimmermann89 (approved):
Supi!🎉

Mal schauen ob uns dann probleme im betrieb auffallen - aktuell sieht fur mich alles top aus.

PR #440, opened 2024-10-09
Sparsify documentation

fzimmermann89:
denoising by removal of autosummary methods and imported members


fzimmermann89:
This is one way to solve the docu issues and have an automatic API documentation.
`__all__ `could also be filled in automatically with some magic in the top level init.

fzimmermann89:
@ckolbPTB (once the docu has been rebuilt)

fzimmermann89:
```python
def populate_all_for_inits():
    """
    Populates the __all__ attribute for the package's __init__.py files based on imported members.
    """
    import sys
    import pkgutil

    package_name = __name__
    package = sys.modules[package_name]

    def finalize_all(module):
        initial_symbols = set(module.__dict__)
        current_symbols = set(module.__dict__)
        imported_symbols = current_symbols - initial_symbols
        
        module.__all__ = [
            name for name in imported_symbols
            if not name.startswith('_') and name != package_name.split('.')[0]
        ]

    finalize_all(package)

    for _, modname, ispkg in pkgutil.walk_packages(package.__path__, package_name + '.'):
        __import__(modname)
        if ispkg:
            finalize_all(sys.modules[modname])

    del sys.modules[package_name].populate_all_for_inits

populate_all_for_inits()

```
before the first import in the top level init could work...

fzimmermann89:
fun fact:sphinx-build -b and sphinx make save .doctrees in different paths.
took me a while to understand while our github action uploads .doctrees

fzimmermann89:
i would just delete github-pages and recreate it to solve #413 
any objections @schuenke @ckolbPTB ?

ckolbPTB:
> any objections @schuenke @ckolbPTB ?

no objections, but recreating it or creating it somewhere else?

schuenke:
> This is one way to solve the docu issues and have an automatic API documentation. `__all__ `could also be filled in automatically with some magic in the top level init.

In general I think defining `__all__` is a good idea. Was actually thinking about it when adding the version in the top-level init. Not sure if an automatic generation is needed. For me it would definitely not be high priority.

ckolbPTB:
Annoying that we have to repeat everything twice in each init but at least if it is done there we are less likely to forget about it. Some further suggestions:
- throw an error when there are warnings in sphinx build: `sphinx-build -W --keep-going` this should flag problems in the docu during the tests.
- there is still a problem (types are not shown and linked correctly) if the default value of an attribute is more complex, e.g. dataclasses.field(default_factory=dict), this affects e.g IHeader or KData

schuenke:
And personally I would prefer to have classmethods directly below the init and not under the regular methods.
For example here: https://zimf.de/zipserve/https://github.com/PTB-MR/mrpro/actions/runs/11355196439/artifacts/2060612604/_autosummary/mrpro.operators.FourierOp.html#mrpro.operators.FourierOp

fzimmermann89:
> > any objections @schuenke @ckolbPTB ?
> 
> no objections, but recreating it or creating it somewhere else?

for now, just recreating it.
the notebooks do not change that often, the fonts and other static files almost never..
so by removal of the .doctrees that changes quite often, github-pages should also be reasonably small again.
if it gets to big ain, be can aways reevalute

fzimmermann89:
I failed to get the classmethods to be sorted first while spending >2hrs. I would for now just ignore that, even though it would be nice to have a better order of stuff. we could use bysource ordering, but that also breaks with inheritance..

fzimmermann89:
pre-commit.ci autofix

fzimmermann89:
> Annoying that we have to repeat everything twice in each init but at least if it is done there we are less likely to forget about it. Some further suggestions:
> 
> * throw an error when there are warnings in sphinx build: `sphinx-build -W --keep-going` this should flag problems in the docu during the tests.

 There are still a few issues, that I think you already addressed. I would for this PR just leave it as is and we fix the remaining issues in a second PR?
 
> * there is still a problem (types are not shown and linked correctly) if the default value of an attribute is more complex, e.g. dataclasses.field(default_factory=dict), this affects e.g IHeader or KData


I have no suggestion how to solve this. Would suggest moving this to a third PR.

fzimmermann89:
This now uses github actions directly to publish the pages, without having a github-pages branch.

schuenke (approved):
✔️

fzimmermann89:
I merge it now to avoid forgetting __all__ in the other PRs.
this already solves some of the issues and the remaining problems can be addressed by a different PR

PR #441, opened 2024-10-09
Add improper Rotations / Rotoflections

fzimmermann89:
Extend Rotations to be able to represent also Drehspiegelungen / Rotation + Mirroring 

This is useful for Siemens data, which has mirrored acquisition directions: 
Although MRD says "Directional cosine of slice normal, i.e. cross-product of read_dir and phase_dir", for siemens it is the cross product of phase_dir and read_dir (opposite sign/mirrored).

Main issues is what the mean over a proper rotation and an improper rotation is -- there is no reasonable mean of mirror / non-mirror.
Currently, I do a majority vote and warn.
Also, what a non-integer power of an improper rotation is, for example (No Rotation + Mirroring)**(0.5).
It should be that applying this twice would result in Mirroring without Rotation.

The as_quats, as_euler etc now have a keyword only option to also return reflections.
The call signature is still compatible with scipy.

Other changes:
- added a scipy compatible apply and type hints for call.
- added a reshape
- added a align_vectors in pytorch instead of moving to numpy and calling scipy. This still does not support batching, but the preparation logic and the actual logic is already split if we ever want to implement a vmap based batching


ckolbPTB:
I had a look through and tried it on some data -> the interfacte looks fine to me and can deal with the Siemens data.

src/mrpro/data/Rotation.py
```diff
@@ -296,19 +467,56 @@ def __init__(self, quaternions: torch.Tensor | NestedSequence[float], normalize:
...
+        reflection: torch.Tensor | NestedSequence[bool] | bool = False,
+    ) -> Self:
         """Initialize from quaternions.
 
         3D rotations can be represented using unit-norm quaternions [QUAa]_.
+        As an extension to the standard, this class also supports improper rotations,
+        i.e. rotations with reflection with respect to the plane perpendicular to the rotation axis
+        or inversion of the coordinate system.
+
+        Note: If inversion != reflection, the rotion will be improper and concerted to a rotation
```
ckolbPTB: the ro**ta**tion. what do you mean with _concerted_ here?

src/mrpro/data/Rotation.py
```diff
@@ -327,22 +541,26 @@ def from_quat(cls, quaternions: torch.Tensor | NestedSequence[float]) -> Self:
...
+        return cls(quaternions, normalize=True, copy=True, inversion=inversion, reflection=reflection)
 
     @classmethod
-    def from_matrix(cls, matrix: torch.Tensor | NestedSequence[float]) -> Self:
+    def from_matrix(cls, matrix: torch.Tensor | NestedSequence[float], allow_improper: bool = True) -> Self:
         """Initialize from rotation matrix.
 
         Rotations in 3 dimensions can be represented with 3 x 3 proper
         orthogonal matrices [ROTa]_. If the input is not proper orthogonal,
         an approximation is created using the method described in [MAR2008]_.
```
ckolbPTB, line 553: Is this also done for improper rotations?

src/mrpro/data/Rotation.py
```diff
@@ -353,24 +571,92 @@ def from_matrix(cls, matrix: torch.Tensor | NestedSequence[float]) -> Self:
...
+    ):
+        """Initialize from basis vectors as SpatialDimensions.
+
+        Parameters
+        ----------
+        *basis
+            3 Basis vectors of the new coordinate system, i.e. the columns of the rotation matrix
+        allow_improper
+            If true, the rotation is considered as improper if the determinant of the matrix is negative
+            snd the sign will be preserved. If false, an ValueError is raised if the determinant is negative.
```
ckolbPTB: snd -> and

ckolbPTB: an ValueError -> a ValueError

src/mrpro/data/Rotation.py
```diff
@@ -383,28 +669,56 @@ def from_rotvec(cls, rotvec: torch.Tensor | NestedSequence[float], degrees: bool
         degrees
             If True, then the given angles are assumed to be in degrees,
             otherwise radians.
+        reflection
+            If True, the resulting transformation will contain a reflection
+            about a plane perpendicular to the rotation axis, resulting in a rotoflexion
```
ckolbPTB: rotoflection?

ckolbPTB (approved):
minor comments

src/mrpro/data/Rotation.py
fzimmermann89: Why doesn't the typo check find this..?

fzimmermann89:
```suggestion
        Note: If inversion != reflection, the rotation will be improper and save as a rotation followed by an inversion.
```

src/mrpro/data/Rotation.py
fzimmermann89: Yes, it will be done whenever the vectors are not true orthogonal (which they should be for improper and proper rotations)

src/mrpro/data/Rotation.py
fzimmermann89: If you want to do me a favor, you can directly use the change suggestion for these small things :)

fzimmermann89:
```suggestion
            and the sign will be preserved. If false, a ValueError is raised if the determinant is negative.
```

src/mrpro/data/Rotation.py
fzimmermann89:
```suggestion
            about a plane perpendicular to the rotation axis, resulting in a rotoflection
```

fzimmermann89:
pre-commit.ci autofix

PR #443, opened 2024-10-13
Add gram to LinearOperator

fzimmermann89:
Some code I have to obtain the gram matrix/operator of a linear operator, i.e. A -> A.H@A
This has the advantages over writing  A.H@A that certain linear operators can implement a different .gram.

For example, the NUFFT-Op might. This is the ground work for it, currently no operator makes use of it. 
(To implement it for the FourierOp we might need to change FourierOp to be a composition of a NufftOp, FFTOp and CartesianSampling Op. This would then allow us to skip doing the frequency encoding ffts an A.H@A and using toeplitz for nufft....) 

I started doing this some time ago, but might not have to time right now to finish it right now -- at least the ground work should maybe already be merged to avoid loosing it

I am open to a name change (normal, hessian, selfadjoint, aha, adjoint_forward, psd,...) @koflera @guastinimara

### DIFF
--- a/tests/operators/test_operators.py
+++ b/tests/operators/test_operators.py
@@ -245,3 +245,52 @@ def test_adjoint_sum():
     v = rng.complex64_tensor(3)
     linear_op_sum = a + b
     dotproduct_adjointness_test(linear_op_sum, u, v)
+
+
+def test_gram_shortcuts():
+    """Test that .gram for composition and sclalar multiplication results in shortcuts."""
+
+    class GramOnlyOperator(LinearOperator):
+        """Operator-Wrapper that only has a working .gram property."""
+
+        def __init__(self, op: LinearOperator):
+            super().__init__()
+            self.op = op
+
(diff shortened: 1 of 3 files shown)


tests/operators/test_operators.py
```diff
@@ -245,3 +245,52 @@ def test_adjoint_sum():
     v = rng.complex64_tensor(3)
     linear_op_sum = a + b
     dotproduct_adjointness_test(linear_op_sum, u, v)
+
+
+def test_gram_shortcuts():
+    """Test that .gram for composition and sclalar multiplication results in shortcuts."""
```
ckolbPTB: scalar

ckolbPTB (approved):

tests/operators/test_operators.py
fzimmermann89:
```suggestion
    """Test that .gram for composition and scalar multiplication results in shortcuts."""
```

PR #444, opened 2024-10-17
Enable numpy version > 2.0

schuenke:
Few more things to fix than expected...

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ classifiers = [
     "Programming Language :: Python :: 3 :: Only",
 ]
 dependencies = [
-    "numpy>=1.23, <2.0",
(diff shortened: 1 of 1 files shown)


schuenke:
ok, seems like I used a wrong env when I tested it locally and everything passed 🙄 

I think we should still tackle this, but I won't be able to do it today. But we can also merge this PR after #439 ...

fzimmermann89:
Puh, too bad. This looks unfortunately like quite some detective work. Maybe put it on our to-do list for some time in the future..

schuenke:
Quite funny actually: https://github.com/numpy/numpy/issues/26620

schuenke:
superseded by #628

PR #446, opened 2024-10-17
Add check-added-large-files to precommit

fzimmermann89:
#413 

> This option prevents big files from being committed.
> The default for "too big" is 500kB.
> This should be adjustable in the pre-commit hook using args: ['--maxkb=123']


schuenke (approved):
For future me and interested ppl:

This option prevents big files from being committed.
The default for "too big" is 500kB.
This should be adjustable in the pre-commit hook using `args: ['--maxkb=123']`

PR #449, opened 2024-10-18
Examples typos

### LINKED ISSUE #459: Old dataset downloaded from zenodo for T1_mapping_with_grad_acq
In 

https://github.com/PTB-MR/mrpro/blob/d396143f62bfbaed5dd9b66b86be3182f3467192/examples/t1_mapping_with_grad_acq.py#L59

we are downloading v1 of the dataset instead of v2. Easy but not safe fix would be to set dataset to "13207352". This will download v2 but not v3 if there is one.

### LINKED ISSUE #462: QMRI notebooks broken
At some points the results of our QMRI notebooks must have taken a turn for the worse:
https://ptb-mr.github.io/mrpro/notebook_qmri_sg_challenge_2024_t1.html
https://ptb-mr.github.io/mrpro/notebook_qmri_sg_challenge_2024_t2_star.html

Both the values of the T1 and T2* map are incorrect.

lrlunin:
Fix issues caused by the switch to SI units.
Use new qmri challenge dataset
Use latex in notebooks  

Closes #459 
Closes #462

> Hello,
> 
> there are some very minor changes, but you have to start somewhere :-)
> 
> 1. The first one is using T1 instead of subscript as done before. Only this example has used it.
> 2. Use of `\ln` and `\cos` TeX commands instead of plain text.
> 
> They are not urgent at all.
> Thanks in advance.
>


fzimmermann89:
good idea to make the use of T_1 consistent in the examples -- but in all equations it should be `$T_1$.`

lrlunin:
> good idea to make the use of T_1 consistent in the examples -- but in all equations it should be `$T_1$.`

Okay, I'll replace the notation in all examples then.

lrlunin:
I also added subscript notations for the graphs. Should be ready to merge so far.

ckolbPTB:
maybe not change the title of the notebooks because these titles are used in the docu as doctree entries which does not support latex formatting.

lrlunin:
> maybe not change the title of the notebooks because these titles are used in the docu as doctree entries which does not support latex formatting.

They actually kinda do, but not on the top level. I guess there is some option to force the TeX display mode.
In the above level:
image

If you check on of the examples.
image

So I believe there is some page-specific display-/parse-mode which was automatically enabled in the page with the markdown notebook and was not enabled in the examples' table of content. I'll check if there is some magic hint to force the page be displayed as markdown.

lrlunin:
I found a much better solution. With the `nbsphinx` all works as expected and all converting, running, including as html in the `docs.yml` can removed. I'll make a commit till monday I guess.

ckolbPTB:
> I found a much better solution. With the `nbsphinx` all works as expected and all converting, running, including as html in the `docs.yml` can removed. I'll make a commit till monday I guess.

They way it is know allows for all the notebooks to be run in parallel. Maybe nbspinx offers this now too?

lrlunin:
> > I found a much better solution. With the `nbsphinx` all works as expected and all converting, running, including as html in the `docs.yml` can removed. I'll make a commit till monday I guess.
> 
> 
> 
> They way it is know allows for all the notebooks to be run in parallel. Maybe nbspinx offers this now too?

[They promise](http://nbsphinx.readthedocs.org/en/0.2.1/usage.html) to allow threads with `-j` parameter analog to `make` parameter.

docs/source/conf.py
```diff
@@ -20,8 +21,10 @@
 # -- Project information -----------------------------------------------------
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
 
+# TODO: need to copy .py/.ipynb files from examples to source/_notebooks or insert them
+
 project = name
-copyright = '2023, Physikalisch-Technische Bundesanstalt (PTB) Berlin'
+copyright = f'{datetime.now().year}, Physikalisch-Technische Bundesanstalt (PTB) Berlin'
```
fzimmermann89: This should be left at 2023

fzimmermann89:
Btw: We will still need a conversion from py to ipynb & commit, as we want to keep the notebooks available in the repo as well.

lrlunin:
> > > I found a much better solution. With the `nbsphinx` all works as expected and all converting, running, including as html in the `docs.yml` can removed. I'll make a commit till monday I guess.
> > 
> > 
> > They way it is know allows for all the notebooks to be run in parallel. Maybe nbspinx offers this now too?
> 
> [They promise](http://nbsphinx.readthedocs.org/en/0.2.1/usage.html) to allow threads with `-j` parameter analog to `make` parameter.

This seems not to be true. The execution of the notebooks is indeed not parallel and lasts approx. 3 minutes on my pc. Then it would be more reasanoble to perform the step with the `.py` -> `.ipynb` conversion and further `.ipynb` evaluation in the action. Then copy and include the evaluated notebooks into the docs with the `nbsphinx_execute = 'never' ` setting in the `conf.py`.

fzimmermann89:
My 2cents: 
Sphinx -j option is not the same as parallel jobs.
Both download speed for the example data and CPU should be exhausted by a single example.
The parallel jobs run afaik on separate worker machines.

On your local machine at home, the total time might be dominated by the data download..

lrlunin:
This is done so far and awaits https://github.com/PTB-MR/mrpro/pull/469 to pass the github action since `pandoc` is missing

fzimmermann89:
For me, the requirement of a non-pip binary is a blocker here. Does the switch to nbshinx really have such advantages that it is worth the hassle?

Can you maybe first fix the issues with the notebooks you noticed (and maybe also change the code to s instead of ms, this would fix and there open issue #462).

And then we can push the overhaul of the action a bit into the future and talk about it and the advantages your approach has first.

fzimmermann89:
Try pypandoc-binary

fzimmermann89:
Do I understand this correctly that remote builds still use papermill to execute (potentially giving us a way to include non public data), but local builds also work and execute the notebooks using nbshinx?

I would add the pandoc binaries as a requirement instead of apt to the docker. this way building just works after a pip install locally.

fzimmermann89:
Still, consider splitting this in two parts:
- Fixing the notebooks (the fitting and formattin
 - Switching to nbshinx

docs/source/conf.py
```diff
@@ -43,6 +46,9 @@
 exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store']
 source_suffix = {'.rst': 'restructuredtext', '.txt': 'restructuredtext', '.md': 'markdown'}
 
+# do not execute notebooks before including in the examples
+# assuming they were executed before
+nbsphinx_execute = 'never'
```
fzimmermann89: why not auto?

lrlunin: We expect the notebooks to be executed in the separate `nb-runner` routine and be executable in `mrpro[notebook]` environment. The docs runs in `mrpro[docs]` environment though and would not only probably crash in execution but also interfere with the result of the runners. This way we would not be able to check where the origin of the problem is. So we expect the runners to deliver a 100% correct executed notebooks for the docs.

fzimmermann89:
consider adding yourself to the pyproject.toml author list

fzimmermann89 (approved):
looks good!

- add yourself to the author list in pyproject.toml
-  minor nitpick: in the plots, add a unit to the colorbars for T1 and T2*

ckolbPTB:
The T1 mapping using radial golden angle data does not lead to good results anymore.

ckolbPTB:
The other two quantitative notebooks also don't work anymore..

-> wait for #535 to fix this

lrlunin:
> add yourself to the author list in pyproject.toml

I don't consider the changes done to worth it. They are primary cosmetic.

> minor nitpick: in the plots, add a unit to the colorbars for T1 and T2*

TBD

fzimmermann89:
Of course your decison. 
I'd say it would be justified.
You can also do it in any future pr if you want to.

fzimmermann89:
@lrlunin:
Small hint for next time: Please use imperative mood and a bit more descriptive name for the commit ;)

PR #452, opened 2024-10-20
Add Jacobian of Operator

### LINKED ISSUE #291: Feature Request: LinearizedOperator
Suggestion:
A LinearOperator that linearizes a function f at a point x_0, so the adjoint is the adjoint of the Jacobian. at x_0

fzimmermann89:
Closes #291 

Status: It only deals with a single input/output. Extension to multiple inputs or outputs will be a different PR.


fzimmermann89:
@koflera Please take a look at the docstrings (and rest of code, ofc)

fzimmermann89:
Die monatliche Frage: can someone review this please?

koflera:
> Die monatliche Frage: can someone review this please?

yes, I am almost done with the "make everything 5d", then I'll have a look at that

guastinimara:
I'm also reading this at the moment

guastinimara:
I have a doubt about the name, but I don’t know if there was already some discussion about it. If I see an operator called Jacobian, I would expect it to return the Jacobian matrix when applied, given a function and a position... Here instead the forward returns the Jacobian vector product directly (right? as we're interested in the linear approximation of an operator). So I think I'd change the name to something like JacobianVectorProduct or JacobianLinearization idk, and I’d still write it in depth in the docstrings what it’s doing. 

edit: or at least this is what I understood torch.func.jvp does?? let me know maybe I read it wrong

I can write now some comments for the docstring to make it clearer 😊

koflera:
> I have a doubt about the name, but I don’t know if there was already some discussion about it. If I see an operator called Jacobian, I would expect it to return the Jacobian matrix when applied, given a function and a position... Here instead the forward returns the Jacobian vector product directly (right? as we're interested in the linear approximation of an operator). So I think I'd change the name to something like JacobianVectorProduct or JacobianLinearization idk, and I’d still write it in depth in the docstrings what it’s doing.
> 
> edit: or at least this is what I understood torch.func.jvp does?? let me know maybe I read it wrong
> 
> I can write now some comments for the docstring to make it clearer 😊

no no, this is the implementation of the Jacobian as a LinearOperator, which can then be applied to new tensors. I.e. given a non-linear operatr q:V -> W, J(q)[x0] : V -> W , with x \mapsto (J(g)[x0])(x)

src/mrpro/operators/Jacobian.py
```diff
@@ -0,0 +1,111 @@
...
+import torch
+
+from mrpro.operators.LinearOperator import LinearOperator
+from mrpro.operators.Operator import Operator
+
+
+class Jacobian(LinearOperator):
+    """Jacobian of an Operator.
+
+    This operator computes the Jacobian of an operator at a given point x0, i.e. a linearization of the operator.
```
koflera:
```suggestion
    This operator implements the Jacobian of a (non-linear) operator at a given point x0 as a LinearOperator, i.e. a linearization of the operator at the point `x0.`
```

koflera: maybe even include an example? e.g. q(p) = sin(p) linearized x0=0 or so?

src/mrpro/operators/Jacobian.py
```diff
@@ -0,0 +1,111 @@
...
+        Returns
+        -------
+            output tensor
+        """
+        self._f_x0, jvp = torch.func.jvp(self._operator, self._x0, x)
+        return jvp
+
+    @property
+    def value_at_x0(self) -> tuple[torch.Tensor, ...]:
+        """Value of the operator at x0."""
```
koflera:
```suggestion
        """Evaluation of the operator at the point x0."""
```

src/mrpro/operators/Jacobian.py
```diff
@@ -0,0 +1,111 @@
...
+        -------
+            Value of the Taylor approximation at x
+        """
+        delta = tuple(ix - ix0 for ix, ix0 in zip(x, self._x0, strict=False))
+        self._f_x0, jvp = torch.func.jvp(self._operator, self._x0, delta)
+        assert self._f_x0 is not None  # noqa: S101 (hint for mypy)
+        f_x = tuple(ifx + ijvp for ifx, ijvp in zip(self._f_x0, jvp, strict=False))
+        return f_x
+
+    def gauss_newton(self, *x: torch.Tensor) -> tuple[torch.Tensor, ...]:
```
koflera, line 98: isn't this the gram operator?

tests/operators/test_jacobian.py
```diff
@@ -0,0 +1,51 @@
...
+    x0 = rng.float32_tensor(3)
+    x = x0 + 1e-2 * rng.float32_tensor(3)
+    op = L2NormSquared()
+    jacobian = Jacobian(op, x0)
+    (actual,) = jacobian.gauss_newton(x)
+    expected = torch.vdot(x, x0) * 4 * x0  # analytical solution for L2NormSquared
+    torch.testing.assert_close(actual, expected, rtol=1e-3, atol=1e-3)
+
+
+def test_jacobian_valueatx0():
```
koflera:
```suggestion
def test_jacobian_value_at_x0():
```

src/mrpro/operators/Jacobian.py
```diff
@@ -0,0 +1,111 @@
...
+            input tensor
+
+        Returns
+        -------
+            Value of the Taylor approximation at x
+        """
+        delta = tuple(ix - ix0 for ix, ix0 in zip(x, self._x0, strict=False))
+        self._f_x0, jvp = torch.func.jvp(self._operator, self._x0, delta)
+        assert self._f_x0 is not None  # noqa: S101 (hint for mypy)
+        f_x = tuple(ifx + ijvp for ifx, ijvp in zip(self._f_x0, jvp, strict=False))
```
guastinimara: maybe I'd add a comment to remind of the taylor expansion formula here

```suggestion
# compute f(x) = f(x_0) + J_f(x_0)(x - x_0)
f_x = tuple(ifx0 + ijvp for ifx0, ijvp in zip(self._f_x0, jvp, strict=False))

File: tests/operators/test_jacobian.py [hunk 6]
```diff
@@ -0,0 +1,51 @@
+import torch
+from mrpro.operators import Jacobian
+from mrpro.operators.functionals import L2NormSquared
+
+from tests import RandomGenerator
+from tests.helper import dotproduct_adjointness_test
+
+
+def test_jacobian_adjointness():
```
guastinimara: we said tests should always be with None? Like here and in the others

```suggestion
def test_jacobian_adjointness() -> None:

guastinimara (approved):

koflera (commented):
a few small comments

src/mrpro/operators/Jacobian.py
fzimmermann89: yes, Jacobian.gram returns the same (and calls the same)
But the gauss newton approximation of the hessian is a thing i want to give a better name than jacobian.gram -- and a matching docstring,

PR #460, opened 2024-10-21
Add multiple axes unsqueeze and broadcasting to the right

fzimmermann89:
Both are not implemented in pytorch
https://github.com/pytorch/pytorch/issues/9410
https://github.com/pytorch/pytorch/issues/130529
https://github.com/pytorch/pytorch/issues/62148

I moved the usage in the signal models already over.
There might be more uses of this pattern in our code base and future uses. This implementation works with torch.jit.

### DIFF
--- a/src/mrpro/utils/reshape.py
+++ b/src/mrpro/utils/reshape.py
@@ -27,7 +27,7 @@ def unsqueeze_left(x: torch.Tensor, n: int) -> torch.Tensor:
     """Unsqueze multiple times in the leftmost dimension.
 
     Example:
-        tensor with shape (1,2,3) and n=2 would result in tensor with shape (1,1,1,1,2,3)
+        tensor with shape (1,2,3) and n=2 would result in tensor with shape (1,1,1,2,3)
 
 
     Parameters
--- a/src/mrpro/utils/reshape.py
+++ b/src/mrpro/utils/reshape.py
@@ -36,7 +36,7 @@ def unsqueeze_left(x: torch.Tensor, n: int) -> torch.Tensor:
         tensor to unsqueeze
     n
         number of times to unsqueeze
-    
+
     Returns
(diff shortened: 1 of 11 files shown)


fzimmermann89:
This should be merged before #461

src/mrpro/utils/reshape.py
```diff
@@ -0,0 +1,61 @@
...
+        number of times to unsqueeze
+    """
+    return x.reshape(*x.shape, *(n * (1,)))
+
+
+def unsqueeze_left(x: torch.Tensor, n: int) -> torch.Tensor:
+    """Unsqueze multiple times in the leftmost dimension.
+
+    Example:
+        tensor with shape (1,2,3) and n=2 would result in tensor with shape (1,1,1,1,2,3)
```
ckolbPTB: I think there is one "1" too many in the resulting shape

tests/utils/test_reshape.py
```diff
@@ -0,0 +1,25 @@
+import torch
```
ckolbPTB, line 3: Tests should also have doc string as the first line

src/mrpro/utils/reshape.py
```diff
@@ -0,0 +1,61 @@
+"""Tensor reshaping utilities."""
```
ckolbPTB, line 1:
```
Returns
--------
```

is missing

ckolbPTB (approved):
just a few small things

src/mrpro/utils/reshape.py
fzimmermann89:
```suggestion
        tensor with shape (1,2,3) and n=2 would result in tensor with shape (1,1,1,2,3)
```

fzimmermann89:
pre-commit.ci autofix

PR #461, opened 2024-10-21
Python 3.10 support

### LINKED ISSUE #456: Split Docker into two layers ?
I am wondering if it would make sense to split the `install_mrpro.sh` used in our docker file into such that two layers are created, one layer up to
https://github.com/PTB-MR/mrpro/blob/d396143f62bfbaed5dd9b66b86be3182f3467192/docker/install_mrpro.sh#L1-L36
and one layer for https://github.com/PTB-MR/mrpro/blob/d396143f62bfbaed5dd9b66b86be3182f3467192/docker/install_mrpro.sh#L37
as the latter will change more often than the former. We should be able to use caching for the layers that stay the same -- we just need a way to trigger a rebuild if a new pytorch version is released -- maybe we can specify a version in the docker file (related #244), thus a bump would result in a change of the Dockerfileand a rebuild?

fzimmermann89:
Change our code to support python3.10 again --> colab!

Import all typing stuff from `typing_extension` which backports typing features and in current python versions, reexports from typing.

Replace tuple unpacking in typehints by `Unpack`. This look a bit less nice but might be an ok trade-off.
We can revert this commit if we want to drop python3.10 support again

Add a docker container for python 3.10

Refactor docker files: installing from github via clone will created chicken and egg problem as main was not installable in python 3.10. docker containers are now split in steps that might be cached. 

Biggest difference is, our dependencies in the docker container will only update if we touch the pyproject.toml.
If we want to, we can also update them on updates to VERSION, for example.

Closes #456

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ include-package-data = true
 name = "mrpro"
 description = "MR image reconstruction and processing package specifically developed for PyTorch."
 readme = "README.md"
-requires-python = ">=3.11,<3.14"
+requires-python = ">=3.10,<3.14"
 dynamic = ["version"]
 keywords = ["MRI, reconstruction, processing, PyTorch"]
 authors = [
@@ -38,6 +38,7 @@ authors = [
 classifiers = [
     "License :: OSI Approved :: Apache Software License",
     "Programming Language :: Python",
+    "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
     "Programming Language :: Python :: 3 :: Only",
@@ -52,6 +53,7 @@ dependencies = [
     "torchkbnufft>=1.4.0",
     "scipy>=1.12",
     "ptwt>=0.1.8",
+    "typing-extensions>=4.12",
 ]
 
 [project.optional-dependencies]
--- a/src/mrpro/algorithms/optimizers/lbfgs.py
+++ b/src/mrpro/algorithms/optimizers/lbfgs.py
@@ -7,11 +7,11 @@ import torch
 from torch.optim import LBFGS
 
 from mrpro.algorithms.optimizers.OptimizerStatus import OptimizerStatus
-from mrpro.operators.Operator import Operator
+from mrpro.operators.Operator import OperatorType
 
 
 def lbfgs(
-    f: Operator[*tuple[torch.Tensor, ...], tuple[torch.Tensor]],
+    f: OperatorType,
     initial_parameters: Sequence[torch.Tensor],
     lr: float = 1.0,
     max_iter: int = 100,
--- a/src/mrpro/data/_kdata/KDataRemoveOsMixin.py
+++ b/src/mrpro/data/_kdata/KDataRemoveOsMixin.py
@@ -1,9 +1,9 @@
 """Remove oversampling along readout dimension."""
 
 from copy import deepcopy
-from typing import Self
 
 import torch
+from typing_extensions import Self
 
 from mrpro.data._kdata.KDataProtocol import _KDataProtocol
 from mrpro.data.KTrajectory import KTrajectory
@@ -49,7 +49,7 @@ class KDataRemoveOsMixin(_KDataProtocol):
         start_cropped_readout = (self.header.encoding_matrix.x - self.header.recon_matrix.x) // 2
         end_cropped_readout = start_cropped_readout + self.header.recon_matrix.x
 
-        def crop_readout(data_to_crop: torch.Tensor):
+        def crop_readout(data_to_crop: torch.Tensor) -> torch.Tensor:
             # returns a cropped copy
             return data_to_crop[..., start_cropped_readout:end_cropped_readout].clone()
 
@@ -61,7 +61,7 @@ class KDataRemoveOsMixin(_KDataProtocol):
         ks = [self.traj.kz, self.traj.ky, self.traj.kx]
         # only cropped ks that are not broadcasted/singleton along k0
         cropped_ks = [crop_readout(k) if k.shape[-1] > 1 else k.clone() for k in ks]
-        cropped_traj = KTrajectory(*cropped_ks)
+        cropped_traj = KTrajectory(cropped_ks[0], cropped_ks[1], cropped_ks[2])
 
         # Adapt header parameters
         header = deepcopy(self.header)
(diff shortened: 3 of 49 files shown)


fzimmermann89:
Should be merged after #460

fzimmermann89:
@ckolbPTB 
did you have a docker file for python3.10? I think we need stupid if case in the install script:
image

fzimmermann89:
I am getting hit by a chicken and egg problem:
During build of the docker image, mrpro cannot be installed with python 3.10 as we currently install from github main (and there the min version is python 3.11)
This results in missing requirements later on
This results in the tests not passing.
So I have to split this into two PRs.

One PR will enable 3.10 compatibility
Second PR will enable tests for 3.10

...or we would have to fix the docker logic, but this would also be a second PR as I want to keep this one only python3.10 related to be able to just revert it later on

Any other suggestions? @ckolbPTB @schuenke @JoHa0811 ?

fzimmermann89:
This should be merged as two commits to make undo of the python3.10 changes easier in the future.

schuenke:
A few questions out of curiosity:

1. Any idea why https://github.com/PTB-MR/mrpro/actions/runs/11451086815 is still running after ~6h ?
2. Any idea why for the build of py310 and py311 the "Cached" value is ~7%, but for py312 ~28%? [see here](https://github.com/PTB-MR/mrpro/actions/runs/11452222619)
3. In general, splitting the docker build is a good idea imo. However, it seems like it's not resulting in a faster build yet, no? Would this change with the next run?

pyproject.toml
schuenke, line 1: We should add the py310 classifier:
Programming Language :: Python :: 3.10"

src/mrpro/data/_kdata/KDataRemoveOsMixin.py
```diff
@@ -61,7 +61,7 @@ def crop_readout(data_to_crop: torch.Tensor):
         ks = [self.traj.kz, self.traj.ky, self.traj.kx]
         # only cropped ks that are not broadcasted/singleton along k0
         cropped_ks = [crop_readout(k) if k.shape[-1] > 1 else k.clone() for k in ks]
-        cropped_traj = KTrajectory(*cropped_ks)
+        cropped_traj = KTrajectory(cropped_ks[0], cropped_ks[1], cropped_ks[2])
```
schuenke, line 64: actually not that bad IMO

src/mrpro/operators/EndomorphOperator.py
```diff
@@ -172,7 +173,7 @@ def __call__(
         torch.Tensor,
         torch.Tensor,
         torch.Tensor,
-        *tuple[torch.Tensor, ...],
+        Unpack[tuple[torch.Tensor, ...]],
```
schuenke, line 176: fine for me as well

src/mrpro/algorithms/optimizers/lbfgs.py
```diff
@@ -7,11 +7,11 @@
 from torch.optim import LBFGS
 
 from mrpro.algorithms.optimizers.OptimizerStatus import OptimizerStatus
-from mrpro.operators.Operator import Operator
+from mrpro.operators.Operator import TypeAlias
 
 
 def lbfgs(
-    f: Operator[*tuple[torch.Tensor, ...], tuple[torch.Tensor]],
+    f: TypeAlias,
```
schuenke: Shouldn't this be `mrpro.operators.Operator.OperatorType` as in adam?

schuenke (approved):
LGTM in general. Just a few comments. Also check my comment above with the 3 questions please.

ckolbPTB:
Seems to me a lot of work for something which is outdated and we will have to revert soon again. Would it not make more sense to try to overcome the problem of python 3.10 in colab? We are not the only ones complaining about this. Are there no options to upgrade python or run one of our containers in colab (maybe something like this: https://github.com/drengskapur/docker-in-colab)?

fzimmermann89:
> Seems to me a lot of work for something which is outdated and we will have to revert soon again. Would it not make more sense to try to overcome the problem of python 3.10 in colab? We are not the only ones complaining about this. Are there no options to upgrade python or run one of our containers in colab (maybe something like this: https://github.com/drengskapur/docker-in-colab)?

There is no good way to use notebooks in a docker at colab.

When we moved to python 3.11, we wanted to get "Self" to work and I did not know about typing_extensions...

src/mrpro/data/_kdata/KDataRemoveOsMixin.py
fzimmermann89: yeah, this allows mypy to properly check that...

fzimmermann89:
> A few questions out of curiosity:
> 
> 1. Any idea why https://github.com/PTB-MR/mrpro/actions/runs/11451086815 is still running after ~6h ?

I have to manually kill them or wait till they time out. There is a bug in pip that installing packages from '/' does not work and freezes. In docklerfiles, it is common to COPY to '/', run and remove afterwards. Took me a few tries to realize this is the issue.

> 2. Any idea why for the build of py310 and py311 the "Cached" value is ~7%, but for py312 ~28%? [see here](https://github.com/PTB-MR/mrpro/actions/runs/11452222619)

No idea. One possible difference is how far previous runs for this task got. 

> 3. In general, splitting the docker build is a good idea imo. However, it seems like it's not resulting in a faster build yet, no? Would this change with the next run?

It should. Here the pyproject.toml changed, which it will not do most of the time..

fzimmermann89:
image

fzimmermann89:
> > 2. Any idea why for the build of py310 and py311 the "Cached" value is ~7%, but for py312 ~28%? [see here](https://github.com/PTB-MR/mrpro/actions/runs/11452222619)
> 
> No idea. One possible difference is how far previous runs for this task got.
> 
> > 3. In general, splitting the docker build is a good idea imo. However, it seems like it's not resulting in a faster build yet, no? Would this change with the next run?
> 
> It should. Here the pyproject.toml changed, which it will not do most of the time..


Lets try adding a scope for each docker image (py10,py11,py12):
https://docs.docker.com/build/cache/backends/gha/#scope
>Scope is a key used to identify the cache object. By default, it is set to buildkit. If you build multiple images, each build will >overwrite the cache of the previous, leaving only the final cache.

>To preserve the cache for multiple builds, you can specify this scope attribute with a specific name. In the following example, the >cache is set to the image name, to ensure each image gets its own cache:


>GitHub's cache access restrictions, still apply. Only the cache for the current branch, the base branch and the default branch is >accessible by a workflow.

fzimmermann89:
This now also changed when dependencies are updated to newest versions. 
We do no longer update on each push to main all dependencies.
Our normal PRs should not fail because a dependency is updated, this was a bit annoying in the past.

Instead, dependencies are updated if either the pyproject.toml is changed, or the VERSION is changed.

Long term, it would be great to get a dependabot-like action checking for requirements going.

PR #463, opened 2024-10-22
Take NonUniformFastFourierOp out of FourierOp 

ckolbPTB:
- NonUniformFastFourierOp separate to FourierOp
- nufft_typ1/2-functions created in init and used in forward/adjoint (easy to switch to finufft)
- reshaping of trajectory done in init
- data is reshaped to (dim with separate trajectory)(dim with same trajectory)(nufft dims) (possible speed up)
- multiple other dimensions are possible in forward/adjoint
- kzyx and k210 do not have to aligned but if they are not FFTOp and NUFFTOp cannot be used together in FourierOp

Still to do:
- make sure operator can be moved between cpu and gpu


src/mrpro/operators/FourierOp.py
```diff
@@ -87,36 +85,16 @@ def get_traj(traj: KTrajectory, dims: Sequence[int]):
...
-                numpoints=nufft_numpoints,
-                kbwidth=nufft_kbwidth,
+            self._non_uniform_fast_fourier_op = NonUniformFastFourierOp(
+                dim=tuple(self._nufft_dims),
+                recon_matrix=get_spatial_dims(recon_matrix, self._nufft_dims),
+                encoding_matrix=get_spatial_dims(encoding_matrix, self._nufft_dims),
+                traj=traj,
+                nufft_oversampling=nufft_oversampling,
+                nufft_numpoints=nufft_numpoints,
+                nufft_kbwidth=nufft_kbwidth,
```
fzimmermann89: I would remove 
                nufft_numpoints
                nufft_kbwidth
from Fourierop.

It doesn't make sense to change the kbwidth, the number of points seems to be an implementation detail of some nufft implementations, but is not generally available in all. Also, there is no guidance how to choose.

I would only keep the oversampling as the main performance knob and mention in the docstring that it can be lowered to 1.25 to improve performance at the cost of qualitiy.

src/mrpro/operators/FourierOp.py
```diff
@@ -3,16 +3,15 @@
...
 from mrpro.data._kdata.KData import KData
 from mrpro.data.enums import TrajType
 from mrpro.data.KTrajectory import KTrajectory
 from mrpro.data.SpatialDimension import SpatialDimension
 from mrpro.operators.FastFourierOp import FastFourierOp
 from mrpro.operators.LinearOperator import LinearOperator
+from mrpro.operators.NonUniformFastFourierOp import NonUniformFastFourierOp
 
 
 class FourierOp(LinearOperator):
```
fzimmermann89, line 17: change this to use our backward fix

src/mrpro/operators/NonUniformFastFourierOp.py
```diff
@@ -0,0 +1,146 @@
...
+
+import numpy as np
+import torch
+from torchkbnufft import KbNufft, KbNufftAdjoint
+
+from mrpro.data.KTrajectory import KTrajectory
+from mrpro.operators.LinearOperator import LinearOperator
+
+
+class NonUniformFastFourierOp(LinearOperator):
```
fzimmermann89:
```suggestion
class NonUniformFastFourierOp(LinearOperator, adjoint_as_backward=True):
```
should fix torch.func while we continue to use torchkbnufft.

fzimmermann89:
Mid-term, I would still really like to decouple us from torchkbnufft and have a NufftOperator option in the FourierOperator,
where you can either specify give a  TorchKBNufft *class* or a Finufft *class* which all implement a common interface.

Also, long-term, I would like to fix the axes sorting code, such that for example a RPE with fft direction along k1 and kz works, i.e. a mismatch in the order of the axes.

src/mrpro/operators/NonUniformFastFourierOp.py
```diff
@@ -0,0 +1,146 @@
...
+        self,
+        dim: Sequence[int],
+        recon_matrix: Sequence[int],
+        encoding_matrix: Sequence[int],
+        traj: KTrajectory,
+        nufft_oversampling: float = 2.0,
+        nufft_numpoints: int = 6,
+        nufft_kbwidth: float = 2.34,
+    ) -> None:
+        """Non-Uniform Fast Fourier Operator class.
```
fzimmermann89:
```suggestion
        """Initialize Non-Uniform Fast Fourier Operator.
```

fzimmermann89:
> Not sure about if we should pass on traj as KTrajectory 

If not to difficult, yes. Or at least allow both, KTrajectory and tensor inputs. 

> and if it should be the original one or already matching recon_ and encoding_matrix.

In case cartesian fft removes stuff? Does it matter? 
Also, does our current code even work in that case?
Shouldnt for that to work the order of NUFFT and FFT be swapped?

ckolbPTB:
> In case cartesian fft removes stuff? Does it matter?
> Also, does our current code even work in that case?
> Shouldnt for that to work the order of NUFFT and FFT be swapped?

FFT and NUFFT always act on separate dimensions so it does not matter if FFT changes stuff. Only case where it matters would be if the non-uniform trajectory changes along a uniform dimension. But the NUFFT acts along the original k-space uniform dimension so this should be fine.

fzimmermann89:
For this or another PR:
Now that we have both operators separated, I think it would be good to have a test that tests nufft with cartesian coordinates vs an FFT.

fzimmermann89 (approved):

ckolbPTB:
@fzimmermann89 I added a test to verify against the analytic solution. Is this what you had in mind?

fzimmermann89 (commented):
Testing with analytical solution is even better!


Final remarks:

I would remove .forward from the operator calls.
It should generally be __call__ that is called on pytorch modules (which all our operator are) as you can in general set global debugging hooks that only work in that case.
I think we should not give a bad example here.

We could either set fast_fourier_op and nufft op always, just with empty dims, or set them to identity op (which we didn't have when we wrote this)
I am wondering why mypy is not complaining about them possibly being None (or I missed something).

src/mrpro/operators/NonUniformFastFourierOp.py
```diff
@@ -0,0 +1,146 @@
...
+        nufft_oversampling
+            oversampling used for interpolation in non-uniform FFTs
+        nufft_numpoints
+            number of neighbors for interpolation in non-uniform FFTs
+        nufft_kbwidth
+            size of the Kaiser-Bessel kernel interpolation in non-uniform FFTs
+        """
+        super().__init__()
+
+        def get_traj(traj: KTrajectory, dims: Sequence[int]):
```
fzimmermann89: now that the NufftOp can be created separately, there are some cases not handled:
What if dim is empty?
What if dim contains invalid axes?
What if dim contains positive axes?

fzimmermann89: I would say dim is empty, the operator should do nothing.
We should add that dim talks about kx ky kz and not k0 k1 k2 here.
So maybe not dim bit direction?

fzimmermann89 (changes requested):

src/mrpro/operators/NonUniformFastFourierOp.py
fzimmermann89: also, what about fastfourierop with empty dim?

ckolbPTB: > I would say dim is empty, the operator should do nothing.

added

> also, what about fastfourierop with empty dim?

should be fixed in separate PR

ckolbPTB: > What if dim contains invalid axes?
What if dim contains positive axes?

We could transform them to negative axes based on the dimensions of traj and then verify the result is within (-3,-2,-1)?

fzimmermann89: I started to rewrite the dim logic some time ago to fully separate dim und direction, but got side tracked by life.

If I remember and unstand correctly, what we actually mean for nufft by dim is NOT dim, but direction.
So I would maybe make it literal kx ky kz ?

fzimmermann89: It was ok to use numbers while within FourierOp, but as this is now a public interfaces, I would make it a set of literals.

ckolbPTB: I guess we have a similar problem with your FFTOp where we also assume that kz ky kx and k2 k1 k0 align, i.e. if I set dim = -1 then I expect to carry out the FFT along k0 with the parameters of (k)x. 

We could either say that FFTOp and NUFFTOp work in k2 k1 k0 space and we have to provide matrix sizes and trajectory and so on in this space. Then we could do the magic of figuring all this out in the FourierOp. This would leave the FFTOp to be more intuitive but would mean that for the NUFFTOp users would have to figure out the traj by themselves. 

Other option would be to provide dim_kzkykx and dim_k2k1k0 to both operators which would make FFTOp less intuitive but mean that NUFFTOp could get the original trajectory as input.

fzimmermann89: I see basically the same two options here :
- nuffOp gets data and omega in shape (batch_different_traj, batch_same_traj, samples), all reshaping is done in FourierOp. 

- NufftOp to gets 'x','y','z' as a `direction` argument. We transform it any ways always to x,y,z where we use it.
This would mean a minimal change to the code. Nufft does not need dim_k2k1k0 at all.
(Midterm, it would be nice to support in FourierOp also RPE with read-out in z direction, but saved in k0. For this, FFTOp would need both dim and direction. But as I said, this is IMHO a separate piece of work for the future)

Not really sure which I prefer, I think a bit the second for know, as it is easier to implement and switch to the first one if it required later on without losing work. Am I missing substantial downsides of the second option?

ckolbPTB: I think Nufft and Fft need k2 k1 k0 because we need to know which dimensions of kdata to combine to samples and which to use as batch. currently we assume this is the same as kz ky kx and hence do the rearrange of kdata in forward/adjoint using nufft_dims which is in kz ky kx space although we are actually rearranging k2 k1 k0 dimensions.

I would prefer it if FftOp and NufftOp used a similar interface. This I think would be much easier to understand by users.

fzimmermann89: Ok, so option two would be to provide direction Literal['x', 'y', 'z',-3,-2,-1] and dim Literal[-3,-2,-1,'k0,'k1',k2'] to nufft.
FFTOp would still only get dim:int. Maye at some point we can add direction here as well  to support the mixed case.

Option one is still all reshaping in FourierOp

tests/operators/test_non_uniform_fast_fourier_op.py
```diff
@@ -0,0 +1,81 @@
...
+    )
+    (reconstructed_img,) = nufft_op.adjoint(kdata.data)
+
+    # Due to discretisation artifacts the reconstructed image will be different to the reference image. Using standard
+    # testing functions such as numpy.testing.assert_almost_equal fails because there are few voxels with high
+    # differences along the edges of the elliptic objects.
+    assert relative_image_difference(reconstructed_img[0, 0, 0, ...], ismrmrd_cart.img_ref) <= 0.05
+
+
+def test_non_uniform_fast_fourier_op_empgy_dims():
```
ckolbPTB:
```suggestion
def test_non_uniform_fast_fourier_op_empty_dims():
```

src/mrpro/operators/NonUniformFastFourierOp.py
ckolbPTB: A bit more work on this topic:
Now `nufft_directions:Sequence[Literal['x', 'y', 'z',-3,-2,-1]]` is provided by the user to define along which directions the nufft should be performed. Based on this ky, kx and kz are selected for Nufft. In a second step it is checked which k0, k1, k2 dimensions ky, kx and kz define (`nufft_dims`). These are then automatically selected. 

In the forward and adjoint this means that the image is permuted/reshaped based on `nufft_directions`, the k-space trajectory and k-space data based on `nufft_dims`. 

This means users don't have to worry about k2-k1-k0-space but can define everything in x-y-z/kx-ky-kz-space. 

It also means that 
> support in FourierOp also RPE with read-out in z direction, but saved in k0

also works now.

The interface is also now more similar to FFT where we either allow SpatialDimension or Sequence[int] for the recon and encoding matrix.

ckolbPTB:
Still some work to do, but let's decide on the interface first (see comment above on directions and dimensions).

fzimmermann89:
is the [misalignment k210 and kzyx still a problem](https://github.com/PTB-MR/mrpro/pull/463/commits/b11acf09fcaf4687ef0d74f252621cf71d6cc577)  ?

regarding review, let me know if i should have another look at this.
(i will not get to reviewing this today, maybe for next meeting, sorry.)

fzimmermann89:
Can we fix the nufft scaling here as well? I.e manually multiply by
`nufft_scale = torch.tensor(get_spatial_dims(encoding_matrix, nufft_dims)).prod().sqrt().reciprocal()` (or the equivalent with the new names) in both forward in adjoint and disabe nufft the scaling of torchkbnufft?

This makes grid oversampling independent and nufft matches fft.

tests/operators/test_fourier_op.py
```diff
@@ -67,21 +64,50 @@ def test_fourier_op_gram(im_shape, k_shape, nkx, nky, nkz, type_kx, type_ky, typ
...
 @pytest.mark.parametrize(
     ('im_shape', 'k_shape', 'nkx', 'nky', 'nkz', 'type_kx', 'type_ky', 'type_kz'),  # parameter names
     [
-        (  # Cartesian FFT dimensions are not aligned with corresponding k2, k1, k0 dimensions
-            (5, 3, 48, 16, 32),  # im_shape
-            (5, 3, 96, 18, 64),  # k_shape
-            (5, 1, 18, 64),  # nkx
-            (5, 96, 1, 1),  # nky - Cartesian ky dimension defined along k2 rather than k1
-            (5, 1, 18, 64),  # nkz
+        (  # 3d single shot stack of spiral
```
fzimmermann89: Why?

ckolbPTB: Most people don't know what RPE means so I wanted to add an example which is easier to understand, but I guess then the RPE example can be removed.

fzimmermann89: So, does stack of spirals work in general and nky is just wrongly aligned?
I missed that from the comments and thought we now longer support stack of spirals

fzimmermann89: BTW: I thought the plan was now to rearrange in fourierop such that these cases work?

fzimmermann89: Ah, ignore that, Lets fist merge this soon and then improve in the second PR.

fzimmermann89 (approved):
I want to change a couple of things at some point.

but we can also do this in the finufft switch or later. So lets get this in for now.
(I want to use it...)

tests/operators/test_fourier_op.py
fzimmermann89: @ckolbPTB Did I get this right? The issue is only the k2/ky alignment here (I changed the test id and comment)

src/mrpro/operators/NonUniformFastFourierOp.py
```diff
@@ -0,0 +1,428 @@
...
+
+        ```{note}
+        Consider using `~mrpro.operators.FourierOp` instead of this operator. It automatically detects if a non-uniform
+        or regular fast Fourier transformation is required and can also be constructed automatically from
+        a `mrpro.data.KData` object.
+        ````
+
+        ```{note}
+        The NUFFT is scaled such that it matches 'orthonormal' FFT scaling for cartesian trajectories.
+        This from other packages, which apply scaling based on the size of the oversampled grid.
```
ckolbPTB:
```suggestion
        This is different from other packages, which apply scaling based on the size of the oversampled grid.
```

tests/operators/test_fourier_op.py
```diff
@@ -100,20 +141,88 @@ def test_fourier_op_not_supported_traj(im_shape, k_shape, nkx, nky, nkz, type_kx
...
 
-    torch.testing.assert_close(img, img_unsorted)
+    nufft_fourier_op = FourierOp(
+        recon_matrix=recon_matrix,
+        encoding_matrix=encoding_matrix,
+        traj=NufftTrajektory(trajectory.kz, trajectory.ky, trajectory.kx),
+        nufft_oversampling=8.0,
+    )
+
+    (result_normal,) = fourier_op(img)
```
ckolbPTB, line 169:
```suggestion
    (result_fft,) = fourier_op(img)
```

tests/operators/test_fourier_op.py
```diff
@@ -100,20 +141,88 @@ def test_fourier_op_not_supported_traj(im_shape, k_shape, nkx, nky, nkz, type_kx
...
+    (result_nufft,) = nufft_fourier_op(img)
+    torch.testing.assert_close(result_normal, result_nufft, atol=3e-4, rtol=5e-3)
```
ckolbPTB, line 171:
```suggestion
    torch.testing.assert_close(result_fft, result_nufft, atol=3e-4, rtol=5e-3)
```

tests/operators/test_fourier_op.py
```diff
@@ -100,20 +141,88 @@ def test_fourier_op_not_supported_traj(im_shape, k_shape, nkx, nky, nkz, type_kx
...
+
+    nufft_fourier_op = FourierOp(
+        recon_matrix=recon_matrix,
+        encoding_matrix=encoding_matrix,
+        traj=NufftTrajektory(trajectory.kz, trajectory.ky, trajectory.kx),
+        nufft_oversampling=8.0,
+    )
+
+    (k,) = fourier_op(img)
+    (result_normal,) = fourier_op.H(k)
```
ckolbPTB, line 198:
```suggestion
    (result_fft,) = fourier_op.H(k)
```

tests/operators/test_fourier_op.py
```diff
@@ -100,20 +141,88 @@ def test_fourier_op_not_supported_traj(im_shape, k_shape, nkx, nky, nkz, type_kx
...
+    (result_nufft,) = nufft_fourier_op.H(k)
+    torch.testing.assert_close(result_normal, result_nufft, atol=3e-4, rtol=5e-3)
```
ckolbPTB, line 200:
```suggestion
    torch.testing.assert_close(result_fft, result_nufft, atol=3e-4, rtol=5e-3)
```

tests/operators/test_fourier_op.py
```diff
@@ -100,20 +141,88 @@ def test_fourier_op_not_supported_traj(im_shape, k_shape, nkx, nky, nkz, type_kx
...
+    fourier_op = FourierOp(recon_matrix=recon_matrix, encoding_matrix=encoding_matrix, traj=trajectory)
+
+    nufft_fourier_op = FourierOp(
+        recon_matrix=recon_matrix,
+        encoding_matrix=encoding_matrix,
+        traj=NufftTrajektory(trajectory.kz, trajectory.ky, trajectory.kx),
+        nufft_oversampling=8.0,
+    )
+
+    (result_normal,) = fourier_op.gram(img)
```
ckolbPTB, line 226:
```suggestion
    (result_fft,) = fourier_op.gram(img)
```

tests/operators/test_fourier_op.py
```diff
@@ -100,20 +141,88 @@ def test_fourier_op_not_supported_traj(im_shape, k_shape, nkx, nky, nkz, type_kx
...
+    (result_nufft,) = nufft_fourier_op.gram(img)
+    torch.testing.assert_close(result_normal, result_nufft, atol=3e-4, rtol=5e-3)
```
ckolbPTB, line 228:
```suggestion
    torch.testing.assert_close(result_fft, result_nufft, atol=3e-4, rtol=5e-3)
```

ckolbPTB (commented):
lgtm

tests/operators/test_fourier_op.py
fzimmermann89: this is not necessary only fft -- it is using the "normal" fourier op logic, so a mix of fft and nufft.
the result_nufft uses ONLY nufft.

PR #467, opened 2024-10-23
Fix PyTest and Deployment workflow triggers

### LINKED ISSUE #445: PyTest coverage badge in the Readme updated on every pytest run
If I understand it correctly then the pytest coverage badge in our readme is always updated whenever pytest runs:
https://github.com/PTB-MR/mrpro/blob/6c1c8cce8c9aa2553db9ef153f64dd6cfe420fff/.github/workflows/pytest.yml#L87
and hence does not correspond to the current status in main.

We should not update the badge for pull-requests. One option would then be to run the tests again during push to main and only then update the badge. Maybe a better alternative would be to get the badge information during the push to main from the latest Pytest coverage comment of the PR using e.g. https://github.com/marketplace/actions/find-comment

schuenke:
This PR should fix that:
- the coverage badge is created only when PR is merged to main branch
- the CD workflow is triggered when the VERSION file was changed in the main branch, which is then merged into a feature branch

closes #445


schuenke:
IMO this should do what we want without using any additional action etc.

EDIT: thought the pytest workflow would also run on push events. I will adjust it...

fzimmermann89:
This now only discards github ui merges, right? not merges done locally?
Is this an issue?
If yes, we still should consider @ckolbPTB 's suggestions of https://github.com/dorny/paths-filter

schuenke:
> This now only discards github ui merges, right? not merges done locally? Is this an issue? If yes, we still should consider @ckolbPTB 's suggestions of https://github.com/dorny/paths-filter

Liked the idea to do it without the paths-filter action, but forgot about local merges... updated the workflow

schuenke:
Okay, works now as we want I think:
Every change of src/mrpro/VERSION will trigger the deployment workflow, but only if the version differs from the version in main branch, the workflow will run through. We just need to keep in mind that changing the VERSION file is the last thing you should do in a PR or you have to change it another 2 times to trigger it with the same version / date. 

Last question: do we also want to skip the build process as well (as it is implemented now), or do we want to build the package and just dont deploy it? Probably the last option, no? @ckolbPTB @fzimmermann89 

Then I will just delete the `if: steps.changes.outputs.version_changed == 'true'` conditions in the steps of the `build-testpypi-package` job

.github/workflows/deployment.yml
```diff
@@ -12,17 +12,37 @@ jobs:
...
+      - name: Check if VERSION file is modified compared to main
+        uses: dorny/paths-filter@v3
+        id: changes
+        with:
+          base: main
+          filters: |
+            version_changed:
+              - 'src/mrpro/VERSION'
+
+      - name: Check if files in docker or the toml file has been modified
```
ckolbPTB: I don't think this part is needed

.github/workflows/pytest.yml
```diff
@@ -85,8 +88,9 @@ jobs:
           pytest-coverage-path: ./pytest-coverage.txt
           junitxml-path: ./pytest.xml
 
-      - name: Create the Badge
+      - name: Create coverage badge if PR is merged to main
         uses: schneegans/dynamic-badges-action@v1.7.0
+        if: github.event_name == 'push' && github.ref == 'refs/heads/main'
```
ckolbPTB, line 91: Probably `github.event_name == 'push'` is not needed but also not 100% sure.

ckolbPTB (approved):

ckolbPTB:
> Last question: do we also want to skip the build process as well (as it is implemented now), or do we want to build the package and just dont deploy it?

I would just do the build and nothing else

fzimmermann89:
Only skip deployments.

At least whenever pyproect.toml is changed we should try a build, no?

.github/workflows/deployment.yml
schuenke: Copy/Pasted it in for debugging and forgot to remove again. 🙄

.github/workflows/pytest.yml
schuenke: Also not 100% sure, but I think its clearer like this... so I decided to leave it in.

fzimmermann89 (approved):

PR #469, opened 2024-10-23
Add pandoc to docker for docs build

lrlunin:
Sphinx docs build with `nbsphinx` requires `pandoc` to be installed.

Unfortunely, in contrary to `conda`, it's not possible get the `pandoc` binary directly via `pip`.

I guess that dealing with `apt` installation in GitHub actions for docs would not considered as good practice and here is the right place.


lrlunin:
Not required.

PR #470, opened 2024-10-23
Add coil combination using PCA

ckolbPTB:
#181 needs to be merged first


fzimmermann89:
I am pretty sure we decided at some point against using the mixins for KData, as this makes a lot of things more complicated vs just having a short functions directly in KData.py that use existing algorithms and tools. And only use mixing for stuff we reuse.

The only reason we still have the mixing for reaaranging is that I am currently working on the new version and it doesn't make sense to change them if we undo the changes anyways..

src/mrpro/data/_kdata/KData.py
```diff
@@ -250,3 +251,89 @@ def __repr__(self):
             f'{self.header}'
         )
         return out
+
+    def combine_coils(
```
fzimmermann89: should this be compress_coil?
is coil combination not usually the application of the csms, and coilk compression what we do here?

(you are more versed in the terminology of MRI stuff, just asking)

fzimmermann89: should we mention that this is "SCC" in the literature / other software (for example bart ?

tests/data/test_kdata.py
```diff
@@ -469,3 +469,52 @@ def test_KData_remove_readout_os(monkeypatch, random_kheader):
...
+@pytest.mark.parametrize(
+    ('batch_dims', 'joint_dims'),
+    [
+        (None, ...),
+        ((0,), ...),
+        ((-2, -1), ...),
+        (None, (-1, -2, -3)),
+        (None, (0, -1, -2, -3)),
+    ],
+)
```
fzimmermann89, line 555: ids for the tests?

src/mrpro/data/_kdata/KData.py
```diff
@@ -250,3 +251,89 @@ def __repr__(self):
...
+
+        pca_compression_op = PCACompressionOp(data=kdata_coil_combined, n_components=n_combined_coils)
+        (kdata_coil_combined,) = pca_compression_op(kdata_coil_combined)
+
+        # reshape to original dimensions and undo permutation
+        kdata_coil_combined = torch.reshape(
+            kdata_coil_combined, [*permuted_kdata_shape[:-1], n_combined_coils]
+        ).permute(*np.argsort(permute_order))
+
+        return type(self)(self.header, kdata_coil_combined, self.traj)
```
fzimmermann89: should we also copy the header and the traj?

it says it returns a copy in the docstring, but it only copies the data.. traj is still shared.

src/mrpro/data/_kdata/KData.py
ckolbPTB: scc == software channel compression? 

I added the corresponding paper although it is different as most of these approaches use image based ROIs to calculate the compression matrix...

ckolbPTB:
pre-commit.ci autofix

src/mrpro/data/_kdata/KData.py
fzimmermann89: yes, but strangely  in some papers they retrospectively redefine scc to mean  "single coil compression", to distinguish it from methods that use different compression matrices for different spatial location...

fzimmermann89 (approved):

PR #471, opened 2024-10-23
Add pre-commit check for __all__

fzimmermann89:
_Somebody_ wrote a pre-commit check for `__all__`
This now solves the issue of _somebody_ forgetting to add to `__all__`.
The logic now is, that everything imported in an` __init__,py ` will be added alphabetically sorted to `__all__`
There is a` #noqa:ALL[symbol1, symbol2]` option to exclude certain symbols.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,6 +26,13 @@ repos:
     hooks:
       - id: typos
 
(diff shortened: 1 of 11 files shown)


schuenke (approved):
lgtm 👍

PR #475, opened 2024-10-24
Remove superfluous .forward

ckolbPTB:
A little bit too quick with #472 

Some more forwards.

I left this one
https://github.com/PTB-MR/mrpro/blob/96184659d74b7546656dbde6f59cdf99ad11173c/src/mrpro/operators/LinearOperator.py#L399
because I think it makes things clearer.

### DIFF
--- a/.github/workflows/deployment.yml
+++ b/.github/workflows/deployment.yml
@@ -1,4 +1,4 @@
-name: Publish to PyPI
+name: Build and Deploy Package
 
 on:
   push:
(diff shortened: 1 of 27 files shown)


fzimmermann89:
ups.
something strange was going on there on my end.
I removed it but maybe didn't save, but the search still does not find the .forward anymore.
This is somewhat surprising. Search searches in unsaved versions of the files if they are open.

image

fzimmermann89 (approved):

PR #477, opened 2024-10-24
Add dollar signs for latex via mathjax to sphinx

fzimmermann89:
Allows us to use` $math$` latex in sphinx.

Allso, bundle a mathjax for offline use.


fzimmermann89:
Maybe we don't do it, as it would encourage us to write non rst docstrings? @schuenke

schuenke:
> Maybe we don't do it, as it would encourage us to write non rst docstrings? @schuenke

Where do we really need this actually? If it's just for cases like `T_1` instead of `T1` in some example titles, I don't really think we need it. But not sure...

PR #480, opened 2024-10-25
Myst

fzimmermann89:


lrlunin:
Wow, that looks great. The only one small detail (possibly very annoying) is that  `\( ... \)` commands not always rendered perfectly if any notebook (except the 1st or 3d one) in "Example" is chosen:
image

The 1st or 3d:
image

Tested both on Firefox and Chrome last versions. I believe it's a `nb-myst` issue. So we can use it and wait when they'll fix it.

P.S. I could not even think that there is an alternative for `nbsphinx`. Thanks you very much for your research!

fzimmermann89:
For now, I would just use T1 and T2* in the h1 header and title.

fzimmermann89:
and it should be fixed @lrlunin

PR #482, opened 2024-10-25
Add CartesianSamplingOp to FourierOp

### LINKED ISSUE #228: Warning for FourierOp for Cartesian data without CartesianSamplingOp
The FourierOp applies a FFT along Cartesian direction even if they have not previously been resorted correctly with the CartesianSamplingOp. This leads to wrong voxel size/dimensions. 

Suggestion: Check in FourierOp  if FFT can be applied (centered k-space, correct sorting of data points) otherwise use NUFFT. Print warning to suggest that it might be faster to use CartesianSamplingOp before calling FourierOp

Challenge: How to figure out if FFT can be applied...

### LINKED ISSUE #288: DirectReconstruction fails for unsorted Cartesian data
Currently the DirectReconstruction does not utilise the CartesianSamplingOp

### LINKED ISSUE #474: Consider putting Cartesian sampling into FourierOp
We should evaluate creating a Cartesian Sampling op in Fourier Op.

ckolbPTB:
closes #474 
closes #288 
maybe closes #228


ckolbPTB:
need to merge #483 first

src/mrpro/operators/FourierOp.py
```diff
@@ -72,6 +73,7 @@ def get_traj(traj: KTrajectory, dims: Sequence[int]):
                 recon_matrix=get_spatial_dims(recon_matrix, self._fft_dims),
                 encoding_matrix=get_spatial_dims(encoding_matrix, self._fft_dims),
             )
+            self._cart_sampling_op = CartesianSamplingOp(encoding_matrix=encoding_matrix, traj=traj)
```
fzimmermann89: we should maybe set it to None or identity if not required?

fzimmermann89: same with the other ops

src/mrpro/operators/FourierOp.py
```diff
@@ -173,6 +171,11 @@ def forward(self, x: torch.Tensor) -> tuple[torch.Tensor,]:
             shape_nufft_dims = [self._kshape[i] for i in self._nufft_dims]
             x = x.reshape(*permuted_x_shape[: -len(keep_dims)], -1, *shape_nufft_dims)  # -1 is coils
             x = x.permute(*unpermute)
+
+        if len(self._fft_dims):
```
fzimmermann89: no need for len here,
pep8: 
image

tests/conftest.py
```diff
@@ -249,6 +250,19 @@ def create_traj(k_shape, nkx, nky, nkz, sx, sy, sz):
     return trajectory
 
 
+@pytest.fixture(scope='session')
+def ismrmrd_cart(ellipse_phantom, tmp_path_factory):
```
fzimmermann89, line 254: longer name?

fzimmermann89 (approved):
could use a few minor changes and will need a merge after the other, related, PRs -- both I can do.

tests/conftest.py
fzimmermann89: ah, it is just moved. lets keep it as is..

PR #483, opened 2024-10-26
Fix CartesianSamplingOp

ckolbPTB:
Some cases are currently not working for the CartesianSamplingOp

Fixes:
- bug because k2 and k1 were swapped
- if any combination of kz, ky and kx were combined along k2, k1 or k0 no sorting would have have been applied


fzimmermann89:
can you explain in a few words what the issue is and how it is fixed?

fzimmermann89 (approved):
Thanks for the explantion. Lets merge this first, before #485 (in which I will then change the names in the test) and before moving it into the FourierOp

PR #484, opened 2024-10-27
Add arange_

fzimmermann89:
In-place array range along a dimension. 
i.e. filling in the values with the result of a matching arange

Tiny function, but still useful abstraction to describe what is going on. Will be used for setting the acq idx after a reshaping


fzimmermann89:
would also be open to a name change,
fill_range_?
fill_arange_?

fzimmermann89:
anybody? 🥺

tests/utils/test_arange_.py
```diff
@@ -0,0 +1,21 @@
+import pytest
```
ckolbPTB, line 3:
```suggestion
"""Tests for arange_."""

import pytest
```

ckolbPTB (commented):
Probably I would go with `fill_arange_ `

PR #485, opened 2024-10-27
Add human readable ids to test

fzimmermann89:
image

I changed most of the really confusing parameterized names.
There are instances where I in the past used a string in the parameterization just to have a readable test name. In theory, this could also be changed now that I learned about the ids parameter.. 
If anybody notices any other tests where the parameterizations can be named better, feel free to contribute here.

also, rename 'uf' and 'nuf' to uniform and non-uniform etc.


fzimmermann89:
not sure about the signal model names...

fzimmermann89:
@schuenke any idea why here are two coverage comments?

schuenke:
> @schuenke any idea why here are two coverage comments?

hmmm not sure, no. In the checks it says that a previous comment was found and updated for all 3 pytest runs (py 3.10, 3.11 and 3.12). Maybe a bug due to the force-push while a previos test was still running?

tests/conftest.py
```diff
@@ -252,159 +252,161 @@ def create_traj(k_shape, nkx, nky, nkz, sx, sy, sz):
...
-            'uf',  # k0 is uniform
-            'uf',  # k1 is uniform
-            'z',  # k2 is singleton
+        (  # (0) 2d cart mri with 1 coil, no oversampling
+            (1, 1, 1, 96, 128),  # im_shape
+            (1, 1, 1, 96, 128),  # k_shape
+            (1, 1, 1, 128),  # nkx
+            (1, 1, 96, 1),  # nky
+            (1, 1, 1, 1),  # nkz
+            'uniform',  # sx
```
ckolbPTB: I would keep the comments "kx is uniform" and so on because they highlight which of all these parameters determines what.

tests/conftest.py
```diff
@@ -252,159 +252,161 @@ def create_traj(k_shape, nkx, nky, nkz, sx, sy, sz):
...
+            'non-uniform',  # sx
+            'non-uniform',  # sy
+            'uniform',  # sz
+            'non-uniform',  # s0
+            'non-uniform',  # s1
+            'uniform',  # s2
         ),
     ],
+    ids=[
+        '2d_cart_mri_1_coil_no_oversampling',
```
ckolbPTB: `mri` is not needed in the ids

tests/conftest.py
fzimmermann89: Suggestion: renaming sx and s1 to type_kx and type_k1 etc?
Then the "type_k2=uniform" already explains what is going on?

tests/operators/test_cartesian_sampling_op.py
```diff
@@ -69,9 +69,9 @@ def test_cart_sampling_op_fwd_adj(sampling):
     nkx = (2, 1, 1, 60)
     nky = (2, 1, 40, 1)
     nkz = (2, 20, 1, 1)
-    sx = 'uf'
-    sy = 'uf'
-    sz = 'uf'
+    sx = 'uniform'
```
ckolbPTB: I think this should also be type_kx, type_ky and type_kz

ckolbPTB (approved):

PR #488, opened 2024-10-28
UserValues

fzimmermann89:
Make AcqInfo.user_float and .user_int a new dataclass.
This allows all AcqInfo fields to have the same shape and will simplify indexing

PR #489, opened 2024-10-28
Add apply_ to dataclasses

fzimmermann89:
We will need an apply on our dataclasses, both for @ckolbPTB's work as well, and for the rearrangeind und indexing.
As the _to already basically does an recursive apply with memorization, this refactors it such that the _apply can also be used with different functions.

### DIFF
--- a/src/mrpro/data/AcqInfo.py
+++ b/src/mrpro/data/AcqInfo.py
@@ -1,6 +1,6 @@
 """Acquisition information dataclass."""
 
-from collections.abc import Callable, Sequence
+from collections.abc import Sequence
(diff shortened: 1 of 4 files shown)


fzimmermann89:
* #496
* #490
* __->__ #489
* #488

fzimmermann89:
This might be moved outside the movedatamixin at some point, or the mixin will just become a baseclass for our datastuff.
But for now, please already review as-if going int main as is.

ckolbPTB:
I think it would be a good idea to add a test which especially verifies the memo-functionality

fzimmermann89:
moved to #505

PR #490, opened 2024-10-28
acqinfo 5d

fzimmermann89:
Make all tensors in AqcInfo always
(*other=at least (1,) , coils=1, k2, k1, k0=1)


fzimmermann89:
* #496
* __->__ #490
* #489
* #488

PR #492, opened 2024-10-29
Fix unrolled CG autograd

fzimmermann89:
@Stef-Martin wants to backpropagate through cg 🙈
This makes it possible.
Even though it is strictly better to always avoid doing this and instead use a proper gradient for the CG solution, we might still want to support this.

Before, there were in-place modifications of tensors, resulting in anomaly detection.


fzimmermann89:
This has been tested by Stefan -- cherry picking fixes his issues.

Stef-Martin (approved):
Works. Looks good to me

PR #496, opened 2024-10-31
Add CheckDataMixin

fzimmermann89:
The idea of jaxtyping (and its string syntax) applied to our use case:
```python

# Aliases for better readability
Tensor_Idx = Annotated[torch.Tensor, Annotation(dtype=(torch.uint32), shape='*#other 1 #k2 #k1 1')]
""" A torch.Tensor used in AcqIdx"""

Tensor_K1Idx = Annotated[torch.Tensor, Annotation(dtype=(torch.uint32), shape='*#other 1 #k2 k1 1')]
""" A torch.Tensor used in AcqIdx"""

Tensor_K2Idx = Annotated[torch.Tensor, Annotation(dtype=(torch.uint32), shape='*#other 1 k2 #k1 1')]
""" A torch.Tensor used in AcqIdx"""

Tensor_UserIdx = Annotated[torch.Tensor, Annotation(dtype=(torch.uint32), shape='*#other 1 #k2 #k1 1 8')]
""" A torch.Tensor used in AccqIdx"""

# The dataclass
@dataclass
class AcqIdx(CheckedDataMixin, dtype='self.user.dtype', shape='*#other 1 #k2 #k1 1'):
    phase: Tensor_Idx  # a normal idx
    user: Tensor_UserIdx  # has an extra dimension
    k1: Tensor_K1Idx  # does not broadcast along k1
    k2: Tensor_K2Idx  # dos not broadcast along k2
```

for mypy, all fields are just tensors.
there is only some additional information as metadata added to the type hints.
this information is then used in check_invariant to check if the shapes match.
most of the expensive logic is cached, so calling check_invariant mutplie times should not be an issue

### DIFF
--- a/src/mrpro/data/CheckDataMixin.py
+++ b/src/mrpro/data/CheckDataMixin.py
@@ -1,0 +1,489 @@
+"""Dtype and shape checking for dataclasses."""
+
+from collections.abc import Iterator, Mapping, Sequence
+from dataclasses import dataclass, field, fields
+from functools import lru_cache
+from typing import Annotated, Literal, TypeAlias, get_args, get_origin
+
+import torch
+from typing_extensions import Protocol, Self, runtime_checkable
+
+
+@dataclass(frozen=True)
+class _Dim:
+    name: str
+    broadcastable: bool
+
+
+class _NamedDim(_Dim): ...
+
+
+class _FixedDim(_Dim): ...
+
+
+class _NamedVariadicDim(_Dim): ...
+
+
+_anonymous_variadic_dim = '...'
+_anonymous_dim = '_'
+_DimType: TypeAlias = Literal['...', '_'] | _NamedDim | _NamedVariadicDim | _FixedDim
+
+
+class RuntimeCheckError(Exception):
+    """An error in the runtime shape and dtype checks."""
+
+
+class ShapeError(RuntimeCheckError):
+    """A missmatch in the shape of a tensor."""
+
+
+class DtypeError(RuntimeCheckError):
+    """A missmatch in the dtype of a tensor."""
+
+
+class SpecificationError(RuntimeCheckError):
+    """A syntax error in the specification."""
+
+
+@lru_cache
+def parse_dims(dim_str: str) -> tuple[tuple[_DimType, ...], int | None]:
+    if not isinstance(dim_str, str):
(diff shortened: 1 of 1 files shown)


fzimmermann89:
* __->__ #496
* #490
* #489
* #488

fzimmermann89:
needs stil la bit of cleanup and refactoring with regard to caching and the actual mixin class

fzimmermann89:
@ckolbPTB can we chat about this?

@schuenke , the team and I already talked about it shortly in one of the last mrpro meetings.
would be good to get your opinion here as well, before I continue this path

schuenke:
In general, I like the style. However, if ppl are not familiar with it, it might be another burden to contribute. Thus, I am not 100% sure, but also dont have a strong opinion. I guess I am fine with everything you decide as team...

fzimmermann89:
This also applies to Dataclasses, so you can annotate a child dataclass with an expected shape string.
For each dataclass, we could then define also a string describing what its shape is, 
for example for KData` "*other coil k2 k1 k0"`, and on calling `.shape` on a KData object, we would return the filled-in size.

One other reason why this is useful is that for "nice" indexing (and rearranging etc) to work, we could use some description of the axes names: 
If you index in dim -4 in KData, it should be passed on to other objects as indexing in the "coil" dimension, and only data that has a coil dimension will be indexed.

This is a different approach to making everything have the same last 4 dimensions with introducing stuff like something like #488 to store the 16 uservalues.


If we instead go the everything-has-same-last-4 dim approach, it would still be really useful to enforce that acqidx do not differ across coil dimensions.

fzimmermann89:
If it were  only used for checking the shapes in a invariants check after reshaping, it would not be much of a maintenance burden, as you could always just remove it and only use some safety checks.

But if I depdend on it for the rearranging and indexing stuff, it result in some maintenance burden.



I think having good rearranging, indexing etc stuff would really set mrpro apart from just using torchkbnufft on a tensor and make it more intuitive -- and be a good selling point for a paper ;)
I played a bit around, and for know, this approach seems to be the most promising.

Once again, we might be able to get away with only using it for shape checking and manual .shape and __getitem__ on the different dataclasses  -- with the tradeoff of requiring more, but conceptually maybe simpler code.

PR #499, opened 2024-11-05
fix test installation from TestPyPi workflow

schuenke:
I added a retry loop in the installation from TestPyPi step to avoid the workflow to fail because the deployed package from the previous step is not yet fully available for download / installation.

As far as I know, there is no status / availability check available.

### DIFF
--- a/.github/workflows/deployment.yml
+++ b/.github/workflows/deployment.yml
@@ -94,7 +94,15 @@ jobs:
         run: |
           VERSION=${{ needs.build-testpypi-package.outputs.version }}
           SUFFIX=${{ needs.build-testpypi-package.outputs.suffix }}
-          python -m pip install mrpro==$VERSION$SUFFIX --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple/
+          for i in {1..3}; do
+            if python -m pip install mrpro==$VERSION$SUFFIX --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple/; then
+              echo "Package installed successfully."
+              break
+            else
+              echo "Attempt $i failed. Retrying in 10 seconds..."
+              sleep 10
+            fi
+          done
 
   build-pypi-package:
     name: Build Package for PyPI


lrlunin:
In my opinion (which can be potentially wrong) I would rather put this block to the `Publish to TestPyPI` step. This way we would make the upload of the package "smart", in sense that upload will ensure and guarantee that the version is available. It would make more sense in case if цу would like to perform more things related to the package. Then we can just set a trigger to this step ready instead of monitoring the state of the package in each upcoming sub-routine.

schuenke:
> In my opinion (which can be potentially wrong) I would rather put this block to the `Publish to TestPyPI` step. 

I think I would at least make it another step, no? Or am I missing that `pypa/gh-action-pypi-publish` already has this feature implemented?

> This way we would make the upload of the package "smart", in sense that upload will ensure and guarantee that the version is available. 

Assuming `pypa/gh-action-pypi-publish` doesn't support this: what would be your idea to verify the availability? Directly checking in the projects metadata (https://test.pypi.org/pypi/mrpro/json) if the `$VERSION$SUFFIX` is available? 

> It would make more sense in case if цу would like to perform more things related to the package. Then we can just set a trigger to this step ready instead of monitoring the state of the package in each upcoming sub-routine.

Currently, I don't think we will do more checks on the TestPyPi package, but ofc you never know 😄

lrlunin:
> I think I would at least make it another step, no? Or am I missing that `pypa/gh-action-pypi-publish` already has this feature implemented?

Oups, I mean an additonal step in the same group. You're correct.

> Assuming pypa/gh-action-pypi-publish doesn't support this: what would be your idea to verify the availability? Directly checking in the projects metadata (https://test.pypi.org/pypi/mrpro/json) if the $VERSION$SUFFIX is available?

I briefly looked at the existing parameters and didn't find those. I'll [raise an issue](https://github.com/pypa/gh-action-pypi-publish/issues/286) and ask if the the action would like to cover this thing. It sounds very reasonable for me.

fzimmermann89 (approved):

lrlunin (approved):

PR #500, opened 2024-11-05
Revert NamedTemporaryFile ContextManager in example notebooks

schuenke:
Reverts changes to examples from PTB-MR/mrpro#498 because we decided NOT to use a context manager to open the data files and just ignore the RUFF rule in our example notebooks.


fzimmermann89 (approved):

fzimmermann89:
Thanks!

Any objections to merging it?

PR #501, opened 2024-11-07
Add colab badge to notebooks

ckolbPTB:


fzimmermann89:
good idea!

maybe have a 
```
try:
  import mrpro
except ImportError
   !pip install mrpro 
    import mrpro

```
or similar block at the beginning of ll example notebooks?

fzimmermann89:
have you considered writing one script that a) adds the cell and b) calls jupytext? I.e. added the conversion also to the script?
then we could call that script from pre-commit, with the mid-term goal enabling to see all changes to docu/examples also locally without relying on github actions..

ckolbPTB:
Tried to add the 
```
try:
  import mrpro
except ImportError
   !pip install mrpro 
    import mrpro
```
block but in the script it is invalid code and when I add it to the notebook during conversion it is detected as a modification to the notebook which leads to an endless loop of translating py to ipynb -> adding cell -> committing -> translating py ti ipynb -> adding cell -> committing...

lrlunin:
> Tried to add the
> 
> ```
> try:
>   import mrpro
> except ImportError
>    !pip install mrpro 
>     import mrpro
> ```
> 
> block but in the script it is invalid code and when I add it to the notebook during conversion it is detected as a modification to the notebook which leads to an endless loop of translating py to ipynb -> adding cell -> committing -> translating py ti ipynb -> adding cell -> committing...

I would add the following code cell in the `.py` files for the installation of mrpro package.
```python
try:
    import mrpro
except ImportError:
    subprocess.check_call([sys.executable, "-m", "pip", "install", 'mrpro'])
finally:
    import mrpro
```

Unfortunetely it would not be enough only to add this to the notebooks in the docs since the google colab references to the github file representation.

lrlunin:
implemented in #537

PR #503, opened 2024-11-08
Implement Gram Shortcut for FourierOp

fzimmermann89:
Add Gram Operator to the FourierOp and CartesianSamplingOp

FourierOp:
   written as  convolution with kernel, making use of toeplitz structure of nufft. kernel is calculated with 2^n_dim adjoint nuffs, instead of one nufft with twice the size in all dimensions, saving on memory. same approach as done by tensorflow MRI, but with a loop over he combinations instead of recursion. The torchkbnufft adjoint nufft can at some point be replaced by a different nufft operation. Currently, the code prodces the same results as https://torchkbnufft.readthedocs.io/en/stable/_modules/torchkbnufft/_nufft/toep.html#calc_toeplitz_kernel
If we slit nufft op out of FourierOp, we can also move the nufft-related stuff from init and forward 1:1 to new nufft op.

CartesianSamplingOp
   does forward and adjoint pass on an all-ones mask. this could be optimized to instead just set ones in a all-zero mask. but the current form works and is simple.
If we add CartesianSamplingOp to FourierOp, we would also have to include it in FourierOp.gram on the cartesian dimensions. 

Tests compare against Operator.H@Operator as ground-truth.


fzimmermann89:
This will need an adjustment once the CartesianSamplingOp is added to FourierOp. ✔️ 

Should we keep the helper function that creates the convolution kernel next to the Operators or should we move it into algorithms somewhere?

It is IMHO not really an algorithm, just a tiny bit smarter way to do a Nufft with doubled images size in each dimension..

fzimmermann89:
needs #517 ..

fzimmermann89:
We could remove the Gram Operators and instead return directly the composition of the other operators.
But this might be a bit confusing, that FourierOp.gram results in a composition of 2 FourierOps, two multiplications, ...

fzimmermann89:
Also, it could be made a bit more efficient if we add a normalize and fftshift flag to the fftop, we can avoid both by scaling/shifting the kernel instead.

src/mrpro/operators/FourierOp.py
```diff
@@ -223,3 +225,167 @@ def adjoint(self, x: torch.Tensor) -> tuple[torch.Tensor,]:
...
+    Instead, consider using the `gram` property of :class: `mrpro.operators.FourierOp`.
+    """
+
+    _kernel: torch.Tensor | None
+
+    def __init__(self, fourier_op: FourierOp, dcf: torch.Tensor | DcfData | None = None) -> None:
+        """Initialize the gram operator.
+
+        If density compensation weights are provided, they the operator
+        F.H@dcf@F is calculated.
```
koflera, line 319: not directly related to this: but maybe worth checking again what the dcf calculation returns in the algorithms that we use, i.e. is it W^(1/2) or already the entire W... I remember Pierrick and Stefan had some issues with that

src/mrpro/operators/FourierOp.py
```diff
@@ -223,3 +225,167 @@ def adjoint(self, x: torch.Tensor) -> tuple[torch.Tensor,]:
...
+    """Enforce hermitian symmetry on the kernel. Returns only half of the kernel."""
+    flipped = kernel.clone()
+    for d in range(-rank, 0):
+        flipped = flipped.index_select(d, -1 * torch.arange(flipped.shape[d], device=flipped.device) % flipped.size(d))
+    kernel = (kernel + flipped.conj()) / 2
+    last_len = kernel.shape[-1]
+    return kernel[..., : last_len // 2 + 1]
+
+
+def gramm_nufft_kernel(weight: torch.Tensor, trajectory: torch.Tensor, recon_shape: Sequence[int]) -> torch.Tensor:
```
koflera: gram instead of gramm

src/mrpro/operators/FourierOp.py
```diff
@@ -223,3 +225,167 @@ def adjoint(self, x: torch.Tensor) -> tuple[torch.Tensor,]:
...
+    kernel = torch.fft.hfftn(kernel, dim=list(range(-rank, 0)), norm='backward')
+    kernel /= kernel.shape[-rank:].numel()
+    kernel = torch.fft.fftshift(kernel, dim=list(range(-rank, 0)))
+    return kernel
+
+
+class FourierGramOp(LinearOperator):
+    """Gram operator for the Fourier operator.
+
+    Implements the adjoint of the forward operator of the Fourier operator, i.e. the gram operator
```
koflera, line 300: Maybe we could give a very short explanation of what this does, i.e. it constructs an efficient implementation of the operator NUFFT.H @ W @ NUFFT by approximating by F.H @ W' @ F, where  W are the Toeplitz kernels in the Fourier domain, and F.H and F are (I)FFTs (not (I)NUFFTs), respectively.

koflera (approved):
looks good to me

tests/operators/test_fourier_op.py
```diff
@@ -49,6 +49,25 @@ def test_fourier_op_fwd_adj_property(
     dotproduct_adjointness_test(fourier_op, u, v)
 
 
+@COMMON_MR_TRAJECTORIES
+def test_fourier_op_gram(im_shape, k_shape, nkx, nky, nkz, type_kx, type_ky, type_kz, type_k0, type_k1, type_k2):
```
koflera, line 52: maybe also worth testing it when a DCF is also used?

Stef-Martin:
Looks good to me as well, Toeplitz Kernel from tkbnufft and Gram also give very similar results

src/mrpro/operators/FourierOp.py
fzimmermann89: dcf.as_operator() would be W.

PR #505, opened 2024-11-10
Add apply_ to dataclasses

fzimmermann89:
moved from #489 + additional test

### DIFF
--- a/src/mrpro/data/MoveDataMixin.py
+++ b/src/mrpro/data/MoveDataMixin.py
@@ -240,7 +240,11 @@ class MoveDataMixin:
         return new
 
     def apply_(
-        self: Self, function: Callable[[T], T] | None = None, memo: dict[int, Any] | None = None, recurse: bool = True
+        self: Self,
+        function: Callable[[Any], Any] | None = None,
+        *,
+        memo: dict[int, Any] | None = None,
+        recurse: bool = True,
     ) -> Self:
         """Apply a function to all children in-place.
 
--- a/src/mrpro/data/MoveDataMixin.py
+++ b/src/mrpro/data/MoveDataMixin.py
@@ -1,13 +1,13 @@
 """MoveDataMixin."""
 
 import dataclasses
-from collections.abc import Iterator
+from collections.abc import Callable, Iterator
 from copy import copy as shallowcopy
 from copy import deepcopy
-from typing import ClassVar, TypeAlias
+from typing import ClassVar, TypeAlias, cast
 
 import torch
-from typing_extensions import Any, Protocol, Self, overload, runtime_checkable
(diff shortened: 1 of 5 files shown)


src/mrpro/data/MoveDataMixin.py
```diff
@@ -219,26 +223,64 @@ def _mixin_to(obj: MoveDataMixin) -> MoveDataMixin:
...
+    ) -> Self:
+        """Apply a function to all children in-place.
+
+        Parameters
+        ----------
+        function
+            The function to apply to all fields. None is interpreted as a no-op.
+
+        memo
+            A dictionary to keep track of  objects that the function has already been applied to,
```
ckolbPTB:
```suggestion
            A dictionary to keep track of objects that the function has already been applied to,
```

src/mrpro/data/MoveDataMixin.py
```diff
@@ -219,26 +223,64 @@ def _mixin_to(obj: MoveDataMixin) -> MoveDataMixin:
...
+        memo: dict[int, Any] | None = None,
+        recurse: bool = True,
+    ) -> Self:
+        """Apply a function to all children in-place.
+
+        Parameters
+        ----------
+        function
+            The function to apply to all fields. None is interpreted as a no-op.
+
```
ckolbPTB:
```suggestion
```

src/mrpro/data/MoveDataMixin.py
```diff
@@ -219,26 +223,64 @@ def _mixin_to(obj: MoveDataMixin) -> MoveDataMixin:
...
+        memo
+            A dictionary to keep track of  objects that the function has already been applied to,
+            to avoid multiple applications. This is useful if the object has a circular reference.
+
```
ckolbPTB:
```suggestion
```

src/mrpro/data/SpatialDimension.py
```diff
@@ -109,6 +108,16 @@ def from_array_zyx(
 
         return SpatialDimension(z, y, x)
 
+    def apply_(self, function: Callable[[T], T] | None = None, **_) -> Self:
```
ckolbPTB, line 111: why do we need `**_`? For later overloads?

ckolbPTB (approved):

src/mrpro/data/SpatialDimension.py
fzimmermann89: it overloads the one from MoveDataMixin, which has additional keyword only parameters (recurse and memo), which Ithink don#t make sense here. But we could also have them and ignore them..
The only reason for apply_ here is to define it with a function Callable[[T], T], as the function does not have to deal with any other children-types.

PR #506, opened 2024-11-11
Adapt KHeader parameters

fzimmermann89:
- Use orientation instead of directions.
- Remove n_coils
- Use apply_ of SpatialDimension


fzimmermann89:
needs #508

src/mrpro/data/_kdata/KData.py
```diff
@@ -200,10 +200,13 @@ def from_file(
...
         # Finally, reshape and sort the tensors in acqinfo and acqinfo.idx, and kdata.
-        def sort_and_reshape_tensor_fields(input_tensor: torch.Tensor):
-            return rearrange(input_tensor[sort_idx], '(other k2 k1) ... -> other k2 k1 ...', k1=n_k1, k2=n_k2)
-
-        kheader.acq_info = modify_acq_info(sort_and_reshape_tensor_fields, kheader.acq_info)
+        kheader.acq_info.apply_(
+            lambda field: rearrange_acq_info_fields(
+                field[sort_idx], '(other k2 k1) ... -> other k2 k1 ...', k1=n_k1, k2=n_k2
+            )
+            if isinstance(field, torch.Tensor | Rotation)
```
ckolbPTB, line 207: Why do we need `if isinstance` here? There is also the check in rearrange_acq_info_fields...

src/mrpro/data/AcqInfo.py
```diff
@@ -6,23 +6,24 @@
...
+from mrpro.data.Rotation import Rotation
 from mrpro.data.SpatialDimension import SpatialDimension
+from mrpro.utils.unit_conversion import mm_to_m
 
-# Conversion functions for units
-T = TypeVar('T', float, torch.Tensor)
 
+def rearrange_acq_info_fields(field: object, pattern: str, **axes_lengths: dict[str, int]) -> object:
+    """Change the shape of the fields in AcqInfo."""
+    if isinstance(field, Rotation):
```
ckolbPTB, line 20: why only Rotation and torch.Tensor? Will this also cover SpatialDimension and AcqIdx?

src/mrpro/data/_kdata/KData.py
ckolbPTB: Similar comment for the Mixins...

src/mrpro/data/AcqInfo.py
fzimmermann89: Both have a apply_ on its own and we will recurse into them. Only rotation and tensor are things we need to reshape.

src/mrpro/data/_kdata/KData.py
fzimmermann89: Indexing does not work on all fields of a dataclass, only on tensors and rotation

src/mrpro/data/AcqInfo.py
ckolbPTB: and once you recurse into them than the fields of SpatialDimension and AcqIdx will be torch.Tensors?

fzimmermann89: Yes

fzimmermann89: That seems to work 😜

src/mrpro/data/_kdata/KData.py
ckolbPTB: but if it is not a Tensor or a Rotation then rearrange_acq_info_fields simply does nothing, so why do we need to check beforehand too?

fzimmermann89: Because the indexing in the lambda happens first

fzimmermann89: Or we would have to move the indexing also into the rearrange_acq_info function

fzimmermann89: Curently, it is conceptually two chained operations: indexing and reaaranging. Both should pass everything that is neither rotation nor tensor unchanged

ckolbPTB: That makes sense, thanks. We could add SpatialDimension because it also support indexing similiar to the KDataSplitMixin? and then use something similar for KDataSelectMixin?

fzimmermann89: Once again, I think we don't need spatial dimension as we will always recurse into it and only have to deal with x y z if these are tensors.

This also avoids the issue of SpatialDimension [int]

fzimmermann89: I will take another look at these mixins, but won't spent to much time thinking about it as I plan to make them obsolute as soon as possible anyways.
Curently, they work.

ckolbPTB: Also more than happy to use the recurse-feature for SpatialDimension but currently it seems we are handling it differently in different Mixins. would be nice to unify it

ckolbPTB: in SplitMixin we allow Tensor, Rotation and SpatialDimension and in SelectMixin we only allow Tensor

fzimmermann89: Ah, got it. This should all be rotation or tensor

fzimmermann89:
I made the isinstance checks the same now.


There is a bit of an annoying issue with mypy and  typevars bound to unions,
https://github.com/python/mypy/issues/9424
https://github.com/python/mypy/issues/10817#issuecomment-879425161
etc

This makes a typevar  in KDataSplitMixin a bit difficult.
Overloads will not wok due to the different definitions in two code paths not being supported with overloads.

As I said, I don't want to spend to much time on this now, I think it looks ok now ?

ckolbPTB (approved):

Issue #465, opened 2024-10-22
Fast, differentiable fingerprinting dictionary matching


fzimmermann89:
It would be nice to have a Module that does cosine similiary and dotproduct dictionary matching.
It should take care of batched simulation of the data.
It should use a (GPU) efficient matching, maybe LSH or using faiss.
It should implement two different versions of  for the backward:
 - running the signal model again at the match to obtain gradients
 - using a trick such as softmax/gaussian weighting/... over k nearest neighbours or something similar.

This can then be used for MRF or as initial value estimator for other reconstruction methods.

Ideally, it should keep track how often a dictionary entry was matched and have  prune version that removes unused matches.
This would make it really useful for qMRI networks that work with epg signal models

fzimmermann89:
First would be a simple dictionary module.

On creation, we specify a signal function in the __init__.
Then, we have a function to append new simulations to the dictionary.
This function takes in as many tensors as the signal function has parameters, simulates the signals, and stores signals and parameters.
The forward would then do the matching by dotproduct and argmax.
The reason for the split of signal function and creation of the dictionary is to allow here for future expandability.


We should try to use this function in the qMRI example notebooks  and maybe add a fingerprinting example using a phantom.
For this, I think it would make sense to start the development onto of either the epg or the cmrf branch and once it is working nicely, move to code over to main.

(As a warning: Please be carfule not to commit large files to git, such as raw data, images, etc)

fzimmermann89:
We now have a non-differentiable matching operator.

PR #509, opened 2024-11-11
Add DictionaryMatchingOp

fzimmermann89:
Draft of dictionary matching function

Does only non-differentiable argmax matching, but good enough for now.
Needs tests, example, docstrings etc..

Related #465


fzimmermann89:
@rkcatarina : If you join this weeks hackathon, would you be interested to work on this?

examples/example_DictionaryMatchOp.py
```diff
@@ -0,0 +1,154 @@
...
+        self._f = generating_function
+        self.x: list[torch.Tensor] = []
+        self.y = torch.tensor([])
+
+    def append(self, *x: Unpack[Tin]) -> Self:
+        (newy,) = self._f(*x)
+        newy = newy.to(dtype=torch.complex64)
+        newy = newy / torch.linalg.norm(newy, dim=0, keepdim=True)
+        newy = newy.flatten(start_dim=1)
+        newx = x[1]
```
fzimmermann89: Save all the x values, not only the second parameter

So self.x is a list of all the parameters:
Here: [M0,T1]

examples/example_DictionaryMatchOp.py
```diff
@@ -0,0 +1,154 @@
...
+class DictionaryMatchOp(Operator[torch.Tensor, tuple[*Tin]]):
+    def __init__(self, generating_function: Callable[[Unpack[Tin]], tuple[torch.Tensor,]]):
+        super().__init__()
+        self._f = generating_function
+        self.x: list[torch.Tensor] = []
+        self.y = torch.tensor([])
+
+    def append(self, *x: Unpack[Tin]) -> Self:
+        (newy,) = self._f(*x)
+        newy = newy.to(dtype=torch.complex64)
```
fzimmermann89: Keep the dtype of y

examples/example_DictionaryMatchOp.py
```diff
@@ -0,0 +1,154 @@
...
+            self.y = newy
+            return self
+        self.x = [torch.cat(old, new) for old, new in zip(self.x, newx, strict=True)]
+        self.y = torch.cat((self.y, newy))
+        return
+
+    def forward(self, input_signal: torch.Tensor) -> tuple[Unpack[Tin]]:
+        similar = einops.einsum(input_signal, self.y, 't ..., t idx -> idx ...')
+        idx = torch.argmax(torch.abs(similar), dim=0)
+        t1_start_neu = self.x[idx]
```
fzimmermann89: Return instead more than one parameter
x=tuple(xi[idx] for xi in self.x)
Return x

fzimmermann89:
This already looks promising :)

Why did you change from the list of x values (in the initial commit here) to only supporting a single T1?

It would be great if the matching op can be used with multiple parameters (fingerprinting for example)

src/mrpro/operators/DictionaryMatchOp.py
```diff
@@ -53,7 +53,11 @@ def append(self, *x: Unpack[Tin]) -> Self:
 
         """
         (newy,) = self._f(*x)
+        newy = newy.to(dtype=torch.complex64)
```
rkcatarina: The signal_dictionary has to be complex before being used in line 90 in "einops.einsum". Making it complex inside the "einops.einsum" results in wrong results.

src/mrpro/operators/DictionaryMatchOp.py
```diff
@@ -53,7 +53,11 @@ def append(self, *x: Unpack[Tin]) -> Self:
...
         newy = newy / torch.linalg.norm(newy, dim=0, keepdim=True)
+
+        if newy.ndimension() == 1:
```
rkcatarina: Deleted in next commit.

src/mrpro/operators/DictionaryMatchOp.py
```diff
@@ -83,7 +87,7 @@ def forward(self, input_signal: torch.Tensor) -> tuple[Unpack[Tin]]:
         """
         if not self.x:
             raise KeyError('No keys in the dictionary. Please first add some x values using `append`.')
-        similarity = einops.einsum(input_signal, self.y.conj(), 'm ..., m idx  -> idx ...').abs()
+        similarity = einops.einsum(input_signal, self.y, 'm ..., m idx  -> idx ...').abs()
```
rkcatarina: see comment for line 56

src/mrpro/operators/DictionaryMatchOp.py
```diff
@@ -0,0 +1,133 @@
...
+    def __init__(
+        self, generating_function: Callable[[Unpack[Tin]], tuple[torch.Tensor,]], scaling_argument: int | None = None
+    ):
+        """Initialize DictionaryMatchOp.
+
+        Parameters
+        ----------
+        generating_function
+            signal model that takes n inputs and returns a signal y.
+        scaling_argument
```
ckolbPTB: index_of_scaling_parameter
Dictionary matching is insensitive to overall signal scaling as all the signal curves are normalized. A scaling factor (e.g. the equilibrium magnetisation`m0` in `~mrpro.operators.models.InversionRecovery`) is calculated after the dictionary matching if `index_of_scaling_parameter` is not `None`. `index_of_scaling_parameter` then needs to be set to the index of the scaling parameter in the signal model. For `~mrpro.operators.models.InversionRecovery` the parameters are `[m0, t1]` and therefore `index_of_scaling_parameter` should be set to ``0``. The operator will then return `t1` estimated via dictionary matching and `m0` via a post-processing step. If `index_of_scaling_parameter` is `None`, `m0` will be ``1``.

ckolbPTB: A bit long but hopefully understandable..

fzimmermann89: Thanks! Still not perfect, but way better!

src/mrpro/operators/DictionaryMatchOp.py
fzimmermann89: do you have some example code where both ways are compared? 

at least the example currently seems to produce good results

fzimmermann89:
@rkcatarina 
I f'ed up git a bit. So this was force pushed.
When pulling in the changes, make sure to either use `git pull --rebase --autostash`  if you have useful changes locally.
If you don't have useful local changes in the branch, do
```
git fetch
git reset --hard origin/dict
```
to reset your local state to what is on github

tests/operators/test_dictionary_matching_op.py
```diff
@@ -0,0 +1,36 @@
...
+    operator.append(m0, t1)
+
+    m0_matched, t1_matched = operator(y)
+
+    if scaling_argument is not None:
+        torch.testing.assert_close(m0_matched, m0, atol=1e-3, rtol=0.0)
+    torch.testing.assert_close(t1_matched, t1, atol=1e-4, rtol=0.0)
+
+
+# TODO:
```
fzimmermann89: Some todos

tests/operators/test_dictionary_matching_op.py
```diff
@@ -0,0 +1,104 @@
...
+@pytest.mark.parametrize('shape', [(5, 4, 3)])
+@pytest.mark.parametrize('index_of_scaling_parameter', [None, 0], ids=['dont_predict_scale', 'predict_scale'])
+
+def test_dictionaryop_append(shape: tuple[int], dtype: torch.dtype, index_of_scaling_parameter: int | None) -> None:
+    """Test dictionary matching with concatenated entries."""
+    rng_1 = RandomGenerator(2)
+    m0_1 = rng_1.rand_tensor(shape, dtype=dtype, low=0.2, high=1.0)
+    t1_1 = rng_1.rand_tensor(shape, dtype=dtype.to_real(), low=0.1, high=1.0)
+
+    rng_2 = RandomGenerator(3)
```
fzimmermann89: you can reuse the same rng

tests/operators/test_dictionary_matching_op.py
```diff
@@ -0,0 +1,104 @@
...
+    m0_2 = rng_2.rand_tensor(shape, dtype=dtype, low=0.2, high=1.0)
+    t1_2 = rng_2.rand_tensor(shape, dtype=dtype.to_real(), low=0.1, high=1.0)
+
+    #concatenation of the tensors
+    m0_cat = torch.cat((m0_1, m0_2))
```
fzimmermann89, line 57: move this to the     #dictionary matching when appending the individual tensors section

tests/operators/test_dictionary_matching_op.py
```diff
@@ -0,0 +1,104 @@
...
+
+    rng = RandomGenerator(2)
+    model = InversionRecovery(rng.float32_tensor(5))
+    m0 = rng.rand_tensor(shape, dtype=dtype, low=0.2, high=1.0)
+    t1 = rng.rand_tensor(shape, dtype=dtype.to_real(), low=0.1, high=1.0)
+    (y,) = model(m0, t1)
+    operator = DictionaryMatchOp(model, index_of_scaling_parameter=index_of_scaling_parameter)
+    operator.append(m0, t1)
+
+    if index_of_scaling_parameter is not None:
```
fzimmermann89: just remove this test.

fzimmermann89: the cutrrent version does not actually test behavior of the operator.

and in the operator we actually currently seem to support any value for the index, and just wrap it around. 
so even for a two parameter model, you could set it to 1000.
there is no really nice way to check it when creating the operator, only later on...

tests/operators/test_dictionary_matching_op.py
```diff
@@ -0,0 +1,104 @@
...
+
+import pytest
+import torch
+from mrpro.operators import DictionaryMatchOp
+from mrpro.operators.models import InversionRecovery
+
+from tests import RandomGenerator
+
+
+@pytest.mark.parametrize('dtype', [torch.float32, torch.complex64], ids=['float32', 'complex64'])
```
fzimmermann89, line 11: add here a "-2" case for  index_of_scaling_parameter , name it predict_scale_negative_index

fzimmermann89 (commented):
some comments

examples/scripts/qmri_sg_challenge_2024_t1.py
```diff
@@ -222,8 +209,9 @@ def show_images(*images: torch.Tensor, titles: list[str] | None = None) -> None:
 # as $q(TE) = M_0 e^{-TE/T_2^*}$ with the equilibrium magnetization $M_0$, the echo time $TE$, and $T_2^*$.\
 # Give it a try and see if you can obtain good $T_2^*$ maps!
 # ```{note}
-# The echo times $TE$ can be found in `IData.header.te`. A good starting value for $M_0$ is the signal at the shortest
-# echo time. A good starting value for $T_2^*$ is 20 ms.
+# The echo times $TE$ can be found in `IData.header.te`. Instead of using dictionary matching for starting values,
```
ckolbPTB: This sounds as if dictionary matching would not be a good starting value. Maybe write something like "Either dictionary matching can be used to get starting values or ... "

src/mrpro/operators/DictionaryMatchOp.py
```diff
@@ -0,0 +1,141 @@
...
+        self.inverse_norm_y = None if index_of_scaling_parameter is None else torch.tensor([])
+
+    def append(self, *x: Unpack[Tin]) -> Self:
+        """Append `x` values to the dictionary.
+
+        Parameters
+        ----------
+        x
+            points where the signal model will be evaluated. For signal models
+            with n inputs, n Tensors should be provided. Broadcasting is supported.
```
ckolbPTB:
```suggestion
            with n inputs, n tensors should be provided. Broadcasting is supported.
```

src/mrpro/operators/DictionaryMatchOp.py
```diff
@@ -0,0 +1,141 @@
...
+
+        Returns
+        -------
+        Self
```
ckolbPTB: dictionary? Whatever the term, it should be indented

tests/operators/test_dictionary_matching_op.py
```diff
@@ -0,0 +1,89 @@
...
+@pytest.mark.parametrize('shape', [(5, 4, 3)])
+@pytest.mark.parametrize(
+    'index_of_scaling_parameter',
+    [None, -2, 0],
+    ids=['dont_predict_scale', 'predict_scale_negative_index', 'predict_scale'],
+)
+def test_dictionaryop_matching(shape: tuple[int], dtype: torch.dtype, index_of_scaling_parameter: int | None) -> None:
+    """Test dictionary matching with real-valued signals."""
```
ckolbPTB: why real-valued? isn't the input also complex?

tests/operators/test_dictionary_matching_op.py
```diff
@@ -0,0 +1,89 @@
...
+    m0_1 = rng.rand_tensor(shape, dtype=dtype, low=0.2, high=1.0)
+    t1_1 = rng.rand_tensor(shape, dtype=dtype.to_real(), low=0.1, high=1.0)
+
+    m0_2 = rng.rand_tensor(shape, dtype=dtype, low=0.2, high=1.0)
+    t1_2 = rng.rand_tensor(shape, dtype=dtype.to_real(), low=0.1, high=1.0)
+
+    model = InversionRecovery(rng.float32_tensor(5))
+
+    # dictionary matching when appending the individual tensors
+    # concatenation of the tensors
```
ckolbPTB:
```suggestion
```

tests/operators/test_dictionary_matching_op.py
```diff
@@ -0,0 +1,89 @@
...
+    [None, -2, 0],
+    ids=['dont_predict_scale', 'predict_scale_negative_index', 'predict_scale'],
+)
+def test_dictionaryop_no_entry(index_of_scaling_parameter: int | None) -> None:
+    """Test dictionary matching when no entries have been appended before."""
+    rng = RandomGenerator(2)
+    model = InversionRecovery(rng.float32_tensor(5))
+
+    operator = DictionaryMatchOp(model, index_of_scaling_parameter=index_of_scaling_parameter)
+    # create empty signal model
```
ckolbPTB:
```suggestion
```

tests/operators/test_dictionary_matching_op.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+@pytest.mark.parametrize('dtype', [torch.float32, torch.complex64], ids=['float32', 'complex64'])
+@pytest.mark.parametrize('shape', [(5, 4, 3)])
+@pytest.mark.parametrize(
+    'index_of_scaling_parameter',
+    [None, -2, 0],
+    ids=['dont_predict_scale', 'predict_scale_negative_index', 'predict_scale'],
+)
+def test_dictionaryop_matching(shape: tuple[int], dtype: torch.dtype, index_of_scaling_parameter: int | None) -> None:
```
ckolbPTB:
```suggestion
def test_dictionary_matching_op(shape: tuple[int], dtype: torch.dtype, index_of_scaling_parameter: int | None) -> None:
```

tests/operators/test_dictionary_matching_op.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+@pytest.mark.parametrize('dtype', [torch.float32, torch.complex64], ids=['float32', 'complex64'])
+@pytest.mark.parametrize('shape', [(5, 4, 3)])
+@pytest.mark.parametrize(
+    'index_of_scaling_parameter',
+    [None, -2, 0],
+    ids=['dont_predict_scale', 'predict_scale_negative_index', 'predict_scale'],
+)
+def test_dictionaryop_append(shape: tuple[int], dtype: torch.dtype, index_of_scaling_parameter: int | None) -> None:
```
ckolbPTB:
```suggestion
def test_dictionary_matching_op_append(shape: tuple[int], dtype: torch.dtype, index_of_scaling_parameter: int | None) -> None:
```

tests/operators/test_dictionary_matching_op.py
```diff
@@ -0,0 +1,89 @@
...
+        torch.testing.assert_close(m0_matched, m0_matched_cat, atol=1e-3, rtol=0.0)
+    torch.testing.assert_close(t1_matched, t1_matched_cat, atol=1e-4, rtol=0.0)
+
+
+@pytest.mark.parametrize(
+    'index_of_scaling_parameter',
+    [None, -2, 0],
+    ids=['dont_predict_scale', 'predict_scale_negative_index', 'predict_scale'],
+)
+def test_dictionaryop_no_entry(index_of_scaling_parameter: int | None) -> None:
```
ckolbPTB:
```suggestion
def test_dictionary_matching_op_empty_dictionary(index_of_scaling_parameter: int | None) -> None:
```

ckolbPTB (approved):
Just a few small things.

src/mrpro/operators/DictionaryMatchOp.py
fzimmermann89:
```suggestion
            Self
```

tests/operators/test_dictionary_matching_op.py
fzimmermann89:
```suggestion
    """Test dictionary matching."""
```

examples/scripts/qmri_sg_challenge_2024_t1.py
```diff
@@ -225,8 +213,9 @@ def show_images(*images: torch.Tensor, titles: list[str] | None = None) -> None:
...
+# a good starting value for $M_0$ is the signal at the shortest echo time
+# and a good starting value for $T_2^*$ is 20 ms.
```
fzimmermann89:
```suggestion
# The echo times $TE$ can be found in `IData.header.te`. As starting values, either dictionary matching, or
# the signal at the shortest echo time for $M_0$ and 20 ms for $T_2^*$ can be used.
```

PR #521, opened 2024-11-13
Include .ipynb to sphinx-docs via myst-nb

lrlunin:
This pull request includes the previosly executed .ipynb notebooks to the docs via `myst_nb` instead of copying of .html files internals to the .rst file.


fzimmermann89:
can we, for now, just move this to the mrpro repo to have the doc action run once?
or do you think it is feasible to fix the actions to run also run safely for forks?

lrlunin:
> can we, for now, just move this to the mrpro repo to have the doc action run once? or do you think it is feasible to fix the actions to run also run safely for forks?

Yeah, sure. give me a minute.

Issue #487, opened 2024-10-28
Create KTrajectory object without full header


JoHa0811:
It would be beneficial if we could create KTrajectory objects by hand without having to supply all header information.

fzimmermann89:
### Why do we want a it
For prototyping, it is nice to get FourierOp for different trajectories.

### How
one way could be to have two functions
- one to create from KHeader (the current `__call__`)
- one where you have to specify everything manually that is used by the calculator.
   This would have a different signature for each calculator, but will only be used in manual stuff.

We would have to decide which one is `__call__` and what to call the other one. It would be nice to be able to implement the current` __call__` as a wrapper around the new function that only picks required information out of the header. This also will be a different function in all calculators.

JoHa0811:
After discussion in the group:
It might be most straight-forward to implement it for radial and spiral trajectories first. More discussion is needed if a KTraj without header is acutally sensible for Cartesian trajectories.

fzimmermann89:
done in #560

PR #522, opened 2024-11-13
Ktraj no header

JoHa0811:
in regards to #487

### DIFF
--- a/src/mrpro/data/traj_calculators/KTrajectoryRadial2D.py
+++ b/src/mrpro/data/traj_calculators/KTrajectoryRadial2D.py
@@ -51,6 +51,7 @@ class KTrajectoryRadial2D(KTrajectoryCalculator):
         """Generate KTrajectory object without header information.
 
         Args:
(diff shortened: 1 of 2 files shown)


src/mrpro/data/traj_calculators/KTrajectoryRadial2D.py
```diff
@@ -46,3 +46,32 @@ def __call__(self, kheader: KHeader) -> KTrajectory:
         kz = torch.zeros(1, 1, 1, 1)
 
         return KTrajectory(kz, ky, kx)
+
+    def create_manually(self, n_spokes: int = 240, n_k0: int = 256, initial_angle: float = 0) -> KTrajectory:
```
ckolbPTB, line 50: I think I would prefer it if we would give krad:torch.Tensor and kang_index:torch.Tensor as input to this function. Then we could replace the code in `__call__` with this function.

ckolbPTB: or say krad: torch.Tensor | int and if it is int then you calculate it assuming symmetric sampling and something similar for kang_index

PR #525, opened 2024-11-14
Fix CartesianSamplingOp device

### LINKED ISSUE #502: cartesian sampling operator can't be created for trajectory already on gpu
This could also be an issue for other operators.
It should be possible to create them even after moving data to cuda.
Currently, you would have to move to CPU, create, move to GPU again..

bernharb:
Fixes #502


bernharb:
Originally there were 3 commits. Before my 2nd commit I forgot to pre-commit and ended up with some ruff formatting errors. After spending some time on trying to fix them (primarily fighting with git commands and pre-commit messages) I realised that the changes were applied automatically once I pre-committed. This realization came AFTER I had already tried to reset my local branch to ammend my commit among other things. Not the ideal workflow I understand.
Nevertheless, instead of having three commits: change device according to option 1, now change to option 2, formatting corrections
there is now only one commit left that contains the accumulated changes but has message of the last commit.
I hope this clears the confusion. Do we still wish to ammend the message?

src/mrpro/operators/CartesianSamplingOp.py
```diff
@@ -41,28 +41,36 @@ def __init__(self, encoding_matrix: SpatialDimension[int], traj: KTrajectory) ->
 
         # Cache as these might take some time to compute
         traj_type_kzyx = traj.type_along_kzyx
+        traj_device = traj.device
         ktraj_tensor = traj.as_tensor()
 
         # If a dimension is irregular or singleton, we will not perform any reordering
         # in it and the shape of data will remain.
         # only dimensions on a cartesian grid will be reordered.
+        # The device of the input is matched.
```
ckolbPTB:
```suggestion
        # The device of the input trajectory is matched.
```

ckolbPTB:
don't worry about the individual commits, they will be squashed into one at the end anyway

ckolbPTB:
would it make sense to add a test for this following https://github.com/PTB-MR/mrpro/blob/38722bf9c9aad87bb30008d6ada488e69f239cd1/tests/operators/test_cartesian_sampling_op.py#L107 but then moving the trajectory and the data to cuda? We would currently have to label it `@pytest.mark.cuda` but at least we can test it locally and have it for later

src/mrpro/operators/CartesianSamplingOp.py
```diff
@@ -41,28 +41,36 @@ def __init__(self, encoding_matrix: SpatialDimension[int], traj: KTrajectory) ->
...
+        traj_device = traj.device
         ktraj_tensor = traj.as_tensor()
 
         # If a dimension is irregular or singleton, we will not perform any reordering
         # in it and the shape of data will remain.
         # only dimensions on a cartesian grid will be reordered.
+        # The device of the input trajectory is matched.
         if traj_type_kzyx[-1] == TrajType.ONGRID:  # kx
-            kx_idx = ktraj_tensor[-1, ...].round().to(dtype=torch.int64) + sorted_grid_shape.x // 2
+            kx_idx = ktraj_tensor[-1, ...].round().to(dtype=torch.int64, device=traj_device) + sorted_grid_shape.x // 2
```
fzimmermann89: already on the device?

bernharb: you're right, it is. better to remove the redundancy then?

fzimmermann89:
pip install -e .

ckolbPTB (approved):
Looks good. Please make sure you run the tests of the final version on a GPU system.

fzimmermann89:
@bernharb Have you run this on a cuda system?
If yes, anything that needs to be done before merging?

bernharb:
yes, @fzimmermann89 I ran pytest -m cuda on e81351. the only thing that fails is the rotation test that was fixed by a different PR in the meantime, so as long as this is considered, I believe there shouldn't be a problem 
or do you have something in particular in mind I should check?

fzimmermann89:
No, I was just following up on Christophs question and wondering why we haven't merged this yet

So this is ready?

bernharb:
Understood. In my opinion it's ready

PR #535, opened 2024-11-14
Swap order of target and weight in functionals

fzimmermann89:
I believe this is the more natural use, especially for MSE.

Also, forces a target in MSE.
We cant change the default or the type hints though, due to substitution principle.

Lets see if that fixes the notebooks, where @ckolbPTB, @rkcatarina and @fzimmermann89 all missed this bug.

### DIFF
--- a/src/mrpro/operators/Functional.py
+++ b/src/mrpro/operators/Functional.py
@@ -50,8 +50,8 @@ class ElementaryFunctional(Functional):
 
     def __init__(
         self,
-        weight: torch.Tensor | complex = 1.0,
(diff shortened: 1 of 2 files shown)


ckolbPTB:
> Also, forces a target in MSE.

Although useful I would not do that to keep it better aligned with the behaviour of the other functionals and make testing easier. If somebody wants to do MSE on a tensor without a target, let them.

fzimmermann89:
forgot about the tests...

ckolbPTB (approved):

PR #537, opened 2024-11-14
Update .ipynb notebooks from .py in pre-commit hook locally

### LINKED ISSUE #486: Move jupytext from action to pre commit
We should consider moving jupytext from a GitHub action to a pre-commit hook. This would make local  debugging of the documentation easier. I see so far no disadvantages - am I missing something?

### LINKED ISSUE #546: Split examples into scripts & notebooks subfolders
The number of example scripts/notebooks increased quite a lot already and will probably increase further in the future. FI was wondering if it might be a clearer structure if we split the `examples` folder into a `scripts` and a `notebooks` subfolder. Furthermore, it happened to me several times that I clicked on a notebook when I wanted to open a script and vise versa. 

The new structure could look like this:

```
├── ...
├── examples
│   ├── notebooks
│   │   └── *.ipynb files
│   ├── scripts
│   │   └── *.py files
│   └── ruff.toml
└── ...
```

Or we could leave the scripts in the `examples` folder and only move the *.ipynb files into a `notebooks` subfolder:

```
├── ...
├── examples
│   ├── notebooks
│   │   └── *.ipynb files
│   ├── *.py files
│   └── ruff.toml
└── ...
```

Any opinions?

lrlunin:
see #504


pyproject.toml
```diff
@@ -177,6 +177,9 @@ inline-quotes = "single"
 [tool.ruff.lint.pydocstyle]
 convention = "numpy"
 
+[tool.ruff.lint.per-file-ignores]
+"examples/**.py" = ["ERA001"] # py:percent preamble handled as commented out code
```
fzimmermann89: There is a ruff config in examples

lrlunin: Absolutely correct, fixing now.

lrlunin: resolved in 0600a53

examples/scripts/direct_reconstruction.py
```diff
@@ -1,3 +1,19 @@
+# ---
```
ckolbPTB: Having this part in every script is very error prone. 

Can we add it only during the translation process from py to ipynb without committing it?

lrlunin: I think it's a mandatory for `jupytext` because the `.py` file needs to be an exakt representation of `.ipynb` while `.ipynb` do have more metadata.

For now user can apply the changes either in `.py` file or in `.ipynb` file and they will be synced. I beliebe it is also much more user friendly and intuitive to modify the `.ipynb` files.

Actually the `.py` files are not required anymore since the main reason to have them was to be sure that the `.ipynb` notebooks are empty in the repository. In the current pre-commit setup the notebooks are also cleaned and are not affected by user executions.

This metadata preamble is pretty helpful for the "Open in Colab" feature since we can specify the
```
# ---
# jupyter:
#   accelerator: GPU
#   colab:
#     gputype: T4
#     provenance: []
# ---
```
for the Colab to open the CUDA notebooks on the GPU host.

Yes, we can also specify these things once in the notebook metadata and hope that no one overrides it later. However, I think this preamble makes it more transparent at the end.

ckolbPTB: There is a lot of version information in there which would be a lot of work to keep up-to-date going forward.

Jupytext did not need this before but it was provided via command line arguments. Maybe we can try to get something similar working here. 

I would not encourage modifying notebooks and then update the scripts but it should always work as script -> notebook. Otherwise I still believe it is far too dangerous that people commit notebook content and make our repo grow big again.

lrlunin: > There is a lot of version information in there which would be a lot of work to keep up-to-date going forward.

This version informations are only jupytext converter information which is now fixed in `.pre-commit-config.yaml`. The `kernelspec` information is set only as default value once and will not be changed again.

> Otherwise I still believe it is far too dangerous that people commit notebook content and make our repo grow big again.

We have a `nbstripout` in pre-commit which cleans outputs and any other metadata added by user executution of a notebook. This also happes **before** jupytext does any synchronization attemps. So there is no chance by now how the execution of the notebook by the user can affect the .ipynb and .py data.

I cannot see a scenario how to ruin the repo with this workaround.

The only changes in notebok cells will affect the .ipynb and .py files. So if any code line is changed either in .py or .ipynb it will be reflected in the paired representation. Isn't it more intuitive for a developer than:
1) changing the .py file
2) then execute the jupytext to get an updated .ipynb representation
3) run the .ipynb to verify the expected behaviour
4) if the behaviour is then expected -> go to 1.

ckolbPTB: We cannot enforce that people run a pre-commit before committing so it is actually really easy to add the notebook content to the repo. 

We would need to enforce that people can only contribute via a fork but as you have experienced that does not work with our pipeline at the moment.

lrlunin: > We cannot enforce that people run a pre-commit before committing so it is actually really easy to add the notebook content to the repo.

Okay, this makes sense. But in this case there is also no guaranteed option to protect the repo.

> that does not work with our pipeline at the moment.

@fzimmermann89 and me found the issue with the CI, I have fixed it. The other issue with the fork pull-requests was the repositroy on-site procedure with .py -> .ipynb -> bot commit. So with this pull request we would solve the 2nd problem. After this two pull-requests merged the "fork + PR" workaround will work and there the main repo could not be poluted with the notebooks data anymore.

The user without `pre-commit` installed will only polute its own fork and pull-request with poluted notebooks will declined by `pre-commit [ci]` workflow.

examples/add_badge_in_pipe.py
```diff
@@ -0,0 +1,30 @@
+import sys
```
fzimmermann89: we should store this file somehere else and add a docstring in the beginning when and why it is used

fzimmermann89: mabye next to the pre commit config, also with name staring with a dot?

fzimmermann89:
Do we need the _notebooks folder?
Can we store the executed notebooks in the same directory as the non-executed ones?
Then, if we build the docs locally, we would still get the examples, just without results?

lrlunin:
> Do we need the _notebooks folder?
> 
> Can we store the executed notebooks in the same directory as the non-executed ones?
> 
> Then, if we build the docs locally, we would still get the examples, just without results?

As far as I concerned the problem is that the sphinx is unable to list the directories above the `docs` root.

And symlink would not work on windows.

examples/add_badge_in_pipe.py
lrlunin: Usually there is a folder `script(s)` in the root of repo for this kind of things.

fzimmermann89:
two things:
- `%pip install -q ` to hide the install output?
- maybe move your new commit hooks before the mypy hook, mypy is the slowest check, thus I think it would make sense to have it last.

lrlunin:
> two things:
> 
> * `%pip install -q ` to hide the install output?
> * maybe move your new commit hooks before the mypy hook, mypy is the slowest check, thus I think it would make sense to have it last.

1. Would not the supressed output of `%pip install` confuse a user and make an impression that installation did not work at all? Especially since the mrpro installation takes pretty long. I would be suspecious if the cell hangs for 5 minutes and does not have anyy output.
2. Brillant idea (as always)

fzimmermann89 (approved):

PR #539, opened 2024-11-14
Use Gram Operator in iterative reco and operator norm

### LINKED ISSUE #481: Remove DCF from iterative SENSE algorithm
How important is the weighted objective compared to only using a DCF corrected x0?
The one comparison in the Prussmann paper gives a slight advantage for >30 iterations, disadvantage for fewer.

https://arxiv.org/abs/1902.09657 ( Ong, Uecker, Lustig) also argue that the reweighting of the objective function is suboptimal due to the noise reshaping. the maximum likelihood solution would be without the dcf. they suggest proper preconditioning using different algorithms.

_Originally posted by @fzimmermann89 in https://github.com/PTB-MR/mrpro/issues/388#issuecomment-2422275281_

fzimmermann89:
Closes #481


fzimmermann89:
superseeded by #662 and #682

Issue #381, opened 2024-08-13
Undersampled Image Reconstruction using operators


Pierrickkk:
I am generating k-space data using the Ellipses phantom and performing direct reconstruction to obtain a fully sampled image, which serves as the ground truth from which I want to retrospectively simulate k-space data.

To do that, I split the k-space data to create a direct_reconstruction object from the undersampled k-space data.

I then create an acquisition operator to retrospectively generate the k-space data from the ground-truth image.

The image reconstructed from this undersampled data (`xu`) should ideally match the direct reconstruction of the undersampled data (`x_us`). More precisely, they should also match the ground-truth image in terms of their overall identity, such that when I compute the point-wise error images, only artefacts are left.

The problem is that while `x_us` matches the fully sampled ground truth (`x_fullsamp`) and looks correct, `xu` isn't. The issue persists despite using the density compensation function and Fourier transform from the direct reconstruction.

This is a code that you can try  : 
```
# %% NOT IMPORTANT Import and data 

import mrpro
from mrpro.data import SpatialDimension
import numpy as np
import torch
import torch.nn.functional as F
import matplotlib.pyplot as plt
from mrpro.algorithms.reconstruction.DirectReconstruction import DirectReconstruction
from mrpro.data._kdata.KData import KData  
from mrpro.data.KTrajectory import KTrajectory
from mrpro.data.traj_calculators.KTrajectoryPulseq import KTrajectoryPulseq
from mrpro.operators.FourierOp import FourierOp
from mrpro.phantoms.EllipsePhantom import EllipsePhantom  
from mrpro.phantoms.phantom_elements import EllipseParameters
from mrpro.operators.DensityCompensationOp import DensityCompensationOp


base_path = '/home/global/mrpro_issue381/'
h5_path = base_path + 'pulseq_spiral_2D_220k0_128interleaves_golden_angle_vds_with_traj.h5'
seq_path = base_path + '20240319_spiral_2D_256mm_220k0_128interleaves_golden_angle_vds.seq'



def shift_k_space_trajectory(kdatapuls: KData) -> KData:
    
    # Extract k-space trajectory
    ky_pulseq = kdatapuls.traj.ky
    kx_pulseq = kdatapuls.traj.kx
    kz_pulseq = kdatapuls.traj.kz

    # Number of indices and initilisation
    num_indices = ky_pulseq.shape[2]
    shifted_ky = ky_pulseq.clone()
    shifted_kx = kx_pulseq.clone()

[...]

PR #542, opened 2024-11-15
FourierOp Norm

fzimmermann89:
At least two potential issues with Fourier Op:
- I thought to norm of FourierOp should always be 1?
- Forcing use of nufft should result in the same as using the cartesian operator -- just slower, or?

Currently, the answer two both is "No" 
This adds the tests showcasing the potential problems

### DIFF
--- a/src/mrpro/data/SpatialDimension.py
+++ b/src/mrpro/data/SpatialDimension.py
@@ -414,3 +414,7 @@ class SpatialDimension(MoveDataMixin, Generic[T_co]):
             return self.x.shape
         else:
             raise ValueError('Inconsistent shapes')
+
+    def prod(self: SpatialDimension[T_co]) -> T_co:
+        """Calculate the product x*y*z."""
+        return self.x * self.y * self.z
--- a/src/mrpro/operators/DensityCompensationOp.py
+++ b/src/mrpro/operators/DensityCompensationOp.py
@@ -1,6 +1,7 @@
 """Class for Density Compensation Operator."""
 
 import torch
+from typing_extensions import Self
 
 from mrpro.data.DcfData import DcfData
 from mrpro.operators.EinsumOp import EinsumOp
@@ -25,3 +26,9 @@ class DensityCompensationOp(EinsumOp):
         else:
             dcf_tensor = dcf
         super().__init__(dcf_tensor, '... k2 k1 k0 ,... coil k2 k1 k0 ->... coil k2 k1 k0')
+
+    def __pow__(self, exponent: float) -> Self:
+        """Raise the operator to a power."""
+        if not exponent > 0:
+            raise NotImplementedError('Can only raise to positive powers')
+        return type(self)(dcf=self.matrix**exponent)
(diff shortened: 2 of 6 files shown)


fzimmermann89:
Note to myself: 
wondering if this could also cause issues with self-supervised training.

Should check if F(subset of trajectory)(image) = F(image)[subset] -- which might also be a good test.

fzimmermann89:
Also, what was the outcome of #381 ?

fzimmermann89:
@ckolbPTB @schuenke

ckolbPTB:
> I thought to norm of FourierOp should always be 1?

I don't think this holds for arbitrary FourierOps. If a FourierOp only yields an undersampled k-space then the Norm will not be 1. Also if padding or cropping of image-space/k-space is required (mismatch of recon_matrix and encoding_matrix) the norm will not be 1. 

Here I verified that FFT and NUFFT lead to the same result:
https://github.com/PTB-MR/mrpro/blob/3c6d873cd72afe1feda0abba93d677036d783b9f/tests/operators/test_non_uniform_fast_fourier_op.py#L51

for recon_matrix == encoding_matrix and fully-sampled k-space. 

The oversampling-parameter of the NUFFT I am not sure about. All the packages I know do not ensure that the norm is independent of this parameter but this does not mean it is the correct thing to do.

fzimmermann89:
It should also match for all of the common MR trajectories. Otherwise, a tiny trajectory change would result in a scaled image -- which would be quite surprising.
Ill try if I can make these tests pass..

koflera:
> > I thought to norm of FourierOp should always be 1?
> 
> I don't think this holds for arbitrary FourierOps. If a FourierOp only yields an undersampled k-space then the Norm will not be 1. Also if padding or cropping of image-space/k-space is required (mismatch of recon_matrix and encoding_matrix) the norm will not be 1.
> 
> Here I verified that FFT and NUFFT lead to the same result:
> 
> https://github.com/PTB-MR/mrpro/blob/3c6d873cd72afe1feda0abba93d677036d783b9f/tests/operators/test_non_uniform_fast_fourier_op.py#L51
> 
> for recon_matrix == encoding_matrix and fully-sampled k-space.
> 
> The oversampling-parameter of the NUFFT I am not sure about. All the packages I know do not ensure that the norm is independent of this parameter but this does not mean it is the correct thing to do.

For the Fast Fourier operator, the operator norm should always be be one (if norm="ortho" is used, which we always have, right?), regardless of full or undersampling. For the general FourierOp, the norm will indeed not always be one.

fzimmermann89:
My current opinion:

As we calculate the operator norm using the gram, it should be independent of the norm setting in the fft as long as it's the same in fft and ifft (we only use fft.gram for our operator norm)

Nufft and fft should match for on grid trajectories.

Nufft scaling should be independent of grid size.

An undersampled fft should always have norm <=1:
The eigenvektor with energy only at sampled k space points has the highest ew. If and only if this is also ev to ew 1 of the fully sampled fft, the norm will be one. (Imdersamplimg only removes eigenvalues)

A zero padded fft should have norm 1
It's eogencektor will be the zero padded eogencektor of the normal fft.


I think depending on how to normalize, we can either make the zero padded(oversampled) or undersampled fft have norm 1.

What can't be norm 1 are transforms with row sum>1 - repeated samples in the Nufft.
Here, dcf@fourier should have norm 1

fzimmermann89:
closed by #463

PR #547, opened 2024-11-20
Add apply (out of place)

fzimmermann89:
Copies a dataclass, then applies.

@ckolbPTB made a mistake of using .apply_ to get a converted value in #421, changing the original value (if I am not mistaken)
Maybe also having a not-in-place version of apply makes this mistake less likely

### DIFF
--- a/src/mrpro/data/SpatialDimension.py
+++ b/src/mrpro/data/SpatialDimension.py
@@ -109,6 +109,7 @@ class SpatialDimension(MoveDataMixin, Generic[T_co]):
 
         return SpatialDimension(z, y, x)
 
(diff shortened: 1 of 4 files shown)


src/mrpro/data/SpatialDimension.py
```diff
@@ -116,8 +117,20 @@ def apply_(self, function: Callable[[T], T] | None = None, **_) -> Self:
         function
             function to apply
         """
+        # This function is mainly used for type hinting
```
ckolbPTB: What does "this function" refer to?

ckolbPTB (approved):

PR #549, opened 2024-11-20
Omit notebook subfolders in docs

lrlunin:
Since the notebooks have different names wir can optimize the `_notebooks/NOTEBOOK_NAME/NOTEBOOK_NAME.ipynb` structure and omit the unneeded subfolder.


schuenke (approved):
Maybe change the PR title to something like "Omit notebook subfolders in docs" before merge to clarify that this PR only effects the docs and not the general repo structure.

PR #550, opened 2024-11-20
Add python 3.10 to badge and fix project keywords

lrlunin:
A very small fix including adding the now supported python 3.10 version to the badge and fixing the `keywords` of the `pyproject.toml` which are now individual words.


schuenke (approved):
Maybe the right moment to think about additional keywords... maybe we could add something like "Pulseq", "qMRI", "medical imaging", "machine learning", "deep learning"... any further suggestions? I think up to 10 or 15 keywords would be okay.

ckolbPTB:
physics-informed learning, signal models, model-based reconstruction, quantitative

schuenke:
Should we merge this? Any further comments from @fzimmermann89 ?

PR #551, opened 2024-11-20
Fix trajectory scaling in KTrajectoryPulseq

schuenke:
I came across a bug in the `reshape_pulseq_traj` function in our `KTrajectoryPulseq` class. 

In `KTrajectoryPulseq`, we use `seq.calculate_kspace()` to get our k-space trajectory from the seq-file. Due to some rounding errors etc it happens that we get different kz values even for single 2D acquisitions. These values are usually about 10 orders of magnitude smaller compared to typical kx and ky values.

However, in the `reshape_pulseq_traj` function we rescale the values with `encoding_size / (2 * torch.max(torch.abs(k_traj)))`, which can be in the order of 1e8 for very small kz values (caused by mentioned rounding errors). For `encoding_size = 1`, the resulting max value of kz after rescaling was always 0.5 and thus 2 orders of magnitude larger than our `repeat_detection_tolerance`, which is meant to compensate for the small rounding errors.

I actually don't understand why this hasn't been a problem in the past, but @fzimmermann89 and I are sure that the scaling for encoding_size = 1 is a bug, which should be fixed by this PR.


src/mrpro/data/traj_calculators/KTrajectoryPulseq.py
```diff
@@ -54,13 +54,17 @@ def __call__(self, kheader: KHeader) -> KTrajectoryRawShape:
             raise ValueError('We currently only support constant number of samples')
         n_k0 = int(n_samples.item())
 
-        def reshape_pulseq_traj(k_traj: torch.Tensor, encoding_size: int):
-            k_traj *= encoding_size / (2 * torch.max(torch.abs(k_traj)))
+        def rescale_and_reshape_traj(k_traj: torch.Tensor, encoding_size: int):
+            if encoding_size > 1 and torch.max(torch.abs(k_traj)) > 0:
+                scaling_factor = encoding_size / (2 * torch.max(torch.abs(k_traj)))
+            else:
+                scaling_factor = 0
```
ckolbPTB: Maybe add a comment here which describes that setting the value to 0 here is due to potential rounding errors and we don't have enough information to do better.

ckolbPTB (approved):

ckolbPTB:
> I actually don't understand why this hasn't been a problem in the past

I guess for 2D trajectories it does not matter because the k2 dimension would then be singleton and we simply ignore it in the FourierOp. The only one using 3D pulseq-trajectories is probably @JoHa0811 - not sure if he encountered any problems.

fzimmermann89:
Patrick noticed the bug because the reconstruction did not work.
The kz points were not singleton, but some noise, scaled to -.5 ... .5.
So they are only considered singleton if you increase the detection tolerance to 0.5.
(which could also be a sensible default, if the biggest difference between k space points is less than 1, it surely is singleton)

The reconstruction should still have worked, in theory, but, we also have #553 , the fourier op does not work with recon matrix smaller 6.

ckolbPTB:
> Patrick noticed the bug because the reconstruction did not work.

For 2D acquisitions? Strange, because we also have pulseq reconstructions in our examples

schuenke:
> > I actually don't understand why this hasn't been a problem in the past
> 
> I guess for 2D trajectories it does not matter because the k2 dimension would then be singleton and we simply ignore it in the FourierOp. The only one using 3D pulseq-trajectories is probably @JoHa0811 - not sure if he encountered any problems.

It was a 2D trajectory in my case, but the kz values after rescaling differed in the range [-0.3, 0.5] and thus much more than the default `repeat_detection_tolerance` resulting in final kdata.traj dimensions

```
kz=[  1,   1, 402,   1]
ky=[  1,   1, 402, 256]
kx=[  1,   1, 402, 256]
```
insted of `kz = [1, 1, 1, 1]`.

This actually lead to errors in the reco.

fzimmermann89:
> > Patrick noticed the bug because the reconstruction did not work.
> 
> For 2D acquisitions? Strange, because we also have pulseq reconstructions in our examples

Indeed. I think it depends on the sequence if pypulseq returns 0 or something close to zero.
We should ask a pypulseq maintainer for input ;)

fzimmermann89:
> > > Patrick noticed the bug because the reconstruction did not work.
> > 
> > 
> > For 2D acquisitions? Strange, because we also have pulseq reconstructions in our examples
> 
> Indeed. I think it depends on the sequence if pypulseq returns 0 or something close to zero. We should ask a pypulseq maintainer for input ;)

Ok, in the pulseq example, kz is  constant (but not 0). This  avoids the possible division by zero bug and results in a singleton dimension as all kz after scaling are -0.5 ...

src/mrpro/data/traj_calculators/KTrajectoryPulseq.py
```diff
@@ -54,13 +54,17 @@ def __call__(self, kheader: KHeader) -> KTrajectoryRawShape:
...
+            k_traj *= scaling_factor
```
fzimmermann89: dont do inplace here. 
this only made debugging annoying, as running line 66ff changes k_traj_adc ...

schuenke:
We still want to merge this? Or is it superseeded by #560? 
IMHO it's not a good idea to combine a major refactoring like #560 with a bugfix like this...

PR #553, opened 2024-11-20
Fix DCF and FourierOp edgecase bugs

fzimmermann89:
This fixes two bugs noticed by @schuenke (only triggered because of another bug):

Our FourierOp did not work for `image_size<nufft_numpoints` -- so you could not to non-uniform sampling with fewer than 6 points (not that common)

Our DCF vornoi assumed that the k dimension requiring vorinoi did not contain singleton dimensions.

### DIFF
--- a/src/mrpro/data/DcfData.py
+++ b/src/mrpro/data/DcfData.py
@@ -56,7 +56,7 @@ class DcfData(MoveDataMixin):
 
         if ks_needing_voronoi:
             # Handle full dimensions needing voronoi
-            dcfs.append(smap(dcf_2d3d_voronoi, torch.stack(list(ks_needing_voronoi), -4), 4))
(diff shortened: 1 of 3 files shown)


ckolbPTB:
I don't understand which case caused the dcf to fail. Could you please add a test case for this?

fzimmermann89:
> I don't understand which case caused the dcf to fail. Could you please add a test case for this?


    kx = rng.float32_tensor((1, 1, 4, 4))
    ky = rng.float32_tensor((1, 4, 1, 4))

--> both need voronoi. they cant be stacked without broadcasting. this PR added the broadcast.

Not sure which real 3d trajectory would trigger this besides a wrongly read-in pulseq trajectory -- but it should work nevertheless as it is in theory a completely valid trajectory.

ckolbPTB (approved):

PR #555, opened 2024-11-21
Fetch docker names without github token

lrlunin:
This fix in combination with #537 allows user to send pull requests from forks and all workflows be executed. The issue before was the fact the the GH REST API request required the secret token for fetch of the docker images list.

Now this is done matching the filenames in `docker` directory.


fzimmermann89:
would it be easy to make the workflow fail if no docker files were found? just as a small safety net

lrlunin:
> would it be easy to make the workflow fail if no docker files were found? just as a small safety net

As my first commit experience tells - it will fail anyway in the next step when it will try to fetch the docker. But I'll take a look now.

.github/workflows/pytest.yml
```diff
@@ -15,18 +15,24 @@ jobs:
...
+
       - name: Retrieve Docker Image Names
         id: set-matrix
-        env:
-          GH_TOKEN: ${{ secrets.GHCR_TOKEN }}
         run: |
-          imagenames=$(curl -s --request GET \
-          --url "https://api.github.com/orgs/PTB-MR/packages?package_type=container" \
-          --header "Authorization: Bearer $GH_TOKEN"  | jq -r '.[].name')
+          # search for Dockefile_* in the docker directory, replace "Dockerfile" prefix with "mrpro" and to imagenames
```
ckolbPTB:
```suggestion
          # search for Dockerfile_* in the docker directory, replace "Dockerfile" prefix with "mrpro" and to imagenames
```

ckolbPTB:
Do we also have to adapt the docker-workflow?

lrlunin:
> Do we also have to adapt the docker-workflow?

I do not think so. This will not be executed at `pull-request`. It interacts with registry and tokens only on `github.event_name == 'push'` i.e. push on `main`.

ckolbPTB:
It should be triggered if anything related to the docker image (toml file or the docker file itself) changes, shouldn't it?

fzimmermann89:
pushing the test docker images cannot work from forks.
so it can realistically only be done here anyways. same with releases to testpypi.

so release and changes to docker will always be done on the main repo.

fzimmermann89:
we coulld change the docker action to also use sed (or change it here to also use jq), as on both pretty much the same step is done to obtain the imagenames, but with slightly different commands..

lrlunin:
> we coulld change the docker action to also use sed (or change it here to also use jq), as on both pretty much the same step is done to obtain the imagenames, but with slightly different commands..

The sed here is not really replaceable here with jq or jq in later steps is replaceable with sed. For the `$GITHUB_OUTPUT` we need a JSON like array `["one", "two", "three"]` while the first step in the `pytest.yml` is done purely with the bash arrays. The reason we needed the jq before was the JSON answer from the GH API request.

I would change the `cd ... & ls ... ` with the `find ...` in the `docker.yml` and `docs.yml`.

.github/workflows/docker.yml
```diff
@@ -39,12 +39,10 @@ jobs:
...
         if: steps.filter.outputs.docker_toml == 'true' || github.event_name == 'push'
         run: |
-          cd ./docker/
-          ls
-          dockerfiles=$(ls Dockerfile_* | jq -R -s -c 'split("\n")[:-1]')
+          dockerfiles=$(find docker -type f -name 'Dockerfile*' -exec basename {} \; | jq -R -s -c 'split("\n")[:-1]')
           echo "dockerfiles: $dockerfiles"
           echo "dockerfiles=$dockerfiles" >> $GITHUB_OUTPUT
-          imagenames=$(ls Dockerfile_* | sed -e 's/Dockerfile_/ghcr.io\/ptb-mr\/mrpro_/' | jq -R -s -c 'split("\n")[:-1]')
+          imagenames=$(find docker -type f -name 'Dockerfile*' -exec basename {} \; | sed 's/^Dockerfile/ghcr.io\/ptb-mr\/mrpro/' | jq -R -s -c 'split("\n")[:-1]')
```
lrlunin: It's a little bit messy with the folders and filenames. I believe a better way is to change the folder in the single place. Now we take only the filenames here and append the `docker` folder prefix later.

A better way would mean that we don't use `basename` here and including the whole path to the `imagenames` and then get away from `docker` folder in all other steps.

I think it's obviously has advantages that if the change the folder some day we would need to change the related folder in the single line and not in each following step.

I have already done this for the notebooks and `examples` folder that now the `notebooks` list contain the full path to the notebooks.

Let me know if you agree or disagree then I will proceed or revert with the changes.

ckolbPTB:
> so release and changes to docker will always be done on the main repo.

would be good to discuss then in a meeting how we envisage the new workflow

ckolbPTB:
Quick question: If I create a fork and then a PR to main. The PR gets approved and I merge the PR into main. Currently this would trigger our docker-pipeline as "push to main". This will run on main and hence the github token should be available, shouldn't it?

lrlunin:
> Quick question: If I create a fork and then a PR to main. The PR gets approved and I merge the PR into main. Currently this would trigger our docker-pipeline as "push to main". This will run on main and hence the github token should be available, shouldn't it?

I have expected it to do so. Only after `push on branch main` the `push to registry` will be initiated and the only this really needs to have a token for push.

I also think that right now there is a special token generated and placed in secrets due to that API request. After this pull-request It can be deleted if we use a default `GITHUB_TOKEN` with given "write permissions" for workflows to push.

fzimmermann89:
I am looking forward to getting this PR merged -- and then trying it out.

I don't think there is a good way to test if this really fixes all issues with forks. It looks like it _should_  work ;)

lrlunin:
> I am looking forward to getting this PR merged -- and then trying it out.
> 
> I don't think there is a good way to test if this really fixes all issues with forks. It looks like it _should_ work ;)

I can do it with my "mock" repo or you can just create an own if you wish.

Or you can run the workflow locally with the `act` with env `push main` and then check this out ;-)
I'll try this today evening.

schuenke (approved):

PR #557, opened 2024-11-22
Memory efficient reshape

fzimmermann89:
If a tensor is broadcasted along a dimension, i.e. it as stride 0, and  one wants to reshape it, pytorch will first try to create a view.
If that fails, because dimensions merged or split are broadcasted or in the wrong order, pytorch will create a contiguous version of the tensor and reshape it.
This will remove all broadcasting and create a large memory allocation, even if the broadcasting could be preserved.

For example:
``` python
import torch

a=torch.rand(1,2,1).expand(100,2,2)
print("a", a.shape, "stride:", a.stride())
print("storage a", a.untyped_storage().nbytes(), "bytes")

b=a.reshape(-1,4)
print("b", b.shape, "stride:", b.stride())
print("storage b", b.untyped_storage().nbytes(), "bytes")

# instead, pytorch could do this
c=a[:1].reshape(-1,4).expand(100,-1)
print("c", c.shape, "stride:", c.stride())
print("storage c", c.untyped_storage().nbytes(), "bytes")

torch.testing.assert_close(b,c)
```

```
a torch.Size([100, 2, 2]) stride: (0, 1, 0)
storage a 8 bytes

b torch.Size([100, 4]) stride: (4, 1)
storage b 1600 bytes

c torch.Size([100, 4]) stride: (0, 1)
storage c 16 bytes
```

This is a huge issue for our reshape/rearrange, as it would result in large memory allocation.

This code instead tries to figure out which axes will be merged together by the reshape. If all axes in a group are broadcasted (or there is only one broadcasted axis in a group), it will first remove the broadcasting.
Then it will create a new, contiguous copy of the smaller tensor, before restoring the broadcasting.

I think this approach will make it possible to avoid reimplementing the full einops string parsing and still do the "rearrange sheuld behave as-if broadcasted" we agreed upon, while also taking @schuenke and @JoHa0811's remarks about memory efficiency into account:

We broadcast the tensors in the reaarnge, then make einops use our efficient reshape -- either via temporarly  setting the .reshape method on the tensor or by modifying the TorchBackend in einops. Afterwards, we can remove existing broadcast dimensions again. 

The functionality will also be preserved if this function is ever removed -- it will then just be a lot less efficient in memory.
Thus maintenance might be feasible -- it is quite clear what the result of the function should be: Besides strides etc, it should do exactly the same as torch.reshape.


src/mrpro/utils/reshape.py
```diff
@@ -99,3 +99,62 @@ def reduce_view(x: torch.Tensor, dim: int | Sequence[int] | None = None) -> torc
         for d, (oldsize, stride) in enumerate(zip(x.size(), stride, strict=True))
     ]
     return torch.as_strided(x, newsize, stride)
+
+
+# @lru_cache
```
fzimmermann89: Not sure if lru_cache helps here.
-- I will  have to benchmark it

The code is written, such that the parameters to the function can be used as cache keys.

fzimmermann89: I disabled it for now, if a reviewer want to debug it, it is easier without the cache.

fzimmermann89: (lets not resovle this conversation while the commented our line is still present )

src/mrpro/utils/reshape.py
```diff
@@ -99,3 +99,62 @@ def reduce_view(x: torch.Tensor, dim: int | Sequence[int] | None = None) -> torc
...
+    # This function tries to group axes from new_shape and old_shape into the smallest groups that have
+    # the same number of elements, starting from the right.
+    # If all axes of old shape of a group are stride=0 dimensions,
+    # we can reduce them.
+    idx = []
+    pointer_old, pointer_new = len(old_shape) - 1, len(new_shape) - 1  # start from the right
+    while pointer_old >= 0:
+        product_new, product_old = 1, 1
+        group: list[int] = []
+        while product_old != product_new or not group:
```
ckolbPTB, line 132: Could you add some more information here of what this is trying to achieve? I am a bit lost with this code...

fzimmermann89: better?

fzimmermann89:
pre-commit.ci autofix

src/mrpro/utils/reshape.py
```diff
@@ -99,3 +100,85 @@ def reduce_view(x: torch.Tensor, dim: int | Sequence[int] | None = None) -> torc
...
+def _reshape_idx(old_shape: tuple[int, ...], new_shape: tuple[int, ...], old_stride: tuple[int, ...]) -> list[slice]:
+    """Get reshape reduce index (Cached helper function for reshape_broadcasted).
+
+    This function tries to group axes from new_shape and old_shape into the smallest groups that have
+    the same number of elements, starting from the right.
+    If all axes of old shape of a group are stride=0 dimensions, we can reduce them.
+
+    Example:
+        old_shape = (30, 2, 2, 3)
+        new_shape = (6. 5, 4, 3)
```
lrlunin: `,` instead of `.`

tests/utils/test_reshape.py
```diff
@@ -51,3 +52,30 @@ def test_reduce_view():
...
     assert torch.equal(reduced_one_pos.expand_as(tensor), tensor)
+
+
+@pytest.mark.parametrize(
+    ('shape', 'expand_shape', 'permute', 'final_shape', 'expected_stride'),
+    [
+        ((1, 2, 3, 1, 1), (1, 2, 3, 4, 5), (0, 2, 1, 3, 4), (1, 6, 2, 2, 5), (6, 1, 0, 0, 0)),
+        ((1, 2, 1), (100, 2, 2), (0, 1, 2), (100, 4), (0, 1)),
+        ((1, 1, 1, 1), (2, 3, 4, 5), (2, 3, 0, 1), (1, 2, 6, 10, 1), (0, 0, 0, 0, 0)),
+        ((1, 2, 3), (1, 2, 3), (0, 1, 2), (6,), (1,)),
```
lrlunin: Maybe add some tests with negative values in the `shape`? I've just checked that case and it works correct. So just to be sure and cover that optiont too,

lrlunin (approved):

PR #559, opened 2024-11-26
Move functions from mixins into KData

fzimmermann89:
As the mixins will deprecate or be replaced, moving Kdata back into data makes more sense.

This also fixes the issue with the `KData.__init__` documentation

In one of the later PRs I will remove some of the functions or replace them by the new rearrange.

# Stack
* #591
* #580
* #568
* #560
* #567
* __->__ #559


fzimmermann89:
* #594
* #591
* #580
* #568
* #560
* #567
* __->__ #559

fzimmermann89:
The protocol stuff also created issues in the documentation of KData.

I moved the functions all back into KData, before replacing them.
I think this can already be merged, please review

src/mrpro/data/KData.py
```diff
@@ -366,3 +368,287 @@ def compress_coils(
...
+        Raises
+        ------
+        ValueError
+            If the recon matrix along x is larger than the encoding matrix along x.
+
+        References
+        ----------
+        .. [GAD] Gadgetron https://github.com/gadgetron/gadgetron-python
+        """
+        from mrpro.operators.FastFourierOp import FastFourierOp
```
schuenke, line 423: should be possible to move this to the imports at the top of the file now, no ?!

schuenke (approved):
only one minor comment/question... see below

ckolbPTB (approved):

src/mrpro/data/KData.py
fzimmermann89: no, as operators depend on data.

PR #560, opened 2024-11-26
Refactor AcqInfo and Trajectory Calculation

### LINKED ISSUE #548: Channel mask
- Do we want to keep the channel mask in the acq info?
- If we go with the "everything has the same  >5 dimensions", it would be one of the parameters requiring an exception.
- If we  keep it, should we maybe convert it to a boolean tensor? currently, the 1024 channels are saved as 64 uint16 tensors... 1024 bool values would be more usefull

fzimmermann89:
Removes fields from AcqInfo only requiered for trajectory calculation.

Trajecotry calculators no longer get the full header as parameters. 

TODO: Docstrings could be improved


fzimmermann89:
* #683
* #675
* #674
* #673
* #672
* #671
* #660
* #591
* #679
* __->__ #560
* #656
* #655

JoHa0811 (approved):

src/mrpro/data/traj_calculators/KTrajectoryRadial2D.py
```diff
@@ -22,27 +21,35 @@ def __init__(self, angle: float = torch.pi * 0.618034) -> None:
...
         Parameters
         ----------
-        kheader
-           MR raw data header (KHeader) containing required meta data
+        n_k0
+            number of samples in k0
+        k0_center
+            position of k-space center in k0
+        k1_idx
+            indices of k1
```
fzimmermann89: mention angle here

src/mrpro/data/traj_calculators/KTrajectorySunflowerGoldenRpe.py
```diff
@@ -4,90 +4,94 @@
...
         Parameters
         ----------
-        kheader
-            MR raw data header (KHeader) containing required meta data
+        n_k0
+            number of samples in k0
+        k0_center
+            position of k-space center in k0
+        k1_idx
+            indices of k1
```
fzimmermann89: mention angle and radial here

src/mrpro/data/AcqInfo.py
```diff
@@ -140,26 +148,34 @@ class AcqInfo(MoveDataMixin):
     scan_counter: torch.Tensor
```
ckolbPTB: Can we also remove this? @schuenke can you think of something where this is needed?

src/mrpro/data/AcqInfo.py
```diff
@@ -140,26 +148,34 @@ class AcqInfo(MoveDataMixin):
...
+    ) -> Self | tuple[Self, tuple[torch.Tensor, ...]]:
         """Read the header of a list of acquisition and store information.
 
         Parameters
         ----------
-        acquisitions:
+        acquisitions
             list of ismrmrd acquisistions to read from. Needs at least one acquisition.
+        additional_fields
+            if supplied, additional fields with these names will be from the ismrmrd acquisitions
```
ckolbPTB:
```suggestion
            if supplied, additional information from the fields with these names will be extracted from the ismrmrd acquisitions
```

src/mrpro/data/AcqInfo.py
```diff
@@ -228,33 +244,49 @@ def spatialdimension_2d(data: np.ndarray) -> SpatialDimension[torch.Tensor]:
...
             idx=acq_idx,
-            acquisition_time_stamp=tensor_2d(headers['acquisition_time_stamp']),
-            active_channels=tensor_2d(headers['active_channels']),
-            available_channels=tensor_2d(headers['available_channels']),
-            center_sample=tensor_2d(headers['center_sample']),
-            channel_mask=tensor_2d(headers['channel_mask']),
-            discard_post=tensor_2d(headers['discard_post']),
-            discard_pre=tensor_2d(headers['discard_pre']),
-            encoding_space_ref=tensor_2d(headers['encoding_space_ref']),
+            acquisition_time_stamp=tensor_2d(headers['acquisition_time_stamp']).double(),
```
ckolbPTB: I don't think casting to double is a good idea here and for the PhysiologyTimestamps, because these parameters are very likely to be used in signal models leading than to the whole signal model being evaluated in double precision. I would probably leave it as int - this might be another indicator for users that these are not times in [s] but discrete timesteps

src/mrpro/data/traj_calculators/KTrajectoryCalculator.py
```diff
@@ -3,53 +3,93 @@
...
         """Calculate the trajectory for given KHeader.
 
         The shapes of kz, ky and kx of the calculated trajectory must be
         broadcastable to (prod(all_other_dimensions), k2, k1, k0).
+
+        Not all of the parameters will be used by all implementations.
+
+        Parameters
+        ----------
+        n_k0
```
ckolbPTB, line 38: order of parameters should match orde in `__call__`

src/mrpro/data/traj_calculators/KTrajectoryCalculator.py
```diff
@@ -3,53 +3,93 @@
...
+        k2_idx
+            indices of k2
+        k0_center
+            position of k-space center in k0
+        k1_center
+            position of k-space center in k1
+        k2_center
+            position of k-space center in k2
+        reversed_readout_mask
+            boolean tensor indicating reversed redout
```
ckolbPTB:
```suggestion
            boolean tensor indicating reversed readout
```

src/mrpro/data/traj_calculators/KTrajectoryCalculator.py
```diff
@@ -3,53 +3,93 @@
...
+        encoding_matrix
+            encoding matrix, describing the extend of the k-space coordinates
```
ckolbPTB:
```suggestion
            encoding matrix
```
The encoding matrix can be larger or smaller than the actual locations of the k-space sampling points

src/mrpro/data/traj_calculators/KTrajectoryPulseq.py
```diff
@@ -28,13 +28,21 @@ def __init__(self, seq_path: str | Path, repeat_detection_tolerance: None | floa
...
         """Calculate trajectory from given .seq file and header information.
 
         Parameters
         ----------
-        kheader
-           MR raw data header (KHeader) containing required meta data
+        n_k0
+            number of samples in k0
+        encoding_matrix
+            encoding matrix, describing the extend of the k-space coordinates
```
ckolbPTB:
```suggestion
            encoding matrix
```

src/mrpro/data/traj_calculators/KTrajectoryPulseq.py
```diff
@@ -48,19 +56,20 @@ def __call__(self, kheader: KHeader) -> KTrajectoryRawShape:
...
-        n_samples = torch.unique(n_samples)
-        if len(n_samples) > 1:
-            raise ValueError('We currently only support constant number of samples')
-        n_k0 = int(n_samples.item())
-
-        def reshape_pulseq_traj(k_traj: torch.Tensor, encoding_size: int):
-            k_traj *= encoding_size / (2 * torch.max(torch.abs(k_traj)))
+        def reshape(k_traj: torch.Tensor, encoding_size: int) -> torch.Tensor:
+            max_value_range = 2 * torch.max(torch.abs(k_traj))
+            if max_value_range > 1e-9 and encoding_size > 1:
```
ckolbPTB: 1e-9 seems very small for numerical errors of the trajectory in pulseq... @schuenke do we have any values from real-world examples?

src/mrpro/data/traj_calculators/KTrajectoryRpe.py
```diff
@@ -56,80 +55,71 @@ def _apply_shifts_between_rpe_lines(self, krad: torch.Tensor, kang_idx: torch.Te
 
         Parameters
         ----------
-        krad
-            k-space positions along each phase encoding line
-        kang_idx
-            indices of angles to be used for shift calculation
+        k_radial
+            k-space positions along each phase encoding line, zo be shifted
```
ckolbPTB:
```suggestion
            k-space positions along each phase encoding line, to be shifted
```

src/mrpro/data/traj_calculators/KTrajectoryRpe.py
```diff
@@ -56,80 +55,71 @@ def _apply_shifts_between_rpe_lines(self, krad: torch.Tensor, kang_idx: torch.Te
...
+        idx
+            indices used for shift calculation
+
+        Returns
+        -------
+        shifted radial k-space positions
```
ckolbPTB: tab missing

src/mrpro/data/traj_calculators/KTrajectorySunflowerGoldenRpe.py
```diff
@@ -4,90 +4,94 @@
...
-class KTrajectorySunflowerGoldenRpe(KTrajectoryRpe):
+
+class KTrajectorySunflowerGoldenRpe(KTrajectoryCalculator):
     """Radial phase encoding trajectory with a sunflower pattern."""
 
-    def __init__(self, rad_us_factor: float = 1.0) -> None:
+    def __init__(self) -> None:
         """Initialize KTrajectorySunflowerGoldenRpe.
 
         Parameters
```
ckolbPTB: No Parameters

src/mrpro/data/traj_calculators/KTrajectoryRpe.py
```diff
@@ -56,80 +55,71 @@ def _apply_shifts_between_rpe_lines(self, krad: torch.Tensor, kang_idx: torch.Te
...
 
         References
         ----------
         .. [PRI2010] Prieto C, Schaeffter T (2010) 3D undersampled golden-radial phase encoding
         for DCE-MRA using inherently regularized iterative SENSE. MRM 64(2). https://doi.org/10.1002/mrm.22446
         """
+        k_radial, idx = torch.broadcast_tensors(k_radial, idx)
```
ckolbPTB: in the rest of the TrajCalc code we now use radial and not k_radial anymore

src/mrpro/data/AcqInfo.py
fzimmermann89: we should come up with a solution here.
it it is int, it is quite easy to make the mistake  "acquisition_time_stamp*2.5" which will not work as expected..

also, if i am not mistaken, we would anyways assume it is in 2.5ms steps when we use it in signal models.
so maybe we can just convert it when loading from ismrmrd to seconds, assuming 2.5ms steps.
if there ever is another source for ismrmrd files that uses a different step, we should handle that case separately.

So i would change the documentation to 
"usually in seconds since midnight (Siemens). For other vendors, this will be the vendor time stamp times 2.5e-3"

src/mrpro/data/traj_calculators/KTrajectoryPulseq.py
fzimmermann89: the 1e-9 is only to avoid overflow by division by zero (or almost zero), resulting in infs.

src/mrpro/data/AcqInfo.py
ckolbPTB: I think seconds makes most sense but we should check the vendor and print a warning if not Siemens. Soon we will hopefully have similar data from the low-field scanner and it would be nice to have a higher resolution than 2.5ms...

src/mrpro/data/AcqInfo.py
schuenke: Currently not. I would say we remove it and reimplement it in case we realize we do need it.

fzimmermann89: @ckolbPTB did you resolve the conversation because you want to keep it?
i don#t mind either way. pretty sure I will not cause any issues to keep it, but I don't see a use for it

src/mrpro/data/traj_calculators/KTrajectoryCalculator.py
```diff
@@ -3,53 +3,93 @@
...
+        k1_idx
+            indices of k1
+        k2_idx
+            indices of k2
+        k0_center
+            position of k-space center in k0
+        k1_center
+            position of k-space center in k1
+        k2_center
+            position of k-space center in k2
```
fzimmermann89:
```suggestion
        n_k0
            number of samples in k0
        k0_center
            position of k-space center in k0
        k1_idx
            indices of k1
        k1_center
            position of k-space center in k1
        k2_idx
            indices of k2
        k2_center
            position of k-space center in k2
```

src/mrpro/data/traj_calculators/KTrajectoryPulseq.py
fzimmermann89: made obsolete by #582

src/mrpro/data/traj_calculators/KTrajectorySunflowerGoldenRpe.py
fzimmermann89:
```suggestion
```

src/mrpro/data/traj_calculators/KTrajectorySunflowerGoldenRpe.py
```diff
@@ -4,90 +4,94 @@
...
+        n_k0
+            number of samples in k0
+        k0_center
+            position of k-space center in k0
+        k1_idx
+            indices of k1
+        k1_center
+            position of k-space center in k1
+        k2_idx
+            indices of k2
```
fzimmermann89:
```suggestion
            indices of k2 (angle)
```

src/mrpro/data/traj_calculators/KTrajectorySunflowerGoldenRpe.py
fzimmermann89:
```suggestion
            indices of k1 (radial)
```

src/mrpro/data/KData.py
```diff
@@ -136,47 +137,80 @@ def from_file(
...
+            and isinstance(ismrmrd_header.acquisitionSystemInformation.systemVendor, str)
+            and ismrmrd_header.acquisitionSystemInformation.systemVendor.lower() == 'siemens'
+        ):
+            convert_time_stamp = convert_time_stamp_siemens  # 2.5ms time steps
+        # Here we could include more vendors
+        elif (
+            ismrmrd_header.acquisitionSystemInformation is None
+            or ismrmrd_header.acquisitionSystemInformation.systemVendor is None
+        ):
+            warnings.warn('No vendor information found. Assuming Siemens time stamp format.', stacklevel=1)
```
ckolbPTB: I think here you are missing another `convert_time_stamp = convert_time_stamp_siemens # 2.5ms time steps` but I would actually suggest to add another converter which interprets the integer time stamps as time in ms and use that if there is no vendor information or if the vendor is unknown.

src/mrpro/data/KData.py
```diff
@@ -136,47 +137,80 @@ def from_file(
...
                 stacklevel=1,
             )
 
+        shapes = (torch.as_tensor([acq.data.shape[-1] for acq in acquisitions]) - discard_pre - discard_post).unique()
+        if len(shapes) > 1:
+            if discard_pre.any() or discard_post.any():
+                note = 'discard_pre and post were applied.'
+            else:
+                note = 'discard_pre and discard_post were empty.'
+                warnings.warn(
```
ckolbPTB: Should this be moved out of the else statement?

src/mrpro/data/traj_calculators/KTrajectorySunflowerGoldenRpe.py
```diff
@@ -4,90 +4,93 @@
...
+        *,
+        n_k0: int,
+        k0_center: int | torch.Tensor,
+        k1_idx: torch.Tensor,
+        k1_center: int | torch.Tensor,
+        k2_idx: torch.Tensor,
+        reversed_readout_mask: torch.Tensor | None = None,
+        **_,
+    ) -> KTrajectory:
+        """Calculate radial phase encoding trajectory for given KHeader.
```
ckolbPTB:
```suggestion
        """Calculate radial phase encoding trajectory for given header information.
```

src/mrpro/data/traj_calculators/KTrajectoryRpe.py
```diff
@@ -56,80 +55,71 @@ def _apply_shifts_between_rpe_lines(self, krad: torch.Tensor, kang_idx: torch.Te
...
+        *,
+        n_k0: int,
+        k0_center: int | torch.Tensor,
+        k1_idx: torch.Tensor,
+        k1_center: int | torch.Tensor,
+        k2_idx: torch.Tensor,
+        reversed_readout_mask: torch.Tensor | None = None,
+        **_,
+    ) -> KTrajectory:
         """Calculate radial phase encoding trajectory for given KHeader.
```
ckolbPTB:
```suggestion
        """Calculate radial phase encoding trajectory for given header information.
```

src/mrpro/data/traj_calculators/KTrajectoryRadial2D.py
```diff
@@ -22,27 +21,35 @@ def __init__(self, angle: float = torch.pi * 0.618034) -> None:
...
+    def __call__(
+        self,
+        *,
+        n_k0: int,
+        k0_center: int | torch.Tensor,
+        k1_idx: torch.Tensor,
+        reversed_readout_mask: torch.Tensor | None = None,
+        **_,
+    ) -> KTrajectory:
         """Calculate radial 2D trajectory for given KHeader.
```
ckolbPTB:
```suggestion
        """Calculate radial 2D trajectory for given header information.
```

src/mrpro/data/traj_calculators/KTrajectoryRpe.py
```diff
@@ -56,80 +55,71 @@ def _apply_shifts_between_rpe_lines(self, krad: torch.Tensor, kang_idx: torch.Te
...
+        k1_center
+            position of k-space center in k1
+        k2_idx
+            indices of k2
+        reversed_readout_mask
+            boolean tensor indicating reversed readout
 
         Returns
         -------
             radial phase encoding trajectory for given KHeader
```
ckolbPTB:
```suggestion
            radial phase encoding trajectory for given header information
```

ckolbPTB:
The doc string here is IMO also wrong/outdated:
https://github.com/PTB-MR/mrpro/blob/8ee4daf0fd79725d7c4056b20a803d4d3cc401d0/src/mrpro/data/traj_calculators/KTrajectoryIsmrmrd.py#L16

src/mrpro/data/KData.py
fzimmermann89: fixed that, but will keep it at 2.5ms as currently all ismrmrd files use this as the timer resolution and this is the example in the spec, so it seems a sensible default for a file with missing vendor information

src/mrpro/data/AcqInfo.py
```diff
@@ -26,140 +27,199 @@ def rearrange_acq_info_fields(field: object, pattern: str, **axes_lengths: dict[
...
+def convert_time_stamp_siemens(
+    timestamp: torch.Tensor,
+    _: str,
+) -> torch.Tensor:
+    """Convert Siemens time stamp to seconds."""
+    return timestamp.double() * 2.5e-3
+
+
+def _int_factory() -> torch.Tensor:
+    # TODO: check dtype
```
fzimmermann89: TODO

ckolbPTB: Still missing?

src/mrpro/data/AcqInfo.py
ckolbPTB: Sorry my mistake - it is fine to remoce scan counter

src/mrpro/data/AcqInfo.py
```diff
@@ -1,165 +1,213 @@
...
 
-    encoding_space_ref: torch.Tensor
-    """Indexed reference to the encoding spaces enumerated in the MRD (xml) header."""
+    idx: AcqIdx = field(default_factory=AcqIdx)
+    """Indices describing acquisitions (i.e. readouts)."""
 
-    flags: torch.Tensor
+    acquisition_time_stamp: torch.Tensor = field(default_factory=_float_factory)
+    """Clock time stamp. Usually in seconds (Siemens: seconds since midnight)"""
+    # TODO: check dtype
```
ckolbPTB: Missing?

src/mrpro/data/AcqInfo.py
```diff
@@ -1,165 +1,213 @@
...
+    orientation: Rotation = field(default_factory=lambda: Rotation.identity((1, 1, 1, 1)))
     """Rotation describing the orientation of the readout, phase and slice encoding direction."""
 
-    patient_table_position: SpatialDimension[torch.Tensor]
+    patient_table_position: SpatialDimension[torch.Tensor] = field(default_factory=_position_factory)
     """Offset position of the patient table, in LPS coordinates [m]."""
 
-    physiology_time_stamp: torch.Tensor
+    physiology_time_stamps: PhysiologyTimestamps = field(default_factory=PhysiologyTimestamps)
     """Time stamps relative to physiological triggering, e.g. ECG. Not in s but in vendor-specific time units"""
```
ckolbPTB: This is now in seconds, right?

src/mrpro/data/__init__.py
```diff
@@ -32,7 +31,7 @@
     "KHeader",
     "KNoise",
     "KTrajectory",
-    "KTrajectoryRawShape",
+    "KTrajectory",
```
ckolbPTB:
```suggestion
```

src/mrpro/data/traj_calculators/KTrajectoryIsmrmrd.py
```diff
@@ -1,28 +1,41 @@
...
 class KTrajectoryIsmrmrd:
     """Get trajectory in ISMRMRD raw data file.
 
     The trajectory in the ISMRMRD raw data file is read out [TRA]_.
 
-    The value range of the trajectory in the ISMRMRD file is not well defined. Here we simple normalize everything
-    based on the highest value and ensure it is within [-pi, pi]. The trajectory is in the shape of the unsorted
-    raw data.
+    The value range of the trajectory in the ISMRMRD file is not well defined, thus we normalize
+    based on the highest value and ensure it is within [-pi, pi].
```
ckolbPTB: We do not normalize at all but take the values as they are.

src/mrpro/data/traj_calculators/KTrajectoryIsmrmrd.py
```diff
@@ -35,18 +48,26 @@ def __call__(self, acquisitions: Sequence[ismrmrd.Acquisition]) -> KTrajectoryRa
...
-            ktraj = KTrajectoryRawShape(
-                kz=torch.zeros_like(ktraj_mrd[..., 1]),
-                ky=ktraj_mrd[..., 1],
-                kx=ktraj_mrd[..., 0],
-            )
+        if self.filename is None:
+            ktraj_mrd = torch.stack([torch.as_tensor(acq.traj, dtype=torch.float32) for acq in acquisitions])
         else:
-            ktraj = KTrajectoryRawShape(kz=ktraj_mrd[..., 2], ky=ktraj_mrd[..., 1], kx=ktraj_mrd[..., 0])
+            with ismrmrd.File(self.filename, 'r') as file:
```
ckolbPTB: This seems very experimental. In most cases it won't even work if you use the same ISMRMRD file for kdata and the trajectory. 

This could simply be done by reading in the ismrmrd file and then passing the trajectory to KData.from_file of the other data. The the problem of mismatch is up to the user.

src/mrpro/data/traj_calculators/KTrajectorySunflowerGoldenRpe.py
```diff
@@ -4,90 +4,93 @@
...
-        kang
-            angles of the radial phase encoding lines
-        kheader
-            MR raw data header (KHeader) containing required meta data
+        radial
+            position along radial direction
+        angles
+            angle of spokes
+        k2_idx
+            indices in k2
```
ckolbPTB, line 37: Returns missing?

ckolbPTB (commented):
no ":" here:
https://github.com/PTB-MR/mrpro/blob/31798980a236b7275e35f974ff287bcf2361b8f3/src/mrpro/data/traj_calculators/KTrajectoryIsmrmrd.py#L43

src/mrpro/data/traj_calculators/KTrajectoryIsmrmrd.py
fzimmermann89: this is for reading in a traj saved in a separate header ismrmd file

src/mrpro/data/AcqInfo.py
fzimmermann89: this was meant as:
should the counters default to uint32, int32 or int64?
uint support in torch is spotty, indexing should most of the time be done with int64.
i think we stick with int64 and I remove the comment?

fzimmermann89:
For the timestamps: what vender to you set at the low field?
then I can add a timestamp converter for it.

I would still prefer to go for default with the example given in the ismrmrd documentation.

src/mrpro/data/traj_calculators/KTrajectoryIsmrmrd.py
fzimmermann89: this would be useful really useful for openrecon and to avoid the combine script.

we could add an acquisition filter here as well.

src/mrpro/data/traj_calculators/KTrajectoryIsmrmrd.py
fzimmermann89: I don't know.

This has been like this for some time.

I think it should be scaled to encoding matrix size, as we do it for pulseq?

src/mrpro/data/traj_calculators/KTrajectoryIsmrmrd.py
fzimmermann89: Solution: add to Ktrajectory

src/mrpro/data/traj_calculators/KTrajectoryIsmrmrd.py
```diff
@@ -5,48 +5,61 @@
...
     References
     ----------
     .. [TRA] ISMRMRD trajectory https://ismrmrd.readthedocs.io/en/latest/mrd_raw_data.html#k-space-trajectory
     """
 
-    def __call__(self, acquisitions: Sequence[ismrmrd.Acquisition]) -> KTrajectoryRawShape:
+    def __init__(self, normalize: bool = False) -> None:
+        """Initialize KTrajectoryIsmrmrd.
+
+        The ISMRMRD specification deos not define a normalization for the trajectory. If the trajectory is not
```
ckolbPTB, line 29:
```suggestion
        The ISMRMRD specification does not define a normalization for the trajectory. If the trajectory is not
```

src/mrpro/data/traj_calculators/KTrajectoryIsmrmrd.py
```diff
@@ -5,48 +5,61 @@
...
+        normalized to the encoding_matrix, please set ``normalize=True`` `to enable normalization. Otherwise,
```
ckolbPTB, line 30:
```suggestion
        normalized to the encoding_matrix, please set ``normalize=True`` to enable normalization. Otherwise,
```

src/mrpro/data/traj_calculators/KTrajectorySunflowerGoldenRpe.py
```diff
@@ -4,90 +4,97 @@
...
     """Radial phase encoding trajectory with a sunflower pattern."""
 
-    def __init__(self, rad_us_factor: float = 1.0) -> None:
+    def __init__(self) -> None:
         """Initialize KTrajectorySunflowerGoldenRpe.
 
-        Parameters
         ----------
-        rad_us_factor
+        radial_undersampling_factor
```
ckolbPTB: should be removed

ckolbPTB (approved):

ckolbPTB:
> For the timestamps: what vender to you set at the low field?

We could use `osi2` @schote what do you think?

ckolbPTB:
please don't forget about:
https://github.com/PTB-MR/mrpro/blob/31798980a236b7275e35f974ff287bcf2361b8f3/src/mrpro/data/traj_calculators/KTrajectoryIsmrmrd.py#L43

PR #562, opened 2024-11-27
Example for Basics and Cartesian reconstruction

### LINKED ISSUE #518: Cartesian Reconstruction example

ckolbPTB:
This example shows some basics features of MRpro and then a step by step guide to reconstruct fully sampled Cartesian scan, Cartesian scan with partial echo and partial Fourier and Cartesian scan with regular undersampling and self-calibrated csm calculation.

probably supersedes #233 
closes #518


schuenke:
You might wanna have a look at it again @ckolbPTB , added quite a few changes.

schuenke (approved):

examples/cartesian_reconstruction.py
```diff
@@ -0,0 +1,387 @@
...
+print(kdata.header.lamor_frequency_proton)
+
+# %% [markdown]
+# ## Reconstruction of fully sampled acquisition
+#
+# For the reconstruction of a fully sampled Cartesian acquisition, we can use a simple Fast Fourier Transform (FFT).
+#
+# Let's create an FFT-operator (called `FastFourierOp` in MRpro) and apply it to our `KData` object. Please note that
+# all MRpro operators currently only work on PyTorch tensors and not on the MRpro objects directly. Therefore, we have
+# have to call the operator on kdata.data. One other important feature of MRpro operators is that they always return a
```
ckolbPTB:
```suggestion
# to call the operator on kdata.data. One other important feature of MRpro operators is that they always return a
```

examples/cartesian_reconstruction.py
```diff
@@ -0,0 +1,387 @@
...
+# tuple PyTorch tensors, even if the output is only a single tensor. This is why we use the `(img,)` syntax below.
```
ckolbPTB:
```suggestion
# tuple of PyTorch tensors, even if the output is only a single tensor. This is why we use the `(img,)` syntax below.
```

examples/cartesian_reconstruction.py
```diff
@@ -0,0 +1,387 @@
...
+
+# %% [markdown]
+# Let's have a look at the shape of the obtained tensor.
+
+# %%
+print(img.shape)
+
+# %% [markdown]
+# We can see that the second dimension, which is the coil dimension, is 16. This means we still have a coil resolved
+# dataset. We can use a simply root-sum-of-squares approach to combine them into one. Later, we will do something a bit
```
ckolbPTB:
```suggestion
# dataset (i.e. one image for each coil element). We can use a simply root-sum-of-squares approach to combine them into one. Later, we will do something a bit
```

examples/cartesian_reconstruction.py
```diff
@@ -0,0 +1,387 @@
...
+# more sophisticated. We can also see that the x-dimension is 512. This is because in MRI we commonly oversample the
+# readout direction by a factor 2 leading to a FOV twice as large as we actually need. We can either remove this
+# oversampling along the readout direction or we can simply tell the `FastFourierOp` to remove it by providing the
```
ckolbPTB:
```suggestion
# oversampling along the readout direction or we can simply tell the `FastFourierOp` to crop the image by providing the
```

examples/cartesian_reconstruction.py
```diff
@@ -0,0 +1,387 @@
...
+    kdata_pe_pf.traj.as_tensor()[2, 0, 0, :, :].flatten(), kdata_pe_pf.traj.as_tensor()[1, 0, 0, :, :].flatten(), '+r'
+)
+
+# %% [markdown]
+# We see that for the fully sampled acquisition, the k-space is covered symmetrically from -256 to 255 along the
+# readout direction and from -128 to 127 along the phase encoding direction. For the acquisition with partial Fourier
+# and partial echo acceleration, this is of course not the case and the k-space is asymmetrical.
+#
+# Our FFT-operator does not know about this and simply assumes that the acquisition is symmetric and any difference
+# between encoding and recon matrix need to be zero-padded symmetrically.
```
ckolbPTB:
```suggestion
# between encoding and recon matrix needs to be zero-padded symmetrically.
```

fzimmermann89:
Why does this use the fftop?

Shouldn't we push for using the FourierOp?
Fftop is an implementation detail.

Also, we should not demonstrate trajectory.as_tensor to get Kx.
Instead use trajectory.kx directly

ckolbPTB:
This notebook is meant to provide some background information on what is happening in MRpro. The conclusion is that FourierOp or even better one of the algorithms should be used.

fzimmermann89:
I would just not mention FastFourierOp and only use FourierOp. CartesianSampling Op and FastFourierOp are implementation details.

This notebook is now the "gold standard" recipe for manual reconstruction of cartesian data
And it only shows FastFourierOp (and the all-in-one reconstruction.)

I would never use FastFourierOp for any reconstruction.
I am willing to bet that their will be ppl using FastFourierOp in their code and not even knowing about FourierOp.
This was already an issue in the past -- fft is to "obvious" to not use it. And having this as *the* cartesian reco example is IMHO a bad idea.

I would have removed all the cartesian sampling op stuff, and directly used the fourierop...

PR #567, opened 2024-11-28
Ensure compressed number of coils is not greater than existing number

fzimmermann89:
compression shall be compression, otherwise the code fails.

I noticed, because for performance reasons, I decreased the number of coils in some test data.

While debugging, I also changed some variable names, splitting kdata_coil_compressed into multiple varaibles


fzimmermann89:
* #594
* #591
* #580
* #568
* #560
* __->__ #567
* #559

ckolbPTB (approved):

PR #495, opened 2024-10-31
Add CheckDataMixin

fzimmermann89:

PR #568, opened 2024-11-29
Add CheckDataMixin

fzimmermann89:
This adds a new Mixin for Dataclasses that checks the types of the fields.
If the fields are Annotated with a (newly added) `Annotation`, specifying either its data type or shape, this will also be checked.

After some discussions following #495, this is now optional. None of the core reshaping or indexing functions will depend on these checks, but the will be performed afterwards. If this ever becomes an issue, removing the mixin from the dataclass will only remove the checks.
This is IMHO the best way to check that certain invariants are true after reshaping or indexing in a readable and maintable way.
An alternative I considered was using data class descriptors instead of Annotated and perform the check in the set of the descriptors. But this would make the checking code fundamental to the data classes, instead of just a checking addon that can be removed.

One can specify the expected shape of a field using the syntax of [jaxtyping](https://docs.kidger.site/jaxtyping/api/array/).

There is also a context manager to disable checks temporarily. 
By default, when exiting the context, the suspended checks will be performed. So one can first create a wrongly shaped data class, reshape it, and on exit it is ensured that the fields follow the rules. This behavior can also be disabled to disable checks in a part of the code completely.

The string parsing etc happens on import time, not at creation time.
Parts of the shape checks are cached, so running the same invariants check multiple times without substantial changes to the shapes will be fast.

I also removed some `__future__.annotation ` imports, as they stringify all type annotations. Instead, we should either fix the dependencies, use stringified annotations (these cannot be runtime checked) or wait for  [PEP 649/749](https://peps.python.org/pep-0749/)


fzimmermann89:
* #661
* #660
* #591
* #580
* #657
* __->__ #568
* #560
* #656
* #655

tests/data/test_checkdatamixin.py
```diff
@@ -0,0 +1,259 @@
...
+        elements += ','  # Add trailing comma for single-element tuples
+    return f'\\({elements}\\)'
+
+
+# Tests that these can be defined
+@dataclass
+class CheckedDataClass(CheckDataMixin):
+    """A test dataclass."""
+
+    float_tensor: Annotated[
```
fzimmermann89, line 36: See here for an example of the Annotations

tests/data/test_checkdatamixin.py
```diff
@@ -0,0 +1,259 @@
...
+    not_one = 17
+    with pytest.raises(ShapeError, match=f' the dimension size {not_one} does not equal 1'):
+        CheckedDataClass(
+            float_tensor=torch.ones(*n_other, n_coil, n_k2, n_k1, n_k0),
+            int_tensor=torch.zeros(*n_other, 1, n_k2, 1, not_one, dtype=torch.int),
+            string='test',
+        )
+
+
+def test_suspend_check_success() -> None:
```
fzimmermann89, line 196: see here (and next text) for the suspend checks context manager

PR #571, opened 2024-11-30
Tests for signal models on gpu

ckolbPTB:
These tests verify that signal models can be created and/or run on cuda devices. 

Currently all tests pass except for `WASABITI` and `TransientSteadyStateWithPreparation`. These are the only models with two parameters in the `init` and we do not verify that they are on the same device. 

If I remember correctly then we said that if an operator is created with a cuda-tensor, then the operator should be created on the cuda device. For opertors with multiple `init` parameters we need to have a key-parameter which determines where the operator is created. This could mean that for

- WASABITI: if `offsets` is on a cuda-device, the signal model is created on the cuda-device independent of all the other parameters
- TransientSteadyStateWithPreparation: if `sampling_time` is on a cuda-device, the signal model is created on the cuda-device independent of all the other parameters

Option 2 would be that ALL `init` parameters have to be on the same device and we throw an error if not. I would prefer option 1 as this seems more user friendly. 

Related to #525


ckolbPTB:
Wrong starting branch...

PR #572, opened 2024-11-30
Tests for signal models on gpu

ckolbPTB:
These tests verify that signal models can be created and/or run on cuda devices. 

Currently all tests pass except for `WASABITI` and `TransientSteadyStateWithPreparation`. These are the only models with two parameters in the `init` and we do not verify that they are on the same device. 

If I remember correctly then we said that if an operator is created with a cuda-tensor, then the operator should be created on the cuda device. For opertors with multiple `init` parameters we need to have a key-parameter which determines where the operator is created. This could mean that for

- WASABITI: if `offsets` is on a cuda-device, the signal model is created on the cuda-device independent of all the other parameters
- TransientSteadyStateWithPreparation: if `sampling_time` is on a cuda-device, the signal model is created on the cuda-device independent of all the other parameters

Option 2 would be that ALL `init` parameters have to be on the same device and we throw an error if not. I would prefer option 1 as this seems more user friendly. 

Related to #525


fzimmermann89:
I would not move tensors to different devices, but print a warning.
A signalmodel.cpu() would still fix it afterwards, so no need for an exception.

ckolbPTB:
> I would not move tensors to different devices, but print a warning.

I thought that we decided that if the FourierOp gets the trajectory on cuda, then we want FourierOp also to be on cuda. I.e. not necessarily all input parameters have to be on the same device but we decide based on a key-input parameter or am I remembering this wrongly? 
This would also mean that a DcfOp created from DcfData on cuda will also be on cuda which is most intuitive. 

Otherwise, if we take the WASABI model as an example then I could do WASABI(offsets.cuda(), tp=1.0) which would be a valid and on cuda. But if I use WASABI(offsets.cuda(), tp=torch.ones((1,)) then this would not work and I would have to do WASABI.cuda() before I can use it. This seems IMO not intuitive.

schuenke (approved):
I like the key-input parameter solution. I think more than 90% of the endusers will not care about devices at all. They just want the code to work. And IMO the key-input parameter solution is the most convinient solution for this.

Users that work with different devices usually also know how to check devices and transfer parameters from cpu to gpu and vice versa and are thus fine as well.

PR #577, opened 2024-12-11
Discussion of the indexing rules

fzimmermann89:
This is a suggestion of the indexing rules.

It is not yet polished, but I would like some feedback.

Please just do a "review" and add your comments to the specific point if possible,
This allows us to have discussion threads for the different cases.

In general, there will be some deviations from numpy or torch indexing.
most notably, indexing will not remove dimensions.
this applies to integer indices, vectorized indexing and boolean indexing.

indexing will work as-if-broadcasted.

indexing will return the smallest tensors that can hold the data.
This requires a bit of logic in case a copy is created, but is essential to keep memory consumption of the indices low.
We must avoid having a contiguous coil and k0 dimension in the indices, even if indexing results in a copy of the data.

### DIFF
--- a/docs/source/indexing.rst
+++ b/docs/source/indexing.rst
@@ -1,52 +1,82 @@
-## indexing rules
+Indexing Rules
+==============
 
-the data objects can be indexed, similiar to numpy arrays or tensors
+Data objects can be indexed similarly to NumPy arrays or tensors.
 
+1. **Indexing as tuples**
 
-1. as in numpy, an index is always a tuple of indices. if not, is is converted to a tuple. the elements the tuple are matched to the dimensions of the data left to right.
-   example:
-	data[0,1] is the same as data[(0,1)], as the index is converted to a tuple
-        data[(0,1),] is not the same as  data[(0,1)], but data[((0,1),)]
-2. all indexing operations are perfomred on the fields of a dataclass as the sequence of 
-    broadcasting to the dataclasses shape -> indexing -> replacing stride 0 dimensions by singletons as good as possible.
-    this means each dataobject can be indexed as if all fields had the shape returend by dataobject.shape, even if the fields
-    are only broadcastable to this shape (example: acqidx are singleton in k0)
+   - As in NumPy, an index is always a tuple of indices. If it is not, it is converted to a tuple. The elements of the tuple are matched to the dimensions of the data from left to right.
 
-3. two data objects with the same shape property indexed by the same indexer will result in the same shape property.
-     this explicitly does not mean that all fields of the data class have the same shape after indexing. 
-     only that the fields can be broadcasted to the common shape
+   - Example::
 
-4. slice indexing
-	indexing with a slice object or colon synax, e.g. 0:3 will select a slice of the data
-	it does a view along that dimension if the stride is positive.
-	a[slice1,slice2]=a[slice1,:][:,slice2]=a[:,slice1][slice1,:]
-    as the underlying torch.tensor does not support negative strides, a copy will be performed for negative strides.
-    this is differnet from numpy which always returns a view for slices and from torch, which errors with negative strides.
+       data[0, 1] == data[(0, 1)]  # The index is converted to a tuple
+       data[(0, 1), ] != data[(0, 1)] 
+	   data[(0, 1), ] == data[((0, 1), )]
 
-5. integer indexing
-	integer indexing is the same as indexing with slice index:index+1.
-	thus, it does not remove dimensions. this is different than numpy or pytorch.
-	use squeeze to remove remaining singleton dimensions if requiered.
-	integer indexing always results in a view.
+2. **Broadcasting**
 
-6. advanced indexing with sequences of ints
-	if a single indexer is a sequence of ints, example data[:,(0,5)] 
-	for each value in the sequence indexing will be performed and the result will be conatenated along the indexing dimension.
-	thus, data[:,(0,5)].shape[1]==2.
+   - Indexing operations are performed on the fields of a dataclass in the following sequence:
 
-7. vectorized indexing
-	using more than one integer sequence as index will is done by looping over the values in the sequences, performing
-	indexing and stacking along a new dimension at position 0.
-	example: data[(0,5),:,(2,3)] will result in the same as if one would do stack([data[0,:,2],data[5,:,3],dim=0).
-	perticaulary, at the position of the indexed dimensions a singelton dimenion will remain.  	(different than numpy)
-	this rulke applies irregedless of wheatrher the indexed dimensions are next to each other or not (different than numpy)
+     1. Broadcasting to the dataclass's shape.
+     2. Indexing.
+     3. Replacing stride-0 dimensions with singletons wherever possible.
 
-8. indexing with bolean mask
-        the boolean mask will first be broadcasted to the shape of the indexed object.
-	thus, even though acqidx always are singelton in k0, kdata.header.acqinfo.idx.slice=1 can be used to index kdata.
-	the result is the same as indexing with [mask.nonzero(as_tuple=True)[k] if mask.shape[k]!=1 else slice(None) for k in range(mask.ndim)]
-	in particular, indexing with a single non singleton dimension will not introduce new axes nor change the shape of the data in the singleton axes.
-	but indexing with more than one non-singleton dimension will introduce a new zeroth dimension and result in two singleton dimensins in the data.
-	
-	
-	
+   - This means each data object can be indexed as if all fields had the shape returned by ``dataobject.shape``, even if the fields are only broadcastable to this shape.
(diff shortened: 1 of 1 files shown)


fzimmermann89:
@JoHa0811 @Stef-Martin @ckolbPTB @schuenke (and ofc everybody else)

docs/source/indexing.rst
```diff
@@ -0,0 +1,82 @@
...
+
+       data[0, 1] == data[(0, 1)]  # The index is converted to a tuple
+       data[(0, 1), ] != data[(0, 1)] 
+	   data[(0, 1), ] == data[((0, 1), )]
+
+2. **Broadcasting**
+
+   - Indexing operations are performed on the fields of a dataclass in the following sequence:
+
+     1. Broadcasting to the dataclass's shape.
```
ckolbPTB, line 20: Just to verify, this is broadcasting the shape of the parent dataclass to all children? i.e. if I index kdata along k0 then acqidx is broadcasted first to the shape of kdata and then indexed. But you could not index kdata.header.acq_info along k0 directly, could you?

docs/source/indexing.rst
```diff
@@ -0,0 +1,82 @@
...
+4. **Slice indexing**
+
+   - Indexing with a slice object or colon syntax (e.g., ``0:3``) selects a slice of the data.
+   - A view along the dimension is returned if the stride is positive.
+
+   - Example::
+
+       a[slice1, slice2] == a[slice1, :][:, slice2] == a[:, slice1][slice1, :]
+
+   - For negative strides, a copy is performed because the underlying ``torch.Tensor`` does not support negative strides. This differs from NumPy, which always returns a view, and PyTorch, which raises an error for negative strides.
```
ckolbPTB, line 42: would it not make sense to also return an error? I hate that pytorch does not support negative strides, but I guess the users of MRpro will most likely expect pytorch behaviour as much as possible.

docs/source/indexing.rst
```diff
@@ -0,0 +1,82 @@
...
+
+5. **Integer indexing**
+
+   - Integer indexing is equivalent to slicing with ``index:index+1``, so it does not remove dimensions.
```
ckolbPTB, line 46: This means that kdata[:,0,:,:,:] will be of shape (other 1 k2 k1 k0) rather than (other k2 k1 k0)?

docs/source/indexing.rst
```diff
@@ -0,0 +1,82 @@
...
+
+       data[:, (0, 5)].shape[1] == 2
+
+7. **Vectorized indexing**
+
+   - Using more than one sequence of integers as indices results in vectorized indexing by looping over the values in the sequences, performing indexing, and stacking along a new dimension at position 0.
+
+   - Example::
+
+       data[(0, 5), :, (2, 3)] == stack([data[0, :, 2], data[5, :, 3]], dim=0)
```
ckolbPTB, line 64: What would the shape be here? assume data.shape = (z,y,x) than data[(0, 5), :, (2, 3)].shape = (2,1,y,1)?

docs/source/indexing.rst
```diff
@@ -0,0 +1,82 @@
...
+
+   - Example::
+
+       kdata.header.acqinfo.idx.slice == 1  # Can be used to index `kdata`
+
+   - The result is the same as::
+
+       [mask.nonzero(as_tuple=True)[k] if mask.shape[k] != 1 else slice(None) for k in range(mask.ndim)]
+
+   - Indexing with a single non-singleton dimension will not introduce new axes or change the shape of the singleton dimensions.
```
ckolbPTB, line 81: what do you mean here with "not [...] change the shape of the singleton dimensions?

docs/source/indexing.rst
fzimmermann89: yes.
(other k2 k1 k0) would not be an allowed shape anyways as we enforce  (*other coils k2 k1 k0)

docs/source/indexing.rst
fzimmermann89: yes

fzimmermann89: numpy would instead do (2,y)

docs/source/indexing.rst
fzimmermann89: indexing data with shape (5,4,3) with mask shape (1,4,1) will result in 5,N,3. where N is determined by the mask.
the 5 and 3 shape of data are not changed.

fzimmermann89: I.e. the number of dimensions stays again the same.

docs/source/indexing.rst
fzimmermann89: i was considering that, but would you expect an error or rely on the error?
this is strictly a superset of pytorch.

docs/source/indexing.rst
fzimmermann89: yes.

the shape goes only upwards.the shape of a dataclass is always larger than of its fields.
something to also describe somewhere.

so, an acqidx will have shape k0=1, acq_info will be shape k0=1
but kdata will be the broadcasted shape of all acqinfo, data, traj

I was considering allowing indexing into shape 1 dimensions, so you could index.
so you could index into n_other=1 with (0,1,2),... and you would get n_other=3, indexing with any slice or single integer will keep the shape as n_other=1.
so kdata[0]=kdata[12322] if n_other==1.
(this is how indexing  would be implemented anyways to keep broadcasted dimensions broadcasted, just that there is an explicit check to raise an error if indexed out of bounds.)

docs/source/indexing.rst
```diff
@@ -0,0 +1,82 @@
...
+   - Indexing with more than one non-singleton dimension introduces a new zeroth dimension and results in two singleton dimensions in the data.
```
fzimmermann89:
```suggestion
   - Indexing with a mask with more than one non-singleton dimension introduces a new zeroth dimension and results in singleton dimensions in the data.
  Example: 
        data.shape=(5,4,3) 
        mask.shape=(5,1,3)
        data[mask].shape==N,1,4,1 
        with N determined by the number of True values o the mask.

9. ellipses and slice(None)
     an indexing expression can contain a single ellipsis, which will be expanded to [slice(None)]*(data.ndim-sum(index_dims)). Here, index_dims in 1 for each slice and integer indices and mask.ndim for boolean indices.

10. None
 using None in an index is not supported. Use the rearrange pattern to introduce new axes.
```

fzimmermann89: ...None is forbidden as [...,None] would cause issues with the rule that acqidx are singleton in coils.
introducing new axes can still be done with rearrange, and after the rearrange pattern, the shape rules will be checked anyways. 
I am considering an `unsqueeze()` that can only unsqueeze in "other" dimensions as a simpler way than` kdata.rearrange('...-> 1 ...')` ...

docs/source/indexing.rst
ckolbPTB: my only concern would be that people more used to numpy often use negative strides and would expect a view and it would not be obvious to them that we do a copy. If the user knew that a copy is made then they might use a different approach which does not require a copy. Maybe we could print a warning?

fzimmermann89: Warning is a good idea!

docs/source/indexing.rst
schuenke: data[(0, 4, 5), :, (2, 3)] (sequences of different lengths) would not be allowed I guess?!

fzimmermann89: this could only be interpretet as outer indexing, not vectorized indexing, right? 
different sequence lengths (are integer array shapes) are incompatible with vectorized indexing. or which elements should be indexed by this?

numpy would make  you do data[(0, 4, 5), :, :][:, :, (2, 3)] for outer indexing, or use np.ix_.
(this is also true for two sequences of  the same length,  ` data[(0, 4), :, :][:, :, (2, 3)] != data[(0, 4), :, (2, 3)]
`
I should explicitly mention it and add this as an example, as this is also a common misconception with numpy indexing in  my non-represenative survey of 8.1 PhD students :D

docs/source/indexing.rst
fzimmermann89: I am now considering allowing None at position 0.
Then the rule is: indexing will never remove axes. indexing can introduce new axes in front of the array.
this happens if using None, vectorized indexing, or boolean indexing along more than 1 dimension

docs/source/indexing.rst
fzimmermann89: ..i will skip this for now and just raise an exception

docs/source/indexing.rst
fzimmermann89: This also needs an example that
data.shape=(2,2) 
data[3].shape == (0,2)
(which is what would happen for slices in numpy)

PR #578, opened 2024-12-11
refactor of trajectory type

fzimmermann89:
will most likely never be merged but some parts might be useful for @ckolbPTB


fzimmermann89:
Mostly made absolute by #463.
Will have to pull out the actual trajectory type code at some point.

Issue #320, opened 2024-06-25
Template for issues


ckolbPTB:
Create template for issues, see e.g. https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository

ckolbPTB:
Suggestion: Template for feature, bug and discussion

Suggestion: Template also for corresponding PRs

PR #579, opened 2024-12-12
Add self hosted runner

fzimmermann89:


fzimmermann89:
That works if i run 
`docker run -it -e REPO_NAME=PTB-MR -e TOKEN=ABH....  image`
with the image build from the Runner dockerfile on my laptop.

fzimmermann89:
a worker token can be obtained via
` curl -X POST -H "Authorization: Bearer $PAT" https://api.github.com/orgs/PTB-MR/actions/runners/registration-token | jq -r '.token'` (PAT is a personal access token)

fzimmermann89:
@schuenke @ckolbPTB 
This seems to work with our proxy. The worker is currently running on e....30. 
As long as we don't need artifacts etc that would require access to domains blocked by our firewall.


I am currently for testing using the latest mrpro image as base, so no cuda driver.

Also, we should decide when we run this. either only triggered via a comment by a maintainer or only after the merge to main.
And only approved workflows should be allowed to run on the self-hosted worker (setting in the organisation/settings/actions/worker pane)
It should not be possible to make our worker run code via forking without us approving it. We don't want to provide a crypto miner ;)

fzimmermann89:
@lrlunin said he is interested in continuing here. He will start working using his laptop.

@ckolbPTB after your vacation, lets have a short chat about what system we can use for developing this.

fzimmermann89:
https://github.com/ci-for-research/self-hosted-runners/blob/master/ubuntu-singularity/entrypoint.sh


Should work in singularity. Note the username environment variable.

fzimmermann89:
please add a start-runner.sh script that uses the correct image, singularity call, etc, obtains the registration token etc.

otherwise we will forget how to actually start the runner on our cluster.

lrlunin:
We would need to further decide when we would like to trigger the `pytest-selfhosted`...

fzimmermann89:
This is currently blocked due to office politics.

Before enabling this, we might need some risk assesment.

fzimmermann89:
Some updates here:
- We will try to get this to run on the hpc cluster (to have it running in a secured environment)
- We need to check if it works with the firewall rules of the clsuter
- We need to figure out a way to start a job for the tests. We can either periodically schedule a job that connects a worker and does all the jobs waiting for a worker, or we need to have one container running with the runner client, that starts another job actually performing the tests...
- we need a funktionsaccount

lrlunin:
The current state and design:

- In the head node of slurm cluster the `slurm_dispatcher.py` will be started.
- The dispatcher makes the GitHub API requests and asks for the Pytests selfhosted workflows in `queued` state (permitted by repo's approved members).
- If there is one - inits a time limited CUDA required slurm session with a runner within the isolated singularity container in special (ephemeral) mode that the runner takes a single job and then deactivates. Therefore, there is no GPU waste time. The slurm job will be scheduled only for the run. This terminates after the given time or when the runner has finished its job.

Currently the pytest-selfhosted workflow is triggered by any `pull_request`. There is an option to also be label selective in triggering: https://stackoverflow.com/questions/62325286/run-github-actions-when-pull-requests-have-a-specific-label

I suppose it would be a good solution to avoid the unnecessary triggering. We can run the self-hosted pytest for both pull-requests which suppose to be a release and for any pull-request which are labeled with the `pytest-selfhosted` label. I would like to emphasize that the execution of the workflow would still require a confirmation from a repo's approved persons (see current environment groups and deploy pending).

I guess the labelling design solution is related to #320 to define a issue labelling handling workflow.

lrlunin:
selfhosted run

lrlunin:
According to https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#issue_comment
> This event will only trigger a workflow run if the workflow file exists on the default branch.

PR #580, opened 2024-12-16
Add Indexer

fzimmermann89:
Implements broadcasted indexing on tensors.

Main idea is to parse the indexing expression once, then apply it to all tensor fields.

The indexing rules are such that indexing not violate the shape inveraints of our tensors:
 no axes will be removed. new axes are only introduced at the beginning of the tensor (in the other dimensions)


fzimmermann89:
* #661
* #660
* #591
* __->__ #580
* #657
* #568
* #560
* #656
* #655

PR #581, opened 2024-12-16
Add logo to docu and readme

schuenke:


lrlunin:
Why not SVG?

schuenke:
> Why not SVG?

Because we don't have it as SVG yet

schuenke:
I will reopen another PR directly from the main repo instead of my fork so that we don't have to wait for the workflow fixes.

PR #582, opened 2024-12-16
Move scaling to KTrajectory

fzimmermann89:
There is a bit of duplicaiton for KTrajectoryRawShape and KTrajectory, but inherting one from the other makes KData more complicated. I might remove this duplication in #560 

I also added the ability to load in Trajectories with the wrong, i.e. not zyx, order.


fzimmermann89:
Would be nice to get this in before the release @schuenke

src/mrpro/data/KTrajectory.py
```diff
@@ -85,13 +89,33 @@ def from_tensor(
...
             tolerance to detect if trajectory points are on integer grid positions
+        encoding_matrix
+            if an encoding matrix is supplied, the trajectory is rescaled to fit
+            within the matrix. Otherwise, it is left as-is.
         """
-        kz, ky, kx = torch.unbind(tensor, dim=stack_dim)
+        ks = tensor.unbind(dim=stack_dim)
+        kz, ky, kx = (ks[axes_order.index(axis)] for axis in 'zyx')
+
+        def normalize(k: torch.Tensor, encoding_size: int) -> torch.Tensor:
```
schuenke: rename normalize -> rescale / scale etc

src/mrpro/data/KTrajectory.py
```diff
@@ -85,13 +89,33 @@ def from_tensor(
...
         stack_dim
             The dimension in the tensor the directions have been stacked along.
+        axes_order
+            Order of the axes in the tensor. Our convention usually is zyx order.
         repeat_detection_tolerance
             detects if broadcasting can be used, i.e. if dimensions are repeated.
             Set to None to disable.
         grid_detection_tolerance
             tolerance to detect if trajectory points are on integer grid positions
+        encoding_matrix
```
schuenke:
```suggestion
        scaling_matrix
```

src/mrpro/data/KTrajectory.py
```diff
@@ -85,13 +89,33 @@ def from_tensor(
             This should be a 5-dim tensor, with (kz,ky,kx) stacked in this order along stack_dim
         stack_dim
             The dimension in the tensor the directions have been stacked along.
+        axes_order
+            Order of the axes in the tensor. Our convention usually is zyx order.
         repeat_detection_tolerance
             detects if broadcasting can be used, i.e. if dimensions are repeated.
             Set to None to disable.
```
schuenke:
```suggestion
        repeat_detection_tolerance
            Tolerance for detecting repeated dimensions (broadcasting).
            If trajectory points differ by less than this value, they are considered
            identical. Set to None to disable this feature.
```

src/mrpro/data/KTrajectory.py
```diff
@@ -85,13 +89,33 @@ def from_tensor(
...
         grid_detection_tolerance
             tolerance to detect if trajectory points are on integer grid positions
```
schuenke:
```suggestion
            Tolerance for detecting whether trajectory points align with integer 
            grid positions. This tolerance is applied after rescaling if 
            `scaling_matrix` is provided.
```

src/mrpro/data/KTrajectory.py
```diff
@@ -85,13 +89,33 @@ def from_tensor(
...
+        encoding_matrix
+            if an encoding matrix is supplied, the trajectory is rescaled to fit
+            within the matrix. Otherwise, it is left as-is.
```
schuenke:
```suggestion
            If a scaling matrix is provided, the trajectory is rescaled to fit within 
            the dimensions of the matrix. If not provided, the trajectory remains unchanged.

```

schuenke: and maybe add an example like
`i.e. SpatialDimension(z=1, y=64, x=64) for a single slice acquisition with 64x64 pixels`

schuenke (approved):
See my suggestions as discussed before. Use same/similar docstrings in other class as well. 
But code works nice. Tested with the MAGMA review notebook!

src/mrpro/data/KTrajectory.py
```diff
@@ -85,13 +89,33 @@ def from_tensor(
...
+                # a single encoding point should be at zero
+                # avoid division by zero
+                return torch.zeros_like(k)
+            return k * (encoding_size / max_abs_range)
+
+        if encoding_matrix is not None:
+            kz = normalize(kz, encoding_matrix.z)
+            ky = normalize(ky, encoding_matrix.y)
+            kx = normalize(kx, encoding_matrix.x)
+
```
fzimmermann89:
```suggestion
        def rescale(k: torch.Tensor, size: float) -> torch.Tensor:
            max_abs_range = 2 * k.abs().max()
            if size < 2 or max_abs_range < 1e-6:
                # a single encoding point should be at zero
                # avoid division by zero
                return torch.zeros_like(k)
            return k * (size / max_abs_range)

        if scaling_matrix is not None:
            kz = rescale(kz, scaling_matrix.z)
            ky = rescale(ky, scaling_matrix.y)
            kx = rescale(kx, scaling_matrix.x)

```

PR #584, opened 2024-12-16
Fix and modification for TrajType

ckolbPTB:
The tests for TrajType failed silently after we changed the naming of trajectory types in conftest. Add extra check on this and added some modifications from #578 

@fzimmermann89 : Do you remember why we are setting singleton dimensions to `TrajType.SINGLEVALUE.value | TrajType.ONGRID.value` here:
https://github.com/PTB-MR/mrpro/blob/da02db8c5bc4b951242c69c0d6e92d0f172e9aeb/src/mrpro/data/KTrajectory.py#L153

### DIFF
--- a/src/mrpro/data/enums.py
+++ b/src/mrpro/data/enums.py
@@ -1,6 +1,6 @@
 """All acquisition enums."""
 
-from enum import Enum, Flag, auto
+from enum import Enum, Flag, IntFlag, auto
 
 
 class AcqFlags(Flag):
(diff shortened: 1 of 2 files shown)


fzimmermann89:
A Singleton dimension will always be on grid, right?

The idea was to add more properties in one single place, i.e. fully sampled and sorted.
And for each of them, the flag can be true or false..
And w Singleton dimension will be sorted, fully sampled etc...

fzimmermann89 (approved):

PR #585, opened 2024-12-17
Fix public pull requests coverage and docs workflows comments 

lrlunin:
Split the build steps which trigger on `pull_request` from the steps which make reports in this (parent) repository. The steps which make comments now triggered by the build workflows via `workflow_run`.

Check my playground repository for and make your pull requests to test the action: https://github.com/lunin-inc/amazing-repo


.github/workflows/report_docs.yml
```diff
@@ -0,0 +1,96 @@
...
+    needs: docs_report
+    # maybe should be test that the parent workflow was successful...
+    if: needs.docs_report.outputs.event-name == 'push' && needs.docs_report.outputs.ref == 'refs/heads/main'
+    permissions:
+        pages: write
+        id-token: write
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    steps:
```
lrlunin, line 78: The most critical part I am not sure about

.github/workflows/report_docs.yml
```diff
@@ -0,0 +1,96 @@
...
+      - name: Upload pages
+        id: upload_pages
+        uses: actions/upload-pages-artifact@v3
+        with:
+          path: html_build
+
+      - name: Deploy to GitHub Pages
+        id: deployment
+        uses: actions/deploy-pages@v4
+
```
lrlunin: That's the part I'm most unsure about.

ckolbPTB:
Can we have this as a PR from a fork to ensure it does what we think it should do? Currently I do not see any reports anymore here - or am I missing something?

fzimmermann89:
Leonid has a test repo set up 

And yes, this does not work yet

lrlunin:
> Leonid has a test repo set up 
> 
> And yes, this does not work yet 

It does work. Sync your pull request. There was only an issue with set the Workflow to be failed. The Report of docs and coveragw worked as expected though.

To get this work here you need to have the following permissions set.
image

schuenke:
@lrlunin Maybe use this PR to add yourself to the authors list in the pyproject.toml file

schuenke:
> > Leonid has a test repo set up
> > And yes, this does not work yet
> 
> It does work. Sync your pull request. There was only an issue with set the Workflow to be failed. The Report of docs and coveragw worked as expected though.
> 
> To get this work here you need to have the following permissions set. image

This is something we would need to change in the organization settings. It's not possible to enable it for a single repo within an organization. However, it should also work by granting the permission on the workflow level, no?

.github/workflows/pytest.yml
```diff
@@ -85,31 +85,42 @@ jobs:
...
-          pytest-coverage-path: ./pytest-coverage.txt
-          junitxml-path: ./pytest.xml
+      - name: Read test failures and errors
+        run: |
+          python -c "
+          import os
+          import xml.etree.ElementTree as ET
+          tree = ET.parse('pytest.xml')
+          root = tree.getroot()
+          testsuite = root.find('testsuite') if root.tag != 'testsuite' else root
```
fzimmermann89: does the xml generated have a testsuites root element or not? (why is there the conditional?)
would
`tree.getroot().find(".//testsuite[@name='pytest']")` suffice? (and also make sure that only the pytest results are used)

.github/workflows/pytest.yml
```diff
@@ -85,31 +85,42 @@ jobs:
...
-          namedLogo: python
+          name: pytest-report
+          path: |
+            pytest-coverage.txt
+            pytest.xml
+            pytest_data.json
+          if-no-files-found: error
+          overwrite: true
 
       - name: Set Pipeline Status Based on Test Results
```
fzimmermann89: this was a quick fix because i was to lazy to try out something better,
do you want to check if
https://stackoverflow.com/questions/62045967/is-there-a-way-to-continue-on-error-while-still-getting-correct-feedback
would work here?

it would be nice if the pytest run step would be marked as failure, but the following steps still run (to create the comment)

.github/workflows/report_docs.yml
```diff
@@ -0,0 +1,95 @@
...
+      - name: Parse docs data docs_data.json to outputs
+        id: parse_docs_data
+        run: |
+          echo "artifact-url=$(jq -r '.artifact_url' docs_data.json)" >> $GITHUB_ENV
+          echo "pr-number=$(jq -r '.pr_number' docs_data.json)" >> $GITHUB_ENV
+          echo "event-number=$(jq -r '.event_number' docs_data.json)" >> $GITHUB_ENV
+          echo "event-name=$(jq -r '.event_name' docs_data.json)" >> $GITHUB_ENV
+          echo "ref=$(jq -r '.ref' docs_data.json)" >> $GITHUB_ENV
+
+      - name: Set outputs
```
fzimmermann89: why?

.github/workflows/report_docs.yml
```diff
@@ -0,0 +1,95 @@
...
+        uses: actions/download-artifact@v4
+        with:
+          run-id: ${{ github.event.workflow_run.id }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          name: docs-data
+
+      - name: Parse docs data docs_data.json to outputs
+        id: parse_docs_data
+        run: |
+          echo "artifact-url=$(jq -r '.artifact_url' docs_data.json)" >> $GITHUB_ENV
```
fzimmermann89: you could also just save as a test file instead with name=value, instead of json I think,
and then do
`cat docs_data >> $GITHUB_ENV`
as this is the format that needs to be put into github_env anyways, right?

.github/workflows/report_docs.yml
```diff
@@ -0,0 +1,95 @@
...
+            ### :books: Documentation
+            :file_folder: [Download as zip](${{ env.artifact-url }})
+            :mag: [View online](https://zimf.de/zipserve/${{ env.artifact-url }}/)
+          edit-mode: replace
+
+  deploy:
+    name: Deploy docs
+    runs-on: ubuntu-latest
+    needs: docs_report
+    # maybe should be test that the parent workflow was successful...
```
fzimmermann89: i understand you comment here:
wouldnt the downlaod of the artifact fail if the creation of docs failed?
or will it delete the existing old documentation and not upload something if the build workflow failed?

can you clarify?

fzimmermann89 (approved):
some minor things and questions

.github/workflows/pytest.yml
lrlunin: Actually, it works exactly as you say. The `report_pytest.yml` runs even if the `pytest.yml` fails. The question - what should it report in the message if there is no such output of pytest?

So far, if there is any failures or errors in the `pytest` then the `pytest.yml` marked as failed but the report is still here.

.github/workflows/report_docs.yml
lrlunin: Both of the new workflows `report_docs.yml` or `report_pytest.yml` are always invoked when the `docs.yml` or `pytest.yml` are completed independet from their state.

For the `pytest.yml` I explicitly say that the `pytest.yml` should be marked as failed if any of the tests fails. The pytest report message will be generated anyway. This makes sence and is an expected behaviour.

For the `docs.yml` there is no scenario when the `docs.yml`'s fail makes sence. If the `docs.yml` workflow would fail - then,  presumably, no html will be built and no artifact will be uploaded too. Then, obviosly, `report_docs.yml` will fail at the very first step.

I guess a little bit cleaner solution would be to check if the `docs.yml` is failed and then do not run the `report_docs.yml` at all and have it skipped. An error related to the no existence of empty artifact link will bring not much sense for the user.

.github/workflows/report_docs.yml
fzimmermann89: why?

.github/workflows/pytest.yml
fzimmermann89: The issue originally was, that if pytest failed to to an error in one of the tests, it still write out the XML fail - BUT the following steps did not run and the comment that something failed was never created. Thus, we added an "or true" to the pytest run, to always make the step succeed. and then later on check if the test failed to mark the pipeline as failed (to see it in the checks GUI in a PR).

So, if pytest run but has errors, there should be a comment. and there should be a red x next to the pytest job.

if no pytest XML is created at all, there should be a red x next to the job. ideally the comment would be "Pytest crashed", but this is not that important.


Do you think this can work?

.github/workflows/report_docs.yml
fzimmermann89: or, even better, if docs failed, replace the docs comment by a "Documentation build failed" message?

otherwise, there is the link to a old documentation from a previous commit..

.github/workflows/pytest.yml
lrlunin: This sounds really good. Yeah, this should work.

.github/workflows/report_docs.yml
lrlunin: I believe it works since [7ad8955](https://github.com/PTB-MR/mrpro/pull/585/commits/7ad89559961d3f50c56de0d215cbd590ab99d627) as expected.

For pytest:
1) if the test case fails then is report comment generated, xml is parsed in the pytest step an marked this step as failed.
2) if the pytest command itself fails then this step is aborted (xml parse is not reached), upload artifacts is marked with always. the pytest step is marked as failed due to the pytest exit. the comment workflow gets the artifact anyway, tries to create a normal coverage report. if it is empty - post a comment that the pytest coverage is not available

For docs:
1) if the docu is built succesfully it will be uploaded as artifact and artifact url will be not empty. the report_docs checks if the artifact-url is empty or not.
2) if the docu is not built (sphinx-build command fails) then there is no Documentation uploaded articfact and the url would be empty. if the artifact-url is empty the report_docs post a comment that the documentation build failed.

.github/workflows/pytest.yml
```diff
@@ -77,39 +77,34 @@ jobs:
         run: |
```
fzimmermann89, line 78: Idea: Can you check if instead of the python code, a simple set -o pipefail (before the pytest command, in the same step) works?
Then the whole command is marked as failure if any part of the pipe fails

lrlunin: Yeah, this would be much smarter...

ckolbPTB:
> This is something we would need to change in the organization settings. It's not possible to enable it for a single repo within an organization. However, it should also work by granting the permission on the workflow level, no?

Do we still have to modify the permissions? If we can only change them for the entire organization, it would be better if it could be done on the workflow level.

If we now have docs.yml and report_docs.yml separate, can it not happen that report_docs.yml runs before docs.yml? 

Are the "test PRs" here https://github.com/lunin-inc/amazing-repo/pulls up to date?

.github/workflows/report_pytest.yml
```diff
@@ -0,0 +1,72 @@
+name: Report PyTest
+
+on:
+  workflow_run:
+    workflows:
+      - PyTest
+    types:
+      - completed
```
fzimmermann89, line 8: @ckolbPTB runs after the pytest job

fzimmermann89:
> > This is something we would need to change in the organization settings. It's not possible to enable it for a single repo within an organization. However, it should also work by granting the permission on the workflow level, no?

If we enable it at the org level, we would still need to enable it for each repo.


But I am not convinced we want or need this for this repo.



We should not have to change this settings if I understand the example at https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/ and https://securitylab.github.com/resources/github-actions-building-blocks/ correctlz.

fzimmermann89:
Also, I think the actions should not have
`content: write `
permissions, could you check?

.github/workflows/pytest.yml
```diff
@@ -74,42 +74,28 @@ jobs:
         run: pip install pytest-github-actions-annotate-failures
 
       - name: Run PyTest and Generate Coverage Report
+        # with the "set -o pipefail" the "pytest" exit code won't be suppressed by "tee" command
         run: |
-          pytest -n 4 -m "not cuda" --junitxml=pytest.xml \
-          --cov-report=term-missing:skip-covered --cov=mrpro | tee pytest-coverage.txt
+          (set -o pipefail && pytest -n 4 -m "not cuda" --junitxml=pytest.xml \
```
fzimmermann89: https://github.com/actions/runner-images/issues/4459#issuecomment-1374859960

setting shell to bash also enables  pipefail

fzimmermann89: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference

fzimmermann89 (approved):

schuenke (approved):

PR #588, opened 2024-12-18
Add download badge in docs build

lrlunin:


schuenke (changes requested):
The badge looks like this for me:
grafik
and IMO it's not obvious what to expect when clicking it. Maybe change the text to "Download notebook"

.github/workflows/docs.yml
```diff
@@ -106,6 +106,26 @@ jobs:
         run: |
           echo "current jupyter-notebook: ${{ matrix.notebook }}"
 
+      - name: Add mb-nyst download badge
+        run: |
+          # get name of the notebook
+          notebook=${{ matrix.notebook }}
+          notebook_name=$(basename $notebook)
+          # and the for the myst only the basename for path: reference need to be taken
```
schuenke: I don't understand this comment. Would remove it completely.

.github/workflows/docs.yml
```diff
@@ -106,6 +106,26 @@ jobs:
...
+          download_badge_md="[![Download notebook](https://img.shields.io/badge/Download-notebook-blue?logo=jupyter)](path:$notebook_name)"
+          python_command="import nbformat as nbf\n\
+          nb = nbf.read(open('$notebook'), as_version=4)\n\
+          # if there there is no cell with colab badge - create an empty md cell\n\
```
schuenke: remove one "there"

schuenke (approved):
2 minor things. See my comments.

Did you test it with both a colab and a download badge already?

lrlunin:
> Did you test it with both a colab and a download badge already?

Yep, works as expected.

PR #594, opened 2024-12-29
getnested

fzimmermann89:
idea is to use it for getting the parameters from the ismrmrd header.

have to consider how to deal with type hinting.
alternative is to ignore it and return Any or always force a default.
Or make it a callable generic class that looks like a function but can be called as
getnested[int](...) to tell mypy that it returns an int.


fzimmermann89:
* (to be filled)

PR #596, opened 2025-01-06
Add TV-regularized reconstruction and TV-based denoising

ckolbPTB:
Requires #426


src/mrpro/algorithms/reconstruction/TotalVariationRegularizedReconstruction.py
```diff
@@ -0,0 +1,127 @@
...
+from mrpro.operators import LinearOperatorMatrix, ProximableFunctionalSeparableSum
+from mrpro.operators.FiniteDifferenceOp import FiniteDifferenceOp
+from mrpro.operators.functionals import L1NormViewAsReal, L2NormSquared, ZeroFunctional
+from mrpro.operators.LinearOperator import LinearOperator
+
+
+class TotalVariationRegularizedReconstruction(DirectReconstruction):
+    r"""TV-regularized reconstruction.
+
+    This algorithm solves the problem :math:`min_x \frac{1}{2}||(Ax - y)||_2^2 + ||L\nabla x||_1`
```
koflera: maybe it would be good to distinguish between \lambda * \| \nabla x \|_1 (for scalar reg parameter) and \| \Lambda \nabla x \|_1 for an entire lambda-map. Also, strictly speaking, we would need to introduce the notation to have different regularization parameters for different directions, but this could get too complicated and cumbersome for this here?

src/mrpro/algorithms/reconstruction/TotalVariationRegularizedReconstruction.py
```diff
@@ -0,0 +1,127 @@
...
+    by using the PDHG-algorithm. :math:`A` is the acquisition model (coil sensitivity maps, Fourier operator,
+    k-space sampling), :math:`y` is the acquired k-space data, :math:`L` is the strength of the regularization and
+    :math:`\nabla` is the finite difference operator applied to :math:`x`.
+    """
+
+    n_iterations: int
```
koflera: do we want to have n_iterations or max_iterations with a tolerance? Our PDHG allows for the second as well.

src/mrpro/algorithms/reconstruction/TotalVariationRegularizedReconstruction.py
```diff
@@ -0,0 +1,127 @@
...
+        csm
+            Sensitivity maps for coil combination. If None, no coil combination is carried out, i.e. images for each
+            coil are returned. If a callable is provided, coil images are reconstructed using the adjoint of the
+            FourierOperator (including density compensation) and then sensitivity maps are calculated using the
+            callable. For this, kdata needs also to be provided. For examples have a look at the CsmData class
+            e.g. from_idata_walsh or from_idata_inati.
+        noise
+            KNoise used for prewhitening. If None, no prewhitening is performed
+        dcf
+            K-space sampling density compensation. If None, set up based on kdata. The dcf is only used to calculate a
```
koflera: Do we want to restrict ourselves to this or also allow the problem min_x 1/2*||W^(1/2)(Ax-y)||_2^2?

src/mrpro/algorithms/reconstruction/TotalVariationRegularizedReconstruction.py
```diff
@@ -0,0 +1,127 @@
...
+
+        Returns
+        -------
+            the reconstruced image.
+        """
+        if self.noise is not None:
+            kdata = prewhiten_kspace(kdata, self.noise)
+
+        # Create the acquisition model A = F S if the CSM S is defined otherwise A = F with the Fourier operator F
+        acquisition_operator = self.fourier_op @ self.csm.as_operator() if self.csm is not None else self.fourier_op
```
koflera, line 132: potentially W^(1/2) F C?

src/mrpro/algorithms/reconstruction/TotalVariationDenoising.py
```diff
@@ -0,0 +1,100 @@
...
+from mrpro.operators.FiniteDifferenceOp import FiniteDifferenceOp
+from mrpro.operators.functionals import L1NormViewAsReal, L2NormSquared, ZeroFunctional
+from mrpro.operators.IdentityOp import IdentityOp
+
+
+class TotalVariationDenoising(torch.nn.Module):
+    r"""TV denoising.
+
+    This algorithm solves the problem :math:`min_x \frac{1}{2}||(x - y)||_2^2 + ||L\nabla x||_1`
+    by using the PDHG-algorithm. :math:`y` is the target image, :math:`L` is the strength of the regularization and
```
koflera: maybe better:
 :math:`y` is the given noisy image

src/mrpro/algorithms/reconstruction/TotalVariationDenoising.py
```diff
@@ -0,0 +1,100 @@
...
+    :math:`\nabla` is the finite difference operator applied to :math:`x`.
+    """
+
+    regularization_weight: torch.Tensor
+    """Strength of the regularization :math:`L`."""
+
+    initial_image: torch.Tensor | None
+    """Initial image for PDHG-algorithm"""
+
+    n_iterations: int
```
koflera: max_iterations or n_iterations?

src/mrpro/algorithms/reconstruction/TotalVariationDenoising.py
```diff
@@ -0,0 +1,100 @@
...
+        l2 = 0.5 * L2NormSquared(target=idata.data, divide_by_n=True)
+
+        # Finite difference operator and corresponding L1-norm
+        nabla_operator = [
+            (FiniteDifferenceOp(dim=(dim - len(self.regularization_weight),), mode='forward'),)
+            for dim, weight in enumerate(self.regularization_weight)
+            if weight != 0
+        ]
+        l1 = [
+            weight / len(nabla_operator) * L1NormViewAsReal(divide_by_n=True)
```
koflera: what is the `len(nabla_operator)` for? in the TV-reconstruction it was not there?

examples/tv_regularized_reconstruction.py
```diff
@@ -0,0 +1,164 @@
...
+img_tv = tv_reconstruction(kdata_dynamic)
+show_dynamic_images(img_tv.rss())
+
+# %% [markdown]
+# #### TV-Regularized Reconstruction using ADMM
+# In the above example we need to apply the acquisition operator during the PDHG iterations which is computationally
+# demanding and hence takes a long time. Another option is to use the Alternating Direction Method of Multipliers (ADMM)
+# which solves the general problem
+#
+# $ \min_x f(x) + g(z) \quad \text{subject to} \quad  Ax + Bz = c $
```
koflera: conflict of variables for Ax + Bz = c and y=Ax+noise... maybe call the latter y=Ex+noise?

examples/tv_regularized_reconstruction.py
```diff
@@ -0,0 +1,164 @@
...
+#
+# If we use $f(x) = \lambda \| \nabla x \|_1$, $g(z)= \frac{1}{2}||Az - y||_2^2$, $A = I$, $B= -I$ and $c = 0$
```
koflera: i think the more commonly used identification is actually the following:

`$f(x) = \lambda \| x \|_1$, $g(z)= \frac{1}{2}||Az - y||_2^2$, $A = \nabla$, $B= -I$ and $c = 0$.
`
This avoids the need for PDHG for denoising (instead, only soft-thresholding is needed, i.e. the prox of the l1-norm), and for the regularized iterative SENSE update, one needs the operator `A^H A + \lambda * \nabla^H \nabla` instead of `A^H A + \lambda * Id`.

examples/tv_regularized_reconstruction.py
```diff
@@ -0,0 +1,164 @@
...
+#
+# $u_{k+1} = u_k + x_{k+1} - z_{k+1}$
+#
+# The first step is TV-based denoising of $x$, the second step is a regularized iterative SENSE update of $z$ and the
+# final step updates the helper variable $u$.
+
+
+# %%
+data_weight = 0.5
+n_adam_iterations = 4
```
koflera: `n_admm_iterations `instead of `n_adam_iterations`

examples/tv_regularized_reconstruction.py
```diff
@@ -0,0 +1,164 @@
...
+#
+# If we use $f(x) = \lambda \| \nabla x \|_1$, $g(z)= \frac{1}{2}||Az - y||_2^2$, $A = I$, $B= -I$ and $c = 0$
+#
+# then we can define a scaled form of the ADMM algorithm which solves
+#
+# $ \mathcal{F}(x) = \frac{1}{2}||Ax - y||_2^2 + \lambda \| \nabla x \|_1 $
+#
+# by doing
+#
+# $x_{k+1} = argmin_x \lambda \| \nabla x \|_1 + \frac{\rho}{2}||x - z_k + u_k||_2^2$
```
koflera: \argmin

examples/tv_regularized_reconstruction.py
```diff
@@ -0,0 +1,164 @@
...
+
+# %% [markdown]
+# #### TV-Regularized Reconstruction using PDHG
+# Reconstruct images by solving
+#
+# $ \mathcal{F}(x) = \frac{1}{2}||Ax - y||_2^2 + \lambda \| \nabla x \|_1 $
+#
+# using PDHG.
+#
+# Because we have 2D dynamic images we can apply the TV-regularization along x,y and time.
```
koflera: suggestion:

Because we have 2D dynamic images, we can apply the TV-regularization along x,y, and time.

examples/tv_regularized_reconstruction.py
```diff
@@ -0,0 +1,164 @@
...
+# %%
+tv_reconstruction = TotalVariationRegularizedReconstruction(
+    kdata_dynamic, csm=csm, n_iterations=100, regularization_weight=(0.1, 0, 0, 0.1, 0.1)
+)
+img_tv = tv_reconstruction(kdata_dynamic)
+show_dynamic_images(img_tv.rss())
+
+# %% [markdown]
+# #### TV-Regularized Reconstruction using ADMM
+# In the above example we need to apply the acquisition operator during the PDHG iterations which is computationally
```
koflera: suggestion:

In the above example, PDHG repeatedly applies the acquisition operator and its adjoint during the iterations, which is computationally demanding and hence takes a long time. Another option is to use the Alternating Direction Method of Multipliers (ADMM) [S. Boyd et al, 2011], which solves the general problem

examples/tv_regularized_reconstruction.py
```diff
@@ -0,0 +1,164 @@
...
+# by doing
+#
+# $x_{k+1} = argmin_x \lambda \| \nabla x \|_1 + \frac{\rho}{2}||x - z_k + u_k||_2^2$
+#
+# $z_{k+1} = argmin_z \frac{1}{2}||Az - y||_2^2 + \frac{\rho}{2}||x_{k+1} - z + u_k||_2^2$
+#
+# $u_{k+1} = u_k + x_{k+1} - z_{k+1}$
+#
+# The first step is TV-based denoising of $x$, the second step is a regularized iterative SENSE update of $z$ and the
+# final step updates the helper variable $u$.
```
koflera: updates the dual variable $u$.

examples/tv_regularized_reconstruction.py
ckolbPTB: added

src/mrpro/algorithms/reconstruction/TotalVariationDenoising.py
```diff
@@ -0,0 +1,96 @@
...
+
+from mrpro.algorithms.optimizers.pdhg import pdhg
+from mrpro.data.IData import IData
+from mrpro.operators import LinearOperatorMatrix, ProximableFunctionalSeparableSum
+from mrpro.operators.FiniteDifferenceOp import FiniteDifferenceOp
+from mrpro.operators.functionals import L1NormViewAsReal, L2NormSquared, ZeroFunctional
+from mrpro.operators.IdentityOp import IdentityOp
+
+
+class TotalVariationDenoising(torch.nn.Module):
```
fzimmermann89: Is this a reconstruciton?

examples/scripts/tv_regularized_reconstruction.py
```diff
@@ -0,0 +1,248 @@
...
+# #### TV-regularized reconstruction using ADMM
+# In the above example, PDHG repeatedly applies the acquisition operator and its adjoint during the iterations, which
+# is computationally demanding and hence takes a long time. Another option is to use the Alternating Direction Method
+# of Multipliers (ADMM) [[S. Boyd et al, 2011](http://dx.doi.org/10.1561/2200000016)], which solves the general problem
+#
+# $ \min_x f(x) + g(z) \quad \text{subject to} \quad  Ax + Bz = c $
+#
+# If we use $f(x) = \lambda \| \nabla x \|_1$, $g(z)= \frac{1}{2}||Ez - y||_2^2$, $A = I$, $B= -I$ and $c = 0$
+#
+# then we can define a scaled form of the ADMM algorithm which solves
```
fzimmermann89: can we instead add an admm function?

fzimmermann89 (changes requested):
see comments

also, shouldthe TV reconstrution be joined with the pdgh notebook`
... at least both should reference each other.

src/mrpro/algorithms/reconstruction/TotalVariationRegularizedReconstruction.py
```diff
@@ -0,0 +1,132 @@
...
+from mrpro.operators.FiniteDifferenceOp import FiniteDifferenceOp
+from mrpro.operators.functionals import L1NormViewAsReal, L2NormSquared, ZeroFunctional
+from mrpro.operators.LinearOperator import LinearOperator
+from mrpro.utils import unsqueeze_right
+
+
+class TotalVariationRegularizedReconstruction(DirectReconstruction):
+    r"""TV-regularized reconstruction.
+
+    This algorithm solves the problem :math:`min_x \frac{1}{2}||(Ax - y)||_2^2 + \sum_i l_i ||\nabla_i x||_1`
```
koflera:
```suggestion
    This algorithm solves the problem :math:`min_x \frac{1}{2}||Ax - y||_2^2 + \sum_i l_i ||\nabla_i x||_1`
```

src/mrpro/algorithms/reconstruction/TotalVariationRegularizedReconstruction.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def __init__(
+        self,
+        kdata: KData | None = None,
+        fourier_op: LinearOperator | None = None,
+        csm: Callable | CsmData | None = CsmData.from_idata_walsh,
+        noise: KNoise | None = None,
+        dcf: DcfData | None = None,
+        *,
+        max_iterations: int = 32,
```
koflera: from my experience, PDHG takes much longer to converge, see for example Figure 17 here https://arxiv.org/pdf/2301.05888. 
So the question is somewhat what one aims to do with the method. If one wants to obtain a properly TV-regularized reconstruction (and therefore accepting that there will be most likely  some staircasing artefacts) we should definetely increase max_iterations to say, 128 or so. If the idea is to only approximately solve the TV-problem, max_iterations can also be smaller, but maybe a bit larger than 32? maybe 48 or 64?

koflera: also, currently, the tolerance is set to 0, meaning that the number of iterations is the only stopping criterion for PDHG here. should this parameter then be called `n_iterations`?

koflera: also holds for TV-denoising, btw

src/mrpro/algorithms/TotalVariationDenoising.py
```diff
@@ -0,0 +1,97 @@
...
+
+from mrpro.algorithms.optimizers.pdhg import pdhg
+from mrpro.data.IData import IData
+from mrpro.operators import LinearOperatorMatrix, ProximableFunctionalSeparableSum
+from mrpro.operators.FiniteDifferenceOp import FiniteDifferenceOp
+from mrpro.operators.functionals import L1NormViewAsReal, L2NormSquared, ZeroFunctional
+from mrpro.operators.IdentityOp import IdentityOp
+
+
+class TotalVariationDenoising(torch.nn.Module):
```
koflera: why is this only a torch.nn.Module and not a Reconstruction?

src/mrpro/algorithms/TotalVariationDenoising.py
```diff
@@ -0,0 +1,97 @@
...
+    initial_image: torch.Tensor | None
+    """Initial image for PDHG-algorithm"""
+
+    max_iterations: int
+    """Maximum number of PDHG iterations."""
+
+    def __init__(
+        self,
+        regularization_weights: Sequence[float] | Sequence[torch.Tensor],
+        initial_image: torch.Tensor | None = None,
```
koflera: shouldn't the initial image also be a IData? Analogously as in TotalVariationReconstruction, where KData is used?

src/mrpro/algorithms/TotalVariationDenoising.py
koflera: if yes, then also move the file to "reconstruction" folder?

src/mrpro/algorithms/TotalVariationDenoising.py
koflera: also, shouldn't it actually correspond to IData? Intuitively speaking, I would say it makes more sense to have IData (i.e. the image data that we want to denoise) as parameter for the init and then to have an initial_image in the forward?

src/mrpro/algorithms/reconstruction/TotalVariationRegularizedReconstruction.py
```diff
@@ -0,0 +1,132 @@
...
+        nabla_operator = FiniteDifferenceOp(dim=finite_difference_dim, mode='forward')
+        l1 = L1NormViewAsReal(
+            weight=unsqueeze_right(self.regularization_weights[finite_difference_dim], kdata.data.ndim)
+        )
+
+        f = ProximableFunctionalSeparableSum(l2, l1)
+        g = ZeroFunctional()
+        operator = LinearOperatorMatrix(((acquisition_operator,), (nabla_operator,)))
+
+        # Initial value
```
fzimmermann89:
```suggestion
```

src/mrpro/algorithms/reconstruction/TotalVariationRegularizedReconstruction.py
```diff
@@ -0,0 +1,132 @@
...
+        if self.noise is not None:
+            kdata = prewhiten_kspace(kdata, self.noise)
+
+        # Create the acquisition model A = F S if the CSM S is defined otherwise A = F with the Fourier operator F
+        acquisition_operator = self.fourier_op @ self.csm.as_operator() if self.csm is not None else self.fourier_op
+
+        # L2-norm for the data consistency term
+        l2 = 0.5 * L2NormSquared(target=kdata.data)
+
+        # Finite difference operator and corresponding L1-norm
```
fzimmermann89:
```suggestion
        # TV regularization
```

src/mrpro/algorithms/reconstruction/TotalVariationRegularizedReconstruction.py
```diff
@@ -0,0 +1,132 @@
...
+        -------
+            the reconstruced image.
+        """
+        if self.noise is not None:
+            kdata = prewhiten_kspace(kdata, self.noise)
+
+        # Create the acquisition model A = F S if the CSM S is defined otherwise A = F with the Fourier operator F
+        acquisition_operator = self.fourier_op @ self.csm.as_operator() if self.csm is not None else self.fourier_op
+
+        # L2-norm for the data consistency term
```
fzimmermann89:
```suggestion
        # data consistency term
```

src/mrpro/algorithms/reconstruction/TotalVariationRegularizedReconstruction.py
```diff
@@ -0,0 +1,132 @@
...
+            k-space data to reconstruct.
+
+        Returns
+        -------
+            the reconstruced image.
+        """
+        if self.noise is not None:
+            kdata = prewhiten_kspace(kdata, self.noise)
+
+        # Create the acquisition model A = F S if the CSM S is defined otherwise A = F with the Fourier operator F
```
fzimmermann89:
```suggestion

```

koflera (changes requested):
A few comments

src/mrpro/algorithms/TotalVariationDenoising.py
fzimmermann89: It is not a reconstruction.

src/mrpro/algorithms/TotalVariationDenoising.py
fzimmermann89: Does it make sense to run the algortihm for different images using the same initial value?
Or should initial value be a keyword argument to forward?

src/mrpro/algorithms/TotalVariationDenoising.py
fzimmermann89: Maybe it should be a function?

fzimmermann89:
Maybe the denoising would be better as a function?

I don't see a reason to set it up with weights and initial image for one problem and run it on multiple images. Also, the setup should be fast anyways, or?

ckolbPTB:
I made tv_denoising into a function

koflera (commented):
looks good to me. except that I am not so super sure about the distinction between "reconstruction" and "algorithms"...

as far as I understand it, we have 
- optimizers (general algorithms, e.g. PDHG for min_x f(K(x)) + g(x) )
- reconstructions (eg. PDHG applied to MRI with the usual identification for f, g and K)
but a TV-reconstruction algorithm, which is currently to be found in "reconstruction" is also an algorithm, right?

what about instead of having "algorithms" having "post_processing" (i.e. methods that are merely applied to image tensors/IData) or, perhaps, given that denoising is the simplest inverse problem that can be considered and that gaussian denoisers could be applied within plug n play / regularization by denoising methods, having a "denoising" folder instead, where we could start to collect denoisers?

koflera (approved):

fzimmermann89:
> looks good to me. except that I am not so super sure about the distinction between "reconstruction" and "algorithms"...
> 
> as far as I understand it, we have
> 
> * optimizers (general algorithms, e.g. PDHG for min_x f(K(x)) + g(x) )
> * reconstructions (eg. PDHG applied to MRI with the usual identification for f, g and K)
>   but a TV-reconstruction algorithm, which is currently to be found in "reconstruction" is also an algorithm, right?
> 
> what about instead of having "algorithms" having "post_processing" (i.e. methods that are merely applied to image tensors/IData) or, perhaps, given that denoising is the simplest inverse problem that can be considered and that gaussian denoisers could be applied within plug n play / regularization by denoising methods, having a "denoising" folder instead, where we could start to collect denoisers?

if we ever get 3 more denoisers, one could create a folder. until we do, i woult keep the single denoiser in algorithms.
reconstructios are concrete commonly used named mri reconstruction techniques. I would rather remove some folder complexity instead of creating more. optimizers, reconstrucitons etc are currently within algorithms.

src/mrpro/algorithms/reconstruction/TotalVariationRegularizedReconstruction.py
```diff
@@ -0,0 +1,142 @@
...
+        self,
+        kdata: KData | None = None,
+        fourier_op: LinearOperator | None = None,
+        csm: Callable | CsmData | None = CsmData.from_idata_walsh,
+        noise: KNoise | None = None,
+        dcf: DcfData | None = None,
+        *,
+        max_iterations: int = 100,
+        tolerance: float = 0,
+        regularization_weights: Sequence[float] | Sequence[torch.Tensor],
```
fzimmermann89: should this also support single float?

src/mrpro/algorithms/reconstruction/TotalVariationRegularizedReconstruction.py
```diff
@@ -0,0 +1,142 @@
...
+    ) -> None:
+        """Initialize TotalVariationRegularizedReconstruction.
+
+        Parameters
+        ----------
+        kdata
+            KData. If kdata is provided and fourier_op or dcf are None, then fourier_op and dcf are estimated based on
+            kdata. Otherwise fourier_op and dcf are used as provided.
+        fourier_op
+            Instance of the FourierOperator used for reconstruction. If None, set up based on kdata.
```
fzimmermann89:
```suggestion
            Instance of the `FourierOperator `used for reconstruction. If `None`, set up based on `kdata`.
```

fzimmermann89: (I think this would be enough to make FourierOperator linked in docs, if not, we must use the full name)

fzimmermann89: could you also do this for the other references?

src/mrpro/algorithms/reconstruction/TotalVariationRegularizedReconstruction.py
```diff
@@ -0,0 +1,142 @@
...
+        -------
+            the reconstruced image.
+        """
+        if self.noise is not None:
+            kdata = prewhiten_kspace(kdata, self.noise)
+
+        acquisition_operator = self.fourier_op @ self.csm.as_operator() if self.csm is not None else self.fourier_op
+
+        # data consistency term
+        l2 = 0.5 * L2NormSquared(target=kdata.data)
```
fzimmermann89:
```suggestion
        data_consistency = 0.5 * L2NormSquared(target=kdata.data)
```

fzimmermann89: Also, i prefer to drop the 1/2

src/mrpro/algorithms/reconstruction/TotalVariationRegularizedReconstruction.py
```diff
@@ -0,0 +1,142 @@
...
+
+        # TV regularization
+        finite_difference_dim = [
+            dim - len(self.regularization_weights)
+            for dim, weight in enumerate(self.regularization_weights)
+            if weight != 0
+        ]
+        nabla_operator = FiniteDifferenceOp(dim=finite_difference_dim, mode='forward')
+        l1 = L1NormViewAsReal(
```
fzimmermann89:
```suggestion
        total_variation = L1NormViewAsReal(
```

src/mrpro/algorithms/reconstruction/TotalVariationRegularizedReconstruction.py
```diff
@@ -0,0 +1,142 @@
...
+            weight=unsqueeze_right(self.regularization_weights[finite_difference_dim], kdata.data.ndim)
+        )
+
+        f = ProximableFunctionalSeparableSum(l2, l1)
+        g = ZeroFunctional()
```
fzimmermann89:
```suggestion
```

src/mrpro/algorithms/reconstruction/TotalVariationRegularizedReconstruction.py
```diff
@@ -0,0 +1,142 @@
...
+        operator = LinearOperatorMatrix(((acquisition_operator,), (nabla_operator,)))
+
+        initial_value = acquisition_operator.H(
+            self.dcf.as_operator()(kdata.data)[0] if self.dcf is not None else kdata.data
+        )[0]
+
+        (img_tensor,) = pdhg(
+            f=f,
+            g=g,
```
fzimmermann89:
```suggestion
            f=ProximableFunctionalSeparableSum(data_consistancy, total_variation),
            g=None,
```

src/mrpro/algorithms/total_variation_denoising.py
```diff
@@ -0,0 +1,98 @@
...
+            Tolerance of PDHG for relative change of the primal solution; if zero, `max_iterations` of PDHG are run.
+
+    Returns
+    -------
+        the denoised image.
+    """
+    img_tensor = idata if isinstance(idata, torch.Tensor) else idata.data
+
+    # L2-norm for the data consistency term
+    l2 = 0.5 * L2NormSquared(target=img_tensor, divide_by_n=False)
```
fzimmermann89: again, instead of a comment, please name it data_consistency or dc

src/mrpro/algorithms/total_variation_denoising.py
```diff
@@ -0,0 +1,98 @@
...
+
+    f = ProximableFunctionalSeparableSum(l2, *l1)
+    g = ZeroFunctional()
+    operator = LinearOperatorMatrix(((IdentityOp(),), *nabla_operator))
+
+    initial_image = initial_image if initial_image is not None else img_tensor
+
+    (img_tensor,) = pdhg(
+        f=f,
+        g=g,
```
fzimmermann89: same as above

tests/algorithms/test_total_variation_denoising.py
```diff
@@ -0,0 +1,29 @@
+"""Tests for total variation denoising."""
+
+import torch
+from mrpro.algorithms.total_variation_denoising import total_variation_denoising
+from mrpro.data import IData, SpatialDimension
+from mrpro.utils import RandomGenerator
+from tests.helper import relative_image_difference
+
+
+def idata_single_coil(ph_ellipse, random_kheader):
```
fzimmermann89: local fixture

tests/algorithms/test_total_variation_denoising.py
```diff
@@ -0,0 +1,29 @@
...
+    """Create single-coil image."""
+    image_dimensions = SpatialDimension(z=1, y=ph_ellipse.n_y, x=ph_ellipse.n_x)
+    img = ph_ellipse.phantom.image_space(image_dimensions)
+    return IData.from_tensor_and_kheader(data=img, header=random_kheader)
+
+
+def test_denoising(ellipse_phantom, random_kheader):
```
fzimmermann89: paramerezize and type hints to enable mypy checks:
suggestion (untested) 

```python
@pytest.fixture()
def idata_single_coil(ph_ellipse, random_kheader)->None:
    """Create single-coil image."""
    image_dimensions = SpatialDimension(z=1, y=ph_ellipse.n_y, x=ph_ellipse.n_x)
    img = ph_ellipse.phantom.image_space(image_dimensions)
    return IData.from_tensor_and_kheader(data=img, header=random_kheader)

@pytest.mark.parametrize("tensor_input", [True, False], ids=["tensor", "idata"])
def test_denoising(idata_single_coil, tensor_input: bool) -> None:
    rng = RandomGenerator(seed=0)
    noisy = IData(idata_single_coil.data + rng.rand_like(idata.data), idata.header)
    if tensor_input:
      denoised = total_variation_denoising(noisy.data, regularization_weights=[1.0, 1.0])
    else:
       denoised = total_variation_denoising(noisy regularization_weights=[1.0, 1.0]).data,
    assert relative_image_difference(denoised, idata_single_coil.data) < relative_image_difference(
        noise.data, idata_single_coil
    )
```

src/mrpro/algorithms/reconstruction/TotalVariationRegularizedReconstruction.py
ckolbPTB: Single float would mean regularisation along last dimension?

fzimmermann89: i thought same lambda for the spatial dimensions as a reasonable shortcut for a common usecase, tbh

src/mrpro/algorithms/total_variation_denoising.py
```diff
@@ -0,0 +1,98 @@
...
+    regularization_weights: Sequence[float] | Sequence[torch.Tensor],
+    initial_image: torch.Tensor | None = None,
+    max_iterations: int = 100,
+    tolerance: float = 0,
+) -> torch.Tensor: ...
+
+
+def total_variation_denoising(
+    idata: IData | torch.Tensor,
+    regularization_weights: float | Sequence[float] | Sequence[torch.Tensor],
```
koflera: this is what I mean here.

I just find it a bit strange to be able to provide a single regularization weight and not being able to specify along which dimensions to apply the finite differences operator.
I think i would have had the `dim `of the finite differences operator as an additional argument of the algorithm and then
regularization_weight = either a float, list of floats or list of tensors
with the meaning that if it's only a single float/tensor, then the same for all the directions specified by `dim`, and if one provides a list of floats or tensors, the length of the list must coincide with the length of dim, otherwise we throw an error. 
Having the finite differences operator to be applied over the last dimension in case one provides a single float seems quite arbitrary to me and I would have avoided that.

But I see the point regarding the annoying checks one needs to do.

koflera: ... but I don't have strong opinion on the issue! ;)

ckolbPTB: If we add a `dim` parameter, we could either

1. only allow negative entries in `dim`, i.e. always count from the back
2. allow any entries in `dim` and provide information about the number of dimension of `x` such that we can check that the entries are unique and valid
3. allow any entries in `dim` and leave it up to the user that these entries make sense

Option 1 is most similar to how it is now.

src/mrpro/algorithms/total_variation_denoising.py
```diff
@@ -0,0 +1,98 @@
...
+        the denoised image.
+    """
+    regularization_weights_ = torch.as_tensor(regularization_weights)
+    img_tensor = idata if isinstance(idata, torch.Tensor) else idata.data
+
+    data_consistency = L2NormSquared(target=img_tensor)
+
+    # TV regularization
+    finite_difference_dim = [
+        dim - len(regularization_weights_) for dim, weight in enumerate(regularization_weights_) if weight != 0
```
koflera:
```suggestion
        dim - len(regularization_weights_) for dim, weight in enumerate(regularization_weights_) if (weight != 0).any()
```

to handle the case when regularization parameter maps are provided

src/mrpro/algorithms/total_variation_denoising.py
koflera: I vote for 1. or 3. 
2. is probably not nice given that this goes in the direction of already specifying the domain/range, which is something we decided to avoid.

src/mrpro/algorithms/total_variation_denoising.py
```diff
@@ -0,0 +1,98 @@
...
+    ]
+    nabla_operator = FiniteDifferenceOp(dim=finite_difference_dim, mode='forward')
+    total_variation = L1NormViewAsReal(
```
koflera: because the identification is the following

identification

I would suggest to rename

```suggestion
    l1_norm= L1NormViewAsReal(
```

src/mrpro/algorithms/total_variation_denoising.py
```diff
@@ -0,0 +1,98 @@
...
+            Tolerance of PDHG for relative change of the primal solution; if zero, `max_iterations` of PDHG are run.
+
+    Returns
+    -------
+        the denoised image.
+    """
+    regularization_weights_ = torch.as_tensor(regularization_weights)
+    img_tensor = idata if isinstance(idata, torch.Tensor) else idata.data
+
+    data_consistency = L2NormSquared(target=img_tensor)
```
koflera: also here, maybe better 

```suggestion
    l2_norm_squared= L2NormSquared(target=img_tensor)
```

fzimmermann89 (approved):

fzimmermann89 (approved):

PR #589, opened 2024-12-18
Make tests faster

ckolbPTB:
Closes #564 

I went through some of the tests and reduced matrix sizes to speed up the tests. 

On my MacBook Pro I got similar timing results than @fzimmermann89 reported here: #564 and I got a total run time for most of the tests (not all tests run on my laptop) of 4.5 mins. With these changes this was reduced to 2 mins. 

In the github action it seems to reduce the time from ~2min to ~1min.

@koflera Could you please have a look at the tests for the optimizers: https://github.com/PTB-MR/mrpro/blob/9dc116731154cade0029c8c58ee97013918d1a98/tests/algorithms/test_optimizers.py#L14 
This currently takes quite some time. Maybe we can speed it up by starting closer to the solution? You can time it yourself by e.g. running `pytest tests/algorithms/test_optimizers.py --durations=10`

PR #599, opened 2025-01-07
Release v0.250107

fzimmermann89:
Some of the highlights:

- Move KData functions from mixins back into KData (#559 
- Example for Basics and Cartesian reconstruction (#562)
- Make tests faster (#589)


schuenke (approved):
👍

PR #604, opened 2025-01-13
Fix WaveletOp

fzimmermann89:
Should fix https://github.com/PTB-MR/mrpro/pull/293#issuecomment-2587655604

@guastinimara


guastinimara (approved):
perfect, I checked it and now this fixed the problem I had, and I have no further questions about it

PR #610, opened 2025-01-16
Fix linear operator in regularized iterative SENSE

ckolbPTB:
For the regularized iterative SENSE reconstruction we allow the following to be minimised:
`1/2*||Ex - y||2^2 + reg_weight/2 * ||B x - x_reg||_2^2 `
with `B` a `LinearOperator`.

If we calculate the derivative and rearrange we get
`(E^H E + reg_weight* B^H B) x = E^H y + B^H x_reg`
(c) Andreas Kofler

In our current implementation we were missing the `B^H` part.

### DIFF
--- a/src/mrpro/algorithms/reconstruction/RegularizedIterativeSENSEReconstruction.py
+++ b/src/mrpro/algorithms/reconstruction/RegularizedIterativeSENSEReconstruction.py
@@ -128,7 +128,7 @@ class RegularizedIterativeSENSEReconstruction(DirectReconstruction):
             operator = operator + IdentityOp() @ (
                 self.regularization_weight * self.regularization_op.H @ self.regularization_op
             )
-            right_hand_side += self.regularization_weight * self.regularization_op.H(self.regularization_data)
+            right_hand_side += self.regularization_weight * self.regularization_op.H(self.regularization_data)[0]
 
         img_tensor = cg(
             operator,
--- a/src/mrpro/algorithms/reconstruction/RegularizedIterativeSENSEReconstruction.py
+++ b/src/mrpro/algorithms/reconstruction/RegularizedIterativeSENSEReconstruction.py
@@ -125,9 +125,7 @@ class RegularizedIterativeSENSEReconstruction(DirectReconstruction):
 
         # Add regularization
         if not torch.all(self.regularization_weight == 0):
-            operator = operator + IdentityOp() @ (
-                self.regularization_weight * self.regularization_op.H @ self.regularization_op
-            )
+            operator = operator + self.regularization_weight * self.regularization_op.H @ self.regularization_op
             right_hand_side += self.regularization_weight * self.regularization_op.H(self.regularization_data)[0]
 
         img_tensor = cg(
--- a/src/mrpro/algorithms/reconstruction/RegularizedIterativeSENSEReconstruction.py
+++ b/src/mrpro/algorithms/reconstruction/RegularizedIterativeSENSEReconstruction.py
@@ -37,7 +37,7 @@ class RegularizedIterativeSENSEReconstruction(DirectReconstruction):
     """Regularization data (i.e. prior) :math:`x_0`."""
 
     regularization_weight: torch.Tensor
-    """Strength of the regularization :math:`\lambda`."""
+    r"""Strength of the regularization :math:`\lambda`."""
 
     regularization_op: LinearOperator
     """Linear operator :math:`B` applied to the current estimate in the regularization term."""
@@ -55,7 +55,7 @@ class RegularizedIterativeSENSEReconstruction(DirectReconstruction):
         regularization_weight: float | torch.Tensor,
         regularization_op: LinearOperator | None = None,
     ) -> None:
-        """Initialize RegularizedIterativeSENSEReconstruction.
+        r"""Initialize RegularizedIterativeSENSEReconstruction.
 
         For a unregularized version of the iterative SENSE algorithm the regularization_weight can be set to 0 or
         IterativeSENSEReconstruction algorithm can be used.
--- a/src/mrpro/algorithms/reconstruction/RegularizedIterativeSENSEReconstruction.py
+++ b/src/mrpro/algorithms/reconstruction/RegularizedIterativeSENSEReconstruction.py
@@ -1,4 +1,4 @@
-"""Regularized Iterative SENSE Reconstruction by adjoint Fourier transform."""
+"""Regularized Iterative SENSE Reconstruction."""
 
 from __future__ import annotations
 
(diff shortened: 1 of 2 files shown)


src/mrpro/algorithms/reconstruction/RegularizedIterativeSENSEReconstruction.py
```diff
@@ -125,8 +125,10 @@ def forward(self, kdata: KData) -> IData:
 
         # Add regularization
         if not torch.all(self.regularization_weight == 0):
-            operator = operator + IdentityOp() @ (self.regularization_weight * self.regularization_op)
-            right_hand_side += self.regularization_weight * self.regularization_data
+            operator = operator + IdentityOp() @ (
+                self.regularization_weight * self.regularization_op.H @ self.regularization_op
+            )
```
koflera:
```suggestion
            operator = operator + self.regularization_weight * self.regularization_op.H @ self.regularization_op
```

koflera: i think the identity operator is not needed, is it?

koflera (changes requested):

koflera:
a few other comments:
- in the title https://github.com/PTB-MR/mrpro/blob/4aa5d59c16447a497a1f63b10603571df34778df/src/mrpro/algorithms/reconstruction/RegularizedIterativeSENSEReconstruction.py#L1, the "adjoint Fourier Transform" seems unnecessary?
- in the docstring https://github.com/PTB-MR/mrpro/blob/4aa5d59c16447a497a1f63b10603571df34778df/src/mrpro/algorithms/reconstruction/RegularizedIterativeSENSEReconstruction.py#L27 it should be :math:`H x = b` with :math:`H = A^H W A + L B^H B` and :math:`b = A^H W y + L B^H x_0`. 
- I would also maybe change the capital L to l, to better distinguish scalars from linear operators. currently, the code would not make sense anyway for an entire regularization parameter map. Also, if with capital L it was implicitly intended to be able to have an entire reg parameter map, the docstring is wrong, cause it then would need to be  :math:`min_x \frac{1}{2}||W^\frac{1}{2} (Ax - y)||_2^2 + \frac{1}{2}||L(Bx - x_0)||_2^2`, for which then the solution of the problem is however slightly different than what is currently implemented.
- maybe rename `precondition_op` to `dcf_op` in https://github.com/PTB-MR/mrpro/blob/4aa5d59c16447a497a1f63b10603571df34778df/src/mrpro/algorithms/reconstruction/RegularizedIterativeSENSEReconstruction.py#L120?

ckolbPTB:
> I would also maybe change the capital L to l, to better distinguish scalars from linear operators

@koflera should the regularization strength then actually always be a `float` and not `float | torch.Tensor`?

fzimmermann89:
please keep it as tensor or float.

koflera:
> > I would also maybe change the capital L to l, to better distinguish scalars from linear operators
> 
> @koflera should the regularization strength then actually always be a `float` and not `float | torch.Tensor`?

I think it doesn't hurt to keep the option to have both. If the intention was to make it possible to have an entire Lambda-Map, I would say we can still update the code later on.

koflera (approved):

src/mrpro/algorithms/reconstruction/RegularizedIterativeSENSEReconstruction.py
```diff
@@ -22,11 +22,11 @@ class RegularizedIterativeSENSEReconstruction(DirectReconstruction):
     r"""Regularized iterative SENSE reconstruction.
 
     This algorithm solves the problem :math:`min_x \frac{1}{2}||W^\frac{1}{2} (Ax - y)||_2^2 +
-    \frac{1}{2}L||Bx - x_0||_2^2`
+    \frac{1}{2}l||Bx - x_0||_2^2`
```
koflera:
```suggestion
    \frac{l}{2}||Bx - x_0||_2^2`
```

koflera (approved):
very last comment :)

src/mrpro/algorithms/reconstruction/RegularizedIterativeSENSEReconstruction.py
koflera: or maybe, since `l` and `1` are quite indistinguishable, rename it to `lam `or so?

fzimmermann89:
Why not \lambda?

It is a latex equation.
The parameter is named differently anyways..?

src/mrpro/algorithms/reconstruction/RegularizedIterativeSENSEReconstruction.py
ckolbPTB:
```suggestion
    \frac{1}{2}\lambda||Bx - x_0||_2^2`
```

src/mrpro/algorithms/reconstruction/RegularizedIterativeSENSEReconstruction.py
```diff
@@ -22,11 +22,11 @@ class RegularizedIterativeSENSEReconstruction(DirectReconstruction):
...
     by using a conjugate gradient algorithm to solve
-    :math:`H x = b` with :math:`H = A^H W A + L B` and :math:`b = A^H W y + L x_0` where :math:`A`
+    :math:`H x = b` with :math:`H = A^H W A + l B^H B` and :math:`b = A^H W y + l B^H x_0` where :math:`A`
```
ckolbPTB:
```suggestion
    :math:`H x = b` with :math:`H = A^H W A + \lambda B^H B` and :math:`b = A^H W y + \lambda B^H x_0` where :math:`A`
```

src/mrpro/algorithms/reconstruction/RegularizedIterativeSENSEReconstruction.py
```diff
@@ -22,11 +22,11 @@ class RegularizedIterativeSENSEReconstruction(DirectReconstruction):
...
     is the acquisition model (coil sensitivity maps, Fourier operator, k-space sampling), :math:`y` is the acquired
-    k-space data, :math:`W` describes the density compensation, :math:`L` is the strength of the regularization and
+    k-space data, :math:`W` describes the density compensation, :math:`l` is the strength of the regularization and
```
ckolbPTB:
```suggestion
    k-space data, :math:`W` describes the density compensation, :math:`\lambda` is the strength of the regularization and
```

src/mrpro/algorithms/reconstruction/RegularizedIterativeSENSEReconstruction.py
```diff
@@ -37,7 +37,7 @@ class RegularizedIterativeSENSEReconstruction(DirectReconstruction):
     """Regularization data (i.e. prior) :math:`x_0`."""
 
     regularization_weight: torch.Tensor
-    """Strength of the regularization :math:`L`."""
+    """Strength of the regularization :math:`l`."""
```
ckolbPTB:
```suggestion
    """Strength of the regularization :math:`\lambda`."""
```

src/mrpro/algorithms/reconstruction/RegularizedIterativeSENSEReconstruction.py
```diff
@@ -82,7 +82,7 @@ def __init__(
         regularization_data
             Regularization data, e.g. a reference image (:math:`x_0`).
         regularization_weight
-            Strength of the regularization (:math:`L`).
+            Strength of the regularization (:math:`l`).
```
ckolbPTB:
```suggestion
            Strength of the regularization (:math:`\lambda`).
```

PR #612, opened 2025-01-17
Pdhg format cells

ckolbPTB:


ckolbPTB:
No idea how this is supposed to work

fzimmermann89:
Ah, I haven't tried out the pre-commit bot for autofixes.

We might need another nbstrip action, i.e.
Strip->update->strip.

Otherwise, if the .py is changed, sometimes you currently need to run pre-commit twice to fix everything...

PR #617, opened 2025-01-20
Add T2* as outlook

ckolbPTB:
I had a look at the QMRI Challenge ISMRM 2024 T2* notebook and the results are still the same as what we had before. The only problem was the vmax setting in the final plot was too high. So everything is still working as expected.

My suggestion would be to add the T2* part as a "do-it-yourself" part at the end - would be very useful for training courses and practical sessions.


fzimmermann89 (approved):

fzimmermann89:
Good solution!

PR #621, opened 2025-01-25
Remove unused WASABI(TI) parameter

ckolbPTB:
Seems `larmor_frequency` is not used in the signal model or am I missing something?

### DIFF
--- a/src/mrpro/operators/models/WASABI.py
+++ b/src/mrpro/operators/models/WASABI.py
@@ -14,8 +14,8 @@ class WASABI(SignalModel[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]
         self,
         offsets: torch.Tensor,
         rf_duration: float | torch.Tensor = 0.005,
-        b1_nominal: float | torch.Tensor = 3.70,
-        gamma: float | torch.Tensor = 42.5764,
+        b1_nominal: float | torch.Tensor = 3.70e-6,
+        gamma: float | torch.Tensor = 42.5764e6,
     ) -> None:
         """Initialize WASABI signal model for mapping of B0 and B1 [SCHU2016]_.
 
@@ -27,9 +27,9 @@ class WASABI(SignalModel[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]
         rf_duration
             RF pulse duration [s]
         b1_nominal
-            nominal B1 amplitude [µT]
+            nominal B1 amplitude [T]
         gamma
-            gyromagnetic ratio [MHz/T]
+            gyromagnetic ratio [Hz/T]
 
         References
         ----------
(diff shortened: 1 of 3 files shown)


fzimmermann89:
@schuenke This was required when b0 was and offsets where ppm and no longer needed?

 Can you check if the units/equation as is are correct?

fzimmermann89:
Should gamma really be a parameter?
if yes. should it default to the water proton gamma in unit_conversion? 
I would say No, and make it fixed to that constant.
Or does anybody use wasabi with other nuclei -- seems rather unlikely.

SI units would be s, Hz and Tesla, instead of uT for b1.

fzimmermann89 (approved):

schuenke:
> @schuenke This was required when b0 was and offsets where ppm and no longer needed?
> 
> Can you check if the units/equation as is are correct?

Yeah, larmor freq and b0 were needed when offsets were given in ppm instead of Hz. Can be removed now.

schuenke:
> Should gamma really be a parameter? if yes. should it default to the water proton gamma in unit_conversion? I would say No, and make it fixed to that constant. Or does anybody use wasabi with other nuclei -- seems rather unlikely.
> 
> SI units would be s, Hz and Tesla, instead of uT for b1.

In principle you could use WASABI(TI) for other nuclei. So keeping it a parameter with the proton value as default makes sense IMO. We could default to the value in unit_conversion, but the given value is the one used in the [BMSim Challenge](https://github.com/pulseq-cest/BMsim_challenge) and therefore the one that is probably the most established one in the CEST community. 

The explanation for the value is given in the BMSim Readme:

> The [NIST value of the shielded proton gyromagnetic ratio](https://physics.nist.gov/cgi-bin/cuu/Value?gammapp) is 2.675153151 x 108 s-1 T-1. Dividing this value by 2 Pi yields 42.576384750950949004433240733872 MHz/T, which results in the used value of 42.5764 MHz/T when rounded to 4 digits.

fzimmermann89:
I would make it a float value that is not saved as a Parameter, but just as
an attribute.

And set it the the value in unit_conversion. We can change the value if we
want to. Why is it  42.58 * 1e6 and not  42.5764 * 1e6?
It seem quite strange to have two different values for gamma within mrpro.

PS: would it even matter for wasabi? The difference is 0.008%. Most likely
the error by single precision calculation or uncertainty in b1/data is
higher... right?


Am Do., 30. Jan. 2025 um 11:30 Uhr schrieb Patrick Schuenke <
***@***.***>:

schuenke (approved):
maybe @fzimmermann89 wants to double check that my changes made in https://github.com/PTB-MR/mrpro/pull/621/commits/de18b595068821726196f61d908eb1a740292f80 are what he meant?!

fzimmermann89:
lgtm

PR #591, opened 2024-12-23
IHeader

fzimmermann89:
Changes to IHeader

PR #626, opened 2025-01-29
Support for multi-frame dicom

ckolbPTB:
On our scanner we are exporting data as multi-frame, i.e. a singel dicom file can hold a 3D dataset or all phases of a cardiac cine scan. So far this was not supported. I also fixed a few other issues, such as wrong ordering of SpatialDimensions for the FOV.

Multi-frame data has the advantage that a 3D volume can be easily recognised. For single-frame data (i.e. each file is a 2D slice) this is much more complex because the orientation and position of each slices needs to be verified. IMO supporting this would require far too much work and so currently z-direction of the volume ends up in other for single-frame data.

Once #591 is done we will have to adapt the IHeader part again.


fzimmermann89:
I suggest either 
a) holding this until #591 is done, or 
b) merging it as-is and taking care of any improvements and details when adjusting in #591 (assuming there are no glaring issues and it at least kinda-works. 

I'd lean towards b)
I don't think it would be wise to spend to much time now on details. So maybe @schuenke could focus on the test code

tests/data/_DicomTestImage.py
```diff
@@ -0,0 +1,189 @@
+"""Create 2D dicom image datasets for testing."""
```
schuenke: adj to "2D or 3D"

tests/data/_DicomTestImage.py
```diff
@@ -0,0 +1,189 @@
...
+        Orientation of slice. If None assume transversal orientation.
+    slice_offset
+        Slice offset from isocentre along slice_orientation. The number of slices is determined by the length of this
+        parameter. If the length is greater than 1, the dicom will be saved as a 3D volume.
+    te
+        echo time
+    time_after_rpeak
+        time after R-peak
+    phantom
+        phantom with different ellipses
```
schuenke: move to doc of init?!

tests/data/_DicomTestImage.py
```diff
@@ -0,0 +1,189 @@
...
+        te: float = 0.037,
+        time_after_rpeak: float = 0.333,
+        phantom: EllipsePhantom | None = None,
+    ):
+        if not phantom:
+            phantom = EllipsePhantom()
+
+        self.filename: str | Path = filename
+        self.matrix_size_y: int = matrix_size_y
+        self.matrix_size_x: int = matrix_size_x
```
schuenke: type hints are not needed, are they?

tests/data/_DicomTestImage.py
```diff
@@ -0,0 +1,189 @@
...
+                SpatialDimension(0, 0, 1),
+            ]
+
+        self.slice_orientation: Sequence[SpatialDimension] = (
+            transversal_orientation if slice_orientation is None else slice_orientation.as_directions()
+        )
+        self.slice_offset: torch.Tensor = torch.atleast_1d(torch.as_tensor(slice_offset))
+        self.te: float = te
+        self.time_after_rpeak: float = time_after_rpeak
+        self.phantom: EllipsePhantom = phantom
```
schuenke: type hints are not needed, are they?

tests/data/_DicomTestImage.py
```diff
@@ -0,0 +1,189 @@
...
+
+        elem = pydicom.DataElement(0x00191015, 'FD', [1.0, 2.0, 3.0])
+        dataset.add(elem)
+
+        dataset.BitsAllocated = 16
+        dataset.PixelRepresentation = 0  # uint
+        dataset.SamplesPerPixel = 1
+        dataset.PhotometricInterpretation = 'MONOCHROME2'
+        dataset.BitsStored = 16
+
```
schuenke, line 125: nslices -> n_slices

schuenke (changes requested):
The different 3d_dicom fixtures make sense to me, but there are no tests using them right now, no? Should we add some tests in this PR?

And a few minor things... see my comments

tests/data/test_idata.py
```diff
@@ -51,18 +49,25 @@ def test_IData_from_empty_dcm_file_list():
...
+    original_echo_times = torch.as_tensor([ds.te for ds in dcm_2d_multi_echo_times_multi_folders])
     assert idata.header.te is not None
-    # dicom expects echo times in ms, mrpro in s
-    assert torch.allclose(torch.sort(original_echo_times)[0] / 1000, torch.sort(idata.header.te)[0])
+    assert torch.allclose(torch.sort(original_echo_times)[0], torch.sort(idata.header.te)[0])
     # Verify all images were read in
     assert idata.data.shape[0] == len(original_echo_times)
 
 
+def test_IData_from_3d(dcm_3d):
```
fzimmermann89: @schuenke

ckolbPTB: @schuenke I added some more tests (and fixed a bug)

ckolbPTB:
QData/QHeader will need more work but this should happen after #591

schuenke (approved):
just had a quick look cause I am on vacation today, but looks goot to me now 👍

src/mrpro/data/IData.py
```diff
@@ -93,19 +92,28 @@ def from_single_dicom(cls, filename: str | Path) -> Self:
...
+        return cls.from_dicom_files(filenames=(Path(filename),))
 
     @classmethod
     def from_dicom_files(cls, filenames: Sequence[str] | Sequence[Path] | Generator[Path, None, None]) -> Self:
         """Read multiple DICOM files and return IData object.
 
+        DICOM images can be saved as single-frame or multi-frame images [DCMMF]_.
+
+        If the DICOM files are single-frame, we treat each file separately and stack them along the `other` dimension.
+        If the DICOM files are multi-frame and the MRAcquisitionType is '3D' we treat the frame dimension as the `z`
```
ckolbPTB:
```suggestion
        If the DICOM files are multi-frame and the MRAcquisitionType is ``3D`` we treat the frame dimension as the `z`
```

src/mrpro/data/IData.py
```diff
@@ -93,19 +92,28 @@ def from_single_dicom(cls, filename: str | Path) -> Self:
...
+        dimension. Otherwise, we move the frame dimension to the `other` dimension. Multiple multi-frame DICOM
+        images are stacked along an additional `other` dimension before the frame dimension.
+        Providing the list of files sortey by filename usually leads to a reasonable sorting of the data.
```
ckolbPTB:
```suggestion
        Providing the list of files sorted by filename usually leads to a reasonable sorting of the data.
```

PR #629, opened 2025-02-01
Faster DCF based on pipe

### LINKED ISSUE #82: Faster density compensation calulation
We could implement a DCF function that wraps 
https://github.com/mmuckley/torchkbnufft/blob/main/torchkbnufft/_nufft/dcomp.py
to implement the DCF calculation according to [this paper by Pipe](https://doi.org/10.1002/(sici)1522-2594(199901)41:1%3C179::aid-mrm25%3E3.0.co;2-v)

fzimmermann89:
Not quite pipe/jackson, but would work with any nufft implementation (finuft does not have separate spread kernel)

for a quick draft, I just added it to the direct recon example.

@schuenke, @JoHa0811, @ckolbPTB 
do you have any spiral data you could easily test this on and post a result?

closes #82


Stef-Martin:
image
image

For my data, it looks like this - way way faster but looks worse

fzimmermann89:
This is with single iteration?
Does increasing the iterations help?

Stef-Martin:
whoops sry didn't even see that default iter of the estimate was set to 0 - just ran the script real quick. 
So the one shown above was Jackson, so 0 iter. With pipe it is much better with 1 iteration:
image
By increasing the iterations, it gets worse, this is with 3 iter:
image

fzimmermann89:
Mhm. Maybe there is something wrong with the iterations then.

The one iteration case looks good though

fzimmermann89:
Let's have a look tomorrow.
Maybe compare the dcfs calculated by 1 and 3 iterations -- are there some points that get zero/really high?
just a    
``` 
plt.matshow(dcf_operator2.matrix[0,0],vmax=2) # change vmax
plt.colorbar()
````
might help to identify issues.

Can you share an example file?

fzimmermann89:
... and let's check if the images are good enough to create CSMs from them.

Stef-Martin:
Update: without .abs() in `return weight.abs().squeeze(-4)` it works

fzimmermann89:
finufft now supports gridding only on cpu and gpu. so we can make it more similair to tochkbnufft version

PR #630, opened 2025-02-03
Add Brainweb phantom

fzimmermann89:
Adds a Brainweb Dataset as a phantom.

It can either be used as 20 subjects, 3D 1mm isotropic.
Or as augmented 2D slices.

The phantom provides mask, m0, t1/r1, t2/r2 and tissue class.
The values per class are randomized.

- [x] download
- [x] slices dataset
- [x] volume dateset
- [x] download test
- [x] slices tests
- [x] volume tests


.github/workflows/docs.yml
```diff
@@ -51,7 +51,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Install mrpro and dependencies
-        run: pip install --upgrade --upgrade-strategy "eager" -e .[notebook]
+        run: pip install --index-url=https://download.pytorch.org/whl/cpu --extra-index-url https://pypi.org/simple/ --upgrade --upgrade-strategy "eager" -e .[notebook]
```
schuenke, line 54: Is this related to the brainweb phantom? IMO this should be in a separate PR, no? Same in pytest.yml.

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,577 @@
...
+    '&zip_value=gnuzip'
+    '&download_for_real=%5BStart+download%21%5D'
+)
+
+
+CLASSES = ('bck', 'skl', 'gry', 'wht', 'csf', 'mrw', 'dura', 'fat', 'fat2', 'mus', 'm-s', 'ves')  # noqa: typos
+VERSION = 1
+CACHE_DIR = platformdirs.user_cache_dir('mrpro')
+K = TypeVar('K')
+TClassNames = Literal['skl', 'gry', 'wht', 'csf', 'mrw', 'dura', 'fat', 'fat2', 'mus', 'm-s', 'ves']  # noqa: typos
```
schuenke, line 37:
````suggestion
TClassNames = Literal[*CLASSES[1:]]

File: src/mrpro/phantoms/brainweb.py [hunk 3]
```diff
@@ -0,0 +1,577 @@
...
+    'do_download_alias=subject{subject}_{c}'
+    '&format_value=raw_short'
+    '&zip_value=gnuzip'
+    '&download_for_real=%5BStart+download%21%5D'
+)
+
+
+CLASSES = ('bck', 'skl', 'gry', 'wht', 'csf', 'mrw', 'dura', 'fat', 'fat2', 'mus', 'm-s', 'ves')  # noqa: typos
+VERSION = 1
+CACHE_DIR = platformdirs.user_cache_dir('mrpro')
```
schuenke: maybe add a short comment what this does

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,577 @@
...
+K = TypeVar('K')
+TClassNames = Literal['skl', 'gry', 'wht', 'csf', 'mrw', 'dura', 'fat', 'fat2', 'mus', 'm-s', 'ves']  # noqa: typos
+
+
+@dataclass
+class T1T2M0:
```
schuenke: hmmm... any better class name?

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,577 @@
...
+
+def affine_augment(data: torch.Tensor, size: int = 256, rng: torch.Generator | None = None) -> torch.Tensor:
+    """Apply random affine augmentation.
+
+    Parameters
+    ----------
+    data
+        2D data to augment.
+    size
+        resulting image will be size x size pixels.
```
schuenke:
```suggestion
        resulting image will be (size x size) pixels.

File: tests/phantoms/test_brainweb.py [hunk 6]
```diff
@@ -0,0 +1,209 @@
...
+            compressed_data = gzip.compress(fake_data.numpy().astype(np.uint16).tobytes(), compresslevel=0)
+            return type('MockResponse', (), {'content': compressed_data, 'raise_for_status': lambda _: None})()
+
+        raise ValueError(f'Unexpected URL: {url}')  # Ensure no unexpected requests are made
+
+    monkeypatch.setattr(requests, 'get', mock_get)
+
+
+def test_download_brainweb(tmp_path, mock_requests):
+    """Test download_brainweb"""
```
schuenke:
````suggestion
    """Test download_brainweb using mock data."""

File: .github/workflows/docs.yml (hunk 1 above)
Comment (fzimmermann89) [reply .github/workflows/docs.yml:54]: It is only required to install torch vision, which the dataset depends on.
If torch is installed from the CPU only repo, torch vision also needs to be installed there. So I would not split it out, but consider it as part of "adding torch vision dependency" ?

File: src/mrpro/phantoms/brainweb.py (hunk 2 above)
Comment (fzimmermann89) [reply src/mrpro/phantoms/brainweb.py:37]: Type(s/aliases) cannot depend on variables.

**Review by schuenke — APPROVED** [review]
`BrainwebTissue` >> `T1T2M0` IMO 👍 
I didn't try to use the code, but tests look reasonable to me and seem to cover all relevant functionality.

File: src/mrpro/phantoms/brainweb.py [hunk 7]
```diff
@@ -0,0 +1,582 @@
...
+    '&zip_value=gnuzip'
+    '&download_for_real=%5BStart+download%21%5D'
+)
+
+
+CLASSES = ('bck', 'skl', 'gry', 'wht', 'csf', 'mrw', 'dura', 'fat', 'fat2', 'mus', 'm-s', 'ves')  # noqa: typos
+VERSION = 1
+CACHE_DIR = platformdirs.user_cache_dir('mrpro')  # ~/.cache/mrpro on Linux, %AppData%\Local\mrpro on Windows
+K = TypeVar('K')
+TClassNames = Literal['skl', 'gry', 'wht', 'csf', 'mrw', 'dura', 'fat', 'fat2', 'mus', 'm-s', 'ves']  # noqa: typos
```
ckolbPTB, line 37: why not "bck"?

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,582 @@
...
+
+
+@dataclass
+class BrainwebTissue:
+    """Container for Parameters of a single tissue."""
+
+    t1: float | tuple[float, float]
```
ckolbPTB, line 49: Are these either the (fixed) value of T1 or the range from where T1 is drawn? Maybe add a bit more info in the doc string.

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,582 @@
...
+    t2: float | tuple[float, float]
+    m0_abs: float | tuple[float, float]
+    m0_phase: float | tuple[float, float] = 0.0
+
+    def sample_r1(self, rng: None | torch.Generator = None) -> torch.Tensor:
+        """Get randomized r1 value.
```
ckolbPTB: It is not really randomized, is it?

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,582 @@
...
+            magnitude = torch.tensor(self.m0_abs)
+        if isinstance(self.m0_phase, tuple):
+            phase = torch.empty(1).uniform_(*self.m0_phase, generator=rng)
+        else:
+            phase = torch.tensor(self.m0_phase)
+        return torch.polar(magnitude, phase)
+
+
+def affine_augment(data: torch.Tensor, size: int = 256, rng: torch.Generator | None = None) -> torch.Tensor:
+    """Apply random affine augmentation.
```
ckolbPTB: Similar to above I would add some more info on the augmentation. would be good if the range for the affine transformations were input parameters

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,582 @@
...
+        else:
+            magnitude = torch.tensor(self.m0_abs)
+        if isinstance(self.m0_phase, tuple):
+            phase = torch.empty(1).uniform_(*self.m0_phase, generator=rng)
+        else:
+            phase = torch.tensor(self.m0_phase)
+        return torch.polar(magnitude, phase)
+
+
+def affine_augment(data: torch.Tensor, size: int = 256, rng: torch.Generator | None = None) -> torch.Tensor:
```
ckolbPTB: should the default of size not be the shape of data?

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,582 @@
...
+    """Apply random affine augmentation.
+
+    Parameters
+    ----------
+    data
+        2D data to augment.
```
ckolbPTB: should add a check that it is 2D. will 3D lead to the first dimension be treated as batch?

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,582 @@
...
+    resized data
+    """
+    scale = size / max(data.shape[1:])
+    data = torchvision.transforms.functional.resize(data, [int(scale * data.shape[1]), int(scale * data.shape[2])])
+    data = torchvision.transforms.functional.center_crop(data, [size, size])
+    return data
+
+
+def trim_indices(mask: torch.Tensor) -> tuple[slice, slice]:
+    """Get slices that remove outer masked out values.
```
ckolbPTB:
```suggestion
    """Get slices that contain only data which is contained in mask.
```

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,582 @@
...
+URL_TEMPLATE = (
+    'http://brainweb.bic.mni.mcgill.ca/cgi/brainweb1?'
+    'do_download_alias=subject{subject}_{c}'
+    '&format_value=raw_short'
+    '&zip_value=gnuzip'
+    '&download_for_real=%5BStart+download%21%5D'
+)
+
+
+CLASSES = ('bck', 'skl', 'gry', 'wht', 'csf', 'mrw', 'dura', 'fat', 'fat2', 'mus', 'm-s', 'ves')  # noqa: typos
```
ckolbPTB: maybe add docstring explaining what the class strings mean

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,582 @@
...
+        'fat': BrainwebTissue((0.300, 0.500), (0.060, 0.100), (0.90, 1.00), (-0.2, 0.2)),
+        'fat2': BrainwebTissue((0.400, 0.600), (0.060, 0.100), (0.60, 0.90), (-0.2, 0.2)),
+        'mus': BrainwebTissue((1.200, 1.500), (0.040, 0.060), (0.90, 1.00), (-0.2, 0.2)),
+        'm-s': BrainwebTissue((0.500, 0.900), (0.300, 0.500), (0.90, 1.00), (-0.2, 0.2)),
+        'ves': BrainwebTissue((1.700, 2.100), (0.200, 0.400), (0.80, 1.00), (-0.2, 0.2)),
+    }
+)
+"""Tissue values at 3T."""
+
+DEFAULT_VALUES = {'r1': 0.0, 'm0': 0.0, 'r2': 0.0, 'mask': 0, 'tissueclass': -1}
```
ckolbPTB, line 244: DEFAULT_VALUES_MASKED ?

tests/phantoms/test_brainweb.py
```diff
@@ -0,0 +1,235 @@
...
+            f.attrs['classnames'] = [c for c in CLASSES if c != 'bck']  # noqa: typos
+            f.attrs['subject'] = subject
+            f.attrs['version'] = 1
+    return test_dir
+
+
+def test_brainwebvolumes_init(brainweb_test_data) -> None:
+    """Test BrainwebVolumes dataset initialization."""
+    dataset = BrainwebVolumes(folder=brainweb_test_data, what=('m0', 'r1'))
+    assert len(dataset) == 2  # Two subjects
```
ckolbPTB: Two subjects or two parameters?

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,582 @@
...
+        self, index: int
+    ) -> dict[Literal['r1', 'r2', 'm0', 't1', 't2', 'mask', 'tissueclass'] | TClassNames, torch.Tensor]:
+        """Get data from one subject."""
+        with h5py.File(self.files[index]) as file:
+            data = torch.as_tensor(np.array(file['classes'], dtype=np.uint8))
+            classnames = tuple(file.attrs['classnames'])
+        result: dict[Literal['r1', 'r2', 'm0', 't1', 't2', 'mask', 'tissueclass'] | TClassNames, torch.Tensor] = {}
+        for el in self.what:
+            if el == 'r1':
+                values = torch.stack([self.parameters[k].sample_r1() for k in classnames]) / 255
```
ckolbPTB, line 449: why "/ 255"?

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,582 @@
...
+                continue
+            if (mask_value := self.mask_values.get(key, None)) is not None:
+                value[~mask] = mask_value
+            elif key not in classnames:
+                value[~mask] = torch.nan
+
+        return result
+
+
+class BrainwebSlices(torch.utils.data.Dataset):
```
ckolbPTB, line 487: There seems to be a lot of code duplication with BrainwebVolumes. Do we really need both classes? a "return_slices" function for BrainwebVolumes might also do the job...

src/mrpro/phantoms/brainweb.py
fzimmermann89: all leading dimensiosn are batch

src/mrpro/phantoms/brainweb.py
fzimmermann89: ?What am I missing here?

src/mrpro/phantoms/brainweb.py
fzimmermann89: As the dataset is already processed multiple times, and will be randomly augmented, I don't think it matters to much to keep the original resolution. Instead, this default results in roughly equally scaled slices (due to the trimming and resizing), with a reasonable (and constant) output size for DL stuff
So this dataset is quite opinionated, yes, but it is documented and allows to quickly get something up and running.

src/mrpro/phantoms/brainweb.py
fzimmermann89: added to the class docstrings

tests/phantoms/test_brainweb.py
fzimmermann89: two subjects.

src/mrpro/phantoms/brainweb.py
fzimmermann89: No.

initially had a single dataset for both and "3D" only as a value for the slice direction, but the code got a lot more complicated.
a) to be used with a torch Dataloader, len and getitem have to be different for 2d and 3d. other functions of the datset do not matter.
b) there are some minor differences throughout the code (conv2d vs conv3d for mask dilation, where the scaling from uint8 to 0...1 happens -- saves memory in the dataloaders) etc.

src/mrpro/phantoms/brainweb.py
fzimmermann89: BCK is "no other class", i.e. background.
The only reason why it gets downloaded and used is, that the original data is differently scaled for the different subjects. 
In the final dataset, this would be "mask". We save a bit of memory in not supporting setting m0, t1, t2 for the background, but forcing m0 = 0.

src/mrpro/phantoms/brainweb.py
fzimmermann89: ?

fzimmermann89:
@ckolbPTB

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,673 @@
...
+    ) -> None:
+        """Initialize Dataset.
+
+        Parameters
+        ----------
+        folder:
+            The directory containing Brainweb HDF5 files
+        what
+            What to return for each subject.
+            Psossible values are:
```
fzimmermann89:
```suggestion
            Possible values are:
```

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,673 @@
...
+ALL_CLASSES = ('bck', 'skl', 'gry', 'wht', 'csf', 'mrw', 'dura', 'fat', 'fat2', 'mus', 'm-s', 'ves')  # noqa: typos
+VERSION = 1
+CACHE_DIR = platformdirs.user_cache_dir('mrpro')  #  ~/.cache/mrpro on Linux, %AppData%\Local\mrpro on Windows
+K = TypeVar('K')
+TClassNames = Literal['skl', 'gry', 'wht', 'csf', 'mrw', 'dura', 'fat', 'fat2', 'mus', 'm-s', 'ves']  # noqa: typos
+
+
+@dataclass
+class BrainwebTissue:
+    """Container for Parameters of a single tissue.
```
ckolbPTB:
```suggestion
    """Container for parameters of a single tissue.
```

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,673 @@
...
+
+    Attributes are either single values or ranges.
+    If ranges, the values are sampled uniformly from within this range by the `sample_r1`, `sample_r2`, and `sample_m0`
```
ckolbPTB:
```suggestion
    If ranges, the assigned values are sampled uniformly from within this range by the `sample_r1`, `sample_r2`, and `sample_m0`
```

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,673 @@
...
+    methods. If a single value is given, this value is returned by the respective method.
+    """
+
+    t1: float | tuple[float, float]
```
ckolbPTB, line 49: spaces between different attributes

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,673 @@
...
+    trim: bool = True,
+    max_random_shear: float = 5,
+    max_random_rotation: float = 10,
+    max_random_scaling_factor: float = 0.1,
+    p_horizontal_flip: float = 0.5,
+    p_vertical_flip: float = 0.5,
+) -> Callable[[torch.Tensor, torch.Generator | None], torch.Tensor]:
+    """Get an augmentation function.
+
+    Creates a function that applies augmentation to the input tensor, that is
```
ckolbPTB:
```suggestion
    Creates a function that applies augmentation to the input tensor consisting of rotation, shearing, scaling and horizontal/vertical flipping.
```

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,673 @@
...
+    a random rotation and shear to the input image.
```
ckolbPTB:
```suggestion
```

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,673 @@
...
+    max_random_scaling_factor
+        Strength of the scaling randomization (see above).
+    p_horizontal_flip
+        Probability of horizontal flip.
+    p_vertical_flip
+        Probability of vertical flip.
+
+    Returns
+    -------
+    Callable that applies augmentation to the input tensor
```
ckolbPTB: For Return statements the text should be indented - applies to all Return statements in this file.

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,673 @@
...
+    def augment_fn(data: torch.Tensor, rng: torch.Generator | None = None) -> torch.Tensor:
+        """Apply augmentation to the input tensor."""
+        rand = torch.empty(6).uniform_(-1, 1, generator=rng).tolist()
+
+        shear_x = rand[0] * max_random_shear
+        shear_y = rand[1] * max_random_shear
+        angle = rand[2] * max_random_rotation
+        scale = size / max(data.shape[-2:])
+        scale *= 1 + max_random_scaling_factor * rand[3]
+        translate = rand[4:6]  # single pixel translations
```
ckolbPTB: Why single pixel? Isn't the entire input tensor translated in the same way?

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,673 @@
...
+            angle=angle,
+            scale=scale,
+            shear=[shear_x, shear_y],
+            translate=translate,
+            interpolation=torchvision.transforms.InterpolationMode.BILINEAR,
+            fill=0.0,
+        )
+        rand = torch.empty(2).uniform_(0, 1, generator=rng).tolist()
+        data = torchvision.transforms.functional.hflip(data) if rand[0] < p_horizontal_flip else data
+        data = torchvision.transforms.functional.vflip(data) if rand[0] < p_vertical_flip else data
```
ckolbPTB:
```suggestion
        data = torchvision.transforms.functional.vflip(data) if rand[1] < p_vertical_flip else data
```
Otherwise hflip and vflip are correlated

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,673 @@
...
+    """Get slices that remove fully masked out outer rows and columns.
+
+    Parameters
+    ----------
+    mask
+        Mask indicating valid data.
+
+    Returns
+    -------
+    Two `slice` ohjects, that can be used to index the data
```
ckolbPTB:
```suggestion
    Two `slice` objects, that can be used to index the data
```

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,673 @@
...
+        'mrw': BrainwebTissue((0.400, 0.600), (0.060, 0.100), (0.70, 1.00), (-0.2, 0.2)),
+        'dura': BrainwebTissue((2.000, 2.800), (0.200, 0.500), (0.90, 1.00), (-0.2, 0.2)),
+        'fat': BrainwebTissue((0.300, 0.500), (0.060, 0.100), (0.90, 1.00), (-0.2, 0.2)),
+        'fat2': BrainwebTissue((0.400, 0.600), (0.060, 0.100), (0.60, 0.90), (-0.2, 0.2)),
+        'mus': BrainwebTissue((1.200, 1.500), (0.040, 0.060), (0.90, 1.00), (-0.2, 0.2)),
+        'm-s': BrainwebTissue((0.500, 0.900), (0.300, 0.500), (0.90, 1.00), (-0.2, 0.2)),
+        'ves': BrainwebTissue((1.700, 2.100), (0.200, 0.400), (0.80, 1.00), (-0.2, 0.2)),
+    }
+)
+"""Tissue values at 3T."""
```
ckolbPTB: do we have a reference for these values?

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,673 @@
...
+                        else None
+                    ):
+                        # file is already downloaded and up to date
+                        progressbar.update(len(ALL_CLASSES))
+                        continue
+            download_subject(subject, outfilename, workers, progressbar)
+
+
+class BrainwebVolumes(torch.utils.data.Dataset):
+    """3D Brainweb Dataset.
```
ckolbPTB:
```suggestion
    """Dataset of 3D qMRI parameters based on Brainweb dataset.
```

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,673 @@
...
+
+    This dataset provides 1 mm isotropic 3D brain data of various quantitative MRI (qMRI) parameters.
+
+    References
+    ----------
+    .. [AubertBroche2006] Aubert-Broche, B., Griffin, M., Pike, G.B., Evans, A.C., & Collins, D.L. (2006).
+       Twenty New Digital Brain Phantoms for Creation of Validation Image Data Bases.
+       *IEEE Transactions on Medical Imaging, 25*(11), 1410-1416.
```
ckolbPTB:
```suggestion
       IEEE Transactions on Medical Imaging, 25 (11), 1410-1416.
```

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,673 @@
...
+                        # file is already downloaded and up to date
+                        progressbar.update(len(ALL_CLASSES))
+                        continue
+            download_subject(subject, outfilename, workers, progressbar)
+
+
+class BrainwebVolumes(torch.utils.data.Dataset):
+    """3D Brainweb Dataset.
+
+    This dataset provides 1 mm isotropic 3D brain data of various quantitative MRI (qMRI) parameters.
```
ckolbPTB:
```suggestion
    This dataset provides 1 mm isotropic 3D brain data of various quantitative MRI (qMRI) parameters based on the segmentations provided by Brainweb [AubertBroche2006]_.
```

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,673 @@
...
+        what
+            What to return for each subject.
+            Possible values are:
+                - r1: R1 relaxation rate.
+                - r2: R2 relaxation rate.
+                - m0: M0 magnetization.
+                - t1: T1 relaxation time.
+                - t2: T2 relaxation time.
+                - mask: mask indicating valid data.
+                - tissueclass: (majority) Class index.
```
ckolbPTB:
```suggestion
                - tissueclass: (majority) class index.
```

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,673 @@
...
+                - Brainweb class name: raw percentage for a specific tissue class.
```
ckolbPTB: What does "raw percentage" mean regarding a class name?

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,673 @@
...
+            Parameters for each tissue class.
+            The Brainweb tissue classes are:
+                - skl: skull
+                - gry: gray matter
+                - wht: white matter
+                - csf: cerebrospinal fluid
+                - mrw: bone marrow
+                - dura: dura
+                - fat: fat
+                - fat2: fat and Tissue
```
ckolbPTB:
```suggestion
                - fat2: fat and tissue
```

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,673 @@
...
+                - mus: muscle
+                - m-s: skin
+                - ves: vessels
+        mask_values
+            Default values to use for masked out regions.
+        seed
+            Random seed. Can be an int, the strings ``index`` to use slice index as seed, or ``random`` for random seed.
```
ckolbPTB: does the option `index` makes sense here?

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,673 @@
...
+            if (mask_value := self.mask_values.get(key, None)) is not None:
+                value[~mask] = mask_value
+            elif key not in classnames:
+                value[~mask] = torch.nan
+
+        return result
+
+
+class BrainwebSlices(torch.utils.data.Dataset):
+    """Dataset of 2D qMRI parameter slices based on Brainweb dataset.
```
ckolbPTB, line 488:
```suggestion
    """Dataset of 2D qMRI parameter slices based on Brainweb dataset [AubertBroche2006]_.
```

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,673 @@
...
+
+    References
+    ----------
+    .. [AubertBroche2006] Aubert-Broche, B., Griffin, M., Pike, G.B., Evans, A.C., & Collins, D.L. (2006).
+       Twenty New Digital Brain Phantoms for Creation of Validation Image Data Bases.
+       *IEEE Transactions on Medical Imaging, 25*(11), 1410-1416.
```
ckolbPTB:
```suggestion
       IEEE Transactions on Medical Imaging, 25 (11), 1410-1416.
```

src/mrpro/phantoms/brainweb.py
```diff
@@ -0,0 +1,673 @@
...
+        self.mask_values = mask_values
+
+    def __len__(self) -> int:
+        """Get number of subjects."""
+        return len(self.files)
+
+    def __getitem__(
+        self, index: int
+    ) -> dict[Literal['r1', 'r2', 'm0', 't1', 't2', 'mask', 'tissueclass'] | TClassNames, torch.Tensor]:
+        """Get data from one subject."""
```
ckolbPTB:
```suggestion
        """Get 3D data from one subject."""
```

src/mrpro/phantoms/brainweb.py
fzimmermann89: subpixel.

src/mrpro/phantoms/brainweb.py
fzimmermann89: No. especially not for the ranges.. this is what i used for pinqi.

src/mrpro/phantoms/brainweb.py
fzimmermann89: For each voxel, the brainweb assignes a percentage for each tissue class.
Specifiny 'wht' would return this percentage for wht

src/mrpro/phantoms/brainweb.py
fzimmermann89: It allows to get 20 different 3D subjects, reproduicible.
I replaced 'slice' by subject

PR #632, opened 2025-02-03
Add ignore typos marker

fzimmermann89:
Typos sometimes marks code as having a typo.
Before, we only could add words to the exception list. But if these words are also common spelling errors (such has the abbreviated names of wavelets or of the brainweb classes..) we might not want to globally ignore them.
Typos does not have a "noqa" rule by default, but instead a regex rule system.
I added a rule that makes "#noqa: typos" at the end of a line ignore typos in the line.


schuenke (approved):
lgtm

PR #644, opened 2025-02-09
Disable typos autofix

fzimmermann89:
Typos autofix can silently introduce errors if "spelling mistakes" are intentional.

At least I run pre-commit before committ, and would otherwise have to go back through all code to see what typos might have changed.
This now only reports the mistakes and one has to manually fix them (or manually run `typos -w`)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,6 +25,7 @@ repos:
     rev: v1.29.4
     hooks:
       - id: typos
+        args: [] # empty, to remove write-changes from the default arguments.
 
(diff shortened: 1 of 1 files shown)


fzimmermann89:
recent example: the `wht` class in brainweb,

schuenke (approved):
Fine for me and works as expected. However, we might want to add "typos" as an optional dependency so that users are able to run the `typos -w` command without the need to manually install typos? 

In general, instroducing a `dev` category in the optional-dependencies might be a good idea? Should currently include `pre-commit` and `typos` probably. 

`pre-commit` is currently listed in the `test` category. Do we actually need it there?

PR #645, opened 2025-02-12
Set fixed version of torchkbnufft in notebooks

lrlunin:
`torchkbnufft>1.4.0` requires `numpy>=2.0.0` which updates the pre-installed `numpy` in colab and causes troubles with `h5py`.
You can check this out that presented version of notebooks indeed does not work in colab (unless the environment is restarted) and fails while the h5py tries to parse the dataset from zenodo. 

By the way this PR was made from my own fork and both PyTest and Docs comment reports were posted🥳


fzimmermann89:
If we merge #646 we dont need this here.
If we don't (yet) we do

PR #646, opened 2025-02-13
Switch to finufft

fzimmermann89:
@ckolbPTB


fzimmermann89:
For toeplitz, this could be optimized a bit further by creating a list of the kernel parts first and them summing and stacking them appropriately without ever creating the full sized kernel. This would solve quite a bit of memory issues if we ever differentiated this function.

src/mrpro/operators/NonUniformFastFourierOp.py
```diff
@@ -53,125 +51,104 @@ def __init__(
...
-        if not self._nufft_directions:
+        self._direction_zyx = tuple(direction_dict[d] for d in direction)
+        if len(direction) != len(set(self._direction_zyx)):
+            raise ValueError(f'Directions must be unique. Normalized directions are {self._direction_zyx}')
+        if not self._direction_zyx:
             return
-        nufft_traj = [
-            ks for ks, i in zip((traj.kz, traj.ky, traj.kx), (-3, -2, -1), strict=True) if i in self._nufft_directions
+        if len(self._direction_zyx) != 1 and len(self._direction_zyx) != 2 and len(self._direction_zyx) != 3:
+            raise ValueError('Only 0D, 1D, 2Dn or 3D NUFFT is supported')
```
ckolbPTB: 0D?

src/mrpro/operators/NonUniformFastFourierOp.py
```diff
@@ -53,125 +51,104 @@ def __init__(
...
         if isinstance(recon_matrix, SpatialDimension):
-            im_size: Sequence[int] = [int(astuple(recon_matrix)[d]) for d in self._nufft_directions]
+            im_size = tuple([recon_matrix.zyx[d] for d in self._direction_zyx])
         else:
-            if (n_recon_matrix := len(recon_matrix)) != (n_nufft_dir := len(self._nufft_directions)):
+            if (n_recon_matrix := len(recon_matrix)) != (n_nufft_dir := len(self._direction_zyx)):
                 raise ValueError(f'recon_matrix should have {n_nufft_dir} entries but has {n_recon_matrix}')
-            im_size = recon_matrix
+            im_size = tuple(recon_matrix)
+        assert len(im_size) == 1 or len(im_size) == 2 or len(im_size) == 3  # mypy  # noqa: S101
```
ckolbPTB, line 106: is this assert needed for mypy?

src/mrpro/operators/NonUniformFastFourierOp.py
```diff
@@ -226,22 +201,23 @@ def forward(self, x: torch.Tensor) -> tuple[torch.Tensor,]:
...
-            unflatten_shape = x.shape[: -len(self._nufft_directions)]
+            unflatten_shape = x.shape[: -len(self._direction_zyx)]
             # combine sep_dims
             x = x.flatten(end_dim=len(sep_dims_zyx) - 1) if len(sep_dims_zyx) else x[None, :]
             # combine joint_dims
-            x = x.flatten(start_dim=1, end_dim=-len(self._nufft_directions) - 1)
+            x = x.flatten(start_dim=1, end_dim=-len(self._direction_zyx) - 1)
 
-            x = self._nufft_type2(x)
+            x = torch.vmap(partial(finufft_type2, upsampfac=self.oversampling, modeord=0, isign=-1))(self._omega, x)
```
ckolbPTB, line 220: would be good to be able to set a chunk_size argument, otherwise we will prob not be able to handle 3D data on a GPU...

src/mrpro/operators/NonUniformFastFourierOp.py
fzimmermann89: Do we want to support an empty nufft op?
That does not do anything? We currently do, that's what mean by 0d

src/mrpro/operators/NonUniformFastFourierOp.py
fzimmermann89: We could instead directly  type hint imsize to have 1-3 values

src/mrpro/operators/NonUniformFastFourierOp.py
fzimmermann89: I think finuft does exactly the loop internally for the different trajectories. And supposedly does auto batching for the same trajectories.

But yes, We should try it on GPU.
(Do we have GPU tests for nufft?)

fzimmermann89:
We actually need the "wrong" kind of fft shift here. 
Shifting after the nufft is different than doing the equivalent operation in k space
We shift to avoid wrap around. 

We should check if oversampling=1. Or 1.25 is faster than using 2.0 here.
We only need 1.0, as we alreadyanually double to recon size.

src/mrpro/operators/NonUniformFastFourierOp.py
ckolbPTB:
```suggestion
            raise ValueError('Only 0D, 1D, 2D or 3D NUFFT is supported')
```

src/mrpro/operators/NonUniformFastFourierOp.py
ckolbPTB: might be a bit nicer

src/mrpro/operators/NonUniformFastFourierOp.py
ckolbPTB: we do not have any gpu tests for operators

If x would be 3D with `other` > 1 then vmap would try to apply finufft to all `other` datasets in parallel, wouldn't it? Unless we define a chuncksize here.

fzimmermann89: https://github.com/flatironinstitute/pytorch-finufft/blob/ba4bbc0583fc9d2205572c196eaa3ca26dbada96/pytorch_finufft/functional.py#L162

fzimmermann89: In a perfect finufft implementation, it would handle both batch dimes smart and choose the optimal batching size.

Currently,vit only does this for the second batch dim  and the first batch dim (with different trajectories) is always looped.

ckolbPTB (approved):

src/mrpro/operators/NonUniformFastFourierOp.py
```diff
@@ -24,9 +24,7 @@ def __init__(
         recon_matrix: SpatialDimension[int] | Sequence[int],
         encoding_matrix: SpatialDimension[int] | Sequence[int],
         traj: KTrajectory,
-        nufft_oversampling: float = 2.0,
-        nufft_numpoints: int = 6,
-        nufft_kbwidth: float = 2.34,
+        oversampling: float = 2.0,  # FinufftOversamplingFactors.AUTO,
```
fzimmermann89: must check what happens for invalid value

src/mrpro/operators/NonUniformFastFourierOp.py
```diff
@@ -53,125 +51,104 @@ def __init__(
...
             The k-space trajectories where the frequencies are sampled.
-        nufft_oversampling
+        oversampling
             Oversampling used for interpolation in non-uniform FFTs.
-        nufft_numpoints
-            Number of neighbors for interpolation in non-uniform FFTs.
-        nufft_kbwidth
-            Size of the Kaiser-Bessel kernel interpolation in non-uniform FFTs.
+            On GPU, 2.0 uses an optimized kernel, any value > 1.0 will work.
+            On CPU, there are kernels for 2.0 and 1.25. The latter saves memory. Set to 0.0 for automatic selection.
```
fzimmermann89, line 57: maybe check this?

src/mrpro/operators/NonUniformFastFourierOp.py
```diff
@@ -310,22 +291,17 @@ def gram_nufft_kernel(weight: torch.Tensor, trajectory: torch.Tensor, recon_shap
...
     for flips in list(product([1, -1], repeat=rank)):
-        if all(flip == 1 for flip in flips):
-            # top left ... block already processed before padding
-            continue
         flipped_trajectory = trajectory * torch.tensor(flips).to(trajectory).unsqueeze(-1)
-        kernel_part = adjnufft_ob(weight, flipped_trajectory)
+        kernel_part = adjnufft(flipped_trajectory, torch.polar(weight, (shifts * flipped_trajectory).sum(-2, True)))
         slices = []  # which part of the kernel to is currently being processed
         for dim, flip in zip(range(-rank, 0), flips, strict=True):
             if flip > 0:  # first half in the dimension
```
fzimmermann89, line 314: do we still need hermitify? can we already add them in the correct spot?

.pre-commit-config.yaml
```diff
@@ -90,6 +90,7 @@ repos:
         additional_dependencies:
           - numpy
           - torch>=2.6.0
+          - pytorch-finufft
```
fzimmermann89, line 94: this will only install cpu version.

consider adding cuda optional dependencies?

fzimmermann89: and give a better exception if used without cufionufft installed on a cuda tensor

src/mrpro/operators/NonUniformFastFourierOp.py
ckolbPTB: finufft throws an error:
upsampfac not a value with known Horner poly eval rule (currently 2.0 or 1.25 only)

https://finufft.readthedocs.io/en/latest/error.html

src/mrpro/operators/NonUniformFastFourierOp.py
fzimmermann89: I would need add cast(tuple[int]|tuple[int,int]|tuple[int,int],...) to both assignments.
thus, the assert looks cleaner imho

fzimmermann89:
I added cufinnuft as linux dependency.
Not sure what the best approach here is.
We could also add it as a "cuda" dependency.
But given that compared to a torch installation the 100mb for cufinnuft do not really matter and I assume that on most linux machines a GPU will be available, I think for now this is the best approach.

@ckolbPTB @schuenke Or would you prefer the linux default to not support cuda and require a pip install mrpro[cuda] to install cufinufft?

src/mrpro/operators/NonUniformFastFourierOp.py
fzimmermann89: This is a todo for another day

ckolbPTB:
> Or would you prefer the linux default to not support cuda and require a pip install mrpro[cuda] to install cufinufft?

This would mean that on linux without cuda the install would always fail, wouldn't it? Maybe better to have a cuda option.

fzimmermann89:
No, it just installes cufinufft (approx 100mb) even though it can't be used.

fzimmermann89:
(see the ci tests)

Issue #650, opened 2025-02-19
Sphinx version


fzimmermann89:
This is a reminder for future maintainers to be careful with sphinx version updates.

Our sphinx conf.py overwrites some internal sphinx functions, uses sphinx packages that also overwrite sphinx internals, and sphinx does not have great regression tests.

Each version update could break something.

So after a version bump of the dependencies, remember to do a diff of the final output to see if all changes are intentional and the overall changes are an improvement.

Otherwise, keep the sphinx version-as-is.

@schuenke @ckolbPTB @koflera @schote

PR #651, opened 2025-02-19
Fix formatting of functions as default arguments

fzimmermann89:
Currently this is broken:

https://zimf.de/zipserve/https://github.com/PTB-MR/mrpro/actions/runs/13410186766/artifacts/2615291997/_autosummary/mrpro.data.AcqInfo.html#mrpro.data.AcqInfo.from_ismrmrd_acquisitions

The '<>' in the repr of a function breaks the formatting.

Also, it pins the versions of all sphinx related packages. Reading through the issue tracker, there are many issues caused by updates and changes. I would suggest to only manually update the sphinx versions when required.
Otherwise, subtle breakage as above will go unnoticed for a long time.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -298,7 +298,7 @@ def sync_notebooks(source_folder, dest_folder):
 
 object_description_original = sphinx.util.inspect.object_description
 def object_description_function_repr_overwrite(obj, *, _seen: frozenset[int] = frozenset()) -> str:
-    """Overwrite sphinx default function representation to use functioname instead of <functioname>.
+    """Overwrite sphinx default function representation to use functionname instead of <functionname>.
 
     <> would break interspinx and formatting of the function name."""
     if callable(obj):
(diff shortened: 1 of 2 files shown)


fzimmermann89:
#650

docs/source/conf.py
```diff
@@ -295,11 +296,21 @@ def sync_notebooks(source_folder, dest_folder):
             content.metadata['mystnb'] = {'execution_mode': mode}
             nbformat.write(content, dest_file)
 
+object_description_original = sphinx.util.inspect.object_description
+def object_description_function_repr_overwrite(obj, *, _seen: frozenset[int] = frozenset()) -> str:
+    """Overwrite sphinx default function representation to use functioname instead of <functioname>.
```
fzimmermann89:
```suggestion
    """Overwrite sphinx default function representation to use functionname instead of <functionname>.
```

ckolbPTB (approved):

PR #653, opened 2025-02-20
Update readme on fork and pull request

### LINKED ISSUE #574: Move to fork-based development
Currently we are all working on branches of the main gihub repo at PTB-MR. We want to transition to a workflow where everyone implements code on their own fork and then opens a PR to PTB-MR/mrpro. This will work for everything except changes to the docker files. These files can only be modified via PTB-MR/mrpro by a few people.

Steps to take:
- [x] Ensure github actions using docker images work: review and merge #555
- [ ] Modify contributor guide to highlight fork-based development
- [ ] Restrict write access
- [x] Fix comments of coverage and docs build for the public repos #585 
- [x] Add pre-commit hook to create notebooks also locally: adapt #537
   - [x] create/update notebook from script but do not change script from notebook
   - [x] add notebooks to git ignore and verify that does not create merge conflicts if notebooks are created locally, commit is pushed to origin and then git pull is called
- [x] Subfolders for notebooks and example scripts: probably done in #537 which closes #546
- [x] Add colab support to notebook; done in #537
- [x] Remove binder support and then close #538, close #457

Please try to do this in smaller PRs not in one go such that it is easier to review.

ckolbPTB:
closes #574 

Is this correct?
https://github.com/ckolbPTB/mrpro/blob/c01a6ba0c6d55adfde1c2ea73be15fec8e51ce9a/README.md?plain=1#L102

Any other things which cannot be modified via forks which we should add here?


lrlunin:
I would say it _can be modified_ but changes will not applied until it's merged. Therefore they cannot be tested via pipelines. I need to take a look at the current setup. I guess the workflow will fail due to the lack of access token to the registry.

fzimmermann89:
I would remove that statement and keep it simple. Nobody will read this statement before working on the workflows. 
And worst case, the actions will not work in PRs that change them. And whoever want to change them should be able to deal with that. 

(but I would also have removed the  clone and setup python steps. if you know what they mean, you would also know what do do anyways..)

fzimmermann89:
what about
https://github.com/PTB-MR/mrpro/blob/534f324d71d224eaedb8bcfe7b11aedc263a9984/docs/source/contributor_guide.rst#

?

fzimmermann89 (approved):

PR #655, opened 2025-02-27
Make Rotation einops compatible

fzimmermann89:
Adds support for `einops.rearrange(Rotation() ...) `

### DIFF
--- a/src/mrpro/data/Rotation.py
+++ b/src/mrpro/data/Rotation.py
@@ -2048,6 +2048,20 @@ class Rotation(torch.nn.Module, Iterable['Rotation']):
             self._quaternions.expand(*newshape, 4), inversion=self._is_improper.expand(newshape), copy=True
         )
 
+    def unsqueeze(self, dim: int) -> Self:
+        """Unsqueeze the Rotation object in a batch dimension.
+
+        Add a new dimension to the Rotation object at the specified position.
+
+        Parameters
+        ----------
+        dim
+            The position where the new dimension is to be added.
+        """
+        if dim < 0:
+            dim = dim + self._is_improper.ndim + dim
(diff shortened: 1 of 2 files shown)


fzimmermann89:
* #683
* #675
* #674
* #673
* #672
* #671
* #660
* #591
* #679
* #560
* #656
* __->__ #655

src/mrpro/data/Rotation.py
```diff
@@ -2005,3 +2006,99 @@ def reshape(self, *shape: int | Sequence[int]) -> Self:
...
+        """Perform reduction operation on the Rotation."""
+        if operation == 'mean':
+            return x.mean(dim=reduced_axes)
+        raise NotImplementedError(f'Unknown reduction {operation} for Rotations')
+
+    def transpose(self, x: Rotation, axes: Sequence[int]) -> Rotation:
+        """Permute the axes of the Rotation."""
+        return x.permute(axes)
+
+    def stack_on_zeroth_dimension(self, x: Sequence[Rotation]) -> Rotation:
```
ckolbPTB, line 2090: Is this covered by a test already? I guess it would require a `Sequenze[Rotation]` as input to einops

ckolbPTB (approved):
If we want to use None as return type for test, please make sure it is done for all of them in Rotation (missing eg: test_as_euler_degenerate_asymmetric_axes) and add an issue to do for all other tests.

src/mrpro/data/Rotation.py
fzimmermann89: added

fzimmermann89:
> If we want to use None as return type for test, please make sure it is done for all of them in Rotation (missing eg: test_as_euler_degenerate_asymmetric_axes) and add an issue to do for all other tests.

#680

PR #656, opened 2025-02-27
Add unsqeeze_at

fzimmermann89:
Unsqueeze at a position multiple times and bring tensors to the same number of dimension by adding axes at a position

### DIFF
--- a/src/mrpro/utils/reshape.py
+++ b/src/mrpro/utils/reshape.py
@@ -104,9 +104,12 @@ def unsqueeze_tensors_at(*x, dim: int, ndim: int | None = None) -> tuple[torch.T
     -------
         unsqueezed tensors (views) with the same number of dimensions
     """
-    ndim_ = max(el.ndim for el in x) if ndim is None else ndim
-    if ndim < min(el.ndim for el in x):
+    if ndim is None:
+        ndim_ = max(el.ndim for el in x)
+    elif ndim < min(el.ndim for el in x):
         raise ValueError('ndim must be greater or equal to the minimum number of dimensions of the input tensors')
+    else:
+        ndim_ = ndim
     return tuple(unsqueeze_at(el, dim, n=ndim_ - el.ndim) for el in x)
 
 
--- a/src/mrpro/utils/reshape.py
+++ b/src/mrpro/utils/reshape.py
@@ -83,6 +83,13 @@ def unsqueeze_at(x: torch.Tensor, dim: int, n: int) -> torch.Tensor:
 def unsqueeze_tensors_at(*x, dim: int, ndim: int | None = None) -> tuple[torch.Tensor, ...]:
     """Unsqueeze tensors at a specific dimension to the same number of dimensions.
 
+    Example:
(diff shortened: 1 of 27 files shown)


fzimmermann89:
* #683
* #675
* #674
* #673
* #672
* #671
* #660
* #591
* #679
* #560
* __->__ #656
* #655

src/mrpro/utils/reshape.py
```diff
@@ -50,6 +50,57 @@ def unsqueeze_left(x: torch.Tensor, n: int) -> torch.Tensor:
...
+        raise IndexError(f'Dimension {dim} out of range for tensor of dimension {x.ndim}')
+    if dim < 0:
+        # dim=-1 should index after the last axis, etc. to match unsqueeze
+        dim = x.ndim + dim + 1
+    return x.reshape(*x.shape[:dim], *(n * (1,)), *x.shape[dim:])
+
+
+@endomorph
+def unsqueeze_tensors_at(*x, dim: int, ndim: int | None = None) -> tuple[torch.Tensor, ...]:
+    """Unsqueeze tensors at a specific dimension to the same number of dimensions.
```
koflera, line 84: From the description, I think I would not be able to understand what this should do... maybe add an example like for unsqueeze_at?

koflera:
```suggestion
    """Unsqueeze relevant tensors at a specific dimension such that all tensors in the output have the same number of dimensions.
```

src/mrpro/utils/reshape.py
```diff
@@ -50,6 +50,57 @@ def unsqueeze_left(x: torch.Tensor, n: int) -> torch.Tensor:
...
+    dim
+        dimension to unsqueeze
+    ndim
+        number of dimensions to unsqueeze to. If `None`, unsqueeze to the maximum number of dimensions
+        of the input tensors.
+
+    Returns
+    -------
+        unsqueezed tensors (views)
+    """
```
koflera, line 106:
```
x1 = torch.rand(1,4,5)
x2 = torch.rand(4,5,6,6)
(x1_, x2_) = unsqueeze_tensors_at(x1, x2, dim=2, ndim = 2)
```
raises the ValueError "n must be positive", from which it is not immediately clear what the problem is....
Maybe have a check 
```
if not ndim >= min(el.ndim for el in x):
   raise ValueError(....)
   

```

src/mrpro/utils/reshape.py
koflera: Example:
For `x1 = torch.rand(1,4,5)` and `x2 =torch.rand(4,5,6,6)`, the code `(x1_, x2_) = unsqueeze_tensors_at(x1, x2, dim=2, ndim = 6)`  yields `x1_` with shape `(1, 4, 1, 1, 1, 5)` and `x2_` with shape `(4, 5, 1, 1, 6, 6)`.

koflera (changes requested):
only a few comments

koflera (approved):
looks good to me

PR #683, opened 2025-03-04
Enable indexing of Dataclass

fzimmermann89:
Based on #580, enables the indexing for dataclasses


fzimmermann89:
* #732
* #731
* #730
* #735
* #729
* __->__ #683

tests/data/test_dataclass.py
```diff
@@ -275,3 +276,30 @@ def test_dataclass_repr() -> None:
    booltensor <Tensor>
    doubletensor <Tensor>"""
     assert actual == expected
+
+
+def check_broadcastable(actual_shape, expected_shape):
+    """Raise a Runtime Error is actual is not boradcastable to expected."""
```
ckolbPTB:
```suggestion
    """Raise a Runtime Error if actual is not boradcastable to expected."""
```

tests/data/test_dataclass.py
```diff
@@ -275,3 +276,30 @@ def test_dataclass_repr() -> None:
...
+        (0, (1, 20)),
+        (slice(0, 5), (5, 20)),
+        ((Ellipsis, (2, 3)), (10, 2)),
+        (((0, 1), (0, 1)), (2, 1, 1)),
+    ],
+    ids=['single', 'slice', 'ellipsis', 'vectorized'],
+)
+def test_dataclass_getitem(index, expected_shape: tuple[int, ...]) -> None:
+    """Test the __getitem__ method of the dataclass."""
+    # The indexing itself is alreadytested in test_indexer.py
```
ckolbPTB:
```suggestion
    # The indexing itself is already tested in test_indexer.py
```

tests/data/test_dataclass.py
```diff
@@ -275,3 +276,30 @@ def test_dataclass_repr() -> None:
...
+    # Thus, this test only needs to check that the indexing if performed on the attributes.
```
ckolbPTB:
```suggestion
    # Thus, this test only needs to check that the indexing is performed on the attributes.
```

src/mrpro/data/Dataclass.py
```diff
@@ -573,3 +576,27 @@ def __repr__(self) -> str:
...
         )
         return output
+
+    # region Indexing
+    def __getitem__(self, index: TorchIndexerType | Indexer) -> Self:
+        """Index the dataclass."""
+        indexer = index if isinstance(index, Indexer) else Indexer(self.shape, index)
+        memo: dict[int, Any] = {}
+
+        def apply_index(data: T) -> T:
```
ckolbPTB, line 586: Is the casting here only needed for mypy?

ckolbPTB (approved):

src/mrpro/data/Dataclass.py
fzimmermann89: "cast" is always only for mypy.
it does not do anything but tell mypy the new type

quick explanation:
isinstance(data, torch.Tensor) narros the type down to torch.Tensor, even if data could be a tensor subclass.
but we actually will return the subclass (as we should). so we don't want to narrow the return tpe, but leave it as is

fzimmermann89: this is a common pattern we have in a couple of places already for typvars

Issue #684, opened 2025-03-04
pydicom version


koflera:
pydicom 2.3 from our requirements results in
>       dataset.save_as(self.filename, enforce_file_format=True)
E       TypeError: Dataset.save_as() got an unexpected keyword argument 'enforce_file_format'

in the tests. we might need a newer pydicom version. 

-Felix

#361

koflera:
It works with the current latest pydicom 3.0.1 (and maybe also older versions)

PR #685, opened 2025-03-04
Bump pydicom dependency

koflera:
update pyproject.toml for dependencies


fzimmermann89:
Is this the lowest version we support?

fzimmermann89:
Closes #684

fzimmermann89:
> Is this the lowest version we support?

we at least need 3.0.0, so depending on 3.0.1 seems fine.

fzimmermann89 (approved):

PR #686, opened 2025-03-04
Tests for EPG

ckolbPTB:


fzimmermann89:
Btw: have you checked how many states are required for the cmrf sequence to obtain good results?

This should depend on tr and the maximum t2, right?

PR #687, opened 2025-03-04
Change iteration argument name in optimizers

### LINKED ISSUE #340: Consistent use of name for parameter "number of iterations" / "maximal number of iterations"
Currently, we are using a mix of different variable names for the number of iterations of iterative algorithms.

See also the discussion in #224 .

We seem to have come to the conclusion that we would like to have

`max_iterations` for algorithms that are provided with some stopping critertion, which prevents the algorithm to run for all iterations. Thus, `max_iterations `is an upper bound for the actual number of iterations that the algorithm possibly runs.

`n_iterations ` for algorithms that run ALL iterations.

TODO: 

- [x] CG: uses `max_iterations `  --> includes tolerance criterion to possibly stop before
- [ ] Lbfgs: currently uses `max_iter `instead of `max_iterations` (renaming required)
- [ ] Adam: currently uses `max_iter `instead of `n_iterations` (renaming required)

Stef-Martin:
Change of variable names in optimizers

Closes #340


examples/notebooks/qmri_sg_challenge_2024_t1.ipynb
```diff
@@ -318,7 +318,7 @@
     "lr = 1e-1\n",
     "\n",
     "# Run optimization\n",
-    "result = mrpro.algorithms.optimizers.adam(functional, [m0_start, t1_start], max_iter=max_iter, lr=lr)\n",
+    "result = mrpro.algorithms.optimizers.adam(functional, [m0_start, t1_start], n_iterations=max_iter, learning_rate=lr)\n",
```
koflera: I would change the variables `max_iter `and `lr` in the notebook as well to have
`n_iterations=n_iterations` and `learning_rate=learning_rate`

examples/scripts/qmri_sg_challenge_2024_t1.py
```diff
@@ -158,7 +158,7 @@ def show_images(*images: torch.Tensor, titles: list[str] | None = None) -> None:
 lr = 1e-1
 
 # Run optimization
-result = mrpro.algorithms.optimizers.adam(functional, [m0_start, t1_start], max_iter=max_iter, lr=lr)
+result = mrpro.algorithms.optimizers.adam(functional, [m0_start, t1_start], n_iterations=max_iter, learning_rate=lr)
```
koflera: same comment as above

koflera (dismissed):
looks good to me; only a few small comments

tests/algorithms/test_optimizers.py
```diff
@@ -66,14 +69,18 @@ def test_callback_optimizers(optimizer):
...
-
     # callback function; if the function is called during the iterations, the
     # test is successful
     def callback(optimizer_status: OptimizerStatus) -> None:
         _, _ = optimizer_status['iteration_number'], optimizer_status['solution'][0]
         assert True
 
     # run optimizer
-    _ = optimizer(rosen_brock, parameters, max_iter=max_iter, callback=callback)
+    if optimizer == lbfgs:
```
fzimmermann89: either remove the max_iterations/n_iterations argument here (it does not matter) (preferred)
OR just supply "10" without naming the argument
OR do it optimizer_kwargs in the parameterize, as done in the test above (least favourite)

fzimmermann89: -> dont have this if here.

koflera:
since you are already on it, i think here, it would be better to have

https://github.com/PTB-MR/mrpro/blob/ce66dc58d9140fc1f46f707a2326e1c7c8c56775/examples/scripts/qmri_sg_challenge_2024_t1.py#L89

$E_{\mathrm{relative}} = \sum_{TI}\frac{|(q(M_0, T_1, TI) - x)|}{|x|}$ instead of 

$E_{relative} = \sum_{TI}\frac{|(q(M_0, T_1, TI) - x)|}{|x|}$

koflera:
and here, for better visibility

https://github.com/PTB-MR/mrpro/blob/ce66dc58d9140fc1f46f707a2326e1c7c8c56775/examples/scripts/qmri_sg_challenge_2024_t1.py#L89

\min_{M_0, T_1} \big\| |q(M_0, T_1, TI)| - x\big\|_2^2

instead of 

\min_{M_0, T_1} || |q(M_0, T_1, TI)| - x||_2^2

(in github you don't see the difference, but in the notebook it looks better ;) )

koflera:
typo here 

https://github.com/PTB-MR/mrpro/blob/ce66dc58d9140fc1f46f707a2326e1c7c8c56775/examples/scripts/qmri_sg_challenge_2024_t1.py#L146

"ny moving" --> "by moving"

koflera (commented):
a few comments; typos more than anything else

fzimmermann89:
> since you are already on it, i think here, it would be better to have
image


image

fzimmermann89 (approved):

PR #688, opened 2025-03-04
Read dicom into IHeader

### LINKED ISSUE #654: Image header from dicom
We need to 
- [ ] check that the units of tr, ti, te are at least correct for our scanner
- [ ] load orientation, position and patient position from dicom
- [ ] read timestamp from dicom, maybe timeafterstart.
- [ ] try to fill image idx from dicom (maybe (0021,1106))

All of these should focus in Siemens xa61 dicoms.

See https://github.com/rordenlab/dcm2niix/blob/master/Siemens/README.md for some hints

schote:
closing #654


schote:
Dicom values to be added to the tests:

- [x] InversionTime
- [x] FrameReferenceDateTime
- [x] NominalCardiacTriggerDelayTime
- [x] TemporalPositionIndex
- [x] InStackPositionNumber

src/mrpro/data/IHeader.py
```diff
@@ -275,29 +299,79 @@ def from_dicom(cls, *dataset: Dataset) -> Self:
...
+            basis_y = torch.tensor(dcm_orientation[0][3:])
+            # Calculate third basis vector by cross product
+            basis_z = torch.cross(basis_x, basis_y, dim=-1)
+            # Get orientation from basis vectors
+            orientation = Rotation.from_directions(
+                SpatialDimension.from_array_xyz(basis_x),
+                SpatialDimension.from_array_xyz(basis_y),
+                SpatialDimension.from_array_xyz(basis_z),
+            )
+        else:
```
fzimmermann89: isnt header.orientation defaulted to an identty?

tests/data/test_idata.py
```diff
@@ -168,11 +176,23 @@ def test_IData_to_dicom_folder(dcm_data_fixture, request):
...
     torch.testing.assert_close(idata_reloaded.header.te[0], idata.header.te[0])
-    assert idata_reloaded.header.position == idata.header.position
+    torch.testing.assert_close(idata_reloaded.header.position.x, idata.header.position.x)
+    torch.testing.assert_close(idata_reloaded.header.position.y, idata.header.position.y)
+    torch.testing.assert_close(idata_reloaded.header.position.z, idata.header.position.z)
     assert idata_reloaded.header.orientation == idata.header.orientation
     torch.testing.assert_close(idata_reloaded.data, idata.data)
+
+
+def test_IData_from_kheader_and_tensor_to_dicom_folder(tmp_path_factory, random_kheader, random_test_data):
```
schote, line 187: Should we also check the data?
1. `from_dicom_folder` doesn't know about coils yet
2. `IData.from_tensor_and_kheader` contains complex-valued data,`idata.to_dicom_folder` stores normalized magnitude, i.e. image pixel values

Could add something like:
```
other, coil = idata.shape[:2]
idata_reloaded = idata_reloaded.rearrange('(other coil) 1 z y x -> other coil z y x', other=other, coil=coil)
idata_pixels = idata.data.abs()
idata_pixels = idata_pixels / idata_pixels.max() * (2**16 - 1)
torch.testing.assert_close(idata_pixels, idata_reloaded.data.abs())
```
Dicom provides `(0018,9042) - MR Receive Coil Sequence Attribute`, maybe we could extend `idata.to_dicom_folder` to store individual coil images? 
Any ideas/comments?

ckolbPTB: Testing the image data as suggested above sounds like a good idea to me. 

I would not bother with storing individual coil images with a specific label. There are so many different dicom tags and only very few are really supported by the standard viewers. Better to keep it simple on our end.

ckolbPTB (approved):

PR #694, opened 2025-03-05
CUDA tests of operators

bernharb:
Gradual adding of CUDA tests for the various operators


fzimmermann89:
higher priority:
- [x] nonuniformfastfourierop
- [x] fourierop

lower priority:
- [x] gridsampling
- [x] wavelet
- [x] slice projection

fzimmermann89:
@ckolbPTB : for linear operators, shouldnt we also be testing the adjoint?

maybe, for laziness, we can call operator.H@operator in the for most of the linear operators?

andbut for fourierop, we should also test the .gram (which has a faster implementation than operator.H@operator) explicitly.

tests/operators/test_constraints_op.py
```diff
@@ -156,3 +156,24 @@ def test_autodiff_constraints_operator():
...
+    # Create on CPU, run on CPU
+    constraints_op = ConstraintsOp(bounds)
+    (cx,) = constraints_op(x)
+    assert cx.is_cpu
+
+    # Create on CPU, transfer to GPU, run on GPU
+    constraints_op = ConstraintsOp(bounds)
+    constraints_op.cuda()
+    (cx,) = constraints_op(x.cuda())
+    assert cx.is_cuda
```
fzimmermann89, line 180: it would make sense to also check the inverse operation here

bernharb:
I believe this concludes the cuda tests. Adjoints were added to cuda tests of all linear operators and the other comments (gram of  Fourier and inverse of constraints) were taken care of as well.
As always pytest -m cuda was run on a cuda device and came out successfull after all changes included here.

lrlunin (approved):
With the double confirmation of the successful cuda test in workflow https://github.com/PTB-MR/mrpro/actions/runs/14470063464/job/40581449057 I approve this PR!

PR #695, opened 2025-03-05
Fix batched complex grid sampling

### LINKED ISSUE #513: GridSamplingOp on dynamics
I had problems using the GridSamplingOp on image dynamics, in particular on images in the shape of (other 1 1 y x) with grids of shape (other y x 2), where in other I had the different timesteps, so other was bigger than 1 both for the images and for the grids (and it was the same value)

the error came from the GridSamplingOp rearrange:
` x_real = rearrange(torch.view_as_real(x), '... real_imag  -> real_imag ...') if x.is_complex() else x`
which led to a broadcasting error

For what I need it for, I used it just by changing it into
`x_real = rearrange(torch.view_as_real(x), ' other ... real_imag  -> other real_imag ... ') if x.is_complex() else x`
and
`result = torch.view_as_complex(rearrange(result, 'other real_imag ... -> other ... real_imag').contiguous())`

This worked for me, but it would work only if either other is the same for images and grids, or one of them is equal to one, so maybe there is a better way to generalize this

fzimmermann89:
- [x] Add failing test case
- [x] Fix failign test case 

Fixes #513


fzimmermann89:
@guastinimara your suggested solution would not work for more than one "other" dimension.

Please check if this fixes the issue.
The operator is now applied to real and imag part separately.
This should not be significantly slower but something simplifies the code code a bit.

fzimmermann89:
@guastinimara your suggested solution would not work for more than one "other" dimension.

Please check if this fixes the issue.
The operator is now applied to real and imag part separately.
This should not be significantly slower but something simplifies the code code a bit.

fzimmermann89:
@guastinimara your suggested solution would not work for more than one "other" dimension.

Please check if this fixes the issue.
The operator is now applied to real and imag part separately.
This should not be significantly slower but something simplifies the code code a bit.

The issue was that the tests either considered batch or OR complex values -- never both.

fzimmermann89:
@guastinimara does this fix your issue?

guastinimara:
Sorry I thought I answered here! Yes, the issue is now fixed. I read the code and I had no questions about it

guastinimara (approved):
I have no comments on this

PR #699, opened 2025-03-06
Add stopping criterion to optimizer callback

### LINKED ISSUE #603: Change optimizer callback signature to allow custom stopping criteria
Suggestion: 
We could change the signature of the callback of the optimizers to `Callable[[OptimizerStatus], None|bool]`
Returning `False` form the callback would stop the optimizer. 

This would allow for a custom stopping criterion.

guastinimara:
Fixes #603

I added the option of the callback returning False, thus stopping the optimizers. I made separate tests because I tried making one test only, but it didn't look good because I still had to separate cases... But if you prefer, I can put everything in one


src/mrpro/algorithms/optimizers/adam.py
```diff
@@ -105,13 +105,15 @@ def closure():
...
         objective.backward()
 
-        if callback is not None:
-            callback({'solution': parameters, 'iteration_number': iteration})
-
         return objective
 
     # run adam
-    for iteration in range(max_iter):  # noqa: B007
+    for iteration in range(max_iter):
```
guastinimara, line 112: I just realised this comment got canceled by the pre-commit

fzimmermann89:
Please add the stopping feature to the docstring of the callback argument

guastinimara:
@fzimmermann89 I see here that 1 pytest failed, but I don't really understand it or if it has to do with my changes, for me the tests of optimizers all worked out

fzimmermann89 (approved):

PR #700, opened 2025-03-06
learning of parameter maps for tv reconstruction 

koflera:
we here try to include basic code for being able to run the method described in 

https://arxiv.org/abs/2301.05888

using MRpro in an  "advanced example" or so


koflera:
except that the example is probably not nice yet etc, the main problem is that while training on the GPU, the GPU memory continues to increase, eventually leading to a CUDA out of memory error. 
Is it super evident where the problem lies?

koflera:
Finally, a first draft of the example is ready!
@fzimmermann89 I ended up undoing most of the changes that we had discussed. The most relevant were:
- normalization by std instead of the norm: normalizing by the std had the problem that when at the end of the example, we want to compare the iterative SENSE recon to the PDHG ones, the resulting scaling factor was substantially different when normalizing by the std. normalizing by the norm was much more consistent, so I went for that.
- initial value estimation as part of the network: given that approximately solving the normal equations only makes sense if we have multi-coil problem, I abstained from doing that and instead left that to the user to be done outside. I also think it makes the network more concise.
- creating the dataset with a torch.utils.data.TensorDataset and using the split function, required wo write an additional Wrapper to able to use the method `prepare_data` of the resulting `SubSets`. Additionally, after a few hours of trying to stop mypy from complaining, I gave up and went back to manually creating two different datasets.

src/mrpro/utils/RandomGenerator.py
```diff
@@ -859,10 +859,11 @@ def gaussian_variable_density_samples(
             raise ValueError('n_samples must be <= (high - low)')
         n_random = n_samples - len(always_sample)
         if n_random < 0:
-            raise ValueError('more always sampled points requested than total number of samples')
+            raise ValueError('more "always sampled points" requested than total given number of samples')
         elif n_random == 0:
             return torch.tensor(always_sample).sort().values.broadcast_to(*n_batch, -1)
-        pdf = torch.exp(-torch.tensor(2.0).log() * (2 * torch.arange(low, high) / fwhm) ** 2)
+        pdf = torch.exp(-torch.tensor(2.0).log() * (2 * torch.arange(low - high // 2, low + high // 2) / fwhm) ** 2)
```
fzimmermann89: This looks wrong.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,876 @@
...
+        r"""Apply the network to estimate regularization parameter maps.
+
+        Parameters
+        ----------
+        image
+            the image from which the regularization parameter maps should be estimated.
+
+        """
+        dims = tuple(dim for dim in range(1, image.ndim))
+        image = (image - image.mean(dim=dims, keepdim=True)) / image.std(dim=dims, keepdim=True)
```
fzimmermann89: Please use Instancenorm
https://pytorch.org/docs/stable/generated/torch.nn.InstanceNorm2d.html
or is this something else?

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,876 @@
...
+        regularization_parameter_map = self.cnn_block(image)
+
+        # stack the parameter map channel dimension to share the regularization
+        # between the x- and y-direction of the image gradients
+        regularization_parameter_map = torch.concat(2 * [regularization_parameter_map], dim=1)
+
+        # apply softplus to enforce strict positvity and scale by hand-crafted parameter
+        regularization_parameter_map = torch.nn.functional.softplus(
+            self.t, beta=self.beta_softplus
+        ) * torch.nn.functional.softplus(regularization_parameter_map, beta=self.beta_softplus)
```
fzimmermann89: Add softplus to sequential.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,876 @@
...
+        finite_differences_operator = mrpro.operators.FiniteDifferenceOp(dim=(-2, -1), mode='forward')
+        gradient_operator = (
+            mrpro.operators.RearrangeOp('grad batch ... -> batch grad ... ') @ finite_differences_operator
+        )
+        self.gradient_operator = gradient_operator
+
+        self.g = mrpro.operators.functionals.ZeroFunctional()
+
+        # operator norm of the stacked operator K=[A, \nabla]^T
+        self.stacked_operator_norm = 3.0  # analytically calculated
```
fzimmermann89: I would like to have an analytical  operator_norm for the Finite difference operator and the Fourieroperator  for pure cartesian Fourier.
And use here .operator_norm on the stacked operator.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,876 @@
...
+
+    def estimate_lambda_map(self, image: torch.Tensor) -> torch.Tensor:
+        """Estimate regularization parameter map from image."""
+        # squeeze the dimensions that are one. In particular, the initial image
+        # is a coil-combined image (i.e. coils=1) and because, we only consider 2D problems here,
+        # z can be assumed to be always 1.
```
fzimmermann89: Unnecessary. Please remove or shorten. Use einops for self documenting code.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,876 @@
...
+        regularization_parameter_map = self.lambda_map_network(input_image).unsqueeze(-3).unsqueeze(-3)
+        return regularization_parameter_map
+
+    def forward(
+        self,
+        initial_image: torch.Tensor,
+        csm: torch.Tensor,
+        kdata: torch.Tensor,
+        traj: mrpro.data.KTrajectory,
+        regularization_parameter: torch.Tensor | None = None,
```
fzimmermann89, line 201: Please remove this argument. 
Instead, allow to set a constant parameter in init instead of a prediction network.
As this bypasses the whole network just to get a comparison method, it is a bit strange.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,876 @@
...
+# First, we start by defining some auxiliary functions that we will need for retrospectively simulating undersampled
+# and noisy k-space data.
+
+# %%
+
+
+def add_gaussian_noise(kdata: torch.Tensor, noise_variance: float, seed: int) -> torch.Tensor:
+    """Corrupt given k-space data with additive Gaussian noise."""
+    rng = torch.Generator().manual_seed(seed)
+    kdata = kdata + noise_variance * kdata.std() * torch.randn(
```
fzimmermann89: Use Random generator from utils

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,876 @@
...
+) -> tuple[torch.Tensor, torch.Tensor] | torch.Tensor:
+    """
+    Normalize k-space data and (possibly) target image.
+
+    Divide by std multiply by (an empirically set) factor.
+    """
+    kdata_norm = torch.linalg.norm(kdata)
+
+    kdata /= kdata_norm
+    kdata *= factor
```
fzimmermann89, line 304: As discussed, please get rid of the arbitrary factor and use either Std in image space or l2norm/pixels 
Or anything that is image size independent. (l2 norm scales with the vector lenght).

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,876 @@
...
+
+        Returns
+        -------
+        tuple
+            A tuple containing k-space data, coil sensitivity maps, k-space trajectories adjoint reconstruction,
+            pseudo-inverse solution and the image.
+        """
+        # randomly choose trajectories to define the fourier operator
+        ny, nx = image.shape[-2:]
+        img_shape = mrpro.data.SpatialDimension(z=1, y=ny, x=nx)
```
fzimmermann89: Plase use recon and encoding matrix naming for these to be consistant instead of domain and range or image and kspace.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,876 @@
...
+            assert image is not None
+            kdata, image = normalize_kspace_data_and_image(kdata, image)
+
+            (adjoint_recon,) = forward_operator.H(kdata)
+
+            # compute an approximation of the pseudo-inverse solution
+            # Note that this in general only makes sense if the forward operator can be injective,
+            # i.e. we have multiple coils with acceleration factor < n_coils
+            (pseudo_inverse_solution,) = mrpro.algorithms.optimizers.cg(
+                forward_operator.gram, right_hand_side=adjoint_recon, initial_value=adjoint_recon, max_iterations=16
```
fzimmermann89: As discussed, please have a clear separation what the problem is and what the solution is.
Don't have CG in your dataset.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,876 @@
...
+            kdata = add_gaussian_noise(kdata, noise_variance=noise_variance, seed=seed)
+
+            assert image is not None
+            kdata, image = normalize_kspace_data_and_image(kdata, image)
+
+            (adjoint_recon,) = forward_operator.H(kdata)
+
+            # compute an approximation of the pseudo-inverse solution
+            # Note that this in general only makes sense if the forward operator can be injective,
+            # i.e. we have multiple coils with acceleration factor < n_coils
```
fzimmermann89: Remove.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,876 @@
...
+        """Parameters
+
+        ----------
+        images
+            images to be used in the dataset.
+            Should have the shape (others, coils, z, y, x) with z=1.
+        """
+        self.images = images
+
+    def prepare_data(
```
fzimmermann89: Why is this an instance method of a dataset?
Please refactor as already discussed in length.

fzimmermann89 (changes requested):

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
fzimmermann89: Then, the forward is just the sequential * softplus(t)

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,876 @@
...
+            the image from which the regularization parameter maps should be estimated.
+
+        """
+        dims = tuple(dim for dim in range(1, image.ndim))
+        image = (image - image.mean(dim=dims, keepdim=True)) / image.std(dim=dims, keepdim=True)
+        regularization_parameter_map = self.cnn_block(image)
+
+        # stack the parameter map channel dimension to share the regularization
+        # between the x- and y-direction of the image gradients
+        regularization_parameter_map = torch.concat(2 * [regularization_parameter_map], dim=1)
```
fzimmermann89: move after the softplus and just expand the channel dim.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
fzimmermann89: Not sure if this fits into this PR or should be done separatly

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,876 @@
...
+
+        Returns
+        -------
+            Image reconstructed by the TV-minimization algorithm.
+        """
+        # if no regularization parameter map is provided, compute it with the network
+        if regularization_parameter is None:
+            regularization_parameter = self.estimate_lambda_map(initial_image)
+
+        f_1 = 0.5 * mrpro.operators.functionals.L2NormSquared(target=kdata, divide_by_n=False)
```
fzimmermann89: divide_by_n is False by default.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,876 @@
...
+        fourier_operator = mrpro.operators.FourierOp(
+            recon_matrix=self.img_shape,
+            encoding_matrix=self.k_shape,
+            traj=traj,
+        )
+        csm_operator = mrpro.operators.SensitivityOp(csm)
+        forward_operator = fourier_operator @ csm_operator
+        stacked_operator = mrpro.operators.LinearOperatorMatrix(((forward_operator,), (self.gradient_operator,)))
+
+        primal_stepsize = dual_stepsize = 0.95 * 1.0 / self.stacked_operator_norm
```
fzimmermann89: if we dont calculate the operator_norm, please instead just set the primal and dual stepsize in the init
(same equation). looks a bit nicer than saving the stacked operator norm.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
fzimmermann89: (there is also randn_like)

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
fzimmermann89: If we use some arbitrary scaling for the k-space data, this should be done for all the data -- also for the reference methods. As such, I dont really understand your point wrt SENSE.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,876 @@
...
+    else:
+        return kdata
+
+
+# %% [markdown]
+# Further, we define a torch dataset that will be used for training the network.
+# %%
+
+
+class Cartesian2D(torch.utils.data.Dataset):
```
fzimmermann89: This is just a TensorDataset with a static function added to it. The functional is not rellay part of a "Dataset", right? Maybe separating both would help.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
fzimmermann89: We dont have this blabla for the sense reconstruciton etc. Althoug it is technically correct,  there is no need to "Note" here.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,876 @@
...
+        traj: mrpro.data.KTrajectory,
+        regularization_parameter: torch.Tensor | None = None,
+    ) -> torch.Tensor:
+        """Reconstruct image using an unrolled PDHG algorithm.
+
+        Parameters
+        ----------
+        initial_image
+            initial guess of the solution of the TV problem.
+        csm
```
fzimmermann89: Your argument that moving the initial image estimation inside the network only makes sense for a multicoil problem fails a bit short if the network takes in CSMs.
It is ok to limit this to multicoil with okish g-factor.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
fzimmermann89: Either have a __getitem__ that returns target image, kdata, fourierop with collate funciton for fourier op,
or a tensordataset of images and have a prepare data function or a CartesianProblem class with
n_coils etc set in init and a __call__ that is imagetensor -> (kdata, aquisitionoperator)

and the forward of the network be (kdata, aquisitionoperator) -> image

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,876 @@
...
+        fourier_operator = mrpro.operators.FourierOp(
+            traj=traj,
+            recon_matrix=img_shape,
+            encoding_matrix=k_shape,
+        )
+
+        # generate simualated coil sensitivity maps
+        from mrpro.phantoms.coils import birdcage_2d
+
+        with torch.no_grad():
```
fzimmermann89: why?

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,876 @@
...
+            csm = birdcage_2d(n_coils, img_shape, relative_radius=0.8)
+            csm_operator = mrpro.operators.SensitivityOp(csm)
+
+            forward_operator = fourier_operator @ csm_operator
+
+            (kdata,) = forward_operator(image)
+            kdata = add_gaussian_noise(kdata, noise_variance=noise_variance, seed=seed)
+
```
fzimmermann89: why can image be None?

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
fzimmermann89: As using CG on the normal equation is part of the proposed method , it should be in the Network.

fzimmermann89: The network is only used in this example notbook. Even though you technically could use a differnet intial value or a different kind of problem, this is not helpful for understanding and organization. Just say in the markdown text that we use the initialization as proposed in the paper and that one could also use something else

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,876 @@
...
+        return len(self.images)
+
+    def __getitem__(self, idx: int):
+        """Return the image indexed by idx."""
+        image = self.images[idx]
+
+        return image
+
+
+# %% [markdown]
```
fzimmermann89: consider hiding

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,876 @@
...
+dataset_validation = Cartesian2D(images[20:, ...])
+dataloader_train = torch.utils.data.DataLoader(dataset_train, shuffle=True, batch_size=1)
+dataloader_validation = torch.utils.data.DataLoader(dataset_validation, shuffle=False, batch_size=1)
+
+
+# %% tags=["hide-cell"] mystnb={"code_prompt_show": "Show plotting details"}
+# Define some functions for plotting images.
+import matplotlib.pyplot as plt
+
+plt.ioff()
```
fzimmermann89: ?

src/mrpro/utils/RandomGenerator.py
fzimmermann89: The fix should be in the trajectory calculator which is using this function wrong?

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
koflera: yes, instancenorm seems to be something else. Here, the idea was just to have an input image that is std-normally distributed, i.e. image ~ N(0, Id), but perhaps it could be removed, since the k-space data is already normalized anyway. I'll have a look.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
koflera: maybe I am wrong, but there seems not to be `randn_like `yet?
also, I just found out that, for example, even `rand_like` does not have the expected behaviour. 
I.e. if we have that y.device = cuda --> rng.rand_like(y).device = cuda? which is not the case currently.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
koflera: here it can't, can it?

koflera: do you perhaps mean in the normalization function? 
If yes, then this was for the case, where we don't have a data-pair `(kdata , target_image)`, but only the kdata, as for example at the end of the script.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
fzimmermann89: this is what instancenorm, layernorm, groupnorm etc do.
instancenorm would make each channel ~N(0,Id), groupnorm with groups=1 makes each sample of the batch ~N(0,Id), as you currently do.

koflera: ah, I see....  I overlooked the fact that gamma and beta could also be excluded. 
I'll have a look, cause maybe, the normalization is really not needed.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
koflera: hmmm.... i see what you mean, but I disagree on that. As it is implemented now, one can for example also have a look at what results different regularization parameter maps would give without needing to instantiate different reconstruction networks.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
fzimmermann89: Let me rephrase, why is there an assert?

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
koflera: I agree on having anaytical operator-norms for certain operators. But let's adapt the example once we have that.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
koflera: ah I see... that must have been some residual code where the structure of the code was different and image was indeed allowed to be None.
removed

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
koflera: For this example, basically the following points have to work:
i) train on brain-web data (training must be successful; i.e. validation error must go down)
ii) the application of the network trained "live" (i.e. only for a few epochs) should surpass the best scalar TV-reconstruction

In particular, for ii), we need to be able to compute error between TV-reconstruction and the iterative SENSE from the fully sampled data. These images must have the same scale sucht that the computed error reflects only mistakes coming from TV and not from the different scales. The std didn't work for that cause then the iterative SENSE reconstructions were much brighter than the TV reconstructions (or the other way round, don't remember). And I wanted to use a normalization that does not use the target image for obvious reasons.

For many other normalization techniques that I have tried ii) was fine but then i) was not working.

I think I have found now something for which i) and ii) works.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
koflera: I have now separated the `prepare_data` function, such that using n_coils=1 is easily possible. but I would like to keep the network also sucht that it is directly applicable for single-coil (low-field MRI). 
the network now will take kdata, input image and forward operator as inputs in the forward.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
koflera: I would like to keep that. it's two lines, it's educational and I am planning to hide the `prepare_data` function anyway.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
koflera: I disagree on that. Given that I also refactored now the forward to have the forward operator as input (as suggested) and that one could consider single-coil acquisitions, the preparation of the initial value must take place outside of the network.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,876 @@
...
+
+n_epochs = 10
+validation_loss_values = []
+best_model = None
+noise_variance = 0.05
+acceleration_factor = 4
+
+time_start = time()
+
+if torch.cuda.is_available():
```
fzimmermann89:
```
if not torch.cuda.is_available()
 # without a gpu, skip training
     n_epochs=0
     adaptive_tv_network.load_state_dict(torch.load(data_folder_tv / 'tv_model.pt', map_location='cpu'))
 
```

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
fzimmermann89: i meant torch.rand_like  -- I forgot about randomgenerator.rand_like.
you are right, there is a device missing. could you add it to randomgenerator.rand_like?

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
fzimmermann89: this will break if you sample more k-space lines or switch to non cartesian sampling. you want the image to be z-normalized for network traning. normalizing frobenius norm to be 100 is arbitrary and not a good example.

fzimmermann89: (if you want to speed up traning, change learning rate and optimizer settings instead?)

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
fzimmermann89: ?

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,876 @@
...
+# the network's parameters by gradient-descent.
+# To obtain a somewhat decent estimate of the network parameters, training should take approximately 5-6 minutes
+# on a GPU. If you have GPU and you need a break, start the training now, grab yourself a coffee and come back later
+# to see your network trained from scratch.
+# If you do not have a GPU, we provide a pre-trained model, which you can load and directly apply to the data.
+
+# %%
+from typing import cast
+
+cnn_block = cast(torch.nn.Module, adaptive_tv_network.lambda_map_network.cnn_block)
```
fzimmermann89: why is this without the cast? i would rather have a type ignore than a cast in an example.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,876 @@
...
+                image = image.cuda()
+
+            with torch.no_grad():
+                pdhg_recon = adaptive_tv_network(pseudo_inverse_solution, csm, kdata, traj)
+                loss = torch.nn.functional.mse_loss(torch.view_as_real(pdhg_recon), torch.view_as_real(image))
+
+            running_loss += loss.item()
+        loss_validation = running_loss / len(dataset_validation)
+        validation_loss_values.append(loss_validation)
+        print(validation_loss_values[-1])
```
fzimmermann89: add to tqdm?

fzimmermann89:
sorry, you did not push yet, did you? :D

koflera:
> sorry, you did not push yet, did you? :D

no no, still waiting for the new training to finish... will probably push tomorrow in the morning ;)

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
koflera: I have tried around a bit, but still not success with type ignore. do you have a more concrete suggestion?

koflera: ok, have found something.

src/mrpro/data/traj_calculators/KTrajectoryCartesian.py
```diff
@@ -117,8 +117,13 @@ def gaussian_variable_density(
         k1_center = n_k1 // 2
         center_start = k1_center - n_center // 2
         center_end = center_start + n_center
+
         k1_idx = rng.gaussian_variable_density_samples(
```
fzimmermann89: this still looks wrong. can we move this to a separate PR and discuss there?

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,881 @@
...
+        r"""Initialize Adaptive TV Network.
+
+        Parameters
+        ----------
+        n_filters
+            Number of filters to be applied in the convolutional layers of the network.
+
+        """
+        super().__init__()
+        self.beta_softplus = 10.0
```
fzimmermann89:
```suggestion
```

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,881 @@
...
+        self.cnn_block: torch.nn.Module = torch.nn.Sequential(
+            *[
+                torch.nn.InstanceNorm2d(1, affine=False, track_running_stats=False),
```
fzimmermann89:
```suggestion
                torch.nn.InstanceNorm2d(1),
```

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,881 @@
...
+                torch.nn.Conv2d(in_channels=1, out_channels=n_filters, kernel_size=3, stride=1, padding=1),
+                torch.nn.LeakyReLU(),
+                torch.nn.Conv2d(in_channels=n_filters, out_channels=n_filters, kernel_size=3, stride=1, padding=1),
+                torch.nn.LeakyReLU(),
+                torch.nn.Conv2d(in_channels=n_filters, out_channels=n_filters, kernel_size=3, stride=1, padding=1),
+                torch.nn.LeakyReLU(),
+                torch.nn.Conv2d(in_channels=n_filters, out_channels=1, kernel_size=3, stride=1, padding=1),
+                torch.nn.Softplus(beta=self.beta_softplus),
```
fzimmermann89:
```suggestion
                torch.nn.Softplus(beta=10),
```

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,881 @@
...
+        r"""Apply the network to estimate regularization parameter maps.
+
+        Parameters
+        ----------
+        image
+            the image from which the regularization parameter maps should be estimated.
+        """
+        regularization_parameter_map = self.cnn_block(image)
+        regularization_parameter_map = (
+            torch.nn.functional.softplus(self.t, beta=self.beta_softplus) * regularization_parameter_map
```
fzimmermann89:
```suggestion
            torch.nn.functional.softplus(self.t, beta=10) * regularization_parameter_map
```

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,881 @@
...
+    some processing steps are necessary within the forward method. In particular, we restrict this example,
+    to be used with z=1.
+    """
+
+    def __init__(
+        self,
+        recon_matrix: mrpro.data.SpatialDimension[int],
+        encoding_matrix: mrpro.data.SpatialDimension[int],
+        lambda_map_network: ParameterMapNetwork2D,
+        n_iterations: int = 128,
```
fzimmermann89, line 150: If I remember correclty, even for the radial case Ttrain=Ttest=64  worked fine, maybe this would save some time?

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,881 @@
...
+        """
+        super().__init__()
+
+        self.recon_matrix = recon_matrix
+        self.encoding_matrix = encoding_matrix
+
+        self.lambda_map_network = lambda_map_network
+        self.n_iterations = n_iterations
+
+        finite_differences_operator = mrpro.operators.FiniteDifferenceOp(dim=(-2, -1), mode='forward')
```
fzimmermann89, line 174: side node: could you please chheck if forward actually is the forward finite difference?
Do we take into account that "conv" in pytorch is not a convolution but a correltion? could be that forward and backward are swapped.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,881 @@
...
+        gradient_operator = (
+            mrpro.operators.RearrangeOp('grad batch ... -> batch grad ... ') @ finite_differences_operator
+        )
+        self.gradient_operator = gradient_operator
```
fzimmermann89:
```suggestion
```

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,881 @@
...
+        super().__init__()
+
+        self.recon_matrix = recon_matrix
+        self.encoding_matrix = encoding_matrix
+
+        self.lambda_map_network = lambda_map_network
+        self.n_iterations = n_iterations
+
+        finite_differences_operator = mrpro.operators.FiniteDifferenceOp(dim=(-2, -1), mode='forward')
+        gradient_operator = (
```
fzimmermann89:
```suggestion
        self.gradient_operator = (
```

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,881 @@
...
+            mrpro.operators.RearrangeOp('grad batch ... -> batch grad ... ') @ finite_differences_operator
+        )
+        self.gradient_operator = gradient_operator
+
+        self.g = mrpro.operators.functionals.ZeroFunctional()
```
fzimmermann89, line 183: remove

fzimmermann89, line 178:
```suggestion
```

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,881 @@
...
+
+        # operator norm of the stacked operator K=[A, \nabla]^T
+        stacked_operator_norm = 3.0  # analytically calculated
+        self.primal_stepsize = self.dual_stepsize = 0.95 * (1.0 / stacked_operator_norm)
+
+    def estimate_lambda_map(self, image: torch.Tensor) -> torch.Tensor:
+        """Estimate regularization parameter map from image."""
+        # The initial image is a coil-combined image. Since we only consider 2D problems
+        # here, z can be assumed to be always 1.
```
fzimmermann89:
```suggestion
        Parameters
        ------------
        image
            coil combined 2d image with shape `batch, coils=1, z=1, y x`
        
```

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
fzimmermann89: please remove. Instead, the TV regualrized reconstruction can be used or the lambda network can be set to return a fixed value.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,881 @@
...
+
+        Returns
+        -------
+            Image reconstructed by the PDGH algorithm to solve the weighted TV problem.
+        """
+        # if no regularization parameter map is provided, compute it with the network
+        if regularization_parameter is None:
+            regularization_parameter = self.estimate_lambda_map(initial_image)
+
+        f1 = 0.5 * mrpro.operators.functionals.L2NormSquared(target=kdata)
```
fzimmermann89: name data_consistency and total_variation

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,881 @@
...
+        f2 = mrpro.operators.functionals.L1NormViewAsReal(weight=regularization_parameter)
+        f = mrpro.operators.ProximableFunctionalSeparableSum(f1, f2)
+
+        stacked_operator = mrpro.operators.LinearOperatorMatrix(((forward_operator,), (self.gradient_operator,)))
+
+        (solution,) = mrpro.algorithms.optimizers.pdhg(
+            f=f,
+            g=self.g,
```
fzimmermann89, line 234:
```suggestion
            g=None,
```

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
fzimmermann89: why does the code not just use a tensordataset from torch?

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,881 @@
...
+        forward_operator = fourier_operator
+
+    (kdata,) = forward_operator(target_image)
+    kdata = add_gaussian_noise(kdata, noise_variance=noise_variance, seed=seed)
+
+    kdata, target_image = normalize_kspace_data_and_image(kdata, target_image)  # type: ignore[assignment]
+
+    (adjoint_recon,) = forward_operator.H(kdata)
+
+    if n_coils > 1:
```
fzimmermann89: please remove the swithces for single coil and multi coil.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,881 @@
...
+
+import zenodo_get
+
+dataset = '15407890'
+
+tmp = tempfile.TemporaryDirectory()  # RAII, automatically cleaned up
+data_folder_tv = Path(tmp.name)
+zenodo_get.zenodo_get([dataset, '-r', 5, '-o', data_folder_tv])  # r: retries
+
+n_images = 28
```
fzimmermann89: please add a comment here

fzimmermann89: why is the trainingdataset reduced in size?
wouldnt it be better to use all the images in the dataset?
and use the last 8 images as validation, all others as training?

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,881 @@
...
+image_ = next(iter(dataloader_validation))[0]
+
+n_coils = 8
+acceleration_factor = 4.0
+noise_variance = 0.05
+
+kdata_, forward_operator_, adjoint_recon_, pseudo_inverse_solution_, image_ = prepare_data(
+    image_, n_coils=n_coils, acceleration_factor=acceleration_factor, noise_variance=noise_variance
+)
+if n_coils > 1:
```
fzimmermann89: please remove this switch.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,881 @@
...
+    n_iterations=n_iterations,
+)
+
+if n_coils > 1:
+    assert pseudo_inverse_solution_ is not None
+    input_image_ = pseudo_inverse_solution_
+else:
+    input_image_ = adjoint_recon_
+
+if torch.cuda.is_available():
```
fzimmermann89, line 489: can we move most of this into prepare data?

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,881 @@
...
+    adaptive_tv_network = adaptive_tv_network.cuda()
+    input_image_ = input_image_.cuda()
+    kdata_ = kdata_.cuda()
+    forward_operator_ = forward_operator_.cuda()
+
+with torch.no_grad():
```
fzimmermann89: lets remove the no_grad and instead detach and move to cpu the result here.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,881 @@
...
+# Let us now train the unrolled PDHG network. We set up the optimizer with an appropriate learning rate, choose a
+# number of epochs and write a simple training loop, in which we present the network input-target pairs and update
+# the network's parameters by gradient-descent.
+# To obtain a somewhat decent estimate of the network parameters, training should take approximately 5-6 minutes
+# on a GPU. If you have GPU and you need a break, start the training now, grab yourself a coffee and come back later
+# to see your network trained from scratch.
+# If you do not have a GPU, we provide a pre-trained model, which you can load and directly apply to the data.
+
+# %%
+# cnn_block = cast(torch.nn.Module, adaptive_tv_network.lambda_map_network.cnn_block)
```
fzimmermann89: please remove

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,881 @@
...
+# cnn_block =
+optimizer = torch.optim.Adam(
+    [
+        {'params': adaptive_tv_network.lambda_map_network.cnn_block.parameters(), 'lr': 1e-4},
+        {'params': [adaptive_tv_network.lambda_map_network.t], 'lr': 1e-2},
+    ],
+    weight_decay=1e-8,
```
fzimmermann89: does the tiny weight_decay here matter?
1e-8 is basically "no weight decay".
Either remove it or set it to something more reasonable like 1e-4

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,881 @@
...
+import datetime
+from time import time
+
+from tqdm import tqdm
+
+n_epochs = 10
+validation_loss_values = []
+best_model = None
+
+if not torch.cuda.is_available():
```
fzimmermann89: can we flatten this conditional to remove one level of indentation in the notebook?
set n_epoochs to zero if no gpu is availeble and load the weight-

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,881 @@
...
+    [
+        {'params': adaptive_tv_network.lambda_map_network.cnn_block.parameters(), 'lr': 1e-4},
+        {'params': [adaptive_tv_network.lambda_map_network.t], 'lr': 1e-2},
+    ],
+    weight_decay=1e-8,
+)
+
+lr_scheduler = torch.optim.lr_scheduler.ExponentialLR(optimizer, gamma=0.9)
+
+import copy
```
fzimmermann89, line 517: should this be a new cell?

can we give some text in between the code to have it in a similar style as the other notebooks please?

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,881 @@
...
+
+
+(coil_images,) = fourier_operator_undersampled.H(kdata_undersampled)
+
+csm_undersampled = mrpro.algorithms.csm.walsh(coil_images.squeeze(0), smoothing_width=5).unsqueeze(0)
+
+forward_operator_undersampled = fourier_operator_undersampled @ mrpro.operators.SensitivityOp(csm_undersampled)
+(adjoint_recon_undersampled,) = forward_operator_undersampled.H(kdata_undersampled)
+
+(pseudo_inverse_solution_undersampled,) = mrpro.algorithms.optimizers.cg(
```
fzimmermann89: please just call it initial_image or something.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
```diff
@@ -0,0 +1,881 @@
...
+# %% [markdown]
+# Additionally, we also perform a quick line search to see what the best possible TV-reconstruction using a
+# scalar regularization parameter would be. Note that in practice, you would obviously not be able
+# to obtain this reconstruction since the target image is not available. However, the comparison
+# is useful to assess how much improvement one can expect to obtain when employing
+# spatially adaptive regularization parameter maps.
+
+
+# %%
+def line_search(
```
fzimmermann89, line 621: please remove the linesearch from the notebook or at least hide it.
use TV regularized reconstruciton instead of the network in the linesearch.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
koflera: the new idea was now (also to flatten the condition below that was used to distinguish between the cases of having a GPU or not) to only train for a few epochs (e.g. 4) on one single image (i.e. n_images = 2, with 1 image for training and 1 for validation) and show that the error goes down. Then, we take the pre-trained model and apply it to the real data.

--> probably, n_images=2 will be used.

Does that sound good?

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
koflera: i'll use l2_norm_squared and l1_norm. data_consistency might be ok and correct, but total_variation is wrong, cause that would be ||\nabla x||_1, while here it's only ||x||_1.

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
koflera: the following code 

```
import mrpro
shape = mrpro.data.SpatialDimension(z=1, y=256, x=128)
image = mrpro.phantoms.EllipsePhantom().image_space(shape)
grad_op = mrpro.operators.FiniteDifferenceOp(dim = (-2,-1))
(gx,) = grad_op(image)
fig,ax = plt.subplots(1,3)
ax[0].imshow(image.abs().squeeze())
ax[1].imshow(gx[-1,...].abs().squeeze())
ax[2].imshow(gx[-2,...].abs().squeeze())
```

gives me the following image

grad_ops_results

so I guess it makes sense, cause it stacks the application of the derivative along the last dimension at the last dimension and proceeds in that that order.

koflera: also, now that I am re-reading your question, not sure I understand it :D I thought you meant the order of the application of the kernels.
if forward and backward (you mean the adjoint?) were swapped, it would not be possible to apply the currently implemented forward to an image, cause the mapping grad_2d has to be grad_2d : C^N --> C^{2N} (which seems to be case) and, accordingly, grad_2d ^T: C^{2N} --> C^N

fzimmermann89: We have forward finite difference, central and backward (different kernels) 
It think it would be correct, if conv actually was a convolution. But it isn't... So I think it might be switched

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
koflera: I think I have never done that, but (also for simplicity) I would not go for non-Cartesian here. Also, if non-Cartesian, one would need to compute the operator-norm in the forward if one wants to use different trajectories.
But given the new "story" of the notebook, we do not aim for a proper training anyway

examples/scripts/learning_regularization_parameter_maps_for_tv_reconstruction.py
fzimmermann89: The mode setting in your code.

koflera: ah, yes... yep, they seem to be swapped :)
```
import torch
import mrpro
x = torch.arange(0,5).view(1,1,1,1,-1)
grad_op_1d = mrpro.operators.FiniteDifferenceOp(dim = (-1,), mode="forward")
(gx_1d,) = grad_op_1d(x)
print(x)
print(gx_1d)
```

PR #701, opened 2025-03-06
Add Dataclass baseclass

### LINKED ISSUE #736: Reduce_repeat treats each dimension component separately
As far as I can tell, reduce_repeat looks at the z, y and x component of a SpatialDimension separately and does not treat them as spatial locations:

https://github.com/PTB-MR/mrpro/blob/e40bb6b115a99f801582a4e0aea3c30dd00dcea8/src/mrpro/data/IHeader.py#L64

fzimmermann89:
Instead of MovedataMixin, etc, introduce a Dataclass base class that implements moving to device and will implement indexing, etc.

Closes #744 
closes #736


fzimmermann89:
* #732
* #731
* #730
* #735
* #729
* #683
* __->__ #701

src/mrpro/data/EncodingLimits.py
```diff
@@ -33,7 +33,7 @@ def length(self) -> int:
         return self.max - self.min + 1
 
 
-@dataclass(slots=True)
+@dataclasses.dataclass
```
fzimmermann89: needs slots again

src/mrpro/data/KHeader.py
```diff
@@ -1,11 +1,9 @@
 """MR raw data / k-space data header dataclass."""
 
-from __future__ import annotations
-
 import dataclasses
-import datetime
+import datetime as dt
```
fzimmermann89, line 4: when did this happen?

fzimmermann89: we need this, otherwise it clashses with the datetime field. this will now causes issues because I removed to future annotation flag, i.e. type annotations are code and not strings.

src/mrpro/data/Dataclass.py
```diff
@@ -470,3 +501,49 @@ def is_cpu(self) -> bool:
...
+        Returns
+        -------
+            The broadcasted shape of all fields.
+        """
+        shapes = []
+        for _, data in self.items():
+            if not hasattr(data, 'shape'):
+                continue
+            shapes.append(data.shape)
+        return torch.broadcast_shapes(*shapes)
```
fzimmermann89: Ad a better exception here

src/mrpro/data/Dataclass.py
```diff
@@ -1,4 +1,4 @@
-"""MoveDataMixin."""
+"""Base class for all dataclasses in the `mrpro` package."""
```
ckolbPTB:
```suggestion
"""Base class for all dataclasses."""
```

src/mrpro/data/Dataclass.py
```diff
@@ -28,19 +28,117 @@ def __init__(self, *devices):
...
+    - a `apply` method to apply a function to all fields
+    - a `~Dataclass.clone` method to create a deep copy of the object
+    - `~Dataclass.to`, `~Dataclass.cpu`, `~Dataclass.cuda` merhods to move all tensor fields to a device
+
+    It is intended to be used as a base class for all dataclasses in the `mrpro` package.
+    """
 
     __dataclass_fields__: ClassVar[dict[str, dataclasses.Field[Any]]]
 
+    def __init_subclass__(cls, no_new_attributes: bool = True, *args, **kwargs):
```
ckolbPTB: doc string missing

src/mrpro/data/Dataclass.py
```diff
@@ -470,3 +501,49 @@ def is_cpu(self) -> bool:
...
+    @property
+    def dtype(self) -> torch.dtype | None:
+        """Return the dtype of the tensors.
+
+        Looks at each field of a dataclass implementing a dtype attribute,
+        such as `torch.Tensor` or `Dataclass` instances.
+        Returns the promoted dtype of all fields. If no field implements a dtype attribute,
+        `None` is returned.
+        """
+        dtype: torch.dtype | None = None
```
ckolbPTB: Just for my understanding - we would expect this to be None for SpatialDimension[int,float], right? Anywhere else?

src/mrpro/data/Dataclass.py
```diff
@@ -470,3 +501,49 @@ def is_cpu(self) -> bool:
...
+            if not hasattr(data, 'shape'):
+                continue
+            shapes.append(data.shape)
+        return torch.broadcast_shapes(*shapes)
+
+    @property
+    def dtype(self) -> torch.dtype | None:
+        """Return the dtype of the tensors.
+
+        Looks at each field of a dataclass implementing a dtype attribute,
```
ckolbPTB: I am not sure I fully undestand where this would be useful. If I call `kdata.dtype` - what does this tell me? See also comment in `__repr__`

src/mrpro/data/Dataclass.py
```diff
@@ -470,3 +501,49 @@ def is_cpu(self) -> bool:
...
+    # endregion Properties
+
+    def __repr__(self):
+        """Representation method for Dataclass."""
+        try:
+            device = str(self.device)
+        except RuntimeError:
+            device = 'mixed'
+        name = type(self).__name__
+        return f'{name} with (broadcasted) shape: {list(self.shape)!s} and dtype {self.dtype}\nDevice: {device}.'
```
ckolbPTB: For some random test data self.dtype yields complex128 even though the data is actually complex64.

tests/data/test_dataclass.py
```diff
@@ -1,10 +1,10 @@
-"""Tests the MoveDataMixin class."""
+"""Tests the move to device/dtype functionality of  dataclasses."""
```
ckolbPTB: If you would like this test file to only contain parts of the dataclass tests then we should probably rename the file.

fzimmermann89: I am still going back and forth wether to split it or not.
Will change the comment for now and leave the file name.

fzimmermann89:
Re dtype:
The half-baked idea was to have this as a fallback and implement for certain dataclasses a different logic.
The complex128 is caused by acq time being float64. 

Which also causes problems if used in signal models..

We could split it in a offset and a value?
so each acquisition uses a float32 value and we have an offset that describes the starting point if one want to combine different datasets...?

on the other hand, float32 precision is just 1.5e-7ish, so only  for 10minutes, the resolution  stays below 0.1ms..

If we go back to integers, we could do 0.1ms resolution (but how to name it such that it is not confusing with Siemens using 2.5ms)

We could also store as timestamps and have acq_time just as a property that returns the time with the offset removed? in 99% of the cases this should be what we need anyways. In the 1% that we need high resolution time deltas between datasets, we would then have to use the raw timestamp value...?


Or, we leave it at float64 and add a warning to the signal if used with float64 inversion time, for example?
And ignore the time stamp dtype in the dtype of acq info ;)

ckolbPTB:
> We could also store as timestamps and have acq_time just as a property that returns the time with the offset removed?

I think I like this solution best. Not sure how to define the offset though. Should it be simply the time of the first acquisition? This would mean that it changes if I take a subset and the first acquisition always has t=0. Or should it be set once when the data is read in and not modified even I remove acquisitions at the beginning. This would be useful for retrospective gating but probably not very intuitive.

fzimmermann89:
If we do this, i would suggest to use the time of first acquisition when reading in the data and keep the offset constant when indeing etc.

then, when reading in the data the conversion function could be set as "siemens with 124124124 as offset" 



or we keep it as 64bit and add a warning to the signal models if inversion_time is 64bit?

fzimmermann89:
I removed the dtype from the dataclasses and added it to spatialdimension.
Thats to only place we would actually need it.

I will keep the acq_time as float64. All alternatives are bad as well.
I think we should print a warning and/or  convert to float in the signal models?

ckolbPTB (approved):

PR #703, opened 2025-03-06
Reduce RAM usage in tests

fzimmermann89:
This should fix tests randomly failing in github actions.


lrlunin (approved):

fzimmermann89:
Oops. I should fix the test

lrlunin:
> Oops. I should fix the test

I expected this to be fixed in the separate PR since it's not an issue with the RAM.

PR #708, opened 2025-03-08
Unify naming of user and physiology parameters

### LINKED ISSUE #707: Inconsistency in AcqInfo parameters
For the AcqIdx user parameters we start with index 0:
https://github.com/PTB-MR/mrpro/blob/d87ab355c7fa79b3a4fdde87867c654a64f98c63/src/mrpro/data/AcqInfo.py#L92

for the UserValues and the physiology time stamps we start with index 1:
https://github.com/PTB-MR/mrpro/blob/d87ab355c7fa79b3a4fdde87867c654a64f98c63/src/mrpro/data/AcqInfo.py#L131

https://github.com/PTB-MR/mrpro/blob/d87ab355c7fa79b3a4fdde87867c654a64f98c63/src/mrpro/data/AcqInfo.py#L153

ckolbPTB:
closes #707

### DIFF
--- a/src/mrpro/data/AcqInfo.py
+++ b/src/mrpro/data/AcqInfo.py
@@ -128,6 +128,7 @@ class AcqIdx(MoveDataMixin):
 class UserValues(MoveDataMixin):
     """User Values used in AcqInfo."""
 
+    float0: torch.Tensor = field(default_factory=_float_factory)
     float1: torch.Tensor = field(default_factory=_float_factory)
     float2: torch.Tensor = field(default_factory=_float_factory)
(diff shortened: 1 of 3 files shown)


fzimmermann89 (approved):

ckolbPTB:
ISMRMRD is using `user_0` instead of `user0`. Do we want to also use the "underscore"-naming or better to adapt the ISMRMRD field names to our spelling?

fzimmermann89:
As long as we keep k0 and k1 etc,  I think  I don't mind either way

PR #709, opened 2025-03-08
Clean dependencies and add minimal versions testing

### LINKED ISSUE #648: Dependencies
- [x] change test, notebook to tests, notebooks
- [x] verify pip-tools and sigpy is still needed
- [ ] use include groups to add notebooks to docs and create dev which does everything

lrlunin:
Resolves #648


docker/install_dependencies_minimal.sh
```diff
@@ -0,0 +1,18 @@
...
+                                     pydicom==3.0.1 \
+                                     pypulseq==1.4.2 \
+                                     pytorch-finufft==0.1.0 \
+                                     cufinufft==2.3.1 \
+                                     scipy==1.12 \
+                                     ptwt==0.1.8 \
+                                     typing-extensions==4.12
+
+#clean up
+rm -rf /root/.cache
```
fzimmermann89: btw, why do we need this? what is saved there even with --no-cache-dir?

.github/workflows/pytest_min.yml
```diff
@@ -0,0 +1,38 @@
...
+        uses: actions/checkout@v4
+
+      - name: Install MRpro and Dependencies
+        # do not update dependencies
+        run: pip install .[tests]
+
+      - name: Install PyTest GitHub Annotation Plugin
+        run: pip install pytest-github-actions-annotate-failures
+
+      - name: Run PyTest and Generate Coverage Report
```
fzimmermann89: this is then never used again, right?
do we still want to create it?

docker/install_dependencies_minimal.sh
fzimmermann89, line 1: Suggestion: can we have this as a minimal-versions.txt in the requirements.txt format?
than we run pip install in the dockerfile

pyproject.toml
```diff
@@ -61,7 +61,7 @@ classifiers = [
 ]
 dependencies = [
     "numpy>=1.23, <3.0",
-    "torch>=2.3",
+    "torch>=2.3.1",
     "ismrmrd>=1.14.1",
     "einops",
```
fzimmermann89: we need at least 0.7.0 of einops I think

docker/install_dependencies_minimal.sh
lrlunin: image
Yes, possibly for some internal stuff.

fzimmermann89: i would still just remove this script and to a 
`python -m pip install --no-cache-dir -r minimal-requirements.txt && rm -rf /root/cache`

docker/Dockerfile_minimal
```diff
@@ -0,0 +1,21 @@
...
+ARG PYTHON="python3.10"
+
+# install python
+COPY docker/install_system.sh .
+RUN bash install_system.sh && rm install_system.sh
+
+# install mrpro dependencies
+# forces rebuild if either the version or the pyproject.toml changes
+COPY docker/minimal-requirements.txt docker/install_dependencies_minimal.sh ./
+RUN bash install_dependencies_minimal.sh && rm install_dependencies_minimal.sh
```
fzimmermann89: `python -m pip install --no-cache-dir -r minimal-requirements.txt && rm -rf /root/cache` ?

fzimmermann89 (commented):
is there a reason to have install_dependnecies_minimal.sh?

fzimmermann89 (approved):

fzimmermann89:
Feel free to merge

PR #712, opened 2025-03-10
added analytical 2D rad dcf

### LINKED ISSUE #154: Feature Request/Discussion: analytical DCF
Do we want to have analytical DCFs?
Where to put them?
Belongs  somewhere close to the trajectory calculation....

mxlutz:
I added an analytical DCF for 2D radial trajectories.

So far, I tried them with the direct_reconstruction example, might make sense to include it somewhere else and also to implement some tests.

Addresses Issue #154

### DIFF
--- a/examples/notebooks/direct_reconstruction.ipynb
+++ b/examples/notebooks/direct_reconstruction.ipynb
@@ -313,14 +313,11 @@
    "cell_type": "code",
    "execution_count": null,
    "id": "24",
-   "metadata": {
-    "lines_to_next_cell": 0
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "kykx = torch.stack((kdata.traj.ky[0, 0], kdata.traj.kx[0, 0]))\n",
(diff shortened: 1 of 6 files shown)


fzimmermann89:
I would leave the examples as-is.

The function should not be in dcf_voronoi, but in dcf_radial.py
you can import the dcf1d from voronoi.

Please also consider if the comments are really useful.
By default, try to have code that is self explanatory and avoid comments that describe what you are doing, such as
"# Compute the angle using atan2(vy, vx) instead of arccos"
instead, describe in a short sentence what the basic idea of the function is (either in docstring or as comment) and then only add comments to explain *why* you do things in a certain way if it is not obvious.

Can     `if not isinstance(center_idx, int):` ever be false?

it would be great to write some unit tests, see `tests` for some examples

mxlutz:
Thanks for the feedback Felix, I changed everything accordingly and implemented a simple test comparing the results to a 2D rad equidistant dcf

PR #714, opened 2025-03-13
cMRF example

rkcatarina:


fzimmermann89:
Please try to upload the data and mask from one scanner (unzipped) to zenodo.

Then, use the same code as in the other notebooks to download from zenodo.

Why is the reconstruction a function?

Do you think it is possible to add a bit more high level explanation between in code? Like how you would describe it in a thesis, for example. Less practical, programming, more "use extended phase graphs (reference).." if you know what I mean.


Could we add the results of the reference scans, like what the expected t1 and t2 values are for this phantom? I would hard-code the reference results. use the mask to obtain the values from the fingerprinting measurement. And then do a plot of fingerprinting over reference in the end.

This would also be a great opportunity to add a assertion, that the cmrf results match the pre-calculated reference values.
This way, if epg ever breaks, for example, this example would fail and we would notice.

fzimmermann89:
any updates here?

rkcatarina:
These are the modifications that I added to the example:
- High-level explanation at the beginning of the example and some more detailed explanation inbetween the code lines
- Comparison to hard-coded reference values (I chose to take the hard-coded values since then no additional reconstruction of the reference scans is needed)
- Download of the two color maps for the T1 and T2 maps
- Assertion at the end of the example

The following thing is still missing:
- Zenodo file of the data (in progress)
- I have to add some docstrings (and some other things) to make the example pre-commit-compatible

And I have some questions regarding the example:
- Should the open-source cMRF paper be referenced at the beginning of the example. If yes, how? Since it is not yet published.
- I am not sure if my "high-level"-explanation is ok, especially the part about the sliding window averaging and the signal averaging operator. Any suggestions are welcome:) 
- The reconstruction is a function, since it had already been like this in the example for the paper. Probably to make it "cleaner", or @ckolbPTB ? Should the reconstruction still be a function, since you asked this @fzimmermann89 ?

fzimmermann89 (dismissed):
i did some clean up.

please get the data on zenodo and update the dataset id.

fzimmermann89:
I fixed the text wrt the rough content.
What is missing is fixing the formatting -- currently the overview section uses rst style for the references and equations. This should be fixed to markdown. And some proof reading. 
@rkcatarina consider having a look how the rendered noteook looks in the documentation (see the link in the first comment above)

fzimmermann89 (approved):
fixed

PR #716, opened 2025-03-14
Add AveragingOp

fzimmermann89:
Calculates the avarage over a dimenion or over subets aong a dimension
Useful for qMRI if the images hve been reconstruced using a sliding windows.
Example cmrf (pseudocode):
````
kdata = kdata.split_k1_into_other(split_indices)
mrf_model = SignalAverage(split_indices) @ CardiacFingerprinting()
````

### DIFF
--- a/src/mrpro/operators/AveragingOp.py
+++ b/src/mrpro/operators/AveragingOp.py
@@ -18,7 +18,7 @@ class AveragingOp(LinearOperator):
     the average of the elements in that group is computed.
 
     For example, this operator can be used to simulate the effect of a sliding window average
-    in a signal model.
+    on a signal model.
     """
 
     def __init__(
--- a/src/mrpro/operators/AveragingOp.py
+++ b/src/mrpro/operators/AveragingOp.py
@@ -24,7 +24,7 @@ class AveragingOp(LinearOperator):
     def __init__(
         self,
         dim: int,
-        idx: Sequence[Sequence[int] | torch.Tensor | slice] | torch.Tensor,
+        idx: Sequence[Sequence[int] | torch.Tensor | slice] | torch.Tensor = (slice(None),),  # noqa: B008
         domain_size: int | None = None,
     ) -> None:
         """Initialize the averaging operator.
--- a/src/mrpro/operators/AveragingOp.py
+++ b/src/mrpro/operators/AveragingOp.py
@@ -1,0 +1,83 @@
+"""Averaging operator."""
+
+from collections.abc import Sequence
+from warnings import warn
+
+import torch
+
+from mrpro.operators.LinearOperator import LinearOperator
+
+
+class AveragingOp(LinearOperator):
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by fzimmermann89** [comment]
Not sure if this is the best abstraction - open for suggestions and docstring changes

We could have a reduction op supporting min, max, mean etc - but only mean is linear.

We could use the maximum index for the domain size if domain_sze has not been set (still ofc with a warning) - assuming that you want to use all data somewhere. But the cmrf reco already breaks that assumption.

File: src/mrpro/operators/AveragingOp.py [hunk 1]
```diff
@@ -0,0 +1,86 @@
...
+    """Averaging operator.
+
+    This operator averages the input tensor along a specified dimension.
+    The averaging is performed over groups of elements defined by the `idx` parameter.
+    The output tensor will have the same shape as the input tensor, except for the `dim` dimension,
+    which will have a size equal to the number of groups specified in `idx`. For each group,
+    the average of the elements in that group is computed.
+
+    For example, this operator can be used to simulate the effect of a sliding window average
+    in a signal model.
```
ckolbPTB:
```suggestion
    on a signal model.
```

src/mrpro/operators/AveragingOp.py
```diff
@@ -0,0 +1,86 @@
...
+
+        placeholder = (slice(None),) * (self.dim % x.ndim)
+        averaged = torch.stack([x[(*placeholder, i)].mean(self.dim) for i in self.idx], self.dim)
+        return (averaged,)
+
+    def adjoint(self, x: torch.Tensor) -> tuple[torch.Tensor]:
+        """Apply the adjoint of the averaging operator to the input tensor."""
+        if self.domain_size is None:
+            if self._last_domain_size is None:
+                raise ValueError('Domain size is not set. Please set it explicitly.')
```
ckolbPTB:
```suggestion
                raise ValueError('Domain size is not set. Please set it explicitly or run forward first.')
```

tests/operators/test_averagingop.py
```diff
@@ -0,0 +1,47 @@
...
+from tests import RandomGenerator
+from tests.helper import dotproduct_adjointness_test
+
+
+def test_averageingop_adjointness() -> None:
+    """Test the adjointness of the averaging operator."""
+    rng = RandomGenerator(seed=0)
+    u = rng.complex64_tensor(size=(5, 15, 10))
+    v = rng.complex64_tensor(size=(5, 3, 10))
+    idx = [0, 1, 2], slice(0, 6, 2), torch.tensor([-1, -2, -3])
```
ckolbPTB, line 16: could we also add the case for a boolean Tensor? sth like [False, True, False, True, True] which in `test_averageingop_forward` should be u[(1,3,4),].mean(dim=0), right?

ckolbPTB (approved):

PR #718, opened 2025-03-14
Remove inplace operations from prox and operator norm

### LINKED ISSUE #717: in place operation in prox_conv_conj and operator norm
In

https://github.com/PTB-MR/mrpro/blob/704fecdf1dd4e62ab27525c873c3407c2d0ad0a7/src/mrpro/operators/Functional.py#L172

and

https://github.com/PTB-MR/mrpro/blob/704fecdf1dd4e62ab27525c873c3407c2d0ad0a7/src/mrpro/operators/LinearOperator.py#L203

we have in-place operations that cause problems when trying to train our unrolled TV method.

koflera:
fixes #717

### DIFF
--- a/src/mrpro/operators/Functional.py
+++ b/src/mrpro/operators/Functional.py
@@ -169,7 +169,7 @@ class ProximableFunctional(Functional, ABC):
         if not isinstance(sigma, torch.Tensor):
             sigma = torch.as_tensor(1.0 * sigma)
         self._throw_if_negative_or_complex(sigma)
-        sigma[sigma < 1e-8] += 1e-6
+        sigma = torch.where(sigma < 1e-8, sigma + 1e-6, sigma)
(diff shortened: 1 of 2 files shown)


src/mrpro/operators/Functional.py
```diff
@@ -169,7 +169,7 @@ def prox_convex_conj(self, x: torch.Tensor, sigma: torch.Tensor | float = 1.0) -
         if not isinstance(sigma, torch.Tensor):
             sigma = torch.as_tensor(1.0 * sigma)
         self._throw_if_negative_or_complex(sigma)
-        sigma[sigma < 1e-8] += 1e-6
+        sigma = torch.where(sigma < 1e-8, sigma + 1e-6, sigma)
```
fzimmermann89:
```suggestion
        sigma = sigma.clamp(min=1e-8)
```
Would this work?

fzimmermann89 (approved):

PR #734, opened 2025-03-16
Epg2 no signal model

ckolbPTB:
Remove EPGSignalModel to allow for EPG models with different number of parameters

#733 should be merged first

### DIFF
--- a/src/mrpro/operators/models/EPG.py
+++ b/src/mrpro/operators/models/EPG.py
@@ -550,6 +550,38 @@ class DelayBlock(EPGBlock):
         """Duration of the block."""
         return self.delay_time
 
+class BlockSequence(torch.nn.ModuleList, EPGBlock):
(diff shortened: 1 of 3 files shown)


fzimmermann89:
What is the idea of EPGSequence and BlockSequence?

fzimmermann89:
block sequence should be deleted, right?

and dowe want to keep the MRF example in the EPGSequence?
Or should we have something actully published and maybe simpler?

ckolbPTB:
I would remove the cardiac MRF example from EPGSequences (if this was the question).

The idea of the BlockSequence was to have something where multiple Blocks are combined in a SuperBlock. This is very useful for timing calculations but not essential. Before this was done with EPGSequence but this is not possible anymore. Therefore I added a new class just for that.

fzimmermann89:
I want to keep the idea that an EPGSequence is just a block that can be combined into another EPGSequence (and get the duration of)
Let's see if I can get it to wok with states being int|torch.Tensor

PR #741, opened 2025-03-21
Fix missing time stamp converter bug

ckolbPTB:


fzimmermann89 (approved):

fzimmermann89:
Oops, I thought I fixed that one after you found it in the review.

PR #742, opened 2025-03-23
Fix non-contiguous data warning in NonUniformFourierOp

### LINKED ISSUE #739: Warning when using nufft operator on cuda
UserWarning: Argument `data` does not satisfy the following requirement: C. Copying array (this may reduce performance)
is raised from
https://github.com/flatironinstitute/finufft/blob/dc2931afd39b2cdd3c7a67570a03cc33fa7bc405/python/cufinufft/cufinufft/_plan.py#L380
due to the data not being contiguous.
we could add a .contiguous() (at least if cuda) or a warning filter to the inner nufft functions.
I think this copy might be unavoidable by our design. So either we do it for CPU and GPU tensors, we do it only for GPU tensors, we let finufft handle it and only suppress this warning.

bernharb:
Fixes #739


bernharb:
I did some more experimenting and found that the warning was triggered for CPU tensors as well so I removed the if clause checking the device

fzimmermann89 (approved):

PR #746, opened 2025-03-26
Add list/tuple tests for unit_conversion

ckolbPTB:


fzimmermann89:
I would either create the data per test or parameterize over tensor, list, tuple.
Either way, I would not have a "random_tensor_list_tuple" function...

PR #748, opened 2025-03-27
Fix memory leak in LinearOperator

### LINKED ISSUE #747: GPU memory allocation in for loop when using the FourierOperator
running this minimal example causes a CUDA out-of-memory error at some point. Notice that the optimizer is not even called here and that everything is within the `with torch.no_grad():`, so the error comes already by the repeatedly applying the adjoint of the fourier-operator.
Training

```
import torch
import mrpro

n_k0 = ny = nx = 160
n_k1 = 160
acceleration_factor = 4
k1_idx = torch.randperm(n_k0)[: int(n_k1 // acceleration_factor)]

traj = mrpro.data.traj_calculators.KTrajectoryCartesian()(
    n_k0=n_k0,
    k0_center=n_k0 // 2,
    k1_idx=k1_idx[:, None],
    k1_center=n_k1 // 2,
    k2_idx=torch.tensor(0),
    k2_center=0,
)
im_shape = mrpro.data.SpatialDimension(z=1, y=ny, x=nx)
k_shape = mrpro.data.SpatialDimension(z=1, y=n_k1, x=n_k0)

fourier_operator = mrpro.operators.FourierOp(
    traj=traj,
    recon_matrix=im_shape,
    encoding_matrix=k_shape,
)


class Network(torch.nn.Module):
    def __init__(
        self,
    ):
        super().__init__()
        self.parameter = torch.nn.Parameter(torch.tensor([1.0]), requires_grad=True)

    def forward(self, forward_operator: mrpro.operators.LinearOperator, kdata: torch.Tensor) -> torch.Tensor:
        (solution,) = forward_operator.H(kdata * self.parameter)

        return solution


network = Network()
optimizer = torch.optim.Adam(network.parameters(), lr=1e-4)

image = torch.randn(1, 1, 1, 160, 160, dtype=torch.complex64)
(kdata,) = fourier_operator(image)

if torch.cuda.is_available():
    network = network.cuda()
    image = image.cuda()
    kdata = kdata.cuda()
    fourier_operator = fourier_operator.cuda()

for _ in range(100000):
    print(_)
    # optimizer.zero_grad()
    with torch.no_grad():
        #     reconstruction = network(fourier_operator, kdata)
        (reconstruction,) = fourier_operator.H(kdata)

    # loss = torch.nn.functional.mse_loss(torch.view_as_real(reconstruction), torch.view_as_real(image))

    # loss.backward()
    # optimizer.step()
    # print(loss.item())


```

fzimmermann89:
Linear Operators that use `adjoint_as_backward` leaked memory in the order of the result size per call.

The `setup_context` of the AutogradWrapper returned the output. This leaks and is not required (https://pytorch.org/docs/stable/notes/extending.html)

Finding the issue took a lot longer than it should have.

 PS: I partially blame copilot:
image

I think we can now, after having switched to finufft, also disable the `adjoint_as_backward`  on the FourierOp.
I would still want to fix this functionality, as  it also allows do non-autograd compatible stuff in forward and adjoint.

Fixed #747

### DIFF
--- a/src/mrpro/operators/LinearOperator.py
+++ b/src/mrpro/operators/LinearOperator.py
@@ -35,8 +35,9 @@ class _AutogradWrapper(torch.autograd.Function):
         ctx: Any,  # noqa: ANN401
         inputs: tuple[Callable[[torch.Tensor], torch.Tensor], Callable[[torch.Tensor], torch.Tensor], torch.Tensor],
         _output: torch.Tensor,
-    ) -> torch.Tensor:
+    ) -> None:
         ctx.fw, ctx.bw, x = inputs
+        return None
 
     @staticmethod
     def backward(ctx: Any, *grad_output: torch.Tensor) -> tuple[None, None, torch.Tensor]:  # noqa: ANN401
--- a/src/mrpro/operators/LinearOperator.py
+++ b/src/mrpro/operators/LinearOperator.py
@@ -34,10 +34,9 @@ class _AutogradWrapper(torch.autograd.Function):
     def setup_context(
         ctx: Any,  # noqa: ANN401
         inputs: tuple[Callable[[torch.Tensor], torch.Tensor], Callable[[torch.Tensor], torch.Tensor], torch.Tensor],
-        output: torch.Tensor,
+        _output: torch.Tensor,
     ) -> torch.Tensor:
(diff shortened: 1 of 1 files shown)


koflera (approved):
this fixed the problem! thanks

PR #750, opened 2025-03-30
Change output format of sliding_window

fzimmermann89:
Previously, sliding_window added new dimensions in the back (as this was convenient for inati).
To be consistant with the other methods and our indexing, we should create the new dimensions in the front of the tensor.
This aöo ads a dilation argument to sliding_window
And switches to einops.einsum in inati, rreplacing the shape comments by einops strings.

PR #751, opened 2025-03-30
Add PatchOp

### LINKED ISSUE #692: Patch-extraction / recomposition operator for dictionary learning-based approaches
To be able to implement dictionary learning-based methods, it would be nice to have a "patch-extraction" / "patch-re-composition" operator as a `LinearOperator`.

At some point, I had found this:

https://github.com/f-dangel/unfoldNd?tab=readme-ov-file

which might be useful?

fzimmermann89:
Needs #750.
Closes #692 

for review, ignore all changes to sliding window and inati, just review patchop and test_patchop please.
edit: I changed the base to sliding_window for easier review, this should be changed to main before merge...


src/mrpro/operators/PatchOp.py
```diff
@@ -0,0 +1,120 @@
+"""Patch (sliding window) operator."""
```
koflera:
```suggestion
"""Patch extraction / reassembling (sliding window) operator."""
```

src/mrpro/operators/PatchOp.py
```diff
@@ -0,0 +1,120 @@
...
+class PatchOp(LinearOperator):
+    """Extracts N-dimensional patches using a sliding window view."""
+
+    def __init__(
+        self,
+        dim: Sequence[int] | int,
+        patch_size: Sequence[int] | int,
+        stride: Sequence[int] | int | None = None,
+        dilation: Sequence[int] | int = 1,
+        domain_size: int | Sequence[int] | None = None,
```
koflera, line 23: for some applications (e.g. the dictionary learning, that I have in mind), one would want to appropriately pad the tensor to be able to extract patches also beyond the border of the image and then use for example circular padding. By doing that, one can then achieve that `x=1/factor * R^T R x`, where factor depends on the specific configuration.

Should we include the possibility to pad and crop the tensor with a `padding` argument or should we leave that up to the user?

src/mrpro/operators/PatchOp.py
```diff
@@ -0,0 +1,120 @@
...
+            dilation=self.dilation,
+        )
+        patches = patches.flatten(start_dim=0, end_dim=len(self.dim) - 1)
+        return (patches,)
+
+    def adjoint(
+        self,
+        patches: torch.Tensor,
+    ) -> tuple[torch.Tensor,]:
+        """Perform the adjoint operation, i.e. assmable the patches."""
```
koflera: missing docstring

koflera: typo in "assemble"

koflera (approved):
Cool! Looks good to me
Looking forward to use that :)

src/mrpro/operators/PatchOp.py
fzimmermann89: we have a `ZeroPadOp`. I would rather change that one to a `PaddingOp `and add circular padding to it?

PR #752, opened 2025-03-31
Fix EPG on cuda

fzimmermann89:
Fixes the issue @bernharb notices with EPG/cMRF on cuda.

(I did not run the cuda tests after the last changes to the epg code -- time to get the self hosted runner action)


bernharb (approved):
passes all tests now

PR #753, opened 2025-03-31
Add convolution operator

fzimmermann89:
A draft of a nd convolution.


fzimmermann89:
this could use 
https://pypi.org/project/torchnd/ (https://github.com/fzimmermann89/torchnd/)

Issue #756, opened 2025-04-01
PDHG GPU issue


Stef-Martin:
When primal_stepsize and dual_stepsize are both None, the operator_norm is created on CPU even though initial_values are on GPU. Therefore, different device error arises if PDHG is supposed to be calculated on GPU.
I would just add a `.to(initial_values[0].device)` at the creation of operator_norm

![Image](https://github.com/user-attachments/assets/a48909ec-c0b2-4c1f-bc42-cd627bb2248a)

fzimmermann89:
The issue is in LinearOperatorMatrix.operator norm. 
instead of `torch.tensor(list_of_list_of_values)` is should use  ` torch.stack([torch.stack(inner) for inner in list_of_lists])`

PR #759, opened 2025-04-02
Fix wrong device in LinearOperatorMatrix.operator_norm

Stef-Martin:
Solves #756


fzimmermann89:
Something broke. why the view?

Instead, either stack in the last dimensions or change the max and sum in the next line to first axes.

Stef-Martin:
Yeah, I used the view because in my example norms then looks like this:
image
I'll take a look :)

fzimmermann89 (approved):

PR #764, opened 2025-04-03
Update adjoint_as_backward setting

fzimmermann89:
- Remove it from FourierOp -- we switched from torchkbnufft to finufft, thus no longer have the backward bugs.
- Add it to PatchOp -- Does not really matter for the forward, can save memory usage in the backward of the adjoint in certain cases (if the input is not already saved for the backward of another layer )

Lets see if tests pass.


fzimmermann89:
Spoke to early. finufft does not yet implement jvp. Should reopen in the future.

PR #766, opened 2025-04-04
Add import/export of dataclasses to hdf5-based mrp files

fzimmermann89:
Draft of a more efficient "mrpro native" data format matching our dataclass structure.
So instead of acquisitions as in ismrmrd, we write kdata.data, kdata.traj, etc as h5 groups.
This allows us to only store what we actually need in reduced form, speeds up loading for use in trainings and will allow us to store additional information not in ismrmrd.

The idea is to use this for temporary storage and caching. So one can load in all ismrmrd fileds required for a training, dump them to a temp drive and load randomly back into memory in a data loader..


fzimmermann89:
* (to be filled)

fzimmermann89:
@ckolbPTB I would like to add this to one of the example notebooks. Then if we break something and the mrp import does not work anymore with the old mrp file, we notice and would have to update the file on zenodo.

unit test here would then just do kdata->mrp>kdata roundtrip. 


 Maybe we can have a short notebook about indexing, rearranging, moving to cuda, importing and exporting data?
without any reconstruction etc, just basics?

PR #767, opened 2025-04-05
Fix callback and iteration logic in lbfgs

fzimmermann89:
#699 completly broke the iteration number setting of our lbfgs.

I added a comment to remind us what lbfgs.step does (more than a step!)

And a test checking if using the callback to stop after one iteration results in the same value as using one iteration.

This also unifies the optimizer arguments by changeing n_iterations in adam to max_iterations and adding a tolerance_change.

I made most of the optimizer arguments keyword only (there was another bug in one of the tests where the iteration stetting ended up in the learning rate... my mistake,)


src/mrpro/algorithms/optimizers/lbfgs.py
```diff
@@ -90,7 +97,7 @@ def lbfgs(
 
     Returns
     -------
-        List of optimized parameters.
+        List of optim\ized parameters.
```
ckolbPTB:
```suggestion
        List of optimized parameters.
```

ckolbPTB (approved):

PR #771, opened 2025-04-07
Allow matrix of operators in CG and add preconditioning

### LINKED ISSUE #544: preconditioning for CG
our CG could support preconditioning by another linear operator, see for the algorithm:

https://en.wikipedia.org/wiki/Conjugate_gradient_method#The_preconditioned_conjugate_gradient_method

and use case in MR:
https://github.com/mikgroup/kspace_precond/blob/master/Sense%20Reconstruction.ipynb

fzimmermann89:
As discussed with @ckolbPTB and @koflera , this breaks backward compatibility (@Stef-Martin and maybe @guastinimara):
CG now always returns a tuple of tensors (same as operators and the other optimizers)

if no x0 is given, it now starts with a zero tensor

- Enables use of LinearOperatorMatrix in cg.
- Adds preconditioning to cg
- Adds cuda test.

closes #544


src/mrpro/algorithms/optimizers/cg.py
```diff
@@ -43,25 +91,28 @@ def cg(
...
     ----------
     operator
-        self-adjoint operator (named :math:`H` above)
+        Self-adjoint operator :math:`H`.
     right_hand_side
-        right-hand-side of the system (named :math:`b` above)
+        Right-hand-side :math:`b`.
     initial_value
-        initial value of the algorithm; if `None`, it will be set to `right_hand_side`
+        Initial guess :math:`x_0`.
```
ckolbPTB: I would leave the comment about what None means.

src/mrpro/algorithms/optimizers/cg.py
```diff
@@ -74,54 +125,57 @@ def cg(
...
-    # for the case where the residual is exactly zero
-    if torch.vdot(residual.flatten(), residual.flatten()) == 0:
-        return solution
+    right_hand_side_ = (right_hand_side,) if isinstance(right_hand_side, torch.Tensor) else tuple(right_hand_side)
+    if initial_value is None:
+        solution = right_hand_side_
+    elif isinstance(initial_value, torch.Tensor):
+        solution = (initial_value,)
+    elif len(initial_value) != len(right_hand_side_):
+        raise ValueError('Length mismatch in initial_value and right_hand_side')
```
ckolbPTB: This check should also be run if initial_value is a Tensor, because right_hand_side could be a Sequence of tensors

src/mrpro/algorithms/optimizers/cg.py
```diff
@@ -74,54 +125,57 @@ def cg(
...
+    residual = tuple(rhs - op_sol for rhs, op_sol in zip(right_hand_side_, operator(*solution), strict=True))
+
+    if preconditioner_inverse is not None:
+        conjugate = preconditioner_inverse(*residual)
+    else:
+        conjugate = residual
 
     # dummy value. new value will be set in loop before first usage
-    residual_norm_squared_previous: torch.Tensor | None = None
+    direction_dot_residual_prev: torch.Tensor | None = None
```
ckolbPTB: previous?

src/mrpro/algorithms/optimizers/cg.py
```diff
@@ -74,54 +125,57 @@ def cg(
...
+        if preconditioner_inverse is not None:
+            direction = preconditioner_inverse(*residual)
+            direction_dot_residual = vdot(residual, direction).real
+        else:
+            direction = residual
+            direction_dot_residual = residual_dot_residual
+
+        if direction_dot_residual_prev is not None:  # not first iteration
+            beta = direction_dot_residual / direction_dot_residual_prev
+            conjugate = tuple(res + beta * con for res, con in zip(direction, conjugate, strict=True))
```
ckolbPTB: dir, con in zip(direction, conjugate)

fzimmermann89: dir would shadow dir(), so d

src/mrpro/algorithms/optimizers/cg.py
fzimmermann89: old?

ckolbPTB: sorry, I meant `direction_dot_residual_previous`

fzimmermann89: And I meant direction_dot_residual_old

src/mrpro/algorithms/optimizers/cg.py
fzimmermann89: yes, i also changed it to match scipy and jax

src/mrpro/algorithms/optimizers/cg.py
```diff
@@ -1,32 +1,79 @@
...
+    operator: LinearOperator | LinearOperatorMatrix | OperatorMatrixLikeCallable | OperatorLikeCallable,
+    right_hand_side: Sequence[torch.Tensor] | torch.Tensor,
+    *,
+    initial_value: Sequence[torch.Tensor] | torch.Tensor | None = None,
+    preconditioner_inverse: LinearOperator
+    | LinearOperatorMatrix
+    | OperatorMatrixLikeCallable
+    | OperatorLikeCallable
+    | None = None,
+    max_iterations: int = 128,
```
koflera, line 70: that sounds like many iterations, did we always use to have this many? I think probably 16 or 32 would be enough already in most cases to be used as the default?

src/mrpro/algorithms/optimizers/cg.py
```diff
@@ -1,32 +1,79 @@
...
+    tolerance: float = 1e-4,
+    callback: Callable[[CGStatus], bool | None] | None = None,
+) -> tuple[torch.Tensor, ...] | tuple[torch.Tensor]:
+    r"""(Preconditioned) Conjugate Gradient for solving :math:`Hx=b`.
+
+     This algorithm solves systems of the form :math:`H x = b`, where :math:`H` is a self-adjoint linear operator
```
koflera: maybe worth mentioning that this implementation also works for matrices containing linear operators and what exactly is meant by that, i.e. the interpretation is to have an operator H = diag(H_1, ..., H_N) with H_j self adjoint. This is what is intended to be done here, right? :D

(Also, maybe highlight the difference between working with a matrix of opeartors and the other case that we already had described in the docstrings?)

koflera (approved):
looks good to me

src/mrpro/algorithms/optimizers/cg.py
fzimmermann89: A linearoperatormatrix is mathematicaly speaking  just a linear operator H, correct?
We make no diagonal assumpion anywhere. And how exactly a linear operator matrix is applied is doumented in the linearoperator matrix..
I dont really see the need to add anything here? Maybe for the parameter docstring, we could call out and reference the MatrixOepratr class ?

koflera: Yeah, sorry, the comment was not quite clear.... I think I was just wondering for which matrices of linear operators one could meaningfully use the CG, given that the CG is supposed to be used for self-adjoint operators and the only example I could think of at first sight was block-diagonal matrices of containing self-adjoint operators in the diagonal.
But giving it a little bit more thought, there are also non-block-diagonal examples, i.e. when all the linear operators in the matrix are self-adjoint and the matrix of linear operators is block-symmetric. But I think these are all the matrices of linear operators that fulfill the self-adjoint property, i.e. "M is self-adjoint iff ... ".
Maybe we could include this in the docstring and point out that, as for the case with only the linear operator, it is up to the user to make sure that the matrix of linear operators fulfills the condition and that this is not checked by the CG.

koflera: which also means that we could maybe at least raise a warning or a value error if the block-symmetry is not fulfilled? is it possible to check whether two linear-operators are the same?

fzimmermann89:
So even in the simple case where we use
image
 to consructed a self adjoint operator
image
we don have a good way to check if to sums of linear operators are adjoint to each other.... Nor if a linear operator is selfadjoint... 
Nor if it is symmetric.....

fzimmermann89:
@koflera Another question: The docstring says "This implementation assumes H is self-adjoint"
Wouldnt it be better to say: "Conjugate gradient works only with self-adjoint positive semi-definite operators. This implementation does not verify these properties. The results will be wrong if the assumptions are not met"

src/mrpro/algorithms/optimizers/cg.py
fzimmermann89: we always had it at 128....

koflera:
> So even in the simple case where we use image to consructed a self adjoint operator image we don have a good way to check if to sums of linear operators are adjoint to each other.... Nor if a linear operator is selfadjoint... Nor if it is symmetric.....

No no, that's not what I meant.... what I meant is that a matrix of linear operators $M$ that is self adjoint (and thus for which it makes sense to use CG) has this somewhat general form
grafik
with i) all $A_j$ and $B_j$ being self adjoint and ii) $M_{ij} = M_{ji}$ for $i\neq j$. While we don't have a method for checking if an operator is self-adjoint (i), i thought that we could maybe check (ii) by checking whether the elements in the off-diagonals are equal. Therefore, I was asking, if we have the notion of two operators being "equal".

koflera:
> @koflera Another question: The docstring says "This implementation assumes H is self-adjoint" Wouldnt it be better to say: "Conjugate gradient works only with self-adjoint positive semi-definite operators. This implementation does not verify these properties. The results will be wrong if the assumptions are not met"

Yes, good point. That would be much clearer.

PR #773, opened 2025-04-07
Add (masked) SSIM

fzimmermann89:
Adds SSIM.


lrlunin:
Just an idea for tests since I have recently extended SSIM for the `torch-ignite` package:
https://github.com/pytorch/ignite/blob/master/ignite/metrics/ssim.py
we can also compare an `img` with `img + c`

src/mrpro/operators/functionals/SSIM.py
```diff
@@ -0,0 +1,182 @@
...
+        data_range: torch.Tensor | None = None,
+        window_size: int = 7,
+        k1: float = 0.01,
+        k2: float = 0.03,
+    ) -> None:
+        """Initialize Volume SSIM.
+
+        The Structural Similarity Index Measure [SSIM]_ is used to measure the similarity between two images.
+        It considers luminance, contrast and structure differences between the images.
+        SSIM values range from -1 to 1, where 1 indicates perfect structural similarity.
```
koflera, line 173: isn't it from 0 to 1?

src/mrpro/operators/functionals/SSIM.py
```diff
@@ -0,0 +1,182 @@
...
+
+import torch
+
+from mrpro.operators.Functional import Functional
+from mrpro.utils.sliding_window import sliding_window
+
+
+def ssim3d(
+    target: torch.Tensor,
+    x: torch.Tensor,
```
koflera: other variable name than x? image?

src/mrpro/operators/functionals/SSIM.py
```diff
@@ -0,0 +1,182 @@
...
+        Constant for SSIM computation. Commonly 0.03.
+    window_size
+        Size of the rectangular sliding window.
+        If any of the last 3 dimensions of the target is of size 1, the window size in this dimension will
+        also be set to 1.
+    data_range
+        Value range if the data. If `None`, peak-to-peak of the target will be used.
+    reduction
+        If True, return the mean SSIM over all pixels. If False, return the SSIM map.
+        The map will be of shape `(... z-window_size, y-window_size, x-window_size)`
```
koflera:
```suggestion
        The map will be of shape `(... z - window_size,  y - window_size, x - window_size)`
```
to avoid confusion... z-window_size could sound like z-direction instead of the "-" denoting a minus.

src/mrpro/operators/functionals/SSIM.py
```diff
@@ -0,0 +1,182 @@
...
+        return torch.where(mask, ssim_map, ssim_map[mask].mean())
+    else:  # reduction and mask is None
+        return ssim_map
+
+
+class SSIM(Functional):
+    """(masked) SSIM functional."""
+
+    def __init__(
+        self,
```
koflera, line 159: don't we also ALWAYS have a parameter "weight" for functionals? if yes, what is the interpretation here?

src/mrpro/operators/functionals/SSIM.py
```diff
@@ -0,0 +1,182 @@
...
+            raise ValueError('Mask does not cover any pixels')
+
+    else:
+        target, x = cast(tuple[torch.Tensor, torch.Tensor], torch.broadcast_tensors(target, x))
+
+    window = tuple(window_size if s > 1 else 1 for s in target.shape[-3:])  # To support 1D and 2D uses
+    target_window = sliding_window(target, window_shape=window, dim=(-3, -2, -1)).movedim((0, 1, 2), (-6, -5, -4))
+
+    if data_range is None and mask is not None:
+        data_range_ = target_window[mask].amax() - target_window[mask].amin()
```
koflera: what if target is for example a tensor with shape (10, 1, 16, 32, 32) corresponding to 10 3D volumes with highly different data_ranges? does it maybe make sense to compute the data_range depending on the max value of the dimensions for which the SSIM is to be calculated? i.e. in this example, we would have 10 different data_range.

src/mrpro/operators/functionals/SSIM.py
```diff
@@ -0,0 +1,182 @@
...
+    c2 = (k2 * data_range_) ** 2
+    a1 = 2 * mean_t * mean_x + c1
+    a2 = 2 * cov_tx + c2
+    b1 = mean_t**2 + mean_x**2 + c1
+    b2 = cov_t + cov_x + c2
+
+    ssim_map = (a1 * a2) / (b1 * b2)
+
+    if reduction and mask is not None:
+        return ssim_map[mask].mean()
```
koflera: similarly, here... wouldn't it be of interest to also be able to compute the mean only over certain dimensions? i.e. to be able to return, for example as above, for a tensor of shape (10, 1, 16, 32,32), 10 different values denoting the SSIM for the 10 different tensors?

src/mrpro/operators/functionals/SSIM.py
```diff
@@ -0,0 +1,182 @@
...
+        target
+            Target volume. At least 3d.
+        mask
+            Boolean mask. Only windows with all values `True` will be considered.
+            `None` means all windows are used.
+        data_range
+            Value range if the data. If None, the peak-to-peak of the target will be used.
+        window_size
+            Size of the windows used in SSIM. Usually 7 or 11.
+            Will be cropped to
```
koflera: will be cropped to ... ?

src/mrpro/operators/functionals/SSIM.py
```diff
@@ -0,0 +1,182 @@
...
+
+        Parameters
+        ----------
+        target
+            Target volume. At least 3d.
+        mask
+            Boolean mask. Only windows with all values `True` will be considered.
+            `None` means all windows are used.
+        data_range
+            Value range if the data. If None, the peak-to-peak of the target will be used.
```
koflera: of the data.

koflera:
```suggestion
            Value range if the data. If None, the max-to-min of the target will be used.
```

src/mrpro/operators/functionals/SSIM.py
```diff
@@ -0,0 +1,182 @@
...
+        self.mask = mask
+        self.k1 = k1
+        self.k2 = k2
+        self.window_size = window_size
+        self.data_range = data_range
+        if mask is not None and mask.dtype != torch.bool:
+            raise ValueError('Mask should be a boolean tensor')
+
+    def forward(self, x: torch.Tensor) -> tuple[torch.Tensor]:
+        """Calculate SSIM of an input."""
```
koflera, line 226: can we implement a default behaviour if we want to compare complex-valued tensors? e.g. by computing the SSIM as the average over the SIMM for real and imag part?

src/mrpro/operators/functionals/SSIM.py
```diff
@@ -0,0 +1,182 @@
...
+    reduction: bool = True,
+) -> torch.Tensor:
+    """Compute SSIM between two 3D volumes.
+
+    Parameters
+    ----------
+    target
+        Ground truth tensor, shape `(... z, y, x)`
+    x
+        Predicted tensor, same shape as gt
```
koflera:
```suggestion
        Predicted tensor, same shape as target.
```

src/mrpro/operators/functionals/SSIM.py
```diff
@@ -0,0 +1,182 @@
...
+    if data_range is None and mask is not None:
+        data_range_ = target_window[mask].amax() - target_window[mask].amin()
+    if data_range is None:
+        data_range_ = target_window.amax() - target_window.amin()
+    else:
+        data_range_ = data_range
+
+    x_window = sliding_window(x, window_shape=window, dim=(-3, -2, -1)).movedim((0, 1, 2), (-6, -5, -4))
+
+    mean_t = target_window.mean(dim=(-3, -2, -1))
```
koflera: consider renaming to mean_target, mean_image, etc to have more consistent and descriptive variable names?

tests/operators/functionals/test_ssim.py
```diff
@@ -0,0 +1,29 @@
...
+from mrpro.operators.functionals.SSIM import SSIM
+from mrpro.utils import RandomGenerator
+
+
+def test_ssim() -> None:
+    """Test the SSIM functional."""
+    rng = RandomGenerator(0)
+    target = rng.float32_tensor((1, 8, 32, 32), low=0.0, high=1.0)
+    ssim = SSIM(target)
+    perfect = target.clone()
```
koflera, line 13: if SSIM values indeed range from [-1,1], is then -target the image that yields SSIM(-target, target) = -1?

koflera (changes requested):
a few comments

src/mrpro/operators/functionals/SSIM.py
fzimmermann89: no, it is -1 to 1.
0 is no correlation.
-1 is perfect anti correlation

tests/operators/functionals/test_ssim.py
fzimmermann89: no, a strong negative value would be 
- same mean
- same variance
- different covaraince in each patch.

which is possible to construct, but not really a useful test, as it does not really cover any edge cases or potential in the ssim function. similar as with the addition of a scalar, etc.

koflera (approved):

PR #776, opened 2025-04-10
Move RandomGenerator to main

fzimmermann89:
Move RandomGenerator to main mrpro from test utils and add proper type hints and docstring.

Also updates some uses of the random generator (remove` getattr f'{dtype}_tensor'` uses)

unifies all (or most?) uses of `generator`, `random_generator`, and  `rng` to `rng`


fzimmermann89:
I cant select copilot as a review here, maybe this should run first..

ckolbPTB (approved):
Can ChatGPT create some tests for this?

fzimmermann89:
No, but grok could

PR #777, opened 2025-04-10
Add Poission variable density sampling Cartesian

fzimmermann89:
Adds a simple way to generate a batched random cartesian trajectory with undersampling

Not 100% satisfied with having at on a KTrajectoryCartesian named trajectory calculator.
But I think it is good enough for now.
(at some point I might want to rename and move the traj_calucluators)


src/mrpro/data/traj_calculators/KTrajectoryCartesian.py
```diff
@@ -53,5 +57,72 @@ def __call__(
...
         kz, ky, kx = unsqueeze_tensors_left(kz, ky, kx, ndim=5)
-
         return KTrajectory(kz, ky, kx)
+
+    @classmethod
+    def gaussian_variable_density(
+        cls,
+        encoding_matrix: SpatialDimension[int] | int,
+        acceleration: float = 2.0,
+        n_center: int = 10,
```
koflera, line 67: in the utils.RandomGenerator, this is parameter is called `always_sample`... stick to one of them for consistency?

koflera: ah ok, i see... it's similar, though not exactly the same

tests/data/test_traj_calculators.py
koflera, line 1: maybe include tests that show that, for appropriately sized tensors, setting acceleration_factor exactly yields the desired acceleration? e.g. for 128 lines, acceleration_factor= 4 yields 32-kspace lines etc...

koflera (changes requested):
very few comments.... looks good to me.

src/mrpro/data/traj_calculators/KTrajectoryCartesian.py
```diff
@@ -53,5 +57,72 @@ def __call__(
...
+            Larger values approach uniform sampling. Set to infinity for uniform sampling.
+        n_other
+            Batch size(s)
+        seed
+            Random seed for reproducibility.
+
+
+        Returns
+        -------
+            1D tensor of sorted selected k-space indices.
```
ckolbPTB: does not return Tensor

src/mrpro/data/traj_calculators/KTrajectoryCartesian.py
```diff
@@ -53,5 +57,72 @@ def __call__(
...
+
+        Raises
+        ------
+        ValueError
```
ckolbPTB, line 99: This error is not raised here

src/mrpro/data/traj_calculators/KTrajectoryCartesian.py
```diff
@@ -53,5 +57,72 @@ def __call__(
...
+            If a single integer, a square k-space is considered.
+        acceleration
+            Acceleration factor (undersampling rate).
+        n_center
+            Number of fully-sampled center lines to always include.
+        fwhm_ratio
+            Full-width at half-maximum of the Gaussian relative to encoding_matrix.y.
+            Larger values approach uniform sampling. Set to infinity for uniform sampling.
+        n_other
+            Batch size(s)
```
ckolbPTB: sampled separately for batch dimension?

src/mrpro/data/traj_calculators/KTrajectoryCartesian.py
fzimmermann89: yeah, it is different on purpose: in general, the "center" does not have to be in the center -- it is not even clear where the "center" would be, on randomgenerator, one can specify a range.

for the cartesian trajectory, this is always in the center to be independent of image size.

src/mrpro/data/traj_calculators/KTrajectoryCartesian.py
fzimmermann89: yes, otherwise the repeat removal would remove it anyways. added to docstring

tests/data/test_traj_calculators.py
fzimmermann89: this is already tested

src/mrpro/data/traj_calculators/KTrajectoryCartesian.py
fzimmermann89: the is moved in the underlying function and passed on. so even though its not directly raised, it will be raised when calling this function. so I think it makes sense to list it here?

ckolbPTB: Seems a bit arbitrary when we list errors of subfunctions and when not. I would remove it.

fzimmermann89: A) if it is an input error to the top level function that is just propagated, I would list it. Ignoring that most of our docstrings currently do not list the exceptions
B) to get a proper error message, I moved the check up. So this discussion is moot anyways ;)

koflera (approved):

PR #778, opened 2025-04-11
Add TSE and T1rho prep to EPG

ckolbPTB:
@rkcatarina could you please check that you get the same T2 values for this EPG-TSE model for one of your phantom scans?


rkcatarina:
I just checked the T2 values and they are not exactly the same. For one phantom scan reconstruction that I looked at, the highest difference between the two T2 maps (one created with the "old" and one created with the "new" EPG-TSE-model) was at around 278 ms for the same pixel. The mean difference for all pixels is at around 2.9 ms +- 5.4 ms for this reconstruction.

fzimmermann89:
now, the check the other way around: could you plot for this single worst pixel
a) the signal curve of the new model with the old result t2, m0, ... paramters
b) the signal cuve of the old model with the old result t2, m0 ... parameters
c) the signal curve of the new model with the new result t2, m0, ... parameters
d) the signal curve of the old model with the new result parameters 
e) the signal cuve of the pixel in the measured data (normalized to same m0 or same std)

rkcatarina:
EPG_comparison

fzimmermann89:
neiter the old nor the new model matchs the actua signal at all.. maybe just a bad pixel.

but for both pramters, old and new, the old and new model differ (seems like offset and scaling?)

did you figure out whre the difference comes from (were both run with the same parameter etc?) and which of those two is more correct?

ckolbPTB:
The difference between the output of the old and new model was due to the fact, that B1-scaling was used for one model but not the other. @rkcatarina repeated the comparison and for B1-scaling=1 the signal models match exactly, for B1-scaling != 1 there is a relative difference between the signals of 1e-7. This is probably due to the fact that in the new model we also scale the 90° excitation pulse which was not done in the old model.

fzimmermann89:
could we also add a gpu test?
I think the as_tensor + broadcasting could break if they are floats that are used with gpu tensors:

```
# works
torch.ones(2,2)*torch.as_tensor(1.).expand(2,2)

# works
torch.ones(2,2).cuda()*torch.as_tensor(1.)

# fails
torch.ones(2,2).cuda()*torch.as_tensor(1.).expand(2,2)
```

ckolbPTB:
Added some more tests. These CUDA tests pass but 
`model = EpgTseModel(refocusing_flip_angles.cuda(), refocusing_rf_phases=0.0, te=0.1)`
would fail with 
_RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!_

Do we want to add checks to ensure this also works, or should this be the problem of a higher order function such as the cMRF model?

fzimmermann89:
could solve this if we dont unsqueeze scalars?

ckolbPTB:
`model = EpgTseModel(refocusing_flip_angles.cuda(), refocusing_rf_phases=0.0, te=0.1)`

works already, see e.g. https://github.com/PTB-MR/mrpro/blob/5127a6f942a1c561c7f7080db37e4415f5581c27/tests/operators/models/test_epg.py#L433

fzimmermann89 (approved):
i think this was fine -- i was sure i had already approved it? (i havent looked again tbh)

PR #779, opened 2025-04-11
Add BiCGstab

fzimmermann89:
Add algorithm for solving Ax=b for *non-symmetric* matrices


fzimmermann89:
any news?

koflera:
> any news?

sorry, but I think on my priority list, this is a bit further down. As we do not really need it for the paper, I will postpone this to a later point ;)

PR #781, opened 2025-04-14
Create grid_sampler from grid/displacement components

### LINKED ISSUE #770: Components of GridSamplingOp
The grid provided to the GridSamplingOp is of shape `*batchdim, z,y,x,3` or `*batchdim, y,x,2`. The last dimension are the components of the grid, i.e. transformations along different spatial directions. 

Currently they need to be provided as `x,y,z` and `x,y`, respectively. This it the convention of torch. Would it make sense to change this to `z,y,x` as we use this convention everywhere else?

ckolbPTB:
Pytorch defines the components of the grid for grid_sampling as `x,y,z` and `x,y`, respectively. To avoid confusion, the individual components of the grid are not provided. Closes #770 

Allow to create grid_sampling from a displacement field (common output of image registration packages).


fzimmermann89:
quick questions: why affine_grid and not a meshgrid? the affine_matrix is always the unit matrix, correct?

ckolbPTB:
> quick questions: why affine_grid and not a meshgrid? the affine_matrix is always the unit matrix, correct?

Because depending on the value for `align_corners` different grids are created/needed. Seemed to be better to leave this to pytorch to figure out than reimplementing this logic.

fzimmermann89:
But does this matter?
How are the replacement fields defined?
I think align corners=True only really makes sense for an affine matrix , not a displacement field (and then, the settings just have to match between grid sample and affine grid)
?

ckolbPTB:
> But does this matter?

Depending on the `align_corners` setting we get different results. We can of course say, `from_displacement` does not have this option but then calculating the grid from meshgrid is probably going to be more lines of code than using affine_grid. I am not sure if `align_corners` will ever be useful for a displacement field but also does not hurt to support this option.

fzimmermann89:
I am still confused what align corners means for a displacement field.
I think the you observe is due to your grid logic only being correct for one of the settings.
In the other one (don't als me which way it is), you should scale by shape+-1 ...

fzimmermann89:
So I would do a mesh grid of the image shape (0 to shape exclusive)
Then the displacement.
And then scale down to -1..1 inclusive. (align corners True).

Or scale down to -1+step ..1-step inclusive and align corners false.

Or how is the displacement supposed to be scaled?

fzimmermann89:
Is it easy for you to apply the same displacement field once with mirtk (or where it comes from) and once in Mrpro?

Otherwise I would try it..

fzimmermann89:
Are the displacement fields forward or backward flow? Either way, we should mention it in the docstring.

ckolbPTB:
I compared our transformation to mirtk:
grafik

fzimmermann89:
so there is a difference, but it mostlz depends on intensitz, not on motion state?
and both use "pull" fields?
or what is your verdict?

ckolbPTB:
Yes, both algorithms lead to the almost identical result. The difference is due to the normalisation (for mirtk the images need to saved as uint16 nifti-files). Both approaches use forward transformations, which afaik are push transformations, so they describe where image information goes to.

fzimmermann89:
Grid saming in torch is a pull operation: you specify for each pixel of the output where to sample it in the input, or?

fzimmermann89:
Which is the easier way to implement it, otherwise you would need to invert the field or would create holes etc.. so maybe mirtk uses the same definition?

Could you add it somewhere to the docstrings?

fzimmermann89 (approved):

ckolbPTB:
> Could you add it somewhere to the docstrings?

I had another look and the displacement definitions are the same between mirtk and grid_sample - when I plotted I made a mistake transforming between mirtk/nifti and tensor orientation. . 

I added a comment explaining the pull operation.

fzimmermann89:
nice, thanks!

PR #784, opened 2025-04-15
Compare cart reco to dicom

### LINKED ISSUE #690: add note about separate/integrated calibration in cartesian reconstruction example
For the code to work as is, you need to select integrated calibration (@ckolbPTB what is the exact name?)
We could also add small code comment that if  using the other option, one must manually adjust the trajectory after reading in the data to obtain the correct phase

ckolbPTB:
Compares the MRpro reconstruction to the scanner dicoms. Also closes #690.

Needs #688 to compare orientation of images.

Missing: Example of how to rotate image from one orientation to another for comparison.

I also added the nifti data for this reconstruction, so we could also use it to evaluate the consistency of MRpro - dicom - nifti - orientation


fzimmermann89:
...Should the image rotation be something that is in mrpro? eiher as a util or on IData?

ckolbPTB:
> ...Should the image rotation be something that is in mrpro? eiher as a util or on IData?

yes, this would be a useful feature and would also allow us to calculate CSM from separate reference scans.

examples/scripts/cartesian_reconstruction.py
```diff
@@ -238,20 +238,19 @@ def show_images(*images: torch.Tensor, titles: list[str] | None = None) -> None:
...
-# But wait a second — something still looks a bit off. In the bottom left corner, it seems like there's a "hole"
-# in the brain. That definitely shouldn't be there.
+# But wait a second — what about all these nice receiver elements of our coil?
 #
-# The issue is that we combined the data from the different coils using a root-sum-of-squares approach.
-# While it's simple, it's not the ideal method. Typically, coil sensitivity maps are calculated to combine the data
-# from different coils. In MRpro, you can do this by calculating coil sensitivity data and then creating a
-# `~mrpro.operators.SensitivityOp` to combine the data after image reconstruction.
+# Here we used a root-sum-of-squares approach which is simple, but it's not the ideal method.
+# Typically, coil sensitivity maps are calculated to combine the data rom different coils. In MRpro, you can do this
```
koflera: typo 
```suggestion
# Typically, coil sensitivity maps are calculated to combine the data from different coils. In MRpro, you can do this
```

examples/scripts/cartesian_reconstruction.py
```diff
@@ -238,20 +238,19 @@ def show_images(*images: torch.Tensor, titles: list[str] | None = None) -> None:
 # %%
 
 fourier_op = mrpro.operators.FourierOp.from_kdata(kdata_pe_pf)
-# no need for and explicit CartesianSamplingOp anymore!
+# no need for an explicit CartesianSamplingOp anymore!
```
koflera, line 241: in
https://github.com/PTB-MR/mrpro/blob/3f8c467808948fc6426c10bb67a1e607292052f7/examples/scripts/cartesian_reconstruction.py#L9

a FFT--> an FFT

examples/scripts/cartesian_reconstruction.py
```diff
@@ -238,20 +238,19 @@ def show_images(*images: torch.Tensor, titles: list[str] | None = None) -> None:
...
 # That was easy!
-# But wait a second — something still looks a bit off. In the bottom left corner, it seems like there's a "hole"
-# in the brain. That definitely shouldn't be there.
+# But wait a second — what about all these nice receiver elements of our coil?
 #
-# The issue is that we combined the data from the different coils using a root-sum-of-squares approach.
-# While it's simple, it's not the ideal method. Typically, coil sensitivity maps are calculated to combine the data
-# from different coils. In MRpro, you can do this by calculating coil sensitivity data and then creating a
-# `~mrpro.operators.SensitivityOp` to combine the data after image reconstruction.
+# Here we used a root-sum-of-squares approach which is simple, but it's not the ideal method.
```
koflera:
```suggestion
# Here we used a simple root-sum-of-squares approach, which is not the ideal method.
```

examples/scripts/cartesian_reconstruction.py
```diff
@@ -366,6 +365,13 @@ def show_images(*images: torch.Tensor, titles: list[str] | None = None) -> None:
     *direct_recon_calib_lines.csm.data[0].abs().squeeze(),
     titles=[f'|CSM {i}|' for i in range(direct_recon_calib_lines.csm.data.size(-4))],
 )
+
+# %% [markdown]
+# ```{note}
+# There are already some other filter criteria available, see `mrpro.data.acq_filters`. You can also implement your own
+# function returning whether to include an acquisition.
```
koflera:
```suggestion
# function describing which acquisitions to include.
```
?

examples/scripts/cartesian_reconstruction.py
```diff
@@ -395,3 +401,25 @@ def show_images(*images: torch.Tensor, titles: list[str] | None = None) -> None:
 # This looks better! More information about the iterative SENSE reconstruction and its implementation in MRpro
 # can be found in the examples <project:iterative_sense_reconstruction_radial2D.ipynb> and
 # <project:iterative_sense_reconstruction_with_regularization.ipynb>.
+# ```{note}
+# In this example we used an "integrated" calibration scan, which means that the data used for calibration are acquired
```
koflera:
```suggestion
# In this example, we used an "integrated" calibration scan, which means that the data used for calibration are acquired
```

examples/scripts/cartesian_reconstruction.py
```diff
@@ -395,3 +401,25 @@ def show_images(*images: torch.Tensor, titles: list[str] | None = None) -> None:
...
+# with the same parameters (e.g. FOV and resolution) as the image data. There is also the option to use separately
+# acquired calibration lines. Most of the time they are acquired with a different resolution and hence have a different
```
koflera:
```suggestion
# acquired calibration lines. Usually, they are acquired with a different resolution and hence have a different
```

koflera (approved):
just a few comments

PR #785, opened 2025-04-15
Add a self-hosted runner

lrlunin:
The current state and design:

In the head node of slurm cluster the slurm_dispatcher.py will be started.
The dispatcher makes the GitHub API requests and asks for the Pytests selfhosted workflows in queued state (permitted by repo's approved members).
If there is one - inits a time limited CUDA required slurm session with a runner within the isolated singularity container in special (ephemeral) mode that the runner takes a single job and then deactivates. Therefore, there is no GPU waste time. The slurm job will be scheduled only for the run. This terminates after the given time or when the runner has finished its job.
The trigger to run this action is the `selfhosted run` comment in the PR (available since the workflow is in the `main` branch). Moreover, there would be a test report comment like done by `Pytest` (see .github/workflows/report_pytest.yml file)

I suppose it would be a good solution to avoid the unnecessary triggering. I would like to emphasize that the execution of the workflow would still require a confirmation from a repo's approved persons (see current environment groups and deploy pending).


.github/workflows/pytest.yml
```diff
@@ -21,8 +21,9 @@ jobs:
       - name: Retrieve Docker Image Names
         id: set-matrix
         run: |
-          # search for Dockerfile* in the docker directory, replace "Dockerfile" prefix with "mrpro" and to imagenames
-          imagenames=$(find docker -type f -name 'Dockerfile*' -exec basename {} \; | sed 's/^Dockerfile/mrpro/')
+          # search for Dockerfile_py* in the docker directory, replace "Dockerfile" prefix with "mrpro" and to imagenames
```
fzimmermann89:
```suggestion
          # search for Dockerfile_py* in the docker directory, replace "Dockerfile" prefix with "mrpro" for image name
```

.github/workflows/report_pytest.yml
```diff
@@ -4,6 +4,7 @@ on:
   workflow_run:
     workflows:
       - PyTest
+      - PyTest-selfhosted
```
fzimmermann89: I think I would remove the pytest-selfhosted from the reporting

.github/workflows/slurm_dispatcher.py
```diff
@@ -0,0 +1,139 @@
...
+    REGISTERED = 'REGISTERED'
+    DISPATCHED = 'DISPATCHED'
+    FINISHED = 'FINISHED'
+
+
+GITHUB_OWNER = os.environ.get('GITHUB_OWNER')
+GITHUB_REPOSITORY = os.environ.get('GITHUB_REPOSITORY')
+GITHUB_PERSONAL_TOKEN = os.environ.get('GITHUB_PERSONAL_TOKEN')
+# this workflow_id corresponds to the self-hosted pytest workflow
+WORKFLOW_ID = '132920098'
```
fzimmermann89: For consitency, we could either have all of them as envronment variables or none?
so either workflow_id becomes env var or the others get hard coded?

.github/workflows/slurm_dispatcher.py
```diff
@@ -0,0 +1,139 @@
...
+# all times in seconds
+
+# how long to wait for the job to be fetched as complete
+# since the github runner possibly ensures that the job is finished before turning off
+# this can be reduced to 0
+GITHUB_API_DELAY = 0.5 * 60
+# period to ask the GitHub API for new jobs
+REQUEST_DELAY_TIME = 1.5 * 60
+# if the GitHub API returns an error, wait for this time before retrying
+EXTENDED_DELAY_TIME = 10 * 60
```
fzimmermann89: use docstring style here:

```
GITHUB_API_DELAY = 0.5 * 60 # this can be reduced to 0
""" how long to wait for the job to be fetched as complete """

REQUEST_DELAY_TIME = 1.5 * 60
 """period to ask the GitHub API for new jobs"""

EXTENDED_DELAY_TIME = 10 * 60
""" if the GitHub API returns an error, wait for this time  before retrying"""
```
then hovering in vscode shows you this docstring

.github/workflows/slurm_dispatcher.py
```diff
@@ -0,0 +1,139 @@
...
+        logger.info(f'[run_id={run_id}]: stderr: {stderr.decode().strip()}')
+    await process.wait()
+    scheduled_runs[run_id]['state'] = JobState.FINISHED
+    logger.info(f'[run_id={run_id}]: finished, pid: {process.pid}')
+    logger.debug(f'[run_id={run_id}]: wait {GITHUB_API_DELAY}s')
+    await asyncio.sleep(GITHUB_API_DELAY)
+    scheduled_runs.pop(run_id)
+
+
+if None in (GITHUB_OWNER, GITHUB_REPOSITORY, GITHUB_PERSONAL_TOKEN):
```
fzimmermann89: move this up the where they are read in?

.github/workflows/slurm_dispatcher.py
```diff
@@ -0,0 +1,139 @@
...
+    'X-GitHub-Api-Version': '2022-11-28',
+}
+
+# check if the is any workflow for WORKFLOW_ID is queued
+runs_query = f'https://api.github.com/repos/{GITHUB_OWNER}/{GITHUB_REPOSITORY}/actions/workflows/{WORKFLOW_ID}/runs'
+params = {'status': 'queued'}
+
+# obtain a token to register a runner
+# the idea to do it here to isolate the GITHUB_PERSONAL_TOKEN from runner container
+runner_token_query = f'https://api.github.com/orgs/{GITHUB_OWNER}/actions/runners/registration-token'
```
fzimmermann89: if this is a constant, consider writing it in capital letters and use docstring comment

fzimmermann89:
Wont the actions runner that is started in the slurm job do multiple jobs?
I would change the logic: instead of starting on slurm job per open job, start a new slurm job if none is scheduled.
Most likely, as the jobs will take some time to atually start on the queue, the first runner will have finised all open jobs before the others start.
Or is the inner job runner actioally limited to only run one perticular task?

lrlunin:
> Wont the actions runner that is started in the slurm job do multiple jobs? I would change the logic: instead of starting on slurm job per open job, start a new slurm job if none is scheduled. Most likely, as the jobs will take some time to atually start on the queue, the first runner will have finised all open jobs before the others start. Or is the inner job runner actioally limited to only run one perticular task?

The parameter `--ephemeral` is responsible for it. If this is set - the runner will take exactly single job and then automatically terminate. If it's not set - then the runner would run until it is stopped.

https://github.com/PTB-MR/mrpro/blob/a05de7633f26e3c9559c69d2046d875f72076917/docker/runner_entrypoint.sh#L7-L15

I guess that `--ephemeral` mode is the optimal here. Firstly, we can easily estimate an upper limit of time required for the single job. There would be no time of GPUs wasted and no race conditions. Moreover, than the container is truly isolated and cannot be interfered from the previous runs.

Sure, there could be other solution when we start a runner for limited amount of time, than check if there were multiple jobs etc but the proper managements of the lifetime of the runner, existing jobs would be really hard to impelement correctly. For example, I have no idea how to avoid the case if the runner just dies due to the slurm task limited time in the middle of running job.

.github/workflows/report_pytest.yml
lrlunin: It would be look like this here:
https://github.com/lunin-inc/amazing-repo/pull/9#issuecomment-2700995850

Without a confusing badge but with the nice information about the runned tests, failed and skipped. I think it is pretty, clearly differs from the usual Pytest and make it more consistent and illustrative for contributors.
image

fzimmermann89: I would still remove it, as it is another comment that does not really add any useful information..

fzimmermann89:
Ah, got it. ephemeral makes sense.

Another questions: Do we need to build a singularity image before hand or can we singualrity pull a docker image in the slurm job?
( we would still need to find a way to get githu build docker images into the hpc registry or get hpc to allow us to pull from github...)

lrlunin:
> Another questions: Do we need to build a singularity image before hand or can we singualrity pull a docker image in the slurm job?

Actually, possibly we do not really need to do it. If `singularity run ... docker://...` would not take the time required for build each time we start a singularity session. Would it cache the pre-built image made from docker?

> ( we would still need to find a way to get githu build docker images into the hpc registry or get hpc to allow us to pull from github...)

In worst case we need to have a script which pulls a docker image from ghcr, applies tag for harbor and pushes it to harbor.

lrlunin:
> Actually, possibly we do not really need to do it. If `singularity run ... docker://...` would not take the time required for build each time we start a singularity session. Would it cache the pre-built image made from docker?

I have just checked this and `singularity exec` is indeed able to automatically build a `.sif` image for the first time the `docker://..` container is executed and then cache and reuse it. So it seems to be a good option. I'll check if this works in the "production" environment with any image from the harbor.

UPD: this also indeed works for the "production" environment and harbor docker registry. So the only question is how to get the ghcr image into the harbor registry without pull, apply tag, push (not a big deal though).

.github/workflows/.ruff.toml
```diff
@@ -0,0 +1,8 @@
+extend = "../../pyproject.toml"
+
+lint.extend-select = ["PT"] # flake8-pytest-style
+
+lint.extend-ignore = [
+    "RUF006", # store a reference to the return value of `asyncio.create_task`
+    "D100", # Missing docstring in public module
```
fzimmermann89: please either fix or only ignore for the dispatcher

.github/workflows/pytest_selfhosted.yml
```diff
@@ -0,0 +1,78 @@
...
+# Hint: use SINGULARITY_CACHE=... environmental variable for cache
+#
+# When build is ready you can launch this singularity container via:
+# NVIDIA_VISIBLE_DEVICES=<GPU_ID> singularity run --nv --env-file runner.env --pwd /actions-runner --writable-tmpfs --contain mrpro-runner.sif
+#
+# If you want to run the singularity container in ephemeral mode (one runner - one job) you need to specify a unique id for each runner
+# as an extra argument to the singularity command:
+# NVIDIA_VISIBLE_DEVICES=<GPU_ID> singularity exec --nv --env-file runner.env --pwd /actions-runner --writable-tmpfs --contain mrpro-runner.sif /actions-runner/entrypoint.sh "RUNNER_TOKEN" "RUN_ID"
+#
+# Use the ./github/slurm_dispatcher.py to dispatch each job to the isolated singularity container
```
fzimmermann89: add update instuctrions

.github/workflows/slurm_dispatcher.py
fzimmermann89, line 1: please remove

.github/workflows/pytest_selfhosted.yml
```diff
@@ -0,0 +1,78 @@
...
+      - main
+
+jobs:
+  test:
+    name: Run Tests and Coverage Report
+    if: >-
+      ${{ (github.event_name == 'push' && github.ref == 'refs/heads/main') ||
+          (github.event_name == 'pull_request' && github.event.label.name == 'pytest-selfhosted') }}
+    runs-on: [self-hosted] #gpu
+    environment: pytest-cuda
```
fzimmermann89, line 46: run in merge to main without the environment

.github/workflows/pytest.yml
```diff
@@ -21,8 +21,9 @@ jobs:
       - name: Retrieve Docker Image Names
         id: set-matrix
         run: |
-          # search for Dockerfile* in the docker directory, replace "Dockerfile" prefix with "mrpro" and to imagenames
-          imagenames=$(find docker -type f -name 'Dockerfile*' -exec basename {} \; | sed 's/^Dockerfile/mrpro/')
+          # search for Dockerfile_py* in the docker directory, replace "Dockerfile" prefix with "mrpro" for image name
+          # ! with the Dockerfile_py* mask we omit Dockerfile_runner !
+          imagenames=$(find docker -type f -name 'Dockerfile_py*' -exec basename {} \; | sed 's/^Dockerfile/mrpro/')
```
lrlunin: Should be in other PR where the problem with the pytest executing `mrpro_mininmal`.

fzimmermann89:
lets merge #850 first (to avoid running the selfhosted docker file), then this. 

its ready, correct?

fzimmermann89 (approved):

docker/runner_entrypoint.sh
```diff
@@ -0,0 +1,35 @@
...
+if [ -z ${1+x} ]; then
+    echo "The runner registration token is missing" 1>&2
+    exit 1
+fi
+
+# the 2nd argument must be the github organization
+# for example PTB-MR
+
+# if the 3d argument for the script is given: run in ephemeral mode (one runner takes one job)
+# constructing the unique name for the runner (see .github/slurm_dispatcher.py)
```
lrlunin: update this

PR #788, opened 2025-04-23
Allow complex inputs to ConstraintOp and add InverseConstraintOp

fzimmermann89:
ConstraintOp failed with complex inputs.

Now constraints are interpreted as constrains on real and imaginary components,
thus constraining to a rectangle in the complex plane.

cleans up the operator and tests a bit.

Adds an inverse property to get the inverse of an constraint operator as an operator (usefull for chaining operators)

### DIFF
--- a/src/mrpro/operators/ConstraintsOp.py
+++ b/src/mrpro/operators/ConstraintsOp.py
@@ -8,6 +8,26 @@ import torch.nn.functional as F  # noqa: N812
 from mrpro.operators.EndomorphOperator import EndomorphOperator, endomorph
 
 
+def sigmoid(x: torch.Tensor, beta: float = 1.0) -> torch.Tensor:
+    """Beta scaled sigmoid function."""
+    return F.sigmoid(beta * x)
+
+
+def sigmoid_inverse(x: torch.Tensor, beta: float = 1.0) -> torch.Tensor:
+    """Inverse of `sigmoid`."""
+    return torch.logit(x) / beta
+
+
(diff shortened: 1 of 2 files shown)


src/mrpro/operators/ConstraintsOp.py
```diff
@@ -50,39 +82,64 @@ def __init__(
...
-
-    @staticmethod
-    def softplus_inverse(x: torch.Tensor, beta: float = 1.0) -> torch.Tensor:
-        """Inverse of `softplus_transformation`."""
-        return x + torch.log(-torch.expm1(-beta * x)) / beta
+        self.lower_bounds = tuple(
+            torch.as_tensor(bound[0]) if bound[0] is not None else -torch.tensor(torch.inf) for bound in bounds
+        )
+        self.upper_bounds = tuple(
+            torch.as_tensor(bound[1]) if bound[1] is not None else torch.tensor(torch.inf) for bound in bounds
```
lrlunin: would `tuple(torch.as_tensor(bound[1] if bound[1] is not None else torch.inf) for bound in bounds)` be a little bit shorter without 2nd `torch.tensor(...)`

fzimmermann89: torch.inf is float, not tensor....

fzimmermann89: ups. misread.

fzimmermann89:
```suggestion
            torch.as_tensor(torch.inf if ub is None else ub) for (lb,ub) in bounds
```

lrlunin (approved):
🚀

PR #738, opened 2025-03-20
Calculate csm on low-res image

ckolbPTB:

PR #731, opened 2025-03-15
Add rearrange, split, concatenate to dataclasses

fzimmermann89:
Rearrange, split and concatenate for dataclasses.
Also adds an `__eq__ `to dataclass and Rotation

PR #790, opened 2025-04-26
Example for MCIR

### LINKED ISSUE #164: Motion corrected image reconstruction
We would like to be able to do motion-corrected image reconstruction, i.e. min ||FMx - y|| with a Fourier operator F and a motion operator M. 

Idea: M is initialised with the motion fields and information which acquisition index (https://github.com/PTB-MR/mrpro/blob/80577fa6c0a994b7fa7c4ce2306e7985b5407997/src/mrpro/data/_AcqInfo.py#L27) encodes the different motion states (default: phase). 

The Fourier operator returns k-space data of shape (other, coils, 1, 1, k0) where the acquisition index for the motion states is part of other.

ckolbPTB:
Still needs:

- #789 
- #781 
- #596 
- #731 
- #738 

Closes #164


ckolbPTB:
This is now ready to review but I am wondering if we should remove the TV-part. For the tests we cannot run TV because it takes too long. I added it mainly for the paper. Opinions?

fzimmermann89:
would it be ok to remove the tv part completly instead of hiding it behind a switch?

fzimmermann89:
> This is now ready to review but I am wondering if we should remove the TV-part. For the tests we cannot run TV because it takes too long. I added it mainly for the paper. Opinions?

agree.

examples/scripts/motion_corrected_reconstruction.py
```diff
@@ -0,0 +1,283 @@
+# %% [markdown]
+# # Motion-corrected image reconstruction
+# Physiological motion (e.g. due to breathing or the beating of the heart) is a common challenge for MRI. Motion during
+# the data acquisition can lead to motion artifacts in the reconstructed image. Mathematically speaking, the motion
+# occurring during data acquisition can be described by linear operators $M_m$ for each motion state $m$:
+#
+# $ y = \sum_m A_m M_m x_{\mathrm{true}} + n, $
```
fzimmermann89: are you sure this is a sum here, not a stacking?

examples/scripts/motion_corrected_reconstruction.py
```diff
@@ -0,0 +1,283 @@
...
+# 4. Estimate non-rigid motion fields from the dynamic images
+# 5. Use the motion fields to obtain a motion-corrected image
+#
+# To achieve high image quality for the different respiratory motion states, a TV-regularized image reconstruction
+# should be used. To safe time we will skip this per default and only use the iterative SENSE reconstruction.
+# The TV-regularized reconstruction can be activated by setting `n_iterations_tv` to a value larger than 0 (100  yields
+# good results).
+
+# %%
+n_iterations_tv = 0
```
fzimmermann89: remove tv

examples/scripts/motion_corrected_reconstruction.py
```diff
@@ -0,0 +1,283 @@
...
+# - Select the SVD component the largest frequency contribution between 0.2 and 0.5 Hz (realistic breathing frequency)
+
+
+# %% tags=["hide-cell"] mystnb={"code_prompt_show": "Show self-navigator calculation"}
+def get_respiratory_self_navigator_from_grpe(
+    kdata: KData, respiratory_frequency_range: tuple[float, float] = (0.2, 0.5)
+) -> torch.Tensor:
+    """Get respiratory self-navigator from GRPE data set."""
+    # Get all readout lines for ky = kz = 0
+    ky0_kz0_idx = torch.where((kdata.traj.ky == 0) & (kdata.traj.kz == 0))
```
fzimmermann89: you can index kdata with boolean masks directly, 
kdata[(kdata.traj.ky == 0) & (kdata.traj.kz == 0)] should work

examples/scripts/motion_corrected_reconstruction.py
```diff
@@ -0,0 +1,283 @@
...
+    navigator_data = kdata.data[ky0_kz0_idx[0], :, ky0_kz0_idx[2], ky0_kz0_idx[3], :]
+    navigator_time = kdata.header.acq_info.acquisition_time_stamp[ky0_kz0_idx[0], 0, ky0_kz0_idx[2], ky0_kz0_idx[3], 0]
+
+    # Apply a 1D FFT along the readout to get the projection of the object
+    fft_op_1d = FastFourierOp(dim=(-1,))
+    navigator_data = torch.abs(fft_op_1d(navigator_data)[0])
```
fzimmermann89: why abs here? doesnt motion change phase more than magnitude? why not keeping it complex`?

examples/scripts/motion_corrected_reconstruction.py
```diff
@@ -0,0 +1,283 @@
...
+
+    # Carry out SVD over all readout points and all coils to get the main signal components
+    navigator_data = rearrange(navigator_data, 'k1k2 coil k0 -> k1k2 (coil k0)')
```
fzimmermann89: again, why not do this on  kdata?
indexing, reahsping etc should also index reshapoe etc data, time_stamps,...

examples/scripts/motion_corrected_reconstruction.py
```diff
@@ -0,0 +1,283 @@
...
+    fft_svd_navigator_data_in_resp_window = fft_svd_navigator_data[
+        (f_hz >= respiratory_frequency_range[0]) & (f_hz <= respiratory_frequency_range[1]), :
+    ]
+    respiratory_navigator = svd_navigator_data[
+        :, torch.argmax(torch.max(fft_svd_navigator_data_in_resp_window, dim=0)[0])
+    ]
+
+    # Interpolate navigator from k-space center ky=kz=0 to all phase encoding points
+    return torch.as_tensor(
+        np.interp(kdata.header.acq_info.acquisition_time_stamp[0, 0, 0, :, 0], navigator_time, respiratory_navigator)
```
fzimmermann89: ... we should add a torch interpolate function to utils. i think we do something similiar also somewhere else iirc...

examples/scripts/motion_corrected_reconstruction.py
```diff
@@ -0,0 +1,283 @@
...
+plt.xlabel('Acquisition time (s)')
+plt.ylabel('Navigator signal (a.u.)')
+
+# %% [markdown]
+# ### 2. Split data into different breathing states
+# The self-navigator is used to split the data into different motion phases. We use a sliding window approach to ensure
+# we have got enough data in each motion state.
+
+# %%
+n_points_per_motion_state = int(kdata.data.shape[-2] * 0.36)
```
fzimmermann89: why not kdata.shape?

fzimmermann89: can you write why 0.36? in the markdocwn or comment? or name it as a constant?

examples/scripts/motion_corrected_reconstruction.py
```diff
@@ -0,0 +1,283 @@
...
+# ### 4. Estimate the motion fields from the dynamic images
+# The motion fields can be estimated with an image registration package such as
+# [mirtk](https://mirtk.github.io/commands/register.html). Here we registered each of the dynamic respiratory phases to
+# the first motion state using the free-form deformation (FFD) registration approach with a control point spacing of 9.
+# Normalized mutual information was used as a similarity metric and a LogJac penalty weight of 0.001 was applied.
+#
+# We load the displacement fields and create a motion operator:
+
+# %%
+mf = torch.as_tensor(np.load(data_folder / 'grpe_t1_free_breathing_displacement_fields.npy'), dtype=torch.float32)
```
fzimmermann89, line 226: motion_field

examples/scripts/motion_corrected_reconstruction.py
```diff
@@ -0,0 +1,283 @@
...
+motion_op = GridSamplingOp.from_displacement(mf[..., 2], mf[..., 1], mf[..., 0])
+
+# %% [markdown]
+# ### 5. Use the motion fields to obtain a motion-corrected image
+# Now we obtain a motion-corrected image $x$ by minimizing the functionl $F$
+#
+# $ F(x) = ||\sum_m(A_mx - y_m)||_2^2 \quad$ with $\quad A_m = F_m M_m C $
```
fzimmermann89: i think this is wrong: 
this would mean errors in one motion state cancel the errors in another? i think sum should be outside abs

examples/scripts/motion_corrected_reconstruction.py
```diff
@@ -0,0 +1,283 @@
...
+# Create acquisition operator
+fourier_op = recon_resp_resolved.fourier_op
+csm_op = SensitivityOp(csm_maps)
+averaging_op = AveragingOp(dim=0)
+acquisition_operator = fourier_op @ motion_op @ csm_op @ averaging_op.H
+
+if recon_resp_resolved.dcf is None:
+    (initial_value,) = acquisition_operator.H(kdata_resp_resolved.data)
+else:
+    dcf_op = recon_resp_resolved.dcf.as_operator()
```
fzimmermann89: i think this now changed after #968 and there is dcf_op on the reconstruction

fzimmermann89: also, i dont think this actually helps as is: 
the adjoint with dcf is usually not correctly normalized compared to the result of iterative reconstrucion
look at how its done in iterative sense reconstruction to get a good starting point

fzimmermann89 (changes requested):
see comments

examples/scripts/motion_corrected_reconstruction.py
fzimmermann89: also, why is this optional?
i would  already specialize this here to the test data.


`image_views = [image[:, 61, :], torch.fliplr(image[:, :, 63]), image[54, :, :]]` is also already specialized.

examples/scripts/motion_corrected_reconstruction.py
```diff
@@ -0,0 +1,283 @@
...
+import matplotlib.pyplot as plt
+
+
+def show_views(*images: torch.Tensor, ylabels: Sequence[str] | None = None) -> None:
+    """Plot coronal, transversal and sagittal view."""
+    if ylabels is not None and len(ylabels) != len(images):
+        raise ValueError(f'Expected {len(images)} ylabels, got {len(ylabels)}')
+    _, axes = plt.subplots(len(images), 3, squeeze=False, figsize=(12, 4 * len(images)))
+    for idx, (image, ylabel) in enumerate(zip(images, ylabels or [''] * len(images), strict=True)):
+        image = torch.squeeze(image / image.flatten().sort()[0][int(image.numel() * 0.98)])
```
fzimmermann89: why not quantile instead of sorting?

examples/scripts/motion_corrected_reconstruction.py
ckolbPTB: yes, should be stacking

examples/scripts/motion_corrected_reconstruction.py
ckolbPTB: yes, you are right

examples/scripts/motion_corrected_reconstruction.py
ckolbPTB: after the 1D FFT we are in image space, not sure how much motion information is in the phase. for the gating we need a real-valued signal. taking the abs here is imo easier to interpret than taking the abs of the svd component.

examples/scripts/motion_corrected_reconstruction.py
ckolbPTB: true, but afterwards I don't need any of the header information anymore

fzimmermann89: Ah. Sorry. This is image space already. Ignore this comment for navigator_data

examples/scripts/motion_corrected_reconstruction.py
ckolbPTB: which part would you rename "motion field"?

fzimmermann89: Doesn't mf mean motion field?

fzimmermann89: But i see whyyou abbreviated - to keep next line short. Ok

examples/scripts/motion_corrected_reconstruction.py
```diff
@@ -0,0 +1,274 @@
...
+    navigator_data = fft_op_1d(kdata_navigator.data)[0].squeeze().abs()
+
+    # Carry out SVD over all readout points and all coils to get the main signal components
+    navigator_data = rearrange(navigator_data, 'coil k1k2 k0 -> k1k2 (coil k0)')
+    svd_navigator_data, _, _ = torch.linalg.svd(navigator_data - navigator_data.mean(dim=0, keepdim=True))
+
+    # Select the SVD component the largest frequency contribution closest to the expected respiratory frequency
+    dt = torch.mean(torch.diff(kdata_navigator.header.acq_info.acquisition_time_stamp.squeeze(), dim=0))
+    fft_svd_navigator_data = torch.abs(torch.fft.fft(svd_navigator_data, dim=0))
+    f_hz = torch.linspace(0, 1 / dt, svd_navigator_data.shape[0])
```
fzimmermann89:
```suggestion
    f_hz = torch.fft.fftfreq(len(svd_navigator_data), dt)
```

fzimmermann89 (approved):
consider using utils.interpolate.interp #983  (once you reviewed it...) instead of np.interp
and maybe fftfreq

examples/scripts/motion_corrected_reconstruction.py
fzimmermann89: (maybe, clearer what is calculated)

fzimmermann89:
I did a small cleanup [5132aeb](https://github.com/PTB-MR/mrpro/pull/790/commits/5132aeb937e63a709faf192f3798a48f03e66329)  pass:
- switch to interp (lets see if that works)
- use same import style as the other examples
- prefer .abs(), .mean() etc over torch.abs(), torch.mean() in some places
- typos

@ckolbPTB is that ok with you?

fzimmermann89:
I added two small fixes to interp that I missed in #983.  I am considering to have them as a seperate PR... ?

ckolbPTB:
@fzimmermann89 great, thanks! Feel free to open a separate PR for the interp changes or just merge it here. Both is fine by me.

fzimmermann89:
Then I am lazy and we keep the interp changes here

PR #795, opened 2025-05-02
Add FastMRI Dataset

fzimmermann89:
Files need to be donwloaded beforehand (as terms and conditions have to be accepted)

There are two datasets:

### FastMRIKDataset
This Dataset just returns a KData object of a single slice or a full volme
The kdata in fastmri is already zero padded. We remove this padding.

Usage example:
```
dataset = FastMRIDataset('/echo/allgemein/resources/publicTrainingData/fastmri/brain_multicoil_train')

kdata = dataset[0]
reco = mrpro.algorithms.reconstruction.DirectReconstruction(kdata)
img = reco(kdata)
```
### FastMRIImageDataset
Returns either coilweise or inati-coil-combined image tensors (320x320).
Skips a few files of the dataset and changes the matrix size of some, keeping resolution and FOV.
This is needed to have a constant image size for batching.

### Example
image


src/mrpro/phantoms/fastmri.py
```diff
@@ -0,0 +1,208 @@
...
+        """Initialize the dataset.
+
+        Parameters
+        ----------
+        path : PathLike
+            Path to the data directory.
+        single_slice : bool
+            Whether to return single slices or stacks of slices.
+        """
+        self._filenames = list(Path(path).rglob('*.h5'))
```
lrlunin: Would it make more sense to set the glob mask for the data as default parameter to allow a user to load only T1/T2 weighted acquisitions?

src/mrpro/phantoms/fastmri.py
```diff
@@ -0,0 +1,208 @@
...
+            `None` means no augmentation.
+        allowed_n_coils
+            List of allowed number of coils. If `None`, all coils are allowed.
+            The knee training set has 15 coils consistently, while the brain dataset has
+            roughly 1300 files with 16 coils, 1100 files with 20 coils and 800 files with 4 coils.
+            Only used if `coil_combine` is `False`.
+        """
+        slices = []
+        self._filenames = []
+        for fn in Path(path).rglob('*.h5'):
```
lrlunin: Same idea for the glob mask as parameter

src/mrpro/phantoms/fastmri.py
```diff
@@ -0,0 +1,215 @@
...
+        coil_combine: bool = False,
+        augment: Callable[[torch.Tensor, int], torch.Tensor] | None = None,
+        allowed_n_coils: Sequence[int] | None = (16, 15),
+    ):
+        """Initialize the dataset.
+
+        Parameters
+        ----------
+        path : PathLike
+            Either a path to a directory containing the FastMRI data as .h5 filesor a Sequence of paths of
```
lrlunin:
```suggestion
            Either a path to a directory containing the FastMRI data as .h5 files or a sequence of paths of
```
fixed typo

lrlunin (approved):
LGTM! 🚀🚀🚀

PR #802, opened 2025-05-09
Normalise csm after upsampling

### LINKED ISSUE #797: Lowres CSMs result in artifacts due to upsampling
Low-Res CSMs result in artifacts in the reconstruction
Image

If using linear upsampling, we need to normalize afterwards: 

consider two coils at  two neighbouring pixels
0|1  and 1|0:
if we add a new interpolated pixel in between, it will be .5|.5 -- which will not be normalized

Alternativly, changing the interpolation mode to nearest would also fix it.

ckolbPTB:
Closes #797


src/mrpro/data/CsmData.py
```diff
@@ -135,6 +135,8 @@ def from_idata_walsh(
             chunk_size=chunk_size_otherdim,
         )
         csm_tensor = csm_fun(idata.data.flatten(end_dim=-5)).reshape(idata.data.shape)
+        # upsampled csm requires normalization
+        csm_tensor /= csm_tensor.abs().square().sum(dim=-4, keepdim=True).sqrt() + 1e-9
```
fzimmermann89, line 139: csm_tensor = torch.nn.functional.normalize(csm_tensor, p=2, dim=-4, eps=1e-9)

fzimmermann89:
```suggestion
        csm_tensor = torch.nn.functional.normalize(csm_tensor, p=2, dim=-4, eps=1e-9)
```

src/mrpro/data/CsmData.py
```diff
@@ -221,6 +223,8 @@ def from_idata_inati(
...
```
fzimmermann89:
```suggestion
        csm_tensor = torch.nn.functional.normalize(csm_tensor, p=2, dim=-4, eps=1e-9)
```

fzimmermann89 (approved):
just in case we ever want to autograd here, should already avoid inplace if possible.

also, you could consider swithcing to F.normalize (see suggestions).

PR #805, opened 2025-05-14
Fix bug in shape of KData parameters

ckolbPTB:
The change of all tensors to 5D meant that the use of discard_pre/discard_post led to wrong dimensions here:
https://github.com/PTB-MR/mrpro/blob/2f9d495eed367780c2f1f9f56c8210e408af85f6/src/mrpro/data/KData.py#L182

and in KTrajectorySunflowerGoldenRpe a k0-dimension was added which is not needed.

### DIFF
--- a/src/mrpro/data/KData.py
+++ b/src/mrpro/data/KData.py
@@ -162,6 +162,9 @@ class KData(Dataclass):
             additional_fields=('center_sample', 'number_of_samples', 'discard_pre', 'discard_post'),
             convert_time_stamp=convert_time_stamp,
         )
+        # discard_pre and discard_post is of shape (N_readouts,1,1,1,1) and we only need it as (N_readouts,)
+        discard_pre = discard_pre[:, 0, 0, 0, 0]
+        discard_post = discard_post[:, 0, 0, 0, 0]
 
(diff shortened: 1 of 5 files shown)


src/mrpro/data/KData.py
```diff
@@ -162,6 +162,9 @@ def from_file(
             additional_fields=('center_sample', 'number_of_samples', 'discard_pre', 'discard_post'),
             convert_time_stamp=convert_time_stamp,
         )
+        # discard_pre and discard_post is of shape (N_readouts,1,1,1,1) and we only need it as (N_readouts,)
+        discard_pre = discard_pre[:, 0, 0, 0, 0]
```
fzimmermann89: instead of the comment, maybe 
`discard_pre.squeeze_()`
or
discard_pre = rearrange("discard_pre, "readout 1 1 1 1 -> readout")
discard_post = rearrange("discard_post, "readout 1 1 1 1 -> readout")

fzimmermann89: if it is easy to add chaging discards to the test data, could you do it? 
if it would take more than 10min / 20 lines, it's ok to leave it

fzimmermann89 (approved):
thx

PR #809, opened 2025-05-20
Fix imports from utils

fzimmermann89:
Some utils improts did not import from the actual file. This is inconsistant and caused an circular import issue.


fzimmermann89:
(looks like mostly i made these mistakes...)

ckolbPTB (approved):

PR #813, opened 2025-05-23
Fix wrong dimension of scaling for initial value of itSENSE

ckolbPTB:


fzimmermann89:
..why did this not break any of the examples?

Maybe we should some simple test for reconstructions as well?

fzimmermann89 (approved):

PR #816, opened 2025-05-24
Fix #815

### LINKED ISSUE #815: Gram for nufft + cartesian sampling
If the FourierOp contains both NonUniformFastFourierOp and CartesianSamplingOp then the implementation of the gram operator is not working. I added a test for this case here: https://github.com/PTB-MR/mrpro/tree/gram_bug_fft_nufft

Currently in our tests for the FourierOp we do not have any tests which require the CartesianSamplingOp and so we have missed this so far.

fzimmermann89:
Fix #815 

Leaves Bug in  964c541 open. Thus, there is no proper test as it would hit the bug. Only the morre common case of reducible mask is tested.


fzimmermann89:
I just want to see the diff and the test result

fzimmermann89:
not sure if my test case is wrong or if there is a bug in the nufft ooperator @ckolbPTB

fzimmermann89:
@ckolbPTB any idea what do do here ?

ckolbPTB:
There are two issues, right?

One is the bug when we try to figure out the joint/separate dims which is currently covered by this test. This I would leave as an open issue for the moment until somebody has data which requires us to fix this.

The other one is from #815. As a first step I would simply not use nufft-gram in `FourierOp.gram` if `CartesianSamplingOp._needs_indexing` is True. This would ensure our code does not break for this case and fixing this could again be a future issue to speed up the code.

fzimmermann89:
the original bug is fixed, but to properly test it, we hit the new issue that the joint/separate dim logic seems to be broken.

fzimmermann89:
but i dont really see through the joint/separate dims logic as it currently is.
could you take a look if you can fix it such that the test in 
964c541 works?

ckolbPTB (approved):
GRPE with partial Echo along the Cartesian readout can be reconstructed.

PR #820, opened 2025-05-31
Change WASABI model: make c scaling parameter

fzimmermann89:
Change the model from
`c-d*something` to` c*(1-a*something)` with` a=d/c`

This makes c a pure scaling parameter (easier for dictionary matching)


schuenke (approved):
Works as expected

PR #822, opened 2025-06-03
Switch to lipari colormap for T1 maps

### LINKED ISSUE #715: colormaps in qmri examples
Should we use lipari and navia colormaps for t1 and t2?
https://cig-utrecht.org/blog/2023/11/27/colormap-relaxometry.html

They are both available for example in the `cmap` package https://github.com/pyapp-kit/cmap
available on pip and conda-forge.

We could switch to these colormaps and add `cmap` as a dependency to the notebook dependency group.

guastinimara:
I modified the colormaps of T1 values in two notebooks so that they are all done with the lipari colormap.


guastinimara:
When I did the pre-commit it seemed that all tests passed, I don't understand why now 5 don't pass

examples/scripts/qmri_sg_challenge_2024_t1.py
```diff
@@ -124,14 +125,16 @@ def show_images(*images: torch.Tensor, titles: list[str] | None = None) -> None:
 # Let's have a look at the starting values for $M_0$ and $T_1$:
 
 # %%
+cmap_t1 = Colormap('lipari')
```
fzimmermann89: I would either not have this variable here and do
`im = axes[0, 1].imshow(t1_start[0, 0], vmin=0, vmax=2.5, cmap=Colormap('lipari').to_mpl())`

OR

move the .to_mpl() call here has well.

guastinimara: ah okay! I just copied how it was done in another notebook, I'll change it to this

fzimmermann89:
> When I did the pre-commit it seemed that all tests passed, I don't understand why now 5 don't pass

Not your fault. There was a pytest update. Should be fixed already

fzimmermann89 (approved):

PR #824, opened 2025-06-03
GridSampling Operator from inputs on the gpu

### LINKED ISSUE #823: GridSamplingOp from displacement on gpu
When initializing a GridSampling Operator from displacements fields on the GPU, it gives an error of tensors not being on the same device due to some grids computed on the CPU. 

I'm currently fixing this

guastinimara:
Now the Gridsampling operator has the grid on the GPU if at least one of the displacements is on the GPU. I also added a test for this and this should fix the problem

Fixes #823

### DIFF
--- a/src/mrpro/operators/GridSamplingOp.py
+++ b/src/mrpro/operators/GridSamplingOp.py
@@ -277,15 +277,6 @@ class GridSamplingOp(LinearOperator):
         padding_mode
             how the input of the forward is padded.
         """
-        # if at least one of the displacements on 'cuda', move everything to cuda
-        if displacement_x.is_cuda or displacement_y.is_cuda or (displacement_z is not None and displacement_z.is_cuda):
-            device = torch.device('cuda')
-            displacement_x = displacement_x.to(device)
-            displacement_y = displacement_y.to(device)
-            if displacement_z is not None:
-                displacement_z = displacement_z.to(device)
-        else:
-            device = torch.device('cpu')
         if displacement_z is not None:  # 3D
             if displacement_x.ndim < 4 or displacement_y.ndim < 4 or displacement_z.ndim < 4:
                 raise ValueError(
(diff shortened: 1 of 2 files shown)


src/mrpro/operators/GridSamplingOp.py
```diff
@@ -277,6 +277,15 @@ def from_displacement(
         padding_mode
             how the input of the forward is padded.
         """
+        # if at least one of the displacements on 'cuda', move everything to cuda
+        if displacement_x.is_cuda or displacement_y.is_cuda or (displacement_z is not None and displacement_z.is_cuda):
```
fzimmermann89: this does not work if you have multiple gpus.
instead, i would do the grid always on cpu as before.
then, when adding grid_x and displacement_x do:
`grid_x = grid_x.to(displacement_x) + displacement_x * 2 / (n_x - 1)`
and so on.

then dtype (float16, float32) and which gpu device will also match.

fzimmermann89 (changes requested):

tests/operators/test_grid_sampling_op.py
```diff
@@ -399,3 +399,38 @@ def test_grid_sampling_op_from_displacement_2d():
...
+
+    # all displacements are on cuda, so the operator grid should also be on cuda
+    operator_all_cuda = GridSamplingOp.from_displacement(
+        displacement_z=displacement_z,
+        displacement_y=displacement_y,
+        displacement_x=displacement_x,
+        interpolation_mode='nearest',
+    )
+
+    # at least one displacement is on cuda, so the operator grid should also be on cuda
```
fzimmermann89: i would remove the mixed cases and consider it user error and undefined behaviour what we do in that case...

fzimmermann89: maybe, add a check if all inputs are on the same device, otherwise throw an error (besides being on cuda, they have to be on the same gpu..)

fzimmermann89:
?

fzimmermann89 (approved):

PR #825, opened 2025-06-03
Fix coverage for pytest 8.4

fzimmermann89:
fixes currently failing pytest ci.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -232,3 +232,6 @@ extend-exclude = [
 
 [tool.coverage.paths]
 source = ["src", "*/site-packages"]
+
+[tool.coverage.run]
+disable_warnings = ["module-not-measured"]


fzimmermann89:
the warning is caused by pytest-cov issues with xdist. removing xdist or settign n=0 would also remove the warning.

in the recent pytest (few days old) this warning then triggesrs an internal error and everything fails.

it seems the warning can safely be ignored and we still get good coverage out (we have had thsi warning for a long time and ignored it).

if we notive that coverage is wrong, we might need to remove parallel test execution via xdist.

fzimmermann89:
btw: this was a whole evening of wild guessing...

fzimmermann89:
https://github.com/pytest-dev/pytest-cov/issues/693

lrlunin (approved):

PR #830, opened 2025-06-04
Add cpu flag for the pytest minimal library installation

### LINKED ISSUE #829: Pytest minimal downloads the CUDA libraries
In the `minimal_requirements.txt` and the `--index-url pytorch-cpu` looks to be ignored or wrongly interpeted in a way that the CUDA version will be installed instead.

lrlunin:
Fixes #829


fzimmermann89 (approved):
thx!

PR #831, opened 2025-06-05
Fix random undersampling trajectory generation

fzimmermann89:
There was an issue in how the sampling probability was used. Basically, the mid-freqency lines where sampled most often (instead of the low frequency lines)

@koflera can you also check if this fixes the issue?
@rkcatarina try this


fzimmermann89:
..and allow "None" as a seed

koflera (approved):
Looks good to me

PR #832, opened 2025-06-05
Fix bug in k-space center along k0

### LINKED ISSUE #827: Possible bug in flip readouts
@ckolbPTB noticed some issue with flipped readouts in Siemens data.

Christoph, once you get around investigating, maybe you can either fix the issue or explain what the iussue is?

Note that I also removed the special case in the ismrmrd export. maybe there needs to be some speicial case, but we were not sure where the error originated from.

ckolbPTB:
Closes #827 

According to the ISMRMRD documentation the `center_sample` describes the "Index of the readout sample corresponing to k-space center (zero indexed)". For a reversed readout (bipolar gradient) we did not take this into consideration which led to a linar phase ramp in image space.
Comparison of two echos acquired with monopolar and bipolar gradient:
before_readout_correction

This PR fixes this:
after_readout_correction

Remaining issue: 
The `center_sample` seems to be off by 1 for the bipolar gradient and the current fix of course is not good:
https://github.com/PTB-MR/mrpro/blob/158b857da02b32531ea5cd76e8937c14c79277ec/src/mrpro/data/traj_calculators/KTrajectoryCalculator.py#L95

My suggestion would be to fix this here
https://github.com/PTB-MR/mrpro/blob/158b857da02b32531ea5cd76e8937c14c79277ec/src/mrpro/data/KData.py#L160
with unfortunately another "if Siemens then" line. Any opinions on that?


fzimmermann89:
Can you post a couple of examples? what k0 index is and where the k-center is in the data?
 for 128 adc samples
- full normal read out
- reversed read out
- partial fourier normal readout
- partial fourier reversed readout

I think there might be a simpler way to write this that doesnt break the existing code but fixes your issues.

can you add a reversed readout test case somewhere?

ckolbPTB:
full normal read out: 512 k0 samples -> k0_centre: 256 -> argmax of k-space data: 256
reversed read out: 512 k0 samples -> k0_centre: 256 -> argmax of k-space data: 255
partial fourier normal readout: 428 k0 samples -> k0_centre: 172 -> argmax of k-space data: 172
partial fourier reversed readout: 428 k0 samples -> k0_centre: 172 -> argmax of k-space data: 171

fzimmermann89:
> full normal read out: 512 k0 samples -> k0_centre: 256 -> argmax of k-space data: 256
> reversed read out: 512 k0 samples -> k0_centre: 256 -> argmax of k-space data: 255

this matches our previous code.
```
k0=torch.arange(512) - 256
assert k0[256] == 0 
reversed = torch.flip(k0,(-1,))
assert reversed[255] == 0
```
but
> partial fourier normal readout: 428 k0 samples -> k0_centre: 172 -> argmax of k-space data: 172
> partial fourier reversed readout: 428 k0 samples -> k0_centre: 172 -> argmax of k-space data: 171
```
k0=torch.arange(428) - 172
assert k0[172] == 0 
reversed = torch.flip(k0,(-1,))
assert reversed[171] == 84
assert reversed[255] == 0
````
so the flipped  partial fourier readout is broken. 

The previous code was already more complex then requiered -- we know how k0 looks at that point.

i.e. to set the nth element of an arange to zero, we can subtract n. 
and to flip it, we we can subtract it from m. if we have already centered it, it would be subtracting from zero (or -1 to match the  offset...)

So no need for broadcasting, masking flipping etc i think. a simple
`            k0 = torch.where(reversed_readout_mask.unsqueeze(-1), -1 - k0, k0)
`
should cover the cases?



I also modfied the bipolar test to check for partial fourier the center point.

fzimmermann89:
@ckolbPTB could you check if this fixes your issue?

ckolbPTB:
Yes, this fixes the issue. 

I believe that according to the ISMRMRD documentation there should not be a need for the `-1` but I am happy to leave it as is for the moment and fix it when we have a case where this is a problem.

fzimmermann89:
Perfect, then maybe let's add a comment why we do it with a link to this discussion.
For clarification: the bipolar gradient part of the sequence you use is a common building block or could it be that there is a mistake in the sequence?

fzimmermann89 (approved):

PR #833, opened 2025-06-10
Fix orientation due to fft used for recon by Siemens

### LINKED ISSUE #828: Wrong orientation for siemens data
At least siemens data seems to be orientated with all fourier dimensions flipped. maybe we need to conjugate the data.

ckolbPTB:
We believe that Siemens provides the  orientation for readout, phase encoding and slice encoding assuming a fft between k-space and image space. We use a ifft to go from k-space to image space in accordance with other image reconstruction packages. To ensure correct orientation we use the complex conjugate of the k-space for Siemens data. 

closes #828


fzimmermann89 (approved):

fzimmermann89:
@rkcatarina Can you take a look here?
The orientation fix changed the orientation of the image. thus, the positions of the tubes changed. can you fix the notebook accordingly?

ckolbPTB:
When we read in Siemens data, we are taking the complex conjugate of the data. 

For saving we could "undo" this and save it with vendor = Siemens or we could leave the data as is and use vendor = mrpro. The latter would also make it clearer that we are not replicating the original raw data (we do not add all the header information and there are also no inverted readouts anymore for a bipolar acquisition).

fzimmermann89:
I think I would keep the acquisition system vendor as Siemens...

rkcatarina:
I now changed the orientation of the reference and mask files on zenodo, adapted the DOI-zenodo-link and changed slightly the TE time.

fzimmermann89:
pytest failure is caused by https://github.com/flatironinstitute/finufft/issues/707

fzimmermann89:
Can i merge this? If I dont here a veto, I will merge it this weekend :)

ckolbPTB:
Can we please wait with this until the MRpro meeting on Wednesday? Once this is merged, the image orientation for every reconstruction will change, which might cause issues for some people.

PR #834, opened 2025-06-10
Patient table position is provided in micrometer

ckolbPTB:


fzimmermann89:
is it always in um? or, do we assume it is aways in um?
then we dont need to have to conversion function as a paramter, or?

ckolbPTB:
for Siemens it seems to be in mum. The documentation of ISMRMRD does not specify anything. Not sure what other vendors do. Also happy to hardcode it to `micrometer_to_m`.

fzimmermann89:
unless we find a different value somehere , i would hardcode it for now.

fzimmermann89 (approved):

PR #838, opened 2025-06-13
Fix #811

### LINKED ISSUE #811: Finite difference kernels might be swapped
We call this the forward differences: https://github.com/PTB-MR/mrpro/blob/ddf60d9c0e6294b8c200839206d2642a09003cc0/src/mrpro/operators/FiniteDifferenceOp.py#L36

We use Pytorch conv, which does a cross correlation, that is  a convolution with a FLIPPED kernel.

Our definition assumes a convolution instead. We should verify this and fix.

fzimmermann89:
Needs update of tests.


fzimmermann89:
ignore this. i think it was correct before.
@koflera why did you say it was mixed up? the test results match the expectred behavour.

koflera:
> ignore this. i think it was correct before. @koflera why did you say it was mixed up? the test results match the expectred behavour.

cause I mixed up forward and backward finite differences. it was, as you say, correct before. Sorry for the chaos.

PR #841, opened 2025-06-16
Add PEX signal model

fzimmermann89:


fzimmermann89:
i removred the example script for now and only keep the pex signal model. i moved t1 to the forward. this allows it tbe be inuded in a dictionary for dcitionary matching.

src/mrpro/operators/models/PEX.py
```diff
@@ -0,0 +1,79 @@
...
+    def forward(self, b1: torch.Tensor, t1: torch.Tensor) -> tuple[torch.Tensor,]:
+        """Apply PEX signal model.
+
+        .. note::
+            Prefer calling the instance of the PEX operator as ``operator(b1, t1)`` over
+            directly calling this method.
+        """
+        ndim = b1.ndim
+        voltages = unsqueeze_right(self.voltages, ndim - self.voltages.ndim + 1)  # +1 are voltages
+        prep_delay = unsqueeze_right(self.prep_delay, ndim - self.prep_delay.ndim + 1)
```
mxlutz, line 69: For me the test_pex_t1_recovery Test was failing due to dimension issues in the signal function (prep_delay and t1 are arrays). This is fixed with moving this to the same dimension as voltages, however I am not sure whether this is a good idea? Usually one of them should only be scalar, could add some assert statement.

mxlutz (commented):
Looks good, I renamed a to b1 (as it basically is b1...) and changed the dimension of prep_delay, about this I am not sure whether it is a good idea

PR #846, opened 2025-06-24
Fix python3.13 dependecies group typo

lrlunin:


fzimmermann89:
was this a bug, is this a bug?  should i ignore it?

lrlunin:
> was this a bug, is this a bug? should i ignore it?

Fixed this in #850 instead

PR #850, opened 2025-06-26
Include minimal version testing in the pytest workflow

lrlunin:
Fixing bugs related to the pytest minimal


.github/workflows/pytest.yml
```diff
@@ -86,6 +97,8 @@ jobs:
           path: |
             pytest-coverage.txt
             pytest.xml
+            # only present for "mrpro_py310_minimal", would only warn if not present
+            pytest.log
```
fzimmermann89: why do we upload pytest.log?

fzimmermann89 (approved):
why the pytest.log upload for minimal?

lrlunin:
> why the pytest.log upload for minimal?

Actually, there is no need to do so. I can simply remove this if there is no objections.

PR #861, opened 2025-07-06
Add stack to dataclasses

fzimmermann89:
Stacking on zeroth-dimension, usefull for batches.

PR #860, opened 2025-07-06
Fix padding in FastMRI

fzimmermann89:

PR #859, opened 2025-07-06
Fix Dataclass error on mixed devices

fzimmermann89:
Should have been a RuntimeError: repr also expects RuntimeError

PR #858, opened 2025-07-06
Fix SSIM mask

fzimmermann89:
2D mask was applied wrongly.

PR #862, opened 2025-07-06
Fix CSMData init typing

fzimmermann89:
CSMData can be created with a KHeader (it will be converted). No mypy accepts this as well.

### DIFF
--- a/src/mrpro/data/Dataclass.py
+++ b/src/mrpro/data/Dataclass.py
@@ -823,6 +823,16 @@ class Dataclass:
         new._reduce_repeats_(recurse=True)
         return new
 
+    def stack(self, *others: Self) -> Self:
+        """Stack other along new first dimension.
+
+        Parameters
+        ----------
+        others
+            other instance to stack.
+        """
+        return self[None].concatenate(*[o[None] for o in others], dim=0)
+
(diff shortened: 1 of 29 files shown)


fzimmermann89:
* #865
* #864
* #863
* __->__ #862
* #861
* #860
* #859
* #858

src/mrpro/data/Dataclass.py
```diff
@@ -100,6 +100,7 @@ def __init_subclass__(  # noqa: D417
         cls,
         no_new_attributes: bool = True,
         auto_reduce_repeats: bool = True,
+        init: bool = True,
```
ckolbPTB, line 103: this should be added to the doc string - should this not be raised by our tests?

ckolbPTB (approved):

src/mrpro/data/Dataclass.py
```diff
@@ -113,7 +114,7 @@ def __init_subclass__(  # noqa: D417
             If `True`, try to reduce dimensions only containing repeats to singleton.
             This will be done after init and post_init.
         """
```
fzimmermann89, line 118:
```suggestion
        init
            If `True`, an automatic init function will be added. Set to `False` to use a custom init.
        """
```

PR #867, opened 2025-07-08
Fix InconsistentDevice test

fzimmermann89:
I forgot to include the changes to the test in #859


fzimmermann89:
@lrlunin the self-hosted on push to main found it :) good job!

If only I had trigged the self-hosted test earliear. Maybe I should set it to requeired.

lrlunin (approved):

PR #871, opened 2025-07-09
Add missing dependencies

fzimmermann89:
I thought it was a built-in package. Instead, 
in most cases, this was already a dependency of our dependencies -- but not in all cases.


lrlunin:
@fzimmermann89
I have added the minimal versions of the new dependencies into the `py310_minimal`.

lrlunin (approved):

PR #875, opened 2025-07-14
Add reflection padding

fzimmermann89:
Enable reflection padding in pad_or_crop.

Also fixes the existing circular padding. 

Ideally, this function would also work for non-constant padding in more than the last 3  dimensions by reshaping and multiple calls But this is still a TODO.


fzimmermann89:
there are some other issues with the pad function, i will fix them first.

edit: this is not "nice" but at least better. proper fix is todo.

src/mrpro/utils/pad_or_crop.py
```diff
@@ -72,15 +73,33 @@ def pad_or_crop(
...
         before = diff - after
-        npad.append(before)
-        npad.append(after)
+        if before or after or npad:
+            npad.append(before)
+            npad.append(after)
+
+    n_extended_dims = 0
+    if mode != 'constant':
+        # See https://docs.pytorch.org/docs/stable/generated/torch.nn.functional.pad.html for the supported shapes
```
schuenke: Maybe add a short explanation directly here before the link?! Like "PyTorch's non-constant padding modes only support (up to) the last 3 dimensions"

src/mrpro/utils/pad_or_crop.py
```diff
@@ -72,15 +73,33 @@ def pad_or_crop(
...
+        while len(npad) // 2 < data.ndim - 2:
+            npad = [0, 0, *npad]
+
+        n_extended_dims = max(0, len(npad) // 2 - (data.ndim - 2))
+        if n_extended_dims:  # We need to extend data such that the padding is supported.
+            data = unsqueeze_left(data, n_extended_dims)
+
+        if len(npad) > 6:  # TODO: reshape and call multiple times
+            raise ValueError('replicate and circular padding are only supported for up to the last 3 dimensions.')
```
schuenke: Mention 'reflect' as well or simply change to "Non-constant padding modes are only supported for up to the last 3 dimensions."

schuenke (approved):
Minor comments. Rest looks fine.
Need to fix and merge #900 first.

PR #843, opened 2025-06-18
Add image export to dicom

ckolbPTB:
Requires #833

PR #880, opened 2025-07-23
Review of save_to_dicom

fzimmermann89:
Three major change suggestions (as PR easier than via comments) for #843 
- Move swapdims to dataclass. 
- Add Datetime to IHeader (already exists on KHeader) and use that instead of "now" if available
- Instead of copying the same dataset and sequence, I think it is cleaner to just create a new dataset/sequence. This is not slower and makes it easier to understand.

also, i think "unique" does not describe the property we need for the parameters: we want them to be singelton, i.e. all the same -- not all unique.


fzimmermann89:
(test failures are due to ptwt, already fixed on main)

ckolbPTB (approved):
Looks good, thanks a lot!

PR #894, opened 2025-09-09
 Add example notebook demonstrating TV vs TGV (denoising + radial MRI)

trung-vt:
# Add example notebook demonstrating TV vs TGV (denoising + radial MRI)

**No API changes**; documentation/education-oriented.

Motivation: Demonstrate how `mrpro` tools can be used to implement TGV minimization and symmetrized gradient for arbitrary dimensions.

As discussed with @koflera, this PR adds an example Jupyter notebook `examples/notebooks/tgv_tv_comparison.ipynb` that showcases **Total Generalised Variation (TGV)** side-by-side with **Total Variation (TV)** on two small problems:

1) **Image denoising** (additive noise on a simple 2D image)
2) **Radial undersampled MRI reconstruction** (Fourier operator + radial mask)

TV and TGV are run with comparable settings. TGV is shown to reduce staircasing relative to TV while preserving details.

## Implementation notes

- The notebook downloads 2 jpeg images and one .mat file from Zenodo.
- Automatically uses cuda if available.
- Other than downloading, most of the runtime is for running TV and TGV for the two examples, 256 iterations per run.
- Parameters are chosen for clarity rather than exhaustive tuning.

### DIFF
--- a/examples/notebooks/tgv_minimization_reconstruction_pdhg.ipynb
+++ b/examples/notebooks/tgv_minimization_reconstruction_pdhg.ipynb
@@ -1,0 +1,872 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "0",
+   "metadata": {
+    "lines_to_next_cell": 0
+   },
+   "source": [
+    "[![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/PTB-MR/mrpro/blob/main/examples/notebooks/tgv_minimization_reconstruction_pdhg.ipynb)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "1",
+   "metadata": {
+    "tags": [
+     "remove-cell"
+    ]
+   },
+   "outputs": [],
+   "source": [
+    "import importlib\n",
+    "\n",
+    "if not importlib.util.find_spec('mrpro'):\n",
+    "    %pip install mrpro[notebooks]"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "2",
+   "metadata": {},
+   "source": [
+    "# Total-generalized-variation (TGV)-minimization reconstruction"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "3",
+   "metadata": {
+    "lines_to_next_cell": 0
+   },
+   "source": [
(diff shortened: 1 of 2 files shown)


fzimmermann89:
Thank you for this great contribution! This is, in my opinion, a better approach than the other PR. Unfortunately, I will need some time to go through it in detail.

After skimming through it, I have a couple of general questions and remarks:

- Would it make sense to include the symmetric gradient operator as an Operator in our library, similar to the finite difference operator? It could still use the same logic, but this might be something to put into mrpro proper (as in your previous PR).

- We are an MR group, so I might be a bit picky about MR stuff: You do Cartesian MRI with a pseudo-radial undersampling pattern. Radial MRI would typically use a NUFFT and not have spokes lined up on the Cartesian grid. My suggestion would be to either just do 1D random Cartesian undersampling, i.e., a mask where lines are randomly removed in one direction, or an actual radial trajectory. This would remove the radial_mask code from this example.

- Data formats: I think it might be nicer to either a) save the data as a reshaped PyTorch tensor already on Zenodo, or, even better, use an ISMRMD file with a proper header, undersampling information, etc. This would remove some of this "boilerplate" from the example notebook. Do you think you have the capacity to try your code on one of the datasets we use in the other examples? There are examples using radial acquisitions, where the FourierOp detects that a non-uniform Fourier operator should be used, and examples using Cartesian data.

- Would you be okay with removing the image denoising example and focusing only on the MR example for the example notebook?

- If not, can image loading using PIL and torchvision be simplified by only using torchvision's decode_image?

- For the comparative TV reconstruction, you could even use the existing TV reconstruction in mrpro.algorithms. We already have a notebook explaining the background behind TV reconstruction, so here I think it would be good to focus on the TGV reconstruction by removing as much as possible of the "non-interesting," i.e., non-TGV code.

I think this would make this example even more educational. If you don't have the bandwidth to do this, we could also postpone some of these suggestions, and I can try to help you at some point (or maybe @koflera can continue to help).

Cheers,
Felix

trung-vt:
Hi Felix, thank you for the review! Would https://github.com/PTB-MR/mrpro/pull/904 be a more suitable addition? There is a new `SymmetrizedGradientOp` class. The example notebook now follows closely the TV reconstruction example: it downloads the ISMRMD file from Zenodo, applies NUFFT for radial MRI, and removes the image denoising example.

Best wishes,
Trung

PR #903, opened 2025-11-17
Add SymmetrizedGradientOp and unit tests

trung-vt:
(https://github.com/PTB-MR/mrpro/pull/904 adds to this PR with an example notebook using `SymmetrizedGradientOp` for TGV image reconstruction.)

## Summary

Add `SymmetrizedGradientOp` as a linear operator, and unit tests for correctness and adjointness.

## Motivation

The symmetrized gradient is required for total generalised variation (TGV) type regularisation, in particular for
second-order TGV as in Bredies (2014). Having a dedicated operator in `mrpro` makes it easier to implement TGV-based
reconstruction methods in a consistent way with the rest of the library.

## Changes

- New class `SymmetrizedGradientOp` in `mrpro.operators`
- Update docstring for `FiniteDifferenceOp` and change default value of `mode` to `forward`.

## Testing

New tests `tests/operators/test_symmetrized_gradient_op.py`. All tests pass locally.

## Notes

No breaking changes expected; existing APIs remain unchanged.

---

> [!NOTE]
> Introduce SymmetrizedGradientOp with forward/adjoint implementations, export it, and add comprehensive unit tests including CUDA coverage.
> 
> - **Operators**:
>   - **New `SymmetrizedGradientOp`**: Implements symmetrized gradient as `0.5 * (1 + transpose) @ FiniteDifferenceOp`; supports `dim`, `mode={'central','forward','backward'}`, and `pad_mode={'zeros','circular'}`; provides `forward` and `adjoint`.
>   - Exported via `mrpro.operators.__init__` and added to `__all__`.
> - **Tests** (`tests/operators/test_symmetrized_gradient_op.py`):
>   - Validates forward on a 2D linear vector field.
>   - Checks adjointness, gradient, and forward-mode autodiff across multiple `dim`/`mode`/`pad_mode` settings.
>   - Verifies CPU/GPU (CUDA) execution.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7faa03d84e415041e0fb6be4581ba41991552801. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/mrpro/operators/SymmetrizedGradientOp.py
```diff
@@ -0,0 +1,99 @@
...
+
+from mrpro.operators.FiniteDifferenceOp import FiniteDifferenceOp
+from mrpro.operators.LinearOperator import LinearOperator
+from mrpro.operators.RearrangeOp import RearrangeOp
+
+
+class SymmetrizedGradientOp(LinearOperator):
+    r"""Symmetrized gradient operator.
+
+    The symmetrized gradient :math:`\mathcal{E}: \mathcal{R}^d \to \mathcal{R}^{d+1}`
```
koflera: Maybe it would be helpful to explicitly state that this here is understood to be point-wise for an image, i.e. the R^d or the C^d is, for example, supposed to denote the space of the derivatives. Potentially, one could think of R^d or C^d denoting the space of the vectorized image and yielding some confusion.

src/mrpro/operators/SymmetrizedGradientOp.py
```diff
@@ -0,0 +1,99 @@
...
+    is defined as:
+
+    .. math::
+
+        \mathcal{E}v = \frac{1}{2}(\nabla v + (\nabla v)^{\top})
+
+    where :math:`(\nabla v)^{\top}` denotes the transpose of the Jacobian matrix :math:`\nabla v`.
```
koflera: Maybe here, one could provide an example, e.g. to say that v = v(x,y) would correspond to a scalar-valued function v: R^2 --> R, and then the connection between nabla as an operator applied to the whole image and the Jacobian being meant as for each point becomes clearer.

src/mrpro/operators/SymmetrizedGradientOp.py
```diff
@@ -0,0 +1,99 @@
...
+    .. [TGV] Bredies, K. Recovering piecewise smooth multichannel images by minimization of convex
+       functionals with total generalized variation penalty. In: Bruhn, A., Pock, T., Tai, X.C. (eds)
+       Efficient Algorithms for Global Optimization Methods in Computer Vision. Lecture Notes in Computer Science,
+       vol. 8293, Springer, Berlin, Heidelberg, 2014, pp. 44-77. https://doi.org/10.1007/978-3-642-54774-4_3
+    """
+
+    def __init__(
+        self,
+        dim: Sequence[int],
+        mode: Literal['central', 'forward', 'backward'] = 'central',
```
koflera: In the reference you provided, "backward" differences is what they use, so maybe use this here as a default?

koflera (commented):
Looks good to me. I would first merge this and then, once this is merged, get rid of the changes regarding the symmetrized gradient op and the related tests in #904, so that in #904 we can focus only on the example notebook.

What do you think @ckolbPTB, @fzimmermann89?

src/mrpro/operators/SymmetrizedGradientOp.py
fzimmermann89: does central lead to checkerpoint patterns as it does for TV? 

maybe lets change the default to backward for both the finite difference op (https://github.com/PTB-MR/mrpro/blob/13df110733ea5e3595a50768c7e1a459dbbce378/src/mrpro/operators/FiniteDifferenceOp.py#L46) and here?

src/mrpro/operators/SymmetrizedGradientOp.py
```diff
@@ -0,0 +1,99 @@
...
+            k elements for k gradients.
+        mode
+            Type of finite difference operator
+        pad_mode
+            Padding to ensure output has the same size as the input
+        """
+        super().__init__()
+        finite_difference_op = FiniteDifferenceOp(dim, mode=mode, pad_mode=pad_mode)
+        transpose_op = RearrangeOp('sym_grad_dim  grad_dim  ...   ->   grad_dim  sym_grad_dim  ...')
+        self.symmetric_gradient_op = 0.5 * (1 + transpose_op) @ finite_difference_op
```
fzimmermann89:
```suggestion
        self._symmetric_gradient_op = 0.5 * (1 + transpose_op) @ finite_difference_op
```

src/mrpro/operators/SymmetrizedGradientOp.py
fzimmermann89: Yes - this operator if a finite difference approximation of this, as we dont use the Jacobian of a function v, but always a finite diffrerence for \nabla v.
Maybe rephrase to make this more more clear? I would even consider avoiding "Jacobian" here completly, as we here always consider discrete images - and have a Jacobian Operator that operates on functions.

Would it be Ok with you to formulate this first as the finite difference version and then mention that this is the discrete version of the Jacobian?

Could you add a link to the finite difference operator in the docstring?

Also, could @kofler or @trung-vt add a similiar short description to finite difference operator?

fzimmermann89:
I would prefer to have similiar docstrings for the finite difference operator and the symmetrized gradient op, i.e. write the description here already for the discrete version of it (as we can only apply it to images, not to functions v)

Also, would be great to use similar notation in both operators and link from Open
 SymmetrizedGradientOp docstring to the finite difference op.

@koflera lgtm. your decision if you want to include the finite difference docstring change here or make a tiny seperate PR. I think it would be great if  @trung-vt or you could do it.

src/mrpro/operators/SymmetrizedGradientOp.py
fzimmermann89: just a tiny suggestion to "hide" the internals, as otherwise it a) might look a bit confusing that symmetricgradientop as an property "symmetric_gradient_op" which you should not call directly or modify.

(you could even call this "_implementation", "_inner" "_operator" etc if you want to)

src/mrpro/operators/SymmetrizedGradientOp.py
koflera: Yes, I also agree with getting rid of the word "Jacobian" here

src/mrpro/operators/SymmetrizedGradientOp.py
koflera: not sure regarding the checkerboard artefacts, but yes, also for TV we should change that. 

Since in the reference they use forward for the TV and backward for the symmetrized gradient (not sure if this is strictly needed), I would go for "backward" here and for "forward" in the finite differences op.

koflera:
> @koflera lgtm. your decision if you want to include the finite difference docstring change here or make a tiny seperate PR. I think it would be great if @trung-vt or you could do it.

@trung-vt I would be super happy if you could also include the suggested changes in the docstring of the finite difference operator in this PR :)

src/mrpro/operators/SymmetrizedGradientOp.py
trung-vt: Thanks for the suggestion! I have changed it to `_operator`

src/mrpro/operators/SymmetrizedGradientOp.py
trung-vt: Thank you! I have set the default of `FiniteDifferenceOp` to `forward` and `SymmetrizedGradientOp` to `backward`

trung-vt:
Thank you for your suggestions! I have updated the docstrings for both classes (for convenience, I attached the renderings below) 

image

image

fzimmermann89:
Mh, this looks nice and mathematical, but I am not convinced that the average user of mrpro will understand it.
image
might be precise, but we usually just talk about a nd-tensor (in the programmer's definition as an array, not as a multilinear map) anoperator is applied on -- think numpy or torch documentation, not math book.

Same for "Assuming unit spaceing", "for a continous function" ... etc
This specific class can only be applied for discrete images.

Can we dumb it down? For an average undergrad phycicist/SWE?

koflera:
> Mh, this looks nice and mathematical, but I am not convinced that the average user of mrpro will understand it. image might be precise, but we usually just talk about a nd-tensor (in the programmer's definition as an array, not as a multilinear map) anoperator is applied on -- think numpy or torch documentation, not math book.
> 
> Same for "Assuming unit spaceing", "for a continous function" ... etc This specific class can only be applied for discrete images.
> 
> Can we dumb it down? For an average undergrad phycicist/SWE?

I wouldn't actually mind being more precise here. But if we have to make it simpler, we could perhaps for the Finite differences operator just keep the following formula (without the h) and say something like: "The finite differences operator implements this operation for the entire tensor for all dimensions i indexed by dim by means of a separable convolution with appropriate filters, where x is a coordinate on the equispacced grid implicitly defined by the tensor. How the boundaries are handled is defined by pad_mode."
grafik

For the symmetrized gradient, we could then perhaps just keep the formula 
grafik
and again point out what x is meant to denote, and mention that the \nabla is implemented by means of the finite differences operator and reference it.

What do you think @fzimmermann89 @trung-vt?

trung-vt:
Thank you for your comments! Would this be a more straightforward approach? 
I assume only discrete settings now and switched the formulas to pseudo-code, hopefully it still keeps things readable and precise.

image

image

koflera:
What's the status on this? i think it would be nice to merge this. to me the docs look good as they are now, but they also looked good before
@ckolbPTB @fzimmermann89 @trung-vt

ckolbPTB:
Is the idea to do https://github.com/PTB-MR/mrpro/pull/904 afterwards also including a TGV reconstruction algorithm like we have for TV?

koflera:
> Is the idea to do #904 afterwards also including a TGV reconstruction algorithm like we have for TV?

yes, I think it makes sense to split the two things. so that we can start using TGV and have the example at a later point.

PR #913, opened 2025-11-28
Configure Dependabot to ignore specific GitHub Actions

fzimmermann89:
Ignore updates for actions in pytest_selfhosted to prevent compatibility issues with the self-hosted runner.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,7 +3,7 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "daily"
+      interval: "monthly"
     groups:
       all-actions:
         patterns: ["*"]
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,10 +1,14 @@
...
-  directory: "/"
-  schedule:
-    interval: "daily"
-  groups:
-    all-actions:
-      patterns: [ "*" ]
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
```
schuenke: do we maybe want to switch to weekly? In the past, we sometimes had PRs every 2 days and not beeing up to date for a week shouldn't be problematic imo.

schuenke (approved):
should do the job 👍

fzimmermann89:
mhm. manifest-file does not seem to be supported in this context. need to find a dfferent approach.

PR #914, opened 2025-11-28
Run selfhosted runner on reserved GPU

fzimmermann89:
as suggested by Florian in May.

This should skip some of the long running jobs on the cluster -- as long as AG Schwabe is not also using all of the reserved GPUs.


ckolbPTB:
superceded by #920 

These changes were already implemented on the uncommited code version on the HPC

PR #916, opened 2025-12-01
Reduce dependabot github action update frequency

fzimmermann89:
these updates are not urgent and regulary break our CI, lets do them less frequently, and properly check if the updates change anything before merging.

kind of a follow up to #913


schuenke (approved):
The jump from daily to quarterly is quite big, but as you said, these updates are usually not very urgent and thus it would be fine for me. 

Still, monthly might be an option as well. What do you think @ckolbPTB @fzimmermann89

PR #922, opened 2025-12-21
Remove DcfData and QData

ckolbPTB:
We always need the `DensityCompensationOp` rather than the `DcfData`, so it makes more sense to create the operator directly rather than doing it as a two-step approach.

`QData` is currently not used anywhere and currently has very limited Dicom support. Better to use `IData` for all data in image space including coil sensitivity maps and parametric maps.


fzimmermann89:
Fun fact: I am actually using DCFData -- for one dataset with a somlicated 3D spiral, there exists a precalucated DCF as part of the dataset. And I use DCFData to subset along the batch dimension... 

But in general, I agree and we should get rid off this.

ckolbPTB:
Superseded by #982

PR #923, opened 2025-12-30
Fix numpy warning in brainweb download normalization

schuenke:
Added `out` parameter to `np.divide` call to avoid UserWarning about uninitialized memory when using `where` without `out`. 

This resolves current test failures due to stricter warnings in numpy >= 2.4.

### DIFF
--- a/src/mrpro/phantoms/brainweb.py
+++ b/src/mrpro/phantoms/brainweb.py
@@ -304,7 +304,7 @@ def download_brainweb(
         sum_values = sum(values)
         values.pop(ALL_CLASSES.index('bck'))  # noqa: typos
         for i, x in enumerate(values):
-            x = np.divide(x, sum_values, where=sum_values != 0)
+            x = np.divide(x, sum_values, out=np.zeros_like(x, dtype=float), where=sum_values != 0)
             x[sum_values == 0] = 0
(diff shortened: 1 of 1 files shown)


src/mrpro/phantoms/brainweb.py
```diff
@@ -304,7 +304,7 @@ def norm_(values: list[np.ndarray]) -> np.ndarray:
         sum_values = sum(values)
         values.pop(ALL_CLASSES.index('bck'))  # noqa: typos
         for i, x in enumerate(values):
-            x = np.divide(x, sum_values, where=sum_values != 0)
+            x = np.divide(x, sum_values, out=np.zeros_like(x, dtype=float), where=sum_values != 0)
```
ckolbPTB, line 307: Is `dtype=float `needed? Should `np.zeros_like` not return the same type as x as default?

schuenke: Not sure if it's needed, but it doesn't hurt imo.

schuenke: @ckolbPTB I had a quick look again and IMO the dtype specification is necessary, because x should be of type `uint16` (before the devision) and if we dont define the type, all devided values will probably get truncated to integers.

ckolbPTB (approved):

PR #932, opened 2026-01-30
Update ACTION_RUNNER_URL to version 2.331.0 and downgrade actions

ckolbPTB:


schuenke (approved):
No idea if this fill fix the failing tests, but we can give it a try.

schuenke:
seems to work 🎉

PR #935, opened 2026-01-30
Fix rotation from mixed dim spatial dimensions

ckolbPTB:
The previous `test_from_to_directions` simply tested for a `SpatialDimensions` of shape `(1,1,1,1,1)` because `reduce_repeat` is not called. 

`reduce_repeat` also means that `SpatialDimensions` for the orientation vectors of read, phase and slice can have different shapes. This led to an error when creating the `RotationMatrix`. Now we broadcast before creating the `RotationMatrix`


fzimmermann89 (approved):
we can use broadcast tensors here, or?

fzimmermann89:
I also added a fix (+test) for the mean of a single rotation

PR #702, opened 2025-03-06
Add export to NIFTI

JoHa0811:
Needs more details in the ISMRMRD exporter and tests for both

PR #938, opened 2026-02-04
Release v0.260204

fzimmermann89:


schuenke (approved):
The last release was done quite a while ago. I agree that it's time for a new one. 

Do we want to release it as it is or merge a few other PRs like #922 first? 
Both is fine for me!

fzimmermann89:
I would do one release before removing DCFData and QData.

We could merge these before 
#702 (really basic nifit export and basic test only)
#778 (t1rho)
#779 (bicgstab)

PR #940, opened 2026-02-06
refactor: single normalize_axes function

fzimmermann89:
we have many places where we normalize axes and through an error. 
we already had a normalize axis function in utils.
lets use that one more consistantly.
i also move it from the indexer helper to utils.reshape  to avoid circular imports ...

### DIFF
--- a/src/mrpro/operators/PatchOp.py
+++ b/src/mrpro/operators/PatchOp.py
@@ -43,7 +43,7 @@ class PatchOp(LinearOperator):
         self.dim = (dim,) if isinstance(dim, int) else dim
 
         if len(set(self.dim)) != len(self.dim):
-            raise ValueError('Duplicate values in axis are not allowed')
+            raise ValueError('must be unique')
 
         def check(param: int | Sequence[int], name: str) -> tuple[int, ...]:
             if isinstance(param, int):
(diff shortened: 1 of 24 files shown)


src/mrpro/operators/PatchOp.py
```diff
@@ -43,7 +43,7 @@ def __init__(
         self.dim = (dim,) if isinstance(dim, int) else dim
 
         if len(set(self.dim)) != len(self.dim):
-            raise ValueError('Duplicate values in axis are not allowed')
+            raise ValueError('must be unique')
```
ckolbPTB:
```suggestion
            raise ValueError('Axis indices must be unique')
```

src/mrpro/utils/reshape.py
```diff
@@ -10,6 +10,70 @@
 from mrpro.utils.typing import endomorph
 
 
+def normalize_index(ndim: int, index: int) -> int:
+    """Normalize possibly negative indices.
```
ckolbPTB:
```suggestion
    """Normalize possibly negative index.
```

src/mrpro/utils/reshape.py
```diff
@@ -10,6 +10,70 @@
...
+    unique
+        If True (default), raise ValueError when normalized indices contain duplicates.
+
+    Returns
+    -------
+    Tuple of normalized indices in [0, ndim).
+
+    Raises
+    ------
+    IndexError
```
ckolbPTB, line 60: Needs updating

ckolbPTB (approved):

PR #968, opened 2026-02-12
Fix Reconstructions on cuda

fzimmermann89:
Replaces .csm and .dcf on the plug and play reconstructions by .csm_op and .dcf_op, respectifly

This is a quick fix for moving a reconstruction to cuda.
currently, 
`reco.cuda()` would not move dcf/csm, and application to a cuda tensor fails.

(..Alterntive would be implementing a custom .to that also moves dataclasses.... but i wont have time for that.)


fzimmermann89:
@schuenke  @ckolbPTB 
your opinion here? 
would be great if reco.cuda() would work.
(still need to fix mypy)

ckolbPTB (approved):

Issue #842, opened 2025-06-18
NUFFT joint / separate dimension bug


fzimmermann89:
There is a bug in the dimension detection in the nufft operator.

See https://github.com/PTB-MR/mrpro/commit/964c541d3d29394eaea557dca27dd7a47c994acc and #816

PR #972, opened 2026-03-02
Allow nufft for traj where nufft/fft splitting fails

ckolbPTB:
For some trajectory (e.g. EPI) the FourierOp cannot be created as a combination of FFTOp and NUFFTOp. So far this led to an error and users had to create their own FourierOp. This PR defaults the FourierOp to NufftOp for all dimensions. Not optimal but at least we get an image.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -17,6 +17,4 @@
     ],
     "python.testing.unittestEnabled": false,
     "python.testing.pytestEnabled": true,
-    "python-envs.defaultEnvManager": "ms-python.python:conda",
-    "python-envs.defaultPackageManager": "ms-python.python:conda",
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -17,4 +17,6 @@
     ],
     "python.testing.unittestEnabled": false,
     "python.testing.pytestEnabled": true,
+    "python-envs.defaultEnvManager": "ms-python.python:conda",
+    "python-envs.defaultPackageManager": "ms-python.python:conda",
 }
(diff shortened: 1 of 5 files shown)


fzimmermann89:
I am a bit confused why epi does not work:
Is the full shot in k0 like a spiral or is each line+blip in k0?If the latter, can you cut by indexing the blip and get an FFT?

Or how does an epi trajectory look like? @schuenke

schuenke:
For [our implementation](https://github.com/PTB-MR/mrseq/pull/70), each line is a separate acquisition and we either use a symmetric trajectory (every 2nd line is acquired in reverse order) or a flyback trajectory (each line is acquired from -k to +k like a typical cartesian acquisition. However, we often use ramp sampling, so that only the center points are on a cartesian grid and the first and last few samples are closer together (adc dwell is constant, but due to non-constant gradients we have samller steps in kx direction). 

Note: Currently, we also use the pp.calculate_kspace() function to calculate the "nominell" trajectory we store in the mrd header file. It's been a while, but as far as I remember the calculated trajectories for the symmetric case where not perfectly matching and the default repeat detection did not recognize all kx points as identical for the different ky positions.

This is a typical trajectory for an 8x8 symmetric EPI with ramp sampling and readout oversampling:
image

In this case, the trajectories in non-reverse and reverse direction match quite well. 

However, if we use the same bandwidth and nx=ny=64, they get shifted as we can see when we zoom in:
image


The reason is that the adc start time has to be aligned to the RF (!) raster time and not the ADC raster time and rf_raster_time >> adc_raster time:
image

fzimmermann89:
I am confused why this does not work with the current code (might be #842 related?)

Do you have an mrd file for debugging?

fzimmermann89 (approved):

PR #973, opened 2026-03-06
Add more EPG signal models

fzimmermann89:
- Add multi echo spin echo / tse sequence (used for t2 mapping in paper by @rkcatarina)
- Add DeliCS neuro mrf signal model (https://zenodo.org/records/7697373)
- broadcast result to input shapes. inside the epg simulator, we use the repeat removal logic, but both @rkcatarina and the dictionarymatchingop expect that the result shape to match the input shapes as with all other signal models.


src/mrpro/operators/models/NeuroMRF.py
```diff
@@ -0,0 +1,129 @@
...
+        self, m0: torch.Tensor, t1: torch.Tensor, t2: torch.Tensor, relative_b1: torch.Tensor | None = None
+    ) -> tuple[torch.Tensor]:
+        """Simulate the NeuroMRF signal.
+
+        .. note::
+            Prefer calling the instance of the NeuroMRF as ``operator(x)`` over
+            directly calling this method. See this PyTorch `discussion <https://discuss.pytorch.org/t/is-model-forward-x-the-same-as-model-call-x/33460/3>`_.
+        """
+        params = Parameters(m0, t1, t2, relative_b1)
+        _, signals = self.sequence(params, states=100)
```
ckolbPTB: Can we make `states` an attribute of all the models rather than hardcode it in the `forward` call?

ckolbPTB (approved):

PR #974, opened 2026-03-19
Ignore real time feedback for image reconstruction

ckolbPTB:


schuenke (approved):
Sounds like it makes sense to ignore rt feedback in recon, but would be interesting to know where you encountered that.

fzimmermann89:
I think this was also present in one of the 7T files ..

ckolbPTB:
> Sounds like it makes sense to ignore rt feedback in recon, but would be interesting to know where you encountered that.

I am using it in one of my pulseq files because "PMC" is one of the supported labels (https://github.com/pulseq/pulseq/blob/master/matlab/%2Bmr/getSupportedLabels.m") which corresponds best with RTFEEDBACK.

fzimmermann89:
merge?

PR #978, opened 2026-03-31
Pytorch with cu128 for self-hosted runner

ckolbPTB:
Added PyTorch installation to the workflow.

### DIFF
--- a/docker/Dockerfile_runner
+++ b/docker/Dockerfile_runner
@@ -40,7 +40,7 @@ RUN bash install_dependencies.sh ${CUDA_VERSION} && rm install_dependencies.sh p
 # forces rebuild on any change in the mrpro directory
 COPY . /mrpro/
 RUN python -m pip install "/mrpro[notebooks]" --no-cache-dir --upgrade --upgrade-strategy "eager" \
-    --extra-index-url https://download.pytorch.org/whl/${CUDA_VERSION}
+    torch --extra-index-url https://download.pytorch.org/whl/${CUDA_VERSION}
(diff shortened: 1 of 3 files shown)


fzimmermann89 (approved):

docker/Dockerfile_runner
```diff
@@ -32,12 +34,14 @@ WORKDIR /
...
 # the 1st argument for the install_dependencies specifies the cuda version 12.4 for torch
-RUN bash install_dependencies.sh cu124 && rm install_dependencies.sh pyproject.toml VERSION
+RUN bash install_dependencies.sh ${CUDA_VERSION} && rm install_dependencies.sh pyproject.toml VERSION
 
 # install mrpro
 # forces rebuild on any change in the mrpro directory
 COPY . /mrpro/
-RUN python -m pip install  "/mrpro[notebooks]" --no-cache-dir --upgrade --upgrade-strategy "eager" && rm -rf /mrpro
+RUN python -m pip install "/mrpro[notebooks]" --no-cache-dir --upgrade --upgrade-strategy "eager" \
+    torch --extra-index-url https://download.pytorch.org/whl/${CUDA_VERSION}
```
fzimmermann89, line 43: why do we need this here?
doesnt install_dependencies already install torch?

fzimmermann89 (approved):

docker/Dockerfile_runner
ckolbPTB: otherwise "eager" would lead to an installation of the latest torch version with the default cuda 13 dependency

PR #982, opened 2026-04-16
Remove qdata

ckolbPTB:
`QData` is currently not used anywhere and currently has very limited Dicom support. Better to use IData for all data in image space including coil sensitivity maps and parametric maps.


fzimmermann89 (approved):

fzimmermann89:
Is here anything missing?

PR #983, opened 2026-04-16
Add interp

fzimmermann89:
Add torch implementation of np.interp

Useful for #790 and #975


src/mrpro/utils/interpolate.py
```diff
@@ -8,6 +8,38 @@
...
+    xp
+        1d tensor of strictly monotonically increasing x coordinates of data points.
+    fp
+        1d tensor of y coordinates matching the length of xp.
+
+    Returns
+    -------
+        The interpolated values matching the shape of x.
+    """
+    x_clamped = torch.clamp(x, min=xp[0], max=xp[-1])
```
ckolbPTB, line 40: Would it make sense to add here:
```
if not torch.all(xp[1:] > xp[:-1]):
    raise ValueError("xp must be strictly monotonically increasing")
```
numpy.interp does not do it but seems an easy check. or do you think it will create to much overhead if the function is used repeatedly?

fzimmermann89: I was not sure about this either and skipped it only due to numpy.
Let's include a check. Should we also just sort if it's unsorted and only warn?

ckolbPTB (approved):

PR #989, opened 2026-05-01
Improve NUFFT Gram for large datasets

fzimmermann89:
This makes it possible to do iterative reconstruction on large non cartesian 3D datasets by
1. memory efficient in-place operations 
1. avoiding fftshifts
1. avoiding unnesesarry intermedate allocations

Additionally, the optional DCF weighting is now exposed in NonUniformFastFourierOpGramOp. It was already implemented in the backend, but there was no way to use it. Fourier.gram will still not use it, only if created manually with an additional argument `weight `

For subpsace compressed reconstruction, one needs PCA^H Fourier^H Fourier PCA
So for, for example, 500 different trajectories in the Fourier operator and 5 subspace coefficients, we would have to do 500 FFTs and save 500 Toeplitz kernels. This would be to slow for iterative recon inside a neural network.
Instead, we can pull the subspace compressin into the kernel. The kernel is not longer a real valued 2D image, but includes off-diagonal mixing; in this example 5x5x(2*recon_y)x(2*recon_x) and we only require 5 FFTs

Finally, it adds a .toeplitz function on the nufft operator as a .gram with more control about density compensation and subsapce compression. .gram uses it with the default parameters. Gram is a property and cannot take arguments, gram gets chained through, whereas topelitz needs to be called with the options on the nufft operator for more granular control

### DIFF
--- a/src/mrpro/operators/NonUniformFastFourierOp.py
+++ b/src/mrpro/operators/NonUniformFastFourierOp.py
@@ -377,6 +377,9 @@ class NonUniformFastFourierOpGramOp(LinearOperator, adjoint_as_backward=True):
 
     This should not be used directly, but rather through the `~NonUniformFastFourierOp.gram` method of a
     `NonUniformFastFourierOp` object.
+
+    .. note::
+        Consider calling .half() on the operator to save memory at the cost of precision.
     """
 
     _kernel: torch.Tensor | None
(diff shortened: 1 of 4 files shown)


src/mrpro/operators/NonUniformFastFourierOp.py
```diff
@@ -381,47 +381,112 @@ class NonUniformFastFourierOpGramOp(LinearOperator):
...
+        if nufft_op._omega.shape[-2] == 3:
+            # Special case for memory-hungry 3D case
+            # We loop over the batch/coil samples to reduce peak memory usage
+            # and we create the kernel in a non-contiguous way, such that after the final
+            # permute, we are back at a contiguous tensor.
+            kernel_shape = list(weight_shape)
+            for d, s in zip(nufft_op._direction_zyx, padded_shape, strict=True):
+                kernel_shape[d] = s
+            try:
+                kernel = weight.new_empty(kernel_shape, dtype=torch.float16)
```
ckolbPTB: Why are we using `torch.float16 `here?

src/mrpro/operators/NonUniformFastFourierOp.py
```diff
@@ -381,47 +381,112 @@ class NonUniformFastFourierOpGramOp(LinearOperator):
...
-        weight = torch.ones(
-            [*nufft_op._traj_broadcast_shape[:-4], 1, *nufft_op._traj_broadcast_shape[-3:]],
-        ).to(nufft_op._omega)
+        if isinstance(weight, DcfData):
+            weight = weight.data
+
+        weight_shape = [*nufft_op._traj_broadcast_shape[:-4], 1, *nufft_op._traj_broadcast_shape[-3:]]
 
-        # We rearrange weight into (sep_dims, joint_dims, nufft_dims)
+        if weight is None:
```
ckolbPTB, line 438: Don't we need to define the dtype here?

ckolbPTB:
Can we think of a better way to select the loop case? Doing the loop for all 3D reconstructions could make many of these reconstructions unneccessarily slow.

src/mrpro/operators/NonUniformFastFourierOp.py
fzimmermann89: In my tests it always has enough precision to STORE the kernel. all computations then still have to be done in float32/complex64, otherwise the accuracy of the result of bad. but for storage, float16 has enough precision and saves 50% of vram.

fzimmermann89:
I tried it on cpu and a6000 gpu. for 3d non cartesian reconstruction with 128x128x128 voxels, the loop in the forward saturates the gpu and it is not slower, i.e. batching more compute at once dpes not help.
are there realistically trajectories that are non cartesian on all 3 directions and use less voxels?

The issue is more on the other side, that for some 2D reconstructions it could also make sense to loop (I have not tried that)

src/mrpro/operators/NonUniformFastFourierOp.py
fzimmermann89: no, torch's defaults (float32) is fine. does not really matter what dtype it is

fzimmermann89:
mnhm... i reconsidered the loop case. I think i will remove this and instead, one will have to loop outside the operator. 
This might be achieved by unpacking and using the LinearOperatorMatrix, which also loops over the operator.

src/mrpro/operators/NonUniformFastFourierOp.py
fzimmermann89: but we can also create it in float32, then one can later call .half() on it.

fzimmermann89:
For subpsace compressed reconstruction, one needs PCA^H Fourier^H Fourier PCA
So for, for example, 500 different trajectories in the Fourier operator and 5 subspace coefficients, we would have to do 500 FFTs and save 500 Toeplitz kernels. This would be to slow for iterative recon inside a neural network.
Instead, we can pull the subspace compressin into the kernel. The kernel is not longer a real valued 2D image, but includes off-diagonal mixing; in this example 5x5x(2recon_y)x(2recon_x) and we only require 5 FFTs

Finally, it adds a .toeplitz function on the nufft operator as a .gram with more control about density compensation and subsapce compression. .gram uses it with the default parameters. Gram is a property and cannot take arguments, gram gets chained through, whereas topelitz needs to be called with the options on the nufft operator for more granular control

ckolbPTB (approved):

PR #991, opened 2026-05-03
Improve noise whitening and coil compression

fzimmermann89:
- fix memory usage in coil whitening. this used >1TB for 3D datasets...
- add whitening function to KData
- add varimax option for PCA and coil compression to "rotate" thesubspace such that each of the components has more variance, instead of decreasing information content. makes for example coils a bit more statistically similiar to "real" coils.
- add option to coil compression to return a compression operator. usefull for "uncompressing" the data or applying same compression to different data
- improve tests
- improve exceptino message in PCAOperator: it also failed by mixed devices errors, but always showed a shape missmatch message.

### DIFF
--- a/src/mrpro/data/KData.py
+++ b/src/mrpro/data/KData.py
@@ -558,7 +558,7 @@ class KData(Dataclass):
         Parameters
         ----------
         knoise
-            Noise measurements. Either a KNoise object or a tensor of shape (coils, k0).
+            Noise measurements. Either a KNoise object or a tensor of shape ``(..., coils, k0)``.
         scale_factor
             Square root is applied on the noise covariance matrix. Used to adjust for effective noise bandwidth
             and difference in sampling rate between noise calibration and actual measurement:
--- a/src/mrpro/data/KData.py
+++ b/src/mrpro/data/KData.py
@@ -7,14 +7,17 @@ import warnings
 from collections.abc import Callable, Mapping, Sequence
 from importlib.metadata import version
 from types import EllipsisType
-from typing import Literal, cast
+from typing import TYPE_CHECKING, Literal, cast
 
 import h5py
 import ismrmrd
 import numpy as np
 import torch
 from einops import rearrange
-from typing_extensions import Self, TypeVar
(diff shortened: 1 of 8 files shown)


src/mrpro/data/KData.py
```diff
@@ -478,38 +513,72 @@ def compress_coils(
...
+
+    def prewhiten(self, knoise: KNoise | torch.Tensor, scale_factor: float | torch.Tensor = 1.0) -> Self:
+        """Calculate noise prewhitening matrix and decorrelate coils.
+
+        This is a convenience wrapper around :func:`mrpro.algorithms.prewhiten_kspace`.
+
+        Parameters
+        ----------
+        knoise
+            Noise measurements. Either a KNoise object or a tensor of shape (coils, k0).
```
schuenke: it can actually be of shape (... , coils, k0) as well , no?

tests/data/test_kdata.py
```diff
@@ -419,6 +454,23 @@ def test_KData_compress_coils_error_n_coils(consistently_shaped_kdata: KData) ->
         consistently_shaped_kdata.compress_coils(existing_coils + 1)
 
 
+def test_KData_prewhiten(random_kheader) -> None:
```
schuenke: if we accept (... , coils, k0), we should test it as well

src/mrpro/operators/PCACompressionOp.py
```diff
@@ -76,9 +84,9 @@ def forward(self, data: torch.Tensor) -> tuple[torch.Tensor,]:
             result = (self._compression_matrix @ data.unsqueeze(-1)).squeeze(-1)
         except RuntimeError as e:
             raise RuntimeError(
-                'Shape mismatch in adjoint Compression: '
-                f'Matrix {tuple(self._compression_matrix.shape)} '
-                f'cannot be multiplied with Data {tuple(data.shape)}.'
+                'Shape or devicemismatch in Compression: '
```
schuenke: missing space

schuenke (approved):

ckolbPTB (approved):

PR #993, opened 2026-05-04
Fix #906:  inati should be iterative and walsh should phase align

### LINKED ISSUE #906: Inati's method for csm is walsh / inefficient phase correction
We implement an algorithm called "inati" and refernce both the 2013 and 2014 abstract.
The 2013 abstract is mostly walsh method +  using the a local phase refernce. (According to the abstract Walsh should use a reference coil for the phase, which we also dont do in walsh)

The two algorithms we implement also differ slightly in how they acutally perform the task - via filtering or sliding windows. But this should not matter.

Bart, tensorflow mri, ismrmrd-pythontools and most papers i find on scholar (as well as the LLMs) instead refer to his 2014 abstract as "inati's method".

This confusion stems from gadetron, which implements two algorithms, inati_iter (~2014 abstract) and inati (~2013).

I would check which variant (inati or walsh) uses less memory and is faster.
Both should provide similar results if one comments out the "mrpro inati" phase correction.
Then we should keep the faster/memory efficient variant and add a switch for the 2013 inati phase correction.

And add a proper inati method that uses the inati 2014 approach to get smother non-singular phases of images and sensitivity maps.

fzimmermann89:
This fixes #906 
Inati should be the iterative inati method (2014), not walsh + phase alignment.
Also, it adds phase alignment (can be disabled for pure walsh) to walsh, making it inati 2013.

### DIFF
--- a/src/mrpro/algorithms/csm/inati.py
+++ b/src/mrpro/algorithms/csm/inati.py
@@ -10,7 +10,7 @@ from mrpro.utils.filters import uniform_filter
 def inati(
     coil_img: torch.Tensor,
     smoothing_width: SpatialDimension[int] | int,
-    n_iterations: int = 5,
+    n_iterations: int = 10,
 ) -> torch.Tensor:
     """Calculate a coil sensitivity map (csm) using the iterative Inati method [INA2014]_.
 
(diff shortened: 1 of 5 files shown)


schuenke (approved):
The implementation, tests and documentation look good to me. 

Maybe @ckolbPTB wants to have a quick look as well?!

fzimmermann89:
It would also be great to test it on another challenging dataset. 
For the mrf data I did this change for, all three codes result in slightly different csms but unfortunately none seems clearly superior.
Nevertheless, I think this would be the proper csm Walsh and Inati variants to use

ckolbPTB:
I compared the old and the new methods on a few datasets and all show a similar behaviour as this:

csm

The new Walsh implementation with phase alignment leads to a reasonable homogenous phase. 

Nevertheless, the new Inati now leads to the same results as Walsh. I could not find a dataset where there was any noticeable difference and the number of iterations also does not seem to have an effect (i.e. no difference between first and last iteration). The result do not have a flat phase (more pronounced in cardiac datasets) whereas the old implementation leads to a completely constant phase. 

I could not find a bug so my suggestion would be to only change the Walsh implementation but leave the old Inati method and clearly label it as non-iterative.  @fzimmermann89 @schuenke what are your thoughs on this?

We should really work on the ESPIRT coil estimation...

PR #999, opened 2026-06-17
fix linear combination in PDHG and convergent variant of PGD

koflera:
fixes the update step by correcting the sign of $\theta$
grafik

### DIFF
--- a/src/mrpro/algorithms/optimizers/pdhg.py
+++ b/src/mrpro/algorithms/optimizers/pdhg.py
@@ -198,7 +198,8 @@ def pdhg(
         )
         primals_new = g_sum.prox(*primals_new, sigma=primal_stepsize_)
         primals_relaxed = tuple(
-            torch.lerp(primal, primal_new, relaxation) for primal, primal_new in zip(primals, primals_new, strict=False)
+            torch.lerp(primal, primal_new, -relaxation)
+            for primal, primal_new in zip(primals, primals_new, strict=False)
         )
 
         # check if the solution is already accurate enough
--- a/src/mrpro/algorithms/optimizers/pdhg.py
+++ b/src/mrpro/algorithms/optimizers/pdhg.py
@@ -198,7 +198,7 @@ def pdhg(
         )
         primals_new = g_sum.prox(*primals_new, sigma=primal_stepsize_)
         primals_relaxed = tuple(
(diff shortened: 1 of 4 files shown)


src/mrpro/algorithms/optimizers/pdhg.py
```diff
@@ -167,14 +167,17 @@ def pdhg(
     if primal_stepsize is None or dual_stepsize is None:
         # choose primal and dual step size such that their product is 1/|operator|**2
         # to ensure convergence
-        operator_norm = operator_matrix.operator_norm(*[torch.randn_like(v) for v in initial_values]).amax()
+        squared_operator_norm = (
+            operator_matrix.operator_norm(*[torch.randn_like(v) for v in initial_values]).amax().square()
+        )
+        safety_factor = 0.97
```
ckolbPTB: could we add a comment there why this is needed?

src/mrpro/algorithms/optimizers/pgd.py
```diff
@@ -156,7 +156,7 @@ def pgd(
                 raise RuntimeError('Stepsize to small.')
 
         if convergent_iterates_variant:
-            t = (iteration + 2) / 3
+            t = (iteration + 3) / 3
```
ckolbPTB, line 164: This is now different to the doc string...

ckolbPTB (changes requested):

src/mrpro/algorithms/optimizers/pgd.py
koflera: I put a short comment explaining that this has to do with the iteration counter starting from 0

src/mrpro/algorithms/optimizers/pdhg.py
koflera: I extended the comment that we had

src/mrpro/algorithms/optimizers/pgd.py
ckolbPTB: can we adapt the equation here to `t_n = \frac{n+a}{a}` and say at the end
Note: n here starts with 0 whereas in [CHAM2015]_ it starts with 1 and hence `\frac{(n-1)+a}{a}` is used.

Then we can remove the comment further down in the code.

ckolbPTB (approved):

PR #1001, opened 2026-07-04
Downgrade self-hosted runner to cuda 12.1

ckolbPTB:
In order to run the cuda tests on our hpc we need very specific cuda versions and therefore pytorch versions. It is quite complicated to get this to work with our `--upgrade --upgrade-strategy eager` approach. 

I think it should be fine, if the self-hosted runner does not use the upgrade because any latest package changes will be tested with the cpu tests anyway. 

There are a few corner cases where the current approach for the cuda tests could fail but they should be caught by the `pip check` part.

### DIFF
--- a/docker/Dockerfile_runner
+++ b/docker/Dockerfile_runner
@@ -7,8 +7,6 @@ FROM ${BASE_IMAGE} AS base
 
 ARG CUDA_VERSION=cu121
 
-RUN echo "CUDA_VERSION='${CUDA_VERSION}' TORCH_VERSION='${TORCH_VERSION}'"
-
 ARG ACTION_RUNNER_URL=https://github.com/actions/runner/releases/download/v2.335.1/actions-runner-linux-x64-2.335.1.tar.gz
 
 ARG DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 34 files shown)


docker/Dockerfile_runner
```diff
@@ -27,21 +27,30 @@ RUN chmod -R 777 /actions-runner
...
 # install mrpro
 # forces rebuild on any change in the mrpro directory
 COPY . /mrpro/
-RUN python -m pip install "/mrpro[notebooks]" --no-cache-dir --upgrade --upgrade-strategy "eager" \
-    torch --extra-index-url https://download.pytorch.org/whl/${CUDA_VERSION}
-RUN rm -rf /mrpro
+RUN python -m pip install  "/mrpro[notebooks]" --no-cache-dir && rm -rf /mrpro
+
+# check for mrpro package errors
+RUN python -m pip check | tee /tmp/pip_check.txt; \
```
fzimmermann89, line 45: This is to check if we installed a version that is lower than the minimum version requiered by mrpro, correct?

docker/Dockerfile_runner
```diff
@@ -27,21 +27,30 @@ RUN chmod -R 777 /actions-runner
...
+# install mrpro
 WORKDIR /
 
+# install mrpro dependencies
+# forces rebuild if either the version or the pyproject.toml changes
 COPY docker/install_dependencies.sh pyproject.toml src/mrpro/VERSION ./
-
-# the 1st argument for the install_dependencies specifies the cuda version 12.4 for torch
+# the 1st argument for the install_dependencies specifies the cuda version
 RUN bash install_dependencies.sh ${CUDA_VERSION} && rm install_dependencies.sh pyproject.toml VERSION
```
fzimmermann89: this provides CUDA_VERSION as the first argument, but this is no longer used in install_depencencies?

fzimmermann89 (approved):

docker/Dockerfile_runner
ckolbPTB: yes, exactly

PR #1005, opened 2026-08-09
Fix ismrmrd backward compatiblity

fzimmermann89:
This allows us to load  ismrmrd 1.14 compatille data with common issues.
 For example our own data from zenoodo and a4im osi2one data from zenodo.

### DIFF
--- a/src/mrpro/data/KData.py
+++ b/src/mrpro/data/KData.py
@@ -7,6 +7,7 @@ from collections.abc import Callable, Mapping, Sequence
 from importlib.metadata import version
 from types import EllipsisType
 from typing import TYPE_CHECKING, Literal, cast
+from xml.etree import ElementTree as ET
 
 import h5py
 import ismrmrd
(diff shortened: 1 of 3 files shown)


koflera (approved):
Can't tell too much about it except that it seems to indeed solve the problem of not being able to load the OSI2 data :)

fzimmermann89:
@schuenke @ckolbPTB 
This is a design decision: do we only accept valid mrd files - or are we backward compatible to our own files?

src/mrpro/data/KData.py
```diff
@@ -121,7 +122,18 @@ def from_file(
         # Can raise FileNotFoundError
         with ismrmrd.File(filename, 'r') as file:
             dataset = file[list(file.keys())[dataset_idx]]
-            ismrmrd_header = dataset.header
+            try:
+                ismrmrd_header = dataset.header
+            except TypeError:
+                root = ET.fromstring(dataset._contents['xml'][0])  # noqa: S314
```
ckolbPTB, line 129:
```suggestion
# Ismrmrd files created with ismrmrd<1.15 can lead to errors regarding patientPosition and trajectory type when reading in with ismrmrd >=1.15.
                root = ET.fromstring(dataset._contents['xml'][0])  # noqa: S314
```

ckolbPTB: I would add here a comment why this is needed.

ckolbPTB (approved):

ckolbPTB:
> @schuenke @ckolbPTB This is a design decision: do we only accept valid mrd files - or are we backward compatible to our own files?

I would prefer to only accpet valid mrd files but this code seems to be a small price to pay to ensure we can read in the low field data.

